PR #88, opened 2023-09-12
Change convention for referring to CoIDs

aeplay:


aeplay:
Turns out this creates lots of problems with types because TS can't backtrack from T['id'] but can from CoID<T>

PR #89, opened 2023-09-12
Stream transactions

### LINKED ISSUE #82: Stream large transactions individually / keep more signatures
Note to self, still todo:

- [x] Write tests! For cases where TD size is like 1/3 max size and where it is over max size. Make sure the correct sequence of signatures is sent
- [x] Store signatures in storages and apply same chunking logic
- [x] Expose max size as constant and use that in jazz-browser when pushing BinaryCoStream chunks

aeplay:


aeplay:
This will also fix #90

PR #127, opened 2023-11-03
fix(jazz-nodejs): also set peersToLoadFrom on newly created accounts

KyleAMathews:


aeplay:
Thanks so much for this! This makes you the first contributor to jazz/cojson 🎉

I added a comment mostly for myself, I'll fix this next time I'm in front of a keyboard

PR #153, opened 2023-12-21
Move to auth to use `localStorage` instead of `sessionStorage`

tobiaslins:
As discussed, I think it's better to have more persistent sessions per default.
After your API changes merge, I'm happy to change this into a more pluggable system?


aeplay (approved):

aeplay:
Looks good, thanks so much!

PR #155, opened 2023-12-22
Further improvements to setup

tobiaslins:
- Move internal dependencies to `peerDependencies`
- Use carets instead of `*`
-


aeplay:
Thanks for your help!

PR #156, opened 2023-12-22
Fix loading prop always being false

mirshko:
This fixes custom Auth components never changing from non loading state


aeplay:
Thank you very much!

PR #189, opened 2024-05-20
Move `effect` to `peerDependencies` to avoid conflicts

tobiaslins:
- Make sure that we don't have 2 effect versions installed in case the user also has effect installed

image


aeplay:
Closing this for now to see if Effect creates more incompatibilities in the future.

For DX it would be great if people didn't have to explicitly install Effect.

PR #200, opened 2024-05-31
fix: improve compatibility with React compiler and concurrent features

gdorsi:
Thanks for creating this framework!

I've tried locally to use the new React Compiler (https://react.dev/learn/react-compiler) and got a couple of issues.

This PR address those problems and should make jazz-react more compliant with the compiler requirements.

### DIFF
--- a/packages/jazz-react/src/index.tsx
+++ b/packages/jazz-react/src/index.tsx
@@ -1,9 +1,8 @@
-import React, { useEffect, useRef, useState } from "react";
+import React, { useEffect, useState } from "react";
 import {
     consumeInviteLinkFromWindowLocation,
     createJazzBrowserContext,
 } from "jazz-browser";
-
 
 import {
     Account,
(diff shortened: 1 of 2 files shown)


packages/jazz-react/src/index.tsx
```diff
@@ -18,7 +18,7 @@ import { AuthState, ReactAuthHook } from "./auth/auth.js";
 
 /** @category Context & Hooks */
 export function createJazzReactContext<Acc extends Account>({
-    auth: authHook,
+    auth: useAuthHook,
```
gdorsi, line 20: Without the `use` prefix the compiler treats this function as a cacheable value, breaking the "order" rule of hooks

packages/jazz-react/src/index.tsx
```diff
@@ -128,24 +128,18 @@ export function createJazzReactContext<Acc extends Account>({
...
             if (!id || !me) return;
-            return subscribeToCoValue(Schema, id, me, depth, (update) => {
-                state.current = update as DeeplyLoaded<V, D>;
-
-                setUpdates((u) => {
-                    return u + 1;
-                });
+        
+            return subscribeToCoValue(Schema, id, me, depth, (value) => {
+                setState({ value })
```
gdorsi: Using an object to hold the value reference should fix the update issues as it will trigger a re-render at each update (pretty much similar to the previous force update hack)

The difference is that now the value is committed inside the state, so it should be more compliant with the react conccurent requirements.

aeplay:
Thank you very much for the fix and detailed explanation @gdorsi!

Released in jazz-react@0.7.7 🎉

PR #218, opened 2024-06-30
jazz-tools: allow to mutate nullable fields into null

Schniz:
when having a `co.encoded(Schema.NullOr(Schema.String))`, construction
with null works well, but mutating a value into null throws.
This commit fixes it and adds a test that verifies it actually works.


aeplay (approved):
Thank you very much!

PR #222, opened 2024-07-08
comap: support undefined values when using co.encoded

Schniz:
All encoded values are expected to be JSON serializable, which is cool, but they might also be `undefined`. So this PR allows `Encoder<T>` to be derived from JsonValue or undefined


Schniz:
superseded by #234

PR #223, opened 2024-07-08
cojson-transport-ws: reuse runtime and use fibers instead of setTimeout

Schniz:
when calling Effect.runFork we don't propagate layers and using fibers can allow us to interrupt ongoing requests when the pings fail

### DIFF
--- a/packages/cojson-transport-ws/src/index.ts
+++ b/packages/cojson-transport-ws/src/index.ts
@@ -8,6 +8,7 @@ import {
     Runtime,
     Console,
     Fiber,
+    GlobalValue,
 } from "effect";
 
 interface WebsocketEvents {
@@ -24,6 +25,14 @@ interface AnyWebSocket {
     close(): void;
     send(data: string): void;
 }
+
+const jazzPings = GlobalValue.globalValue<
+    Array<{
+        received: number;
+        sent: number;
+        dc: string;
+    }>
+>("jazzPings", () => []);
 
 export function createWebSocketPeer(options: {
     id: string;
@@ -90,16 +99,7 @@ export function createWebSocketPeer(options: {
                         return;
                     }
 
-                    const g: typeof globalThis & {
-                        jazzPings?: Array<{
-                            received: number;
-                            sent: number;
-                            dc: string;
-                        }>;
-                    } = globalThis;
-
-                    g.jazzPings ||= [];
-                    g.jazzPings.push({
+                    jazzPings.push({
                         received: Date.now(),
                         sent: msg.time,
                         dc: msg.dc,
--- a/packages/cojson-transport-ws/src/index.ts
+++ b/packages/cojson-transport-ws/src/index.ts
@@ -8,7 +8,6 @@ import {
     Runtime,
     Console,
     Fiber,
-    GlobalValue,
 } from "effect";
 
 interface WebsocketEvents {
@@ -26,13 +25,13 @@ interface AnyWebSocket {
     send(data: string): void;
 }
 
-const jazzPings = GlobalValue.globalValue<
-    Array<{
+const g: typeof globalThis & {
+    jazzPings?: {
         received: number;
         sent: number;
         dc: string;
-    }>
->("jazzPings", () => []);
+    }[];
+} = globalThis;
 
 export function createWebSocketPeer(options: {
     id: string;
@@ -100,7 +99,8 @@ export function createWebSocketPeer(options: {
                     return;
                 }
 
-                jazzPings.push({
+                g.jazzPings ||= [];
+                g.jazzPings.push({
                     received: Date.now(),
                     sent: msg.time,
                     dc: msg.dc,
--- a/packages/cojson-transport-ws/src/index.ts
+++ b/packages/cojson-transport-ws/src/index.ts
@@ -82,29 +82,29 @@ export function createWebSocketPeer(options: {
                     yield* Queue.offer(
                         incoming,
                         Either.left(new PingTimeoutError()),
-                    ).pipe(Effect.forkDaemon);
-                    yield* Effect.sync(() => {
-                        try {
-                            ws.close();
-                        } catch (e) {
-                            console.error(
-                                "Error while trying to close ws on ping timeout",
-                                e,
-                            );
-                        }
-                    });
-
-                    if (msg.type !== "ping") {
-                        yield* Queue.offer(incoming, Either.right(msg));
-                        return;
+                    ).pipe(Effect.forkScoped);
+                    try {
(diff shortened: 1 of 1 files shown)


packages/cojson-transport-ws/src/index.ts
```diff
@@ -106,3 +125,15 @@ export function createWebSocketPeer(options: {
         };
     });
 }
+
+const addEventListener = <Event extends keyof WebsocketEvents, R>(
```
Schniz, line 129: technically I think it's better to make this a `acquireRelease` because if we register an event listener we should also free it when a Scope is closed. wdyt?

Schniz:
converted to draft: breaks the chat app :think: 
edit: solved I believe<picture data-single-emoji=":relieved_galactic:" title=":relieved_galactic:">:relieved_galactic:</picture>:!

packages/cojson-transport-ws/src/index.ts
```diff
@@ -29,72 +47,73 @@ export function createWebSocketPeer(options: {
...
+                            "Error while trying to close ws on ping timeout",
+                            e,
+                        );
+                    }
+                }).pipe(Effect.scoped, Effect.forkDaemon);
+
+                if (msg?.type !== "ping") {
+                    yield* Queue.offer(incoming, Either.right(msg)).pipe(
+                        Effect.forkDaemon,
                     );
```
Schniz, line 98: you did Effect.runPromise so I assumed you wanted to fork this

packages/cojson-transport-ws/src/index.ts
```diff
@@ -29,72 +47,73 @@ export function createWebSocketPeer(options: {
...
-                    Effect.sync(() => ws.send(JSON.stringify(msg))),
-                ),
-                Effect.runPromise,
-            );
-        });
+        yield* addEventListener(ws, "open", () =>
+            Queue.take(outgoing).pipe(
+                Effect.andThen((message) => ws.send(JSON.stringify(message))),
+                Effect.forever,
+            ),
```
Schniz, line 115: no need to convert a queue to a stream to read it sequentially

packages/cojson-transport-ws/src/index.ts
```diff
@@ -29,72 +47,73 @@ export function createWebSocketPeer(options: {
...
+                    ).pipe(Effect.forkScoped);
+                    try {
+                        ws.close();
+                    } catch (e) {
+                        yield* Console.error(
+                            "Error while trying to close ws on ping timeout",
+                            e,
+                        );
+                    }
+                }).pipe(Effect.scoped, Effect.forkDaemon);
```
Schniz, line 93: The scope here allows us to interrupt the Queue.offer as I'm using forkScoped (afaik)

packages/cojson-transport-ws/src/index.ts
```diff
@@ -29,72 +47,73 @@ export function createWebSocketPeer(options: {
...
+                    if (Exit.isFailure(e) || Exit.isInterrupted(e)) {
+                        return Console.warn("Failed closing ws", e);
+                    }
+                    return Effect.void;
+                }),
+            ),
+        );
 
-        let pingTimeout: ReturnType<typeof setTimeout> | undefined;
+        let pingTimeout: Fiber.Fiber<void> | undefined;
```
Schniz, line 66: instead of setTimeout we can use fibers and then Fiber.interrupt!

packages/cojson-transport-ws/src/index.ts
```diff
@@ -29,72 +47,73 @@ export function createWebSocketPeer(options: {
...
 
-        ws.addEventListener("message", (event) => {
-            const msg = JSON.parse(event.data as string);
+        yield* addEventListener(ws, "message", (event) =>
+            Effect.gen(function* () {
+                const msg = yield* Effect.sync(() =>
+                    JSON.parse(event.data as string),
+                );
```
Schniz, line 72: I wonder if we should use Schema here to make sure it's a valid struct, or at least a valid JSON (`Schema.parseJson(Schema.Unknown)`)

packages/cojson-transport-ws/src/index.ts
```diff
@@ -1,20 +1,38 @@
...
+    Effect,
+    Exit,
+    Runtime,
+    Console,
+    Fiber,
+} from "effect";
+
+interface WebsocketEvents {
+    close: { code: number; reason: string };
+    message: { data: string | unknown };
```
datner, line 15: `string | unknown` is just regular `unknown` lol

also, what about using `Data.TaggedEnum` for these?
```ts
type WebsocketEvent = Data.TaggedEnum<{
  close: { code: number; reason: string };
  message: { data: unknown };
  open: {}
}>

// later
interface AnyWebSocket {
  addEventListener<K extends WebsocketEvent["_tag"]>(
    type: K,
    listener: (event: Data.TaggedEnum.Value<WebsocketEvent, K>) => void
  )
}
```

packages/cojson-transport-ws/src/index.ts
```diff
@@ -29,72 +47,73 @@ export function createWebSocketPeer(options: {
...
-                }
-            });
-        });
+            ).pipe(
+                Effect.onExit((e) => {
+                    if (Exit.isFailure(e) || Exit.isInterrupted(e)) {
+                        return Console.warn("Failed closing ws", e);
+                    }
+                    return Effect.void;
+                }),
```
datner, line 62:
```suggestion
                Effect.onError((e) => Console.warn("Failed closing ws", e)),
```
same thing

packages/cojson-transport-ws/src/index.ts
datner: I think that this entire dance can be done using an `Effect.race` or even by managing the `ws` using a scope and using effects timeout utils. Basically avoiding so many hard forking which can create subtle bugs if not properly managed

packages/cojson-transport-ws/src/index.ts
datner: It doesn't make sense, offering is a sync action. I believe that this is related to a situation where the `Queue` is being shutdown, if `offer`ed then it would **interrupt** the effect. In other words, it's either not required or "hiding" a bug

packages/cojson-transport-ws/src/index.ts
datner: this operator is redundant, we can replace it wholesale with `Stream.fromEventListener`

datner (commented):
I think a lot of the work in this file can be greatly simplified by using `Stream` and `PubSub` instead of `Effect` and `Queue`

PR #224, opened 2024-07-08
cojson-transport-ws: reuse runtime and use fibers instead of setTimeout 

datner:
- cojson-transport-ws: reuse runtime and use fibers instead of setTimeout
- jazzPings
- fix bugs because I misindented things
- store jazzPings on global
- do stream stuff

an iteration on #223 

@Schniz I could not open a PR to your PR because we're both forks, but your code inspired me to give it a go as well :)

There are 2 design flaws that makes this entire experience much more difficult than it has to be:
1. We cannot use a `Scope` to make all the resource management trivial, making it... non-trivial :)
2. Instead of talking with `Stream`s we're talking with `Queue`s, which lack proper lifetime semantics. This can lead to missed values unless the active state of the queue is checked constantly.


packages/cojson-transport-ws/src/index.ts
```diff
@@ -1,108 +1,121 @@
...
+} = globalThis;
+
 export function createWebSocketPeer(options: {
     id: string;
     websocket: AnyWebSocket;
     role: Peer["role"];
 }): Effect.Effect<Peer> {
     return Effect.gen(function* () {
         const ws = options.websocket;
+        const ws_ = ws as unknown as EventTarget;
```
datner, line 42: `Stream.fromEventListener` expects an `EventTarget` despite using only the `addEventListener` and `removeEventListener` api. ugly.

packages/cojson-transport-ws/src/index.ts
```diff
@@ -1,108 +1,121 @@
...
-                Stream.runForEach((msg) =>
-                    Effect.sync(() => ws.send(JSON.stringify(msg))),
+        yield* Queue.take(outgoing).pipe(
+            Effect.andThen((message) => ws.send(JSON.stringify(message))),
+            Effect.forever,
+            Effect.whenEffect(
+                fromEventListener(ws, "open").pipe(Effect.as(true)),
+            ),
+            Effect.forkDaemon, // :/
+        );
```
datner: reads as:
> `take` one entry from `outgoing` `andThen` `.send` the message. Do this `forever` `when*` the open event is emitted, which we regard `as` `true` .

I wish I would not need to `forkDaemon` here, but we can't attach to the parent scope since it's unmanaged.

packages/cojson-transport-ws/src/index.ts
```diff
@@ -1,108 +1,121 @@
...
+
+        type E = WebsocketEvents["message"];
```
datner, line 80: just prevents prettier from formatting it into the UGLIEST thing I've seen

packages/cojson-transport-ws/src/index.ts
```diff
@@ -1,108 +1,121 @@
...
-                Effect.runPromise,
-            );
-        });
+            ),
+            Stream.mergeLeft(closed),
+            Stream.map((_) => JSON.parse(_.data as string)),
+            Stream.filter(isSyncMessage),
+            Stream.buffer({ capacity: "unbounded" }),
+            Stream.onDone(() => Queue.shutdown(outgoing)),
+        );
```
datner: creates a `Stream` `fromEventListener`, it must emit *any* value **at most** 2.5 seconds apart or we `timeout` and fail. If we do, we print a warning and `try` to close the socket. If it doesn't work, we `catch` the `UnknownError` and report it to the console. 

Then `merge` it with the `closed` stream, ignore anything that comes out of that stream--- here I'm relying on the error handling logic of effect. `closed` is a 0-element `Stream` that if it manages to evaluate, meaning that the `close` event was emitted, it would just immediately fail with `DisconnectedError`. Practically speaking, this error and the fancy way to create it is only required because we can't use scopes.

We then `filter` out all the pings (and save them on the globalThis as a side effect) and put everything into an unlimited `buffer`. This is a bad practice. But its the same thing we had before. Important to note, I've added a behavior change here (everything else works logically exactly the same). The previous impl would listen for messages from t0 and until the stream was closed by the consumer or the socket was closed by anything and there was a lot of hope going around. This impl only listens for messages once the `incoming` stream is evaluated. That can be in t0, t1, or any tn. I make the assumption that it would not be any later than t1 and if thats the case the behavior should be entirely non-breaking. The tests pass too.

Regardless, once this stream is evaluated once and is done, we shut down the `outgoing` queue so the consumer won't attempt to offer data that will never be read.

packages/cojson-transport-ws/src/index.ts
Schniz: Seems like it was fixed on Effect 3.4.0, maybe worthwhile to upgrade?

packages/cojson-transport-ws/src/index.ts
Schniz: woooooah took me a while to understand this, specifically the `whenEffect` as it is changes the direction of reading in a way

What about making it a bit more explicit with a gen so "read as" comments won't be necessary?

```ts
// s/ fromEventListener/ takeFromEventListener/g

yield* Effect.gen(function*() {
  yield* fromEventListener(ws, "open");
  yield* Queue.take(outgoing).pipe(Effect.andThen(...))
});
```

packages/cojson-transport-ws/src/index.ts
```diff
@@ -1,108 +1,121 @@
...
+            Stream.timeoutFail(() => new PingTimeoutError(), "2.5 seconds"),
+            Stream.tapError((_e) =>
+                Console.warn("Ping timeout").pipe(
+                    Effect.zipRight(Effect.try(() => ws.close())),
+                    Effect.catchAll((e) =>
+                        Console.error(
+                            "Error while trying to close ws on ping timeout",
+                            e,
+                        ),
+                    ),
```
Schniz: can we consider gen here too to avoid the jargon?

```ts
Effect.gen(function*() {
  yield* Console.warn("Ping timeout");
  yield* Effect.try(() => ws.close).pipe(
    Effect.catchAll(e => Console.error(...)) // really considering try/catch here
  )
});

File: packages/cojson-transport-ws/src/index.ts (hunk 1 above)
Comment (Schniz) [inline packages/cojson-transport-ws/src/index.ts:42]: can we merge the two lines

File: packages/cojson-transport-ws/src/index.ts [hunk 6]
```diff
@@ -1,108 +1,121 @@
...
-                    ws.close();
-                } catch (e) {
-                    console.error(
-                        "Error while trying to close ws on ping timeout",
-                        e,
-                    );
-                }
-            }, 2500);
+        const isSyncMessage = (msg: unknown): msg is SyncMessage => {
+            // boo side-effects
```
Schniz: let's remove the commentary <picture data-single-emoji=":meow_pensive-pray:" title=":meow_pensive-pray:">:meow_pensive-pray:</picture>

packages/cojson-transport-ws/src/index.ts
```diff
@@ -1,108 +1,121 @@
...
-            if (msg.type === "ping") {
-                // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                (globalThis as any).jazzPings =
-                    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                    (globalThis as any).jazzPings || [];
-                // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                (globalThis as any).jazzPings.push({
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            if ((msg as any)?.type === "ping") {
+                const ping = msg as PingMsg;
```
Schniz, line 59: what if PingMessage was Schema? Then we could decodeOptionUnknown or something similar

packages/cojson-transport-ws/src/index.ts
```diff
@@ -1,108 +1,121 @@
...
-            void Stream.fromQueue(outgoing).pipe(
-                Stream.runForEach((msg) =>
-                    Effect.sync(() => ws.send(JSON.stringify(msg))),
+        yield* Queue.take(outgoing).pipe(
+            Effect.andThen((message) => ws.send(JSON.stringify(message))),
+            Effect.forever,
+            Effect.whenEffect(
+                fromEventListener(ws, "open").pipe(Effect.as(true)),
+            ),
+            Effect.forkDaemon, // :/
```
Schniz: let's remove commentary here too <picture data-single-emoji=":detective_pikachu:" title=":detective_pikachu:">:detective_pikachu:</picture>

packages/cojson-transport-ws/src/index.ts
Schniz: The tests don't test this package, but I verified manually that the chat example works <picture data-single-emoji=":meow_verified:" title=":meow_verified:">:meow_verified:</picture>

packages/cojson-transport-ws/src/index.ts
```diff
@@ -1,108 +1,121 @@
...
-                Stream.mapEffect((either) => either),
-            ),
+            incoming: messages,
             outgoing,
             role: options.role,
         };
     });
 }
+
+const fromEventListener = <Event extends keyof WebsocketEvents>(
```
Schniz: maybe we can rename this to `once`?

packages/cojson-transport-ws/src/index.ts
```diff
@@ -1,108 +1,121 @@
...
-
-            if (pingTimeout) {
-                clearTimeout(pingTimeout);
-            }
+        const closed = fromEventListener(ws, "close").pipe(
+            Effect.map(
+                (event) =>
+                    new DisconnectedError(`${event.code}: ${event.reason}`),
+            ),
+            Effect.flip,
```
Schniz:
```suggestion
            Effect.andThen(
                (event) =>
                    new DisconnectedError(`${event.code}: ${event.reason}`),
            ),
```

a bit shorter don't you think? <picture data-single-emoji=":relieved_galactic:" title=":relieved_galactic:">:relieved_galactic:</picture> or will it infer the wrong types?

packages/cojson-transport-ws/src/index.ts
datner: No, the types are incompatible :/

packages/cojson-transport-ws/src/index.ts
datner: if `PingMessage` was a schema (and the rest were too) it would make things nicer.
Do you want me to add the corresponding schemas?

packages/cojson-transport-ws/src/index.ts
datner: `DisconnectError` is not a `YieldableError`, so I have to `Effect.fail` or `Effect.flip`

packages/cojson-transport-ws/src/index.ts
datner: not my call, @aeplay?

packages/cojson-transport-ws/src/index.ts
datner: I wanted to avoid running the `fromEventListener` more than once, I could also use `Effect.once` for this but thought this was nicer. Guess not haha

packages/cojson-transport-ws/src/index.ts
datner: I can use `.andThen` here as well instead of `zipRight`. I prefer the pipe version because of how aggressive the indents in this project are haha, tell me what you think

packages/cojson-transport-ws/src/index.ts
datner: The chat example works perfectly from what I could tell 😄

packages/cojson-transport-ws/src/index.ts
```diff
@@ -69,21 +67,21 @@ export function createWebSocketPeer(options: {
...
-            Effect.forkDaemon, // :/
-        );
+        yield* Effect.gen(function* () {
+            yield* once(ws, "open");
+            yield* Queue.take(outgoing).pipe(
+                Effect.andThen((message) => ws.send(JSON.stringify(message))),
+                Effect.forever,
+                Effect.forkDaemon,
+            );
+        });
```
Schniz: wdyt about having a managed thingy?

```suggestion
        yield* Effect.gen(function* () {
            yield* once(ws, "open");
            yield* Queue.take(outgoing).pipe(
                Effect.andThen((message) => ws.send(JSON.stringify(message))),
                Effect.forever,
                Effect.forkScoped,
            );
            yield* once(ws, "close");
        }).pipe(Effect.scoped, Effect.forever, Effect.forkDaemon);
```

it might be ugly but what I'm trying to encapsulate is: dont send stuff to the websocket if it's not open, and wait for "open" once it's closed. Maybe it just means that we need to merge the outgoing into the messages stream?

datner: `open` will only ever emit once, when the socket connects. It will never emit multiple times for the same socket. Nothing will emit before it besides an `error` _instead_ of it.
`close` will only ever emit once, when the socket is closing the connection. It will never emit multiple times for the same socket. Nothing will emit after it.

This would be supporting an undefined behavior, it's impossible for a socket to reopen after it has closed. You must create a new socket.

datner: basically, managing it like that would mean that for `n` sockets created and `m<=n` being used, we're raising `n-m` zombies and `m` daemons...
Once the the socket would manage itself instead of needing consumers to manage it, this won't be an issue..
Basically, I'd recommend using `@effect/platform/Socket` and implementing [a client for it](https://github.com/Effect-TS/effect/blob/3429c8fb61dcef45df1d954c18b94851b6a9226f/packages/experimental/src/DevTools/Client.ts#L45). It's already generic, it's builtin, its duplex, and Tim would be maintaining it instead of us so we could focus on more important things. But until then... We don't need to expect more than a single `open` event 😄

aeplay:
Thank you very much for the hard work @Schniz & @datner! 

I'll probably only get around to having a deeper look at this tomorrow.

Happy to upgrade Effect if that simplifies/unlocks some APIs, will let you know once that's in master.

For the long term Effect platforms Socket looks very interesting. Will familiarise myself and then we can switch to that soon I hope

packages/cojson-transport-ws/src/index.ts
aeplay: Hey @datner I updated main to use Effect ^3.5.2 - please may you merge and see if it's possible to avoid the type assertion? Otherwise I'm happy to do it

packages/cojson-transport-ws/src/index.ts
aeplay: Overkill IMO just keep it as is

packages/cojson-transport-ws/src/index.ts
aeplay: Can we make it a YieldableError?

packages/cojson-transport-ws/src/index.ts
aeplay: FWIW I'll have a look and see if switching to `@effect/platform/Socket` makes sense in the near future

aeplay (approved):
LGTM - if you get rid of the type assertion that's hopefully no longer necessary that's great but not necessary

packages/cojson-transport-ws/src/index.ts
aeplay: I think I'm going to switch to 2 spaces soon...

packages/cojson-transport-ws/src/index.ts
datner: Yeah ofc

PR #230, opened 2024-07-13
Add support for property existence

### LINKED ISSUE #194: `in` operator not implemented?
When trying to check if a property exists in a CoMap (mostly when using unions), its always falsy.

```ts
if("someId" in myCoMap) {}
```

I currently fallback to 
```ts
if (myCoMap as any).someId) {}
```

tobiaslins:
Allow to do `'someproperty' in someCoMap` to check for optional properties. 

resolves #194

### DIFF
--- a/packages/jazz-tools/src/coValues/coMap.ts
+++ b/packages/jazz-tools/src/coValues/coMap.ts
@@ -611,10 +611,6 @@ const CoMapProxyHandler: ProxyHandler<CoMap> = {
         }
     },
     has(target, key) {
-        if (key in target) {
-            return true;
-        }
-
         const descriptor = (target._schema[key as keyof CoMap["_schema"]] ||
             target._schema[ItemsSym]) as Schema;
 
--- a/packages/jazz-tools/src/coValues/coMap.ts
+++ b/packages/jazz-tools/src/coValues/coMap.ts
@@ -615,16 +615,14 @@ const CoMapProxyHandler: ProxyHandler<CoMap> = {
             return true;
         }
 
-        if (typeof key === "string") {
-            const descriptor = (target._schema[key as keyof CoMap["_schema"]] ||
-                target._schema[ItemsSym]) as Schema;
+        const descriptor = (target._schema[key as keyof CoMap["_schema"]] ||
+            target._schema[ItemsSym]) as Schema;
 
-            if (descriptor) {
-                return target._raw.get(key) !== undefined;
-            }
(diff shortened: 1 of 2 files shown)


packages/jazz-tools/src/coValues/coMap.ts
```diff
@@ -610,6 +610,22 @@ const CoMapProxyHandler: ProxyHandler<CoMap> = {
             }
         }
     },
+    has(target, key) {
+        if (key in target) {
+            return true;
+        }
+
+        if (typeof key === "string") {
```
aeplay: Could you make this more similar to the implementations of the other proxy traps - i.e. handle the case where it's a string key and in the schema and otherwise use `return Reflect.has(...)`?

aeplay (changes requested):

packages/jazz-tools/src/coValues/coMap.ts
```diff
@@ -610,6 +610,20 @@ const CoMapProxyHandler: ProxyHandler<CoMap> = {
             }
         }
     },
+    has(target, key) {
+        if (key in target) {
```
aeplay: This should be taken care of by Reflect.has (default behaviour)

packages/jazz-tools/src/coValues/coMap.ts
```diff
@@ -610,6 +610,20 @@ const CoMapProxyHandler: ProxyHandler<CoMap> = {
...
+            return true;
+        }
+
+        const descriptor = (target._schema[key as keyof CoMap["_schema"]] ||
+            target._schema[ItemsSym]) as Schema;
+
+        if (typeof key === "string" && descriptor) {
+            return target._raw.get(key) !== undefined;
```
aeplay, line 618: "in" apparently returns true if an object has a property, even if it's set to undefined. I think the closest we can get to this is seeing if it's in the schema, or in a record like comap if it is defined. Thoughts?

aeplay (changes requested):

packages/jazz-tools/src/coValues/coMap.ts
tobiaslins: hmmm. for my usecases I use it to discriminate the type of an union

aeplay: Ok let's keep it like this for now and see if it ever causes problems

aeplay (approved):

PR #234, opened 2024-07-21
add co.optional syntax

Schniz:
This PR adds a new `co.optional` object that acts as an optional constructor for schemas

```ts
export class MyStruct extends CoMap {
  name = co.optional.string;
  birthday = co.optional.encoded(Schema.UndefinedOr(Encoders.Date));
}

myStruct.birthday = undefined; // works
```

### DIFF
--- a/packages/jazz-tools/src/implementation/schema.ts
+++ b/packages/jazz-tools/src/implementation/schema.ts
@@ -17,6 +17,36 @@ export type IfCo<C, R> = C extends infer _A | infer B
     : never;
 export type UnCo<T> = T extends co<infer A> ? A : T;
 
+const optional = {
+    ref: optionalRef,
+    json<T extends JsonValue>(): co<T | undefined> {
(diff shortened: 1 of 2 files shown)


Schniz:
I wonder if `delete struct.encoded` should call `struct.encoded = undefined`, as schemas might receive `undefined` as a value

it might be weird that `delete` will do anything other than setting it to `undefined`, but it's also weird that `x.y = undefined` behaves so much different

wdyt?

aeplay (approved):

aeplay:
> I wonder if `delete struct.encoded` should call `struct.encoded = undefined`, as schemas might receive `undefined` as a value
> 
> 
> 
> it might be weird that `delete` will do anything other than setting it to `undefined`, but it's also weird that `x.y = undefined` behaves so much different
> 
> 
> 
> wdyt?

I need to figure out the difference between `delete` and `= undefined` on a more fundamental level anyways. In many ways delete should actually mean delete the key. But this is out of scope for this concern

PR #244, opened 2024-07-31
Remove effectful API for loading/subscribing

aeplay:


Schniz:
interesting! what's the intent here? I actually thought about the same thing previously and instead it can be a module-level function from `jazz-effect`. Feels more "Effectful" to write `JazzEffect.subscribe(covalue.id)` or `JazzEffect.load(covalue.id)` rather than instance methods

aeplay:
> interesting! what's the intent here? I actually thought about the same thing previously and instead it can be a module-level function from `jazz-effect`. Feels more "Effectful" to write `JazzEffect.subscribe(covalue.id)` or `JazzEffect.load(covalue.id)` rather than instance methods

Yes, this is exactly the plan!

PR #248, opened 2024-08-03
[Inspector] New inspector with improved UX

timolins:
This PR adds a new inspector, with the following key features:
- Fullscreen pages with breadcrumbs instead of a tree view
  - This makes navigating into deeper level easier and less noisy
  - This layout is better suited for making things editable in the future
- Two view modes 👁️
  - Table for lists
  - Cards for properties
- Multi-Account Support
- Option to view and download blobs
- "Inspect my account" button

Some portions of the previous inspector haven't been ported yet (e.g. Sessions), so I left the initial app under `app-legacy.tsx`

image


aeplay (approved):
Looks fantastic

PR #285, opened 2024-08-10
docs: fix the get started links position

gdorsi:
A small style fix regarding the links on the "Get started" section.

|||
|--|--|
|Before|After|
|Screenshot 2024-08-10 at 16 18 34|Screenshot 2024-08-10 at 16 18 40|

### DIFF
--- a/homepage/homepage/app/page.mdx
+++ b/homepage/homepage/app/page.mdx
@@ -268,11 +268,7 @@ Jazz Mesh is currently free &mdash; and it's set up as the default sync & storag
 
 ## Get Started
 
--   <Link href="/docs" target="_blank">
-        Read the docs
-    </Link>
--   <Link href="https://discord.gg/utDMjHYg42" target="_blank">
-        Join our Discord
-    </Link>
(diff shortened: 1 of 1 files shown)


homepage/homepage/app/page.mdx
```diff
@@ -268,11 +268,7 @@ Jazz Mesh is currently free &mdash; and it's set up as the default sync & storag
 
 ## Get Started
 
--   <Link href="/docs" target="_blank">
-        Read the docs
-    </Link>
```
gdorsi, line 273: The newline here drives the MDX processor to add a `p` tag around the text breaking the layout.

Screenshot 2024-08-10 at 16 20 35

The problem is that Prettier automatically inserts this newline so it's very easy to introduce the bug again.

**suggestion:** would be better to disable the Prettier formatting on MDX files as it is easy to incur in this kind of issues

aeplay: Good point! https://github.com/gardencmp/jazz/issues/303

aeplay:
Grazie mille ❤️

PR #302, opened 2024-08-12
feat: added asPlainData(depth?) method for CoMap and CoList

pax-k:


aeplay:
Looks great so far!

Please could you create a sync version `asPlainData()` based on whatever is available and async `asPlainDataLoaded(depth)`?

pax-k:
How should I handle circular references?

For example, here is an example without circular references (taken from the tests):
```
        const innerMap = SimpleMap.create(
            {
                name: "Bob",
                age: 25,
                isActive: false,
            },
            { owner: me }
        );
        const nestedMap = NestedMap.create(
            {
                info: innerMap,
                data: "Some data",
            },
            { owner: me }
        );

        const plainData = nestedMap.asPlainData();
        expect(plainData).toEqual({
            info: {
                name: "Bob",
                age: 25,
                isActive: false,
            },
            data: "Some data",
        });
```

But consider this schema: 
```
export class PasswordItem extends CoMap {
  name = co.string;
  username = co.optional.string;
  username_input_selector = co.optional.string;
  password = co.string;
  password_input_selector = co.optional.string;
  uri = co.optional.string;
  folder = co.ref(Folder);
  deleted = co.boolean;
}

export class PasswordList extends CoList.Of(co.ref(PasswordItem)) {}

export class Folder extends CoMap {
  name = co.string;
  items = co.ref(PasswordList);
}

export class FolderList extends CoList.Of(co.ref(Folder)) {}

export class PasswordManagerAccountRoot extends CoMap {
  folders = co.ref(FolderList);
}

export class PasswordManagerAccount extends Account {
  profile = co.ref(Profile);
  root = co.ref(PasswordManagerAccountRoot);
  ```

We're using `co.ref(Folder)` for both `FolderList` and `PasswordItem.folder`. 

Now, if I want to create a plain object:
```
const { me, logOut } = useAccount();
console.log(me.root?.asPlainData())
```

We get the error:
```
RangeError: Maximum call stack size exceeded
    at Object.ownKeys (coMap.ts:651:12)
    at Function.keys (<anonymous>)
    at Proxy.asPlainDataSync (coMap.ts:543:34)
    at Proxy.asPlainData (coMap.ts:501:25)
    at Proxy.asPlainDataSync (coList.ts:480:39)
    at Proxy.asPlainData (coList.ts:462:25)
    at Proxy.asPlainDataSync (coMap.ts:551:51)
    at Proxy.asPlainData (coMap.ts:501:25)
    at Proxy.asPlainDataSync (coMap.ts:548:51)
    at Proxy.asPlainData (coMap.ts:501:25)
```

One way to fix this is to use a WeakMap to keep track of objects that have already been processed: 
```
if (seen.has(this)) {
   return { _circular: this.id } // or return seen.get(this) to get an infinitely nested object
}
```

And then `console.log(me.root?.asPlainData())` looks like this: 
```
{
    "folders": [
        {
            "name": "Default",
            "items": [
                {
                    "name": "123",
                    "username": "user@gmail.com",
                    "password": "password123",
                    "uri": "https://gmail.com",
                    "folder": {
                        "_circular": "co_zWA4FKAjAUbNc5HT4YiH4d1EE5s"
                    },
                    "deleted": false
                },
                {
                    "name": "Facebook",
                    "username": "user@facebook.com",
                    "password": "facebookpass",
                    "uri": "https://facebook.com",
                    "folder": {
                        "_circular": "co_zWA4FKAjAUbNc5HT4YiH4d1EE5s"
                    },
                    "deleted": false
                }
            ]
        }
    ]
}
```

I'm not sure how it's best to handle this.

**Comment by aeplay** [comment]
Please may you adapt the implementation as suggested by you - so for circular references `asPlainData` returns plain data that is itself circularly references itself?

File: packages/jazz-tools/src/coValues/coList.ts [hunk 1]
```diff
@@ -446,6 +449,77 @@ export class CoList<Item = any> extends Array<Item> implements CoValue {
...
+    asPlainData<L extends CoList, Depth>(
+        this: L,
+        depthOrSeen?:
+            | (Depth & DepthsIn<L>)
+            | WeakMap<CoMap | CoList, RecursiveCoMapInit<L>>,
+    ): RecursiveCoMapInit<L> | Promise<RecursiveCoMapInit<L> | undefined> {
+        if (depthOrSeen instanceof WeakMap) {
+            return this.asPlainDataSync(depthOrSeen);
+        } else if (depthOrSeen === undefined) {
+            return this.asPlainDataSync(new WeakMap());
```
gdorsi, line 467: **question:** why using a `WeakMap` here instead of a `Map`?

`Map` should be a little faster than `WeakMap`, so if we don't have a particular reason to use `WeakMap` we should favor the usage of `Map`

packages/jazz-tools/src/coValues/coList.ts
```diff
@@ -446,6 +449,77 @@ export class CoList<Item = any> extends Array<Item> implements CoValue {
...
+        }
+    }
+
+    private asPlainDataSync<L extends CoList>(
+        this: L,
+        seen: WeakMap<CoMap | CoList, RecursiveCoMapInit<UnCo<L[number]>>[]>,
+    ): RecursiveCoMapInit<L> {
+        if (seen.has(this)) {
+            return seen.get(this) as RecursiveCoMapInit<L>;
+        }
```
gdorsi, line 479: **suggestion:** we can avoid to use the type cast this way:

```
const circularRef = seen.get(this);

if (circularRef) return circularRef;
```

Also since this codepath can become very hot when serializing big chunks of state using this approach we reduce the number of access to the map

packages/jazz-tools/src/coValues/coList.ts
```diff
@@ -446,6 +449,77 @@ export class CoList<Item = any> extends Array<Item> implements CoValue {
...
+        const plainArray: Array<RecursiveCoMapInit<UnCo<L[number]>>> = [];
```
gdorsi: **suggestion:** since we know the size of the array beforehand we can use `new Array(this.length)` the reduce the number of reallocations to 1 when filling the array

Can look like a micro-optimization, but on big sized arrays makes some difference.

packages/jazz-tools/src/coValues/coMap.ts
```diff
@@ -485,6 +489,77 @@ export class CoMap extends CoValueBase implements CoValue {
...
+
+        const plainObject = {} as RecursiveCoMapInit<M>;
+        seen.set(this, plainObject);
+
+        for (const key of Object.keys(this)) {
+            const value = this[key as keyof M];
+            (plainObject as any)[key] = plainDataStrategy.toPlainData(
+                value as CoMap,
+                seen,
+            );
```
gdorsi, line 529: **suggestion:** you can cast the type at his source making it easier to use `key` 

```
        for (const key of Object.keys(this) as Array<keyof M>) {
           const value = this[key] as CoMap;
           plainObject[key] = plainDataStrategy.toPlainData(
                value,
                seen,
          );
```

packages/jazz-tools/src/coValues/coMap.ts
```diff
@@ -617,6 +692,7 @@ const CoMapProxyHandler: ProxyHandler<CoMap> = {
         }
     },
     deleteProperty(target, key) {
+        console.log("delete", key);
```
gdorsi, line 695: **nit:** did you intend to leave this `console.log` here?

packages/jazz-tools/src/coValues/plainData.ts
```diff
@@ -0,0 +1,32 @@
...
+export const plainDataStrategy: PlainDataStrategy = {
+    toPlainData<T>(
+        value: T,
+        seen: WeakMap<CoMap | CoList, RecursiveCoMapInit<any>>,
+    ): RecursiveCoMapInit<T> {
+        if (value instanceof CoMap || value instanceof CoList) {
+            if (seen.has(value)) {
+                return seen.get(value) as RecursiveCoMapInit<T>;
+            }
+            return (value as CoMap).asPlainData(seen) as RecursiveCoMapInit<T>;
```
gdorsi: **question:** why do you need the `as CoMap` cast here?

gdorsi (approved):
Generally the PR looks good to me.

Left some non-blocking suggetions and questions.

packages/jazz-tools/src/coValues/coList.ts
```diff
@@ -446,6 +449,77 @@ export class CoList<Item = any> extends Array<Item> implements CoValue {
...
     }
+
+    asPlainData<L extends CoList>(this: L): RecursiveCoMapInit<L>;
+    asPlainData<L extends CoList, Depth>(
+        this: L,
+        depth: Depth & DepthsIn<L>,
+    ): Promise<RecursiveCoMapInit<L> | undefined>;
+    asPlainData<L extends CoList, Depth>(
+        this: L,
+        depthOrSeen?:
```
aeplay, line 460: this signature is a bit unwieldy and we don't really want to expose the "seen" logic to the user, even in the type. Please may you pull out a helper function that does the recursion, with separate parameters for depth and seen and then we can keep the signature here clean? Thanks!

packages/jazz-tools/src/coValues/coList.ts
aeplay: Since we only use the Map to collect seen instances and don't expose this to the outside after we return, I think we can indeed use just a normal `Map` and it will be GC'd just fine.

packages/jazz-tools/src/coValues/plainData.ts
aeplay: yeah can we have a constraint on `T` to be `CoMap | CoList`?

aeplay (changes requested):
Looks good to me overall, just a couple comments and we sort out the types together and then I think this is a really good first iteration

aeplay:
Closing this for now - we'll likely use it as a basis for a bigger API change in the future

PR #305, opened 2024-08-13
feat(example): add a music player example

gdorsi:
# Description

I'm creating a new example to show how the `BinaryCoStream` can be used to manage files and how the share API can be used to enable collaboration across multiple accounts.


examples/musicPlayer/src/1_schema.ts
```diff
@@ -0,0 +1,60 @@
...
+    co,
+    Profile,
+    Account,
+  } from "jazz-tools";
+  
+  export class MusicTrack extends CoMap {
+    file = co.ref(BinaryCoStream);
+    title = co.string;
+    duration = co.number;
+    waveform = co.json<number[]>();
```
gdorsi: waveform is a list of numbers that won't never change.

Not sure if `co.json` is the better primitive for this kind of data.

examples/musicPlayer/src/3_useMediaPlayer.tsx
```diff
@@ -0,0 +1,108 @@
+import { usePlayMedia } from "@/lib/audio/usePlayMedia";
+import { usePlayState } from "@/lib/audio/usePlayState";
+import { useAccount, useCoState } from "@/lib/jazz";
+import { MusicTrack, Playlist } from "@/1_schema";
+import { BinaryCoStream } from "jazz-tools";
+import { useEffect } from "react";
+
+export function useMediaPlayer() {
+    const { me } = useAccount();
```
gdorsi, line 10: I have some issues with this hook when using the `depth` param.

By calling the hook like this when trying a cold start the account never gets loaded:
```
const { me } = useAccount({
       root: {
            activePlaylist: {
                tracks: [{}],
            },
            activeTrack: {
                file: [],
            },
        },
});
```

Also it was causing some type errors on the below assignements:

```ts
me.root.activeTrack = next;
```

@aeplay is this a wrong usage of the API or a bug to report?

examples/musicPlayer/src/3_useMediaPlayer.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    const activeTrack = useCoState(MusicTrack, me?.root?.activeTrack?.id, {
+        file: [],
```
gdorsi: This depth definition was required to have a non-null value on the `file` attribute.
Also passing an array here doesn't look right, but TS forced me into it.

Is this intentional?

examples/musicPlayer/src/1_schema.ts
aeplay: if that's the case, that's indeed the best way to do it - just how huge is it? Is it worth putting it in a referenced sub-coMap so we can choose to only load it when needed?

gdorsi: > Is it worth putting it in a referenced sub-coMap so we can choose to only load it when needed?

Awesome idea!
Would be great way to show off some load optimizations 🚀

PR #313, opened 2024-08-14
Remove effect from jazz-tools and dependent packages

aeplay:


packages/cojson-storage-indexeddb/src/index.ts
```diff
@@ -53,11 +52,15 @@ export class IDBStorage {
         this.db = db;
         this.toLocalNode = toLocalNode;
```
gdorsi, line 53: **question:** what this property does?

It would be great to have it documented via comments

aeplay: Storage in cojson acts as "just another peer", which is why the storage implementation has a communication channel to the localNode (in memory collection of CoValues).

This fact should indeed be documented somewhere, see https://github.com/gardencmp/jazz/issues/269

PR #327, opened 2024-08-19
fix(CoMapInit): type validation on falsy values for required refs

gdorsi:
Updates the `co.ref` type defintions to flag the following use cases as not valid:

```ts
class TestMap extends CoMap {
  required = co.ref(NestedMap);
}

TestMap.create(
  {
     // @ts-expect-error null can't be passed to a non-optional field
     required: null,
  },
  { owner },
)

TestMap.create(
  // @ts-expect-error Non-optional fields can't be omitted
  {},
  { owner },
)
```

### DIFF
--- a/packages/jazz-tools/src/implementation/schema.ts
+++ b/packages/jazz-tools/src/implementation/schema.ts
@@ -88,6 +88,7 @@ function optionalRef<C extends CoValueClass>(
 
 function ref<C extends CoValueClass>(
     arg: C | ((_raw: InstanceType<C>["_raw"]) => C),
+    options?: never,
 ): co<InstanceType<C> | null>;
 function ref<C extends CoValueClass>(
     arg: C | ((_raw: InstanceType<C>["_raw"]) => C),
--- a/packages/jazz-tools/src/tests/coMap.test.ts
+++ b/packages/jazz-tools/src/tests/coMap.test.ts
@@ -1,4 +1,4 @@
-import { expect, describe, test } from "vitest";
+import { expect, describe, test, expectTypeOf } from "vitest";
 import { connectedPeers } from "cojson/src/streamUtils.js";
 import { newRandomSessionID } from "cojson/src/coValueCore.js";
 import {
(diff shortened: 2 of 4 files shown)


packages/jazz-tools/src/tests/coMap.test.ts
```diff
@@ -734,3 +734,74 @@ describe("CoMap applyDiff", async () => {
...
+            {
+                optional: NestedMap.create({ value: "" }, { owner: me }),
+                // @ts-expect-error This should return error
+                required: null,
+            },
+            { owner: me },
+        );
+
+        // No runtime validation is implemented for now
+        expect(map).toBeTruthy();
```
gdorsi, line 764: Wondering if we also should throw an error here.

From my experience falsy values on non-optional fields can lead to weird situations, so IMHO it would be better to throw early to avoid polluting the store.

packages/jazz-tools/src/implementation/schema.ts
```diff
@@ -88,7 +88,8 @@ function optionalRef<C extends CoValueClass>(
 
 function ref<C extends CoValueClass>(
     arg: C | ((_raw: InstanceType<C>["_raw"]) => C),
-): co<InstanceType<C> | null>;
+    options?: never,
+): co<InstanceType<C>>;
```
aeplay: this means that co.ref fields on the class where they are used are always defined type-wise, right? This may be incorrect, because a value might not be loaded yet and return `null` when get-accessed.

So the type for the field has to be `InstanceType<C> | null` and while we might not be able to do anything about not *assigning* null, we can prevent it in the constructor (`CoMapInit`) at least?

aeplay (changes requested):

packages/jazz-tools/src/implementation/schema.ts
gdorsi: Oops, I missed that. Thanks for catching it!

Added an additional test to ensure the nullability of the ref properties and gone with the solution that you proposed.

aeplay (approved):
Bellissimo 😍

PR #335, opened 2024-08-20
fix(BinaryCoStream): resolve the load request only when the stream is ended

gdorsi:
Currently  `BinaryStream.loadAsBlob` resolves into undefined if the BinaryStream isn't complete.

With this PR we change the behavoir to wait for the BinaryStream to be completed before resolving the promise.

### DIFF
--- a/packages/cojson/src/coValues/coStream.ts
+++ b/packages/cojson/src/coValues/coStream.ts
@@ -96,15 +96,20 @@ export class RawCoStreamView<
     }
 
     getSingleStream(): Item[] | undefined {
-        if (Object.keys(this.items).length === 0) {
+        const streams = Object.values(this.items);
+        const firstStream = streams[0];
+
+        if (!firstStream) {
             return undefined;
-        } else if (Object.keys(this.items).length !== 1) {
+        }
+
+        if (streams.length > 1) {
             throw new Error(
                 "CoStream.getSingleStream() can only be called when there is exactly one stream",
             );
         }
 
-        return Object.values(this.items)[0]?.map((item) => item.value);
+        return firstStream.map((item) => item.value);
     }
 
     sessions(): SessionID[] {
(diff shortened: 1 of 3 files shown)


packages/cojson/src/coValues/coStream.ts
```diff
@@ -96,15 +96,20 @@ export class RawCoStreamView<
...
+        }
+
+        if (streams.length > 1) {
             throw new Error(
                 "CoStream.getSingleStream() can only be called when there is exactly one stream",
             );
         }
 
-        return Object.values(this.items)[0]?.map((item) => item.value);
+        return firstStream.map((item) => item.value);
```
gdorsi, line 112: A light refactoring to avoid the `?.` operator here

packages/cojson/src/coValues/coStream.ts
```diff
@@ -306,20 +320,9 @@ export class RawBinaryCoStreamView<
             const chunk = base64URLtoBytes(
                 item.chunk.slice(binary_U_prefixLength),
             );
```
gdorsi, line 322: **observation:** I can see this sync decoding become a problem with large files

How do you feel about making `getBinaryChunks` async and see how a worklet would perform here? @aeplay

packages/jazz-tools/src/coValues/deepLoading.ts
```diff
@@ -81,8 +86,8 @@ export function fulfillsDepth(depth: any, value: CoValue): boolean {
                           ).optional,
             );
         }
-    } else if (value._type === "BinaryCoStream") {
-        return true;
+    } else if (isBinaryCoStream(value)) {
+        return value.isBinaryStreamEnded();
```
gdorsi, line 90: **observation:** change affects not only `loadAsBlob` but also `load` and `ensureLoaded`

I've gone with this road because:
- `loadAsBlob` already uses `load` interally to check if the stream is loaded
- Resolving `load` and `ensureLoaded` when the stream is completed is kinda what I would expect reading at these methods names

The drawback is that when the user deep loads a file from `useCoState` the hook would return a value only when the file is completely downloaded (am I right?)

Example:
```
// me is returned only when all the tracks files are downloaded
const { me } = useAccount({
  root: { activePlaylist: { tracks: [{ file: [] }] } }
});
```

This isn't probably good for the normal use case, where it would be better to get the JSON data as soon as possible and leave the files to be downloaded in background.

Maybe it would be better to change only `loadAsBlob` and leave `load` and `ensureLoaded` as they are.

WDYT @aeplay?

gdorsi: Writing this comment made me realize that is better to change only `loadAsBlob`

Opened another PR with a better implementation https://github.com/gardencmp/jazz/pull/337

PR #337, opened 2024-08-20
fix(BinaryCoStream): resolve the loadAsBlob call only when the stream id ended

gdorsi:
Currently BinaryStream.loadAsBlob resolves into undefined if the BinaryStream isn't complete.

With this PR we change the behavoir to wait for the BinaryStream to be completed before resolving the promise.

NOTE: This is the second iteration for the bugfix. I've the first one I've also changed the behavoir of `load` but later found out that wasn't the best choice: https://github.com/gardencmp/jazz/pull/335#discussion_r1723661794

### DIFF
--- a/packages/cojson/src/coValues/coStream.ts
+++ b/packages/cojson/src/coValues/coStream.ts
@@ -96,15 +96,20 @@ export class RawCoStreamView<
     }
 
     getSingleStream(): Item[] | undefined {
-        if (Object.keys(this.items).length === 0) {
+        const streams = Object.values(this.items);
+        const firstStream = streams[0];
+
+        if (!firstStream) {
             return undefined;
-        } else if (Object.keys(this.items).length !== 1) {
+        }
+
+        if (streams.length > 1) {
             throw new Error(
                 "CoStream.getSingleStream() can only be called when there is exactly one stream",
             );
         }
 
-        return Object.values(this.items)[0]?.map((item) => item.value);
+        return firstStream.map((item) => item.value);
     }
 
     sessions(): SessionID[] {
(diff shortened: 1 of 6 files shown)


packages/cojson/src/coValues/coStream.ts
```diff
@@ -96,15 +96,20 @@ export class RawCoStreamView<
...
+        }
+
+        if (streams.length > 1) {
             throw new Error(
                 "CoStream.getSingleStream() can only be called when there is exactly one stream",
             );
         }
 
-        return Object.values(this.items)[0]?.map((item) => item.value);
+        return firstStream.map((item) => item.value);
```
gdorsi, line 112: Done a small refactoring to avoid the ?. operator here

packages/cojson/src/coValues/coStream.ts
```diff
@@ -306,20 +320,9 @@ export class RawBinaryCoStreamView<
             const chunk = base64URLtoBytes(
                 item.chunk.slice(binary_U_prefixLength),
             );
```
gdorsi, line 322: **observation:** I can see this sync decoding become a problem with large files

How do you feel about making getBinaryChunks async and see how a worklet would perform here? @aeplay

packages/jazz-tools/src/coValues/coStream.ts
```diff
@@ -540,7 +544,28 @@ export class BinaryCoStream extends CoValueBase implements CoValue {
             allowUnfinished?: boolean;
         },
     ): Promise<Blob | undefined> {
-        const stream = await this.load(id, as, []);
+        let stream = await this.load(id, as, []);
```
gdorsi, line 547: **question:** in which cases `stream` can be `undefined`?

Wondering what to do if the user is passing `allowUnfinished: true` and `stream` is `undefined`

packages/cojson/src/coValues/coStream.ts
```diff
@@ -306,20 +320,9 @@ export class RawBinaryCoStreamView<
...
-            if (Date.now() - lastProgressUpdate > 100) {
-                lastProgressUpdate = Date.now();
-            }
         }
 
-        // const after = performance.now();
-        // console.log(
-        //     "getBinaryChunks bandwidth in MB/s",
-        //     (1000 * totalLength) / (after - before) / (1024 * 1024)
-        // );
```
gdorsi, line 321: Removed this debug code to make it easier for me to understand the logic.

Happy to revert the change if this code is still useful

packages/jazz-tools/src/coValues/coStream.ts
aeplay: When the CoValue for the stream doesn't load at all.

In that case just return undefined.

aeplay (approved):

aeplay (approved):

PR #367, opened 2024-09-01
feat(jazz-run): add a command to spin a local sync server

### LINKED ISSUE #180: cojson-simple-sync doesn't run in node20
With node20 (v20.12.2), installing and running `cojson-simple-sync` (version `0.6.1`, also tried `0.6.1-alpha.0`) results in this error:
```
COJSON sync server listening on port 4200
file:///.../node_modules/cojson-simple-sync/dist/index.js:10
globalThis.crypto = webcrypto;
                  ^

TypeError: Cannot set property crypto of #<Object> which has only a getter
    at file:///.../node_modules/cojson-simple-sync/dist/index.js:10:19
```

I switched to node18, reinstalled, and it runs fine.

gdorsi:
In order to fix #180 and make the `cojson-simple-sync` easier to maintain we are integrating it inside `jazz-run`

### DIFF
--- a/packages/jazz-run/src/startCoJsonSimpleSync.ts
+++ b/packages/jazz-run/src/startCoJsonSimpleSync.ts
@@ -1,0 +1,80 @@
+import { Command, Options } from "@effect/cli";
+import {
+    ControlledAgent,
+    LocalNode,
+    WasmCrypto,
+    cojsonInternals,
+} from "cojson";
+import { WebSocketServer } from "ws";
+
+import { createWebSocketPeer } from "cojson-transport-ws";
+import { Effect } from "effect";
+
+const port = Options.text("port")
+    .pipe(Options.withAlias("p"))
+    .pipe(Options.withDefault("4200"));
+
+export const startCoJsonSimpleSync = Command.make(
+    "cojson-simple-sync",
+    { port },
+    ({ port }) => {
+        return Effect.gen(function* () {
+            const crypto = yield* Effect.promise(() => WasmCrypto.create());
+
+            const wss = new WebSocketServer({ port: parseInt(port) });
+
+            console.log(
+                "COJSON sync server listening on port " + wss.options.port,
+            );
+
+            const agentSecret = crypto.newRandomAgentSecret();
+            const agentID = crypto.getAgentID(agentSecret);
+
+            const localNode = new LocalNode(
+                new ControlledAgent(agentSecret, crypto),
+                cojsonInternals.newRandomSessionID(agentID),
+                crypto,
+            );
+
+            wss.on("connection", function connection(ws, req) {
+                const pinging = setInterval(() => {
+                    ws.send(
+                        JSON.stringify({
+                            type: "ping",
+                            time: Date.now(),
+                            dc: "unknown",
+                        }),
+                    );
+                }, 1500);
+
+                ws.on("close", () => {
+                    clearInterval(pinging);
+                });
+
+                const clientAddress =
+                    (req.headers["x-forwarded-for"] as string | undefined)
+                        ?.split(",")[0]
+                        ?.trim() || req.socket.remoteAddress;
+
+                const clientId = clientAddress + "@" + new Date().toISOString();
+
+                localNode.syncManager.addPeer(
+                    createWebSocketPeer({
+                        id: clientId,
+                        role: "client",
+                        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                        websocket: ws as any, // TODO: fix types
+                        expectPings: false,
+                    }),
+                );
+
+                ws.on("error", (e) =>
+                    console.error(`Error on connection ${clientId}:`, e),
+                );
+            });
+
+            // Indefintely pending promise to keep the server up
+            yield* Effect.promise(() => new Promise(() => {}));
+        });
+    },
+);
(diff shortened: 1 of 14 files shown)


packages/jazz-run/src/startCoJsonSimpleSync.ts
```diff
@@ -0,0 +1,80 @@
...
+                    }),
+                );
+
+                ws.on("error", (e) =>
+                    console.error(`Error on connection ${clientId}:`, e),
+                );
+            });
+
+            // Indefintely pending promise to keep the server up
+            yield* Effect.promise(() => new Promise(() => {}));
```
gdorsi: I'm very new to Effect and haven't found a better way to handle the websocket server.

aeplay: This is fine - we're only using effect for effect-cli here

aeplay (changes requested):
This is perfect and thanks for adjusting the mentions of it in the readme as well.

I'd like to make the name of the command even simpler though: could we make it just `npx jazz-run sync`?

Finally cojson-simple-sync had persistence afaik, can we add cojson-storage-SQLite to this? Maybe behind a default-on flag called --persist

packages/jazz-run/src/startCoJsonSimpleSync.ts
aeplay: I'll have a look what an idiomatic way to wait forever is in effect

gdorsi:
@aeplay 

Added the storage and renamed the command to `sync`

For opting out from the file-based storage I've added an `in-memory` option.

Does it work for you?

packages/jazz-run/src/startSync.ts
```diff
@@ -0,0 +1,121 @@
...
+            );
+
+            /**
+             * In Jazz the communication is decentralized
+             * so the sync server is implemented as "just another peer" in the network
+             *
+             * LocalNode is the class we use to manage the connected peers that could be:
+             * - the storage layer
+             * - the WebSocket connections
+             */
```
gdorsi: @aeplay 

Could you double-check if what I'm saying here is correct?

aeplay: I'd say this kind of stuff belongs into the internal documentation of LocalNode and high level docs of CoJSON as a protocol - and if it's clear there the use of websocket and storage peers here is obvious, so let's remove the comment here for now

aeplay: (your understanding is correct though)

packages/jazz-run/src/startSync.ts
```diff
@@ -0,0 +1,121 @@
...
+                    }),
+                );
+
+                ws.on("error", (e) =>
+                    console.error(`Error on connection ${clientId}:`, e),
+                );
+            });
+
+            // Indefintely pending promise to keep the server up
+            yield* Effect.promise(() => new Promise(() => {}));
```
aeplay: Asking some of the Effect pros I know what to best do here

Schniz: can also do `yield* Effect.never` I guess

fubhy: I'd probably use Layer.launch and launch a SocketServer here (see @effect/experimental SocketServer.layerWebSocket). So basically creating the websocket server through Effect instead of via ws manually.

You'd likely then also want to utilize Effect for the file system interaction (creating the database, etc.) here.

That said, I'm not sure if that's in scope for this PR and if that's leading you on a bunch of tangents and Effect exploration journeys that you might not be ready for / willing to go on for the sake of this PR ;-)

gdorsi: Thanks for the suggestions!

Gone with `yield* Effect.never` as doing more isn't within the scope of this PR (the goal was to land with a minimal implementation of a sync server)

Interested on learning how to use Effect for the file system interactions / websocket server.

Are there any examples or docs to look at to get into it? @fubhy

gdorsi:
@aeplay 

Addressed all the suggestions.

aeplay:
Wonderful, changeset pls. Will setup CI to complain (instead of me) soon

Thanks for the PR 🎉

aeplay (approved):

packages/jazz-run/src/startSync.ts
aeplay: Thanks @Schniz for the quick suggestion

Thanks @fubhy for the more extensive roadmap - out of scope for this PR but exactly what I had in mind when we have the time to re-adopt Effect more meaningfully

fubhy: > Thanks for the suggestions!
> 
> Gone with `yield* Effect.never` as doing more isn't within the scope of this PR (the goal was to land with a minimal implementation of a sync server)
> 
> Interested on learning how to use Effect for the file system interactions / websocket server.
> 
> Are there any examples or docs to look at to get into it? @fubhy

That makes sense.

There are no examples for the SocketServer afaik. But there's an implementation of a SocketServer in the DevTools package within the Effect monorepo.

Either way, I'd suggest to hit our Discord server once you start diving more into Effect. That's the best place to ask questions and learn.

Regarding the FileSystem module there are some tests here that might give you an idea: https://github.com/Effect-TS/effect/blob/main/packages/platform-node-shared/test/FileSystem.test.ts

gdorsi:
> Wonderful, changeset pls. Will setup CI to complain (instead of me) soon

@aeplay Added!

aeplay:
Made it a patch instead :)

aeplay (approved):

PR #376, opened 2024-09-03
Password manager demo

pax-k:


aeplay:
I don't know why some errors only show up in the Vercel build of the homepage (I guess it's not trying to build the password manager example yet) - but please may you address those before we merge it?

PR #378, opened 2024-09-04
fix(inspector): subscribe to latent covalues instead of loading them immediately

pax-k:
- added a new function `subscribeToCoValue()` to [examples/inspector/src/viewer/use-resolve-covalue.ts](https://github.com/gardencmp/jazz/pull/378/files#diff-256a17e2292c8d8014e5055607bc05d67e37ac9dc3fd1e3c4fe4841ee6ee2f50)
- `useResolvedCoValue()` and `useResolvedCoValues()` now use `subscribeToCoValue()` instead of `resolveCoValue()`
- basically using `node.subscribe()` over `node.load()`

### DIFF
--- a/.changeset/thin-olives-fold.md
+++ b/.changeset/thin-olives-fold.md
@@ -1,0 +1,5 @@
+---
+"jazz-inspector": patch
+---
+
+fix(inspector): subscribe to latent covalues instead of loading them immediately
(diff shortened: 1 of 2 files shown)


aeplay:
Looks good to me overall, please can you make it use 4 spaces so the diff is cleaner?

aeplay:
More general comment is that load should behave correctly - but this is a bigger problem we need to address in cojson over the longer term. The nice benefit of subscribing is that we have a live view of the CoValues I guess?

pax-k:
> The nice benefit of subscribing is that we have a live view of the CoValues I guess?

Yes, at least we get this benefit for now.

> More general comment is that load should behave correctly - but this is a bigger problem we need to address in cojson over the longer term. 

Opening a new issue to track this https://github.com/gardencmp/jazz/issues/379

aeplay (approved):

PR #383, opened 2024-09-06
test(deepLoading): fix skipped test

gdorsi:


packages/jazz-tools/src/tests/deepLoading.test.ts
```diff
@@ -138,8 +139,8 @@ describe("Deep loading with depth arg", async () => {
             throw new Error("map4 is undefined");
         }
         expect(map4.list[0]?.stream).not.toBe(null);
-        expect(map4.list[0]?.stream?.[me.id]).toBe(undefined)
-        expect(map4.list[0]?.stream?.byMe?.value).toBe(undefined);
+        expect(map4.list[0]?.stream?.[me.id]).not.toBe(null);
+        expect(map4.list[0]?.stream?.byMe?.value).toBe(null);
```
gdorsi, line 143: Looks like we have improved the load and the result now seems to be more correct

aeplay:
Thanks for catching this!

PR #385, opened 2024-09-06
test: e2e tests for the BinaryCoStream sync

gdorsi:
Created a webapp to use as basis for the e2e tests for BinaryCoStream.

For now we just run a Playwright test on top of it to validate that this works.

Next step will be to compare the performance of our branches with main (planning to use [Tachometer](https://github.com/google/tachometer))

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        example: ["pets"]
+        project: ["examples/pets", "e2e/BinaryCoStream"]
 
     steps:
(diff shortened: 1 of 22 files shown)


.github/workflows/playwright.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        example: ["pets"]
+        project: ["examples/pets"]
```
gdorsi, line 15: NOTE: Had some troubles on running the test on CI because there the sync halts even with small files. Going to enable them when fixing https://github.com/gardencmp/jazz/issues/389

aeplay (approved):
Looks great!

PR #386, opened 2024-09-09
fix(musicPlayer): move the upload of example data outside of the initial migration

gdorsi:
Moved the example upload outside the initial migration because:
- The upload becomes blocking for the account creation flow, making the signup process slower
- By uploading the example data in the intial migration we make it an hard dependency for the account creation, which isn't great since file uploads may fail or take a long amount of time in unstable connections.

### DIFF
--- a/examples/chat/README.md
+++ b/examples/chat/README.md
@@ -39,4 +39,4 @@ If you have feedback, let us know on [Discord](https://discord.gg/utDMjHYg42) or
 
 By default, the example app uses [Jazz Global Mesh](https://jazz.tools/mesh) (`wss://sync.jazz.tools`) - so cross-device use, invites and collaboration should just work.
 
-You can also run a local sync server by running `npx jazz-run sync` and adding the query param `?sync=ws://localhost:4200` to the URL of the example app (for example: `http://localhost:5173/?sync=ws://localhost:4200`), or by setting the `sync` parameter of the `<Jazz.Provider>` provider component in [./src/2_main.tsx](./src/2_main.tsx).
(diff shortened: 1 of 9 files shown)


examples/chat/README.md
```diff
@@ -39,4 +39,4 @@ If you have feedback, let us know on [Discord](https://discord.gg/utDMjHYg42) or
 
 By default, the example app uses [Jazz Global Mesh](https://jazz.tools/mesh) (`wss://sync.jazz.tools`) - so cross-device use, invites and collaboration should just work.
 
-You can also run a local sync server by running `npx jazz-run sync` and adding the query param `?sync=ws://localhost:4200` to the URL of the example app (for example: `http://localhost:5173/?sync=ws://localhost:4200`), or by setting the `sync` parameter of the `<Jazz.Provider>` provider component in [./src/2_main.tsx](./src/2_main.tsx).
+You can also run a local sync server by running `npx jazz-run sync` and adding the query param `?sync=ws://localhost:4200` to the URL of the example app (for example: `http://localhost:5173/?peer=ws://localhost:4200`), or by setting the `sync` parameter of the `<Jazz.Provider>` provider component in [./src/2_main.tsx](./src/2_main.tsx).
```
gdorsi, line 42: Noticed that the real param for the override is `peer` and not `sync`

aeplay (approved):
Good catch - I think in the long run we should be able to deal with failures in the migrations better, right now it's a bit of an Achilles heel but probably best to not do binary uploads in it right now

aeplay:
Please may you have a look why the unrelated playwright tests are failing?

gdorsi:
@aeplay 

Restarting the test made it pass.

We have some flakiness related to the share flow

aeplay:
Ok, merging then. Please may you create a ticket to investigate the share flakiness? Thank you

PR #388, opened 2024-09-09
fix(jazz-react): mark the auth as loading when authState is not ready

gdorsi:
Wanted to implement an "auto-login" but got some issues because `loading` is passed always as `false`.

I guess this can be considered a bug, so I've changed the logic to pass `true`

### DIFF
--- a/.changeset/forty-garlics-punch.md
+++ b/.changeset/forty-garlics-punch.md
@@ -1,0 +1,5 @@
+---
+"jazz-react": patch
+---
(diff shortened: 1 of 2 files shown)


gdorsi:
@aeplay Should I add a test on this change or is this something that will change soon?

aeplay:
This will change anyways with the new auth system but merging for now

aeplay (approved):

PR #394, opened 2024-09-14
fix(BinaryCoStream): set the stream content message as low priority in the WS outgoing channel

gdorsi:
# The problem

Depending on the network quality and the mesh speed a file upload through BinaryCoStream may take minutes and the situation becomes more complicated if we are talking about multiple files download.

This creates congestion on the Websocket outgoing channel blocking the other messages until the file upload is completely processed.

This gets in the way of the "sharing" features of Jazz because the messages required to make the sharing work gets pushed back.

# The solution

I've added a priority system managed by the WS transport that marks the stream content messages as "low priority"

This way updates like the creation of an invitation or even the CoMap that's usually used to contain the BinaryCoStream aren't blocked by the file upload.

# Real world effects

This is a recording of a Playlist sharing before and after the fix.

|||
|--|--|
|Before|After|
| https://github.com/user-attachments/assets/bf450b0e-af4f-4242-a121-f3d4527cb0ef |  https://github.com/user-attachments/assets/b3764871-342e-46e2-afe4-034821247e25 |

## [Loom overview of the changes](https://www.loom.com/share/a0a8b4c945554b359c51fd71b728afb8)


packages/cojson-transport-ws/src/index.ts
```diff
@@ -40,6 +40,24 @@ const g: typeof globalThis & {
...
+    });
+  
+    return {
+      promise,
+      resolve,
+      reject,
+    };
+  }
+
+const BUFFER_LIMIT = 100_000;
```
gdorsi: Using a smaller buffer limit to pause the queue processing earlier.

This way we have more "controlled" buffer and leave to the websocket control only a small amount of messages.

packages/cojson-transport-ws/src/index.ts
```diff
@@ -102,24 +120,77 @@ export function createWebSocketPeer({
...
+        while (highPriorityQueue.length > 0 || lowPriorityQueue.length > 0) {
+            if (websocket.bufferedAmount > BUFFER_LIMIT) {
+                await waitForLessBuffer();
+            }
+
+            if (websocket.readyState !== 1) {
+                return;
+            }
+
+            const entry = highPriorityQueue.shift() ?? lowPriorityQueue.shift();
```
gdorsi, line 148: This way we actually flip the problem, because until we have high priority messages we don't process the low priority ones.

Thinking about using a round robin scheduling, perhaps by pulling 2 high priority messages and 1 low priority.
This way streams would still be sent but we leave more space for the more important updates.

aeplay:
Thank you very much for this experimentation & improvement. I wrote a spec for what I think is the best way to start doing prioritisation within `cojson` itself. @gdorsi please may you use this PR as a base to tackle the more complete vision for this? https://github.com/gardencmp/jazz/issues/396

packages/cojson/src/coValues/coList.ts
```diff
@@ -96,6 +96,8 @@ export class RawCoListView<
         this.afterStart = [];
         this.beforeEnd = [];
 
+        this.core.setPriority(CO_VALUE_PRIORITY.MEDIUM)
```
gdorsi: @aeplay 

I'm setting the priority when the coValue is loaded in a View (e.g. CoMap, Group or BinaryCoStream)

This reduces the effects of this priority system to where the Views are loaded, excluding I guess the mesh servers.

Is there a better way to identify the source View of a coValue?

packages/cojson/src/PeerState.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        ];
+
+        await Promise.all(promises);
+
+        // The first message is pushed directly, the other three are queued because are waiting
+        // for the first push to be completed.
+        expect(peerState["peer"].outgoing.push).toHaveBeenNthCalledWith(1, contentMessageLow); 
+
+        // Load message are managed as high priority messages and having the same priority as the content message
+        // they follow the push order.
```
gdorsi: @aeplay 

I'm considering all the non-content messages as HIGH priority.
Not sure which effect this might have on the system.

My guess was that by giving to the "exchange" message a higher priority we ensure that is processed before any kind of content upload (with the exception of Groups and Accounts) reducing the delay on critical actions like "invitations".

packages/cojson/src/coValues/coList.ts
aeplay: I think we can just always do it on the CoValueCore - we can get the type of covalue from the header and set the default priority based on that - whether it is being created or loaded.

We can worry about how to expose it in the CoValue views (and later Jazz) later

packages/cojson/src/PeerState.test.ts
aeplay: yes, this is good and correct. State info is always more important than content

gdorsi:
Now it works even better than before 🚀 

https://github.com/user-attachments/assets/8f5f1fcc-a1a2-4aa5-9850-3f26c81e93fc

packages/cojson/src/localNode.ts
```diff
@@ -610,7 +610,7 @@ export class LocalNode {
         const groupCoValue = this.createCoValue({
             type: "comap",
             ruleset: { type: "group", initialAdmin: this.account.id },
-            meta: null,
+            meta: { type: "group" },
```
gdorsi: @aeplay 

Used this change to detect the groups and give them an higher priority.

Is it ok?

packages/cojson-transport-ws/src/index.ts
```diff
@@ -102,31 +81,37 @@ export function createWebSocketPeer({
...
     });
 
+    async function pushMessage(msg: SyncMessage) {
+        if (websocket.readyState !== 1) {
+            await websocketOpen;
+        }
+
+        while (websocket.bufferedAmount > BUFFER_LIMIT && websocket.readyState === 1) {
+            await new Promise<void>((resolve) => setTimeout(resolve, BUFFER_LIMIT_POLLING_INTERVAL));
+        }
```
gdorsi, line 91: For now it's ok to leave the buffering loop inside the push.

With the PeerState managing the backpressure we can be sure that there will be one loop active per peer.

packages/cojson-transport-ws/src/index.ts
```diff
@@ -5,41 +5,19 @@ import {
...
+ const g: typeof globalThis & {
     jazzPings?: {
         received: number;
         sent: number;
         dc: string;
     }[];
 } = globalThis;
 
+export const BUFFER_LIMIT = 100_000;
+export const BUFFER_LIMIT_POLLING_INTERVAL = 10;
```
gdorsi, line 19: Lowered the value of the buffering limit to activate the backpressure system earlier.

packages/cojson/src/localNode.ts
aeplay: you could just check the type in the ruleset!

please may you do that?

packages/cojson/src/priority.ts
```diff
@@ -0,0 +1,41 @@
...
+} as const;
+
+export type CoValuePriority = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7;
+
+export function getPriorityFromHeader(header: CoValueHeader | undefined | boolean): CoValuePriority {
+    if (typeof header === "boolean" || !header) {
+        return CO_VALUE_PRIORITY.MEDIUM;
+    }
+
+    if (header.type === "costream") {
```
aeplay: please may you extend the condition with `&& header.meta.type === "binary"` - because non-binary CoStreams should actually be MEDIUM like other CoValues

aeplay (changes requested):

packages/cojson/src/priority.ts
gdorsi: Thanks!

Done!

packages/cojson/src/localNode.ts
gdorsi: Ah, that's better!

Done!

aeplay (approved):

PR #405, opened 2024-09-17
Change mono font to Commit Mono

trishalim:
image


aeplay (changes requested):
Looking very good overall, IBM plex is almost perfect but I'm not a fan of the `{}` (forgot they had so much character)

Please may you change it to Commit Mono? And then we can compare with the next Vercel deploy (just commit on top of this PR)

Thanks!

trishalim:
> Looking very good overall, IBM plex is almost perfect but I'm not a fan of the `{}` (forgot they had so much character)
> 
> Please may you change it to Commit Mono? And then we can compare with the next Vercel deploy (just commit on top of this PR)
> 
> Thanks!

I agree the `{}` was too much. Looks better now.

I can't see the vercel preview because I don't have access.

aeplay (approved):

trishalim:
@aeplay If you're happy with this, feel free to merge.

PR #412, opened 2024-09-19
Fix footer overlaps side nav in docs

trishalim:
before
before

after
after

Clip text that is too long
clipped text


aeplay:
seems to work well already!

trishalim:
Build for the homepage is failing but I only have css changes 🤔

aeplay (approved):
Great fix and thanks for taking care of the horizontal scrolling in the nav container as well

PR #424, opened 2024-09-20
test(BinaryCoStream): add a reconnection e2e test

gdorsi:
Add a e2e test to ensure we handle the offline state and the WS connection drop correctly.


aeplay (approved):
Very cool, thanks!

PR #435, opened 2024-09-23
Fix heading colors in dark mode

trishalim:


trishalim:
We're declaring color palettes in `globals.css` and `tailwind.config.ts` which can cause inconsistencies. For now, this fixes the headings, and I'll sort out the palettes another day.

aeplay (changes requested):
There's still some issues on the Mesh page, I think the list and paragraph components also need to be fixed

PR #444, opened 2024-09-24
Emphasize invite link to join chat in chat example

trishalim:
image

### DIFF
--- a/examples/chat/src/app.tsx
+++ b/examples/chat/src/app.tsx
@@ -19,7 +19,7 @@ export function App() {
   return (
     <AppContainer>
       <TopBar>
-        {me?.profile?.name} · <button onClick={logOut}>Log out</button>
+        <p>{me?.profile?.name}</p> · <button onClick={logOut}>Log out</button>
       </TopBar>
       {useIframeHashRouter().route({
         "/": () => createChat() as never,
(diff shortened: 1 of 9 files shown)


aeplay:
I actually really like this. I have some ideas for the next iteration but right now this separates the app from the surroundings very nicely. And I like the updated styling and how you're making things more consistent with each other

trishalim:
Thanks @aeplay 
Do you think the invite link is clear now?

aeplay:
> Thanks @aeplay Do you think the invite link is clear now?

yes! Was there anything else you wanted to do in this PR? Happy for you to take it out of draft

trishalim:
@aeplay I have to do mobile view and check dark mode

I need to hide the invite bar if there is no chat created
image

aeplay:
So one thing I liked before is that it looked like a little browser. Do you think there's a way to keep the URL in a little URL bar thing, and then once you create a chat maybe have a little note pop up that calls out the shareability of the link?

If that sounds like too much work for this PR, it's fine as is

trishalim:
@aeplay ok will think on that. For now there's some css I'm fixing.

trishalim:
@aeplay what about this?
image

aeplay:
@trishalim love it. Can we make the hint appear only if the URL contains "/#/chat/..."? That way it would pop up on chat creation in a way that draws attention I think

trishalim:
@aeplay great idea! now it's super obvious

PR #449, opened 2024-09-25
Homepage: redesign hero, add diagrams

trishalim:


trishalim:
@aeplay I didn't change the diagram colors. Just couldn't choose a gray shade 😄 Re light mode, will do it another day

trishalim:
Are you happy with this? @aeplay 

> Jazz is an open-source framework that lets you build apps around data that’s synced in real-time between devices and users. It replaces 90% of what frontend + backend stacks do.

I think it needs a bit of polishing. It's kinda bland, and the second sentence feels like it should be in the middle of a paragraph.

What about:

> Jazz is an open-source framework for building apps with real-time data sync across devices and users. Cut out 90% of your stack, and focus on what makes your app your app.

It's an improvement. But the first sentence jumps from one topic to another. Everything else in the hero (aside from the last sentence) talks about the instant sync, then we jump right into how it eliminates a big chunk of the work.

EDIT: let's discuss on Linear

trishalim:
@aeplay Other than the hero copy, what else do we need for this PR? Are you still working on the "hard things are easy now" section, and do you need some help? I think each of those items need a description. Something like this:
image
Move the paragraph up, and have a 4-col grid below it.

aeplay:
@trishalim ooh great idea. The features definitely deserve explanations. Would it be ok if I push an ugly version that has all the content and then you re-style/layout it?

trishalim:
@aeplay Yes, that's great.

PR #451, opened 2024-09-26
Set blue (primary) color palette

trishalim:
image


aeplay (approved):
Looks good to me, once we start using it more in practice we might want to tweak it but it seems like a nice graduation of the brand colour

trishalim:
@aeplay Yes, just needed the 800 for the primary button hover color

PR #453, opened 2024-09-27
Fix scroll issue on chat example app if there are many users

trishalim:


trishalim:
@aeplay in case you missed this

aeplay:
Thanks!

PR #464, opened 2024-10-01
feat: ui refresh for the music player

gdorsi:
Using v0 output as example I've migrated the music-player to a better looking UI

- v0 mockup: https://v0.build/UKqIPEz

Screenshot

Screenshot 2024-10-01 at 16 27 45


bensleveritt:
That looks like a great improvement!

PR #469, opened 2024-10-01
Add book shelf example

trishalim:
[My todo list on Linear](https://linear.app/garden-computing/issue/JAZZ-342/create-book-shelf-example)


examples/book-shelf/src/schema.ts
```diff
@@ -0,0 +1,49 @@
...
+   */
+  migrate(this: JazzAccount, creationProps?: { name: string }) {
+    super.migrate(creationProps);
+    const group = Group.create({ owner: this });
+    group.addMember("everyone", "reader");
+
+    if (!this._refs.root) {
+      this.root = AccountRoot.create(
+        {
+          bookReviews: ListOfBookReviews.create([], { owner: group }),
```
trishalim: When logged in as user A, is it possible to access user B's bookReviews?

I'd like to make a user profile page (see `user/[slug]/page.tsx`) to show user's books publicly.

If it is possible, what am I doing wrong in `UserProfile.tsx`?

aeplay: Just like you can subclass a custom AccountRoot you can also subclass Profile and add more fields to it which will actually be visible to everyone. Your group setup is correct but the root itself is private.


The profile is public by default and you can even re-use its group by setting the bookReviews owner to me.profile._owner

trishalim: This works. Thank you!

trishalim:
@aeplay Any idea what this error means? https://github.com/gardencmp/jazz/actions/runs/11164938640/job/31035438696?pr=469#step:8:453
image

packages/cojson/package.json
```diff
@@ -7,13 +7,13 @@
     "exports": {
         ".": {
             "react-native": "./dist/native/index.native.js",
-            "default": "./dist/web/index.web.js",
-            "types": "./src/index.web.ts"
+            "types": "./src/index.web.ts",
+            "default": "./dist/web/index.web.js"
```
trishalim, line 11: @aeplay @pax-k @bensleveritt @gdorsi 
I was dealing with a build error `Module not found: Error: Default condition should be last one` and going a bit insane, and turns out the fix is to put `default` last here lol I hope this is ok

gdorsi: That should be ok!

Didn't know that the order matters in the exports field! 🤯

aeplay (approved):

packages/cojson/package.json
aeplay: Thanks so much for spotting this! I added a changeset, merged the branch and will do a release so other people can benefit from your fix!

Also thanks for the example!

PR #472, opened 2024-10-01
feat(react): migrate to useSyncExternalStore

gdorsi:
Migrate the React subscription logic to [useSyncExternalStore](https://react.dev/reference/react/useSyncExternalStore).

The main advantage of adopting `useSyncExternalStore` is to follow "The React Way" to subscribe to external stores, which should give us better compatibility and hopefully less surprises with the latest React features.

`useSyncExternalStore` also comes with a dedicated getter which should hopefully help with the SSR integration.

TODO:
- [x] Check how references are managed within Jazz
- [x] Test all the example apps to see if we are breaking some of them
- [x] Add unit tests
- [x] Check the React Compiler compatibility
- [x] Write an issue related to the immutable state and the references graph https://github.com/gardencmp/jazz/issues/565
- [ ] Non blocking: Do some resarch on how we could use Suspense (would probably be very useful for SSR)

### DIFF
--- a/examples/music-player/src/3_HomePage.tsx
+++ b/examples/music-player/src/3_HomePage.tsx
@@ -5,7 +5,7 @@ import { usePlayState } from "./lib/audio/usePlayState";
 import { SidePanel } from "./components/SidePanel";
 import { FileUploadButton } from "./components/FileUploadButton";
 import { Button } from "./components/ui/button";
-import { createNewPlaylist, uploadMusicTracks } from "./4_actions";
+import { createNewPlaylist, updatePlaylistTitle, uploadMusicTracks } from "./4_actions";
 import { useNavigate, useParams } from "react-router";
 import { ID } from "jazz-tools";
 import { Playlist } from "./1_schema";
@@ -62,7 +62,7 @@ export function HomePage({ mediaPlayer }: { mediaPlayer: MediaPlayer }) {
     ) => {
         if (!playlist) return;
 
-        playlist.title = evt.target.value;
+        updatePlaylistTitle(playlist, evt.target.value);
     };
 
     const handlePlaylistShareClick = async () => {
--- a/examples/password-manager/src/3_vault.tsx
+++ b/examples/password-manager/src/3_vault.tsx
@@ -5,9 +5,13 @@ import Table from "./components/table";
 import NewItemModal from "./components/new-item-modal";
 import InviteModal from "./components/invite-modal";
 
-import { saveItem, deleteItem, createFolder, updateItem } from "./4_actions";
+import { saveItem, deleteItem, createFolder, updateItem, addSharedFolder } from "./4_actions";
 import { Alert, AlertDescription } from "./components/alert";
-import { Folder, FolderList, PasswordItem } from "./1_schema";
+import {
+  Folder,
+  FolderList,
+  PasswordItem,
+} from "./1_schema";
 import { useAccount, useCoState } from "./2_main";
 import { CoMapInit, Group, ID } from "jazz-tools";
 import { useNavigate, useParams } from "react-router-dom";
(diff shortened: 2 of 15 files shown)


aeplay:
Thank you very much, I've been meaning to transition to useSyncExternalStore for a while.

CoValues are updated on direct edits because the underlying subscription (to a RawCoValue from cojson and the even deeper CoValueCore) creates a new view object every time the subscription updates.

Forcing recreation of a parent object view on a child update is something that the SubscriptionScope was meant to do but I never got it to work quite right, so I resorted to calling an onUpdate handler from any nested update together with a top-level setState. To fix this, you'll probably need to look into Refs and how they try to cache high-level Jazz CoValues for underlying cojson RawCoValues

packages/jazz-tools/src/implementation/subscriptionScope.ts
```diff
@@ -41,30 +41,29 @@ export class SubscriptionScope<Root extends CoValue> {
...
+            onUpdate(value);
+        };
+
         this.rootEntry.rawUnsub = root._raw.core.subscribe(
             (rawUpdate: RawCoValue | undefined) => {
                 if (!rawUpdate) return;
-                this.rootEntry.value = rootSchema.fromRaw(rawUpdate) as Root;
-                // console.log("root update", this.rootEntry.value.toJSON());
-                subscriptionsScopes.set(this.rootEntry.value, this);
+                this.rootEntry.value = rawUpdate;
```
gdorsi, line 61: @aeplay 

Managed to get the immutable updates working by moving the value view creation inside the `scheduleUpdate`

If this looks a good strategy for you I can proceed with some intensive testing to validate that I'm not breaking anything and there are no major performance regressions.

examples/password-manager/src/3_vault.tsx
```diff
@@ -17,20 +21,19 @@ const VaultPage: React.FC = () => {
...
-    if (!sharedFolderId || !sharedFolder || !me.root?.folders) return;
-    const existsIndex = me.root?.folders.findIndex(
-      (f) => f?.id === sharedFolder.id
-    );
-    if (existsIndex > -1) {
-      me.root?.folders?.splice(existsIndex, 1);
-    }
-    me.root?.folders?.push(sharedFolder);
-    navigate("/vault");
-  }, [sharedFolder, me.root?.folders, sharedFolderId, navigate]);
```
gdorsi, line 33: The `me.root.folders` dependency here now creates an infinite loop because the folders reference changes when the list changes.

Refactored to trigger the effect only on mount and moved the subscription logic into the action.

examples/password-manager/src/4_actions.ts
```diff
@@ -54,3 +55,24 @@ export const shareFolder = (
...
+  sharedFolderId: ID<Folder>,
+  me: PasswordManagerAccount) {
+  const [sharedFolder, account] = await Promise.all([
+    await waitForCoValue(Folder, sharedFolderId, me, Boolean, {}),
+    await waitForCoValue(PasswordManagerAccount, me.id, me, Boolean, {
+      root: {
+        folders: [],
+      },
+    }),
+  ]);
```
gdorsi, line 69: Used this helper to wait for the data to loaded, it works even if the user that triggered the sharing haven't uploaded the folder yet.

gdorsi:
Tested all the example apps!

Only the PasswordManager required a fix because the changes on how we manage references now is actually a breaking change.

Effects with a coValue as dependency are now triggered every time they get an update.

We will probably need to communicate that 🤔 
On the other side this should make us more compatible with the React Compiler, but that's something I need to confirm with more tests.

Next-up, unit tests!

examples/music-player/src/3_HomePage.tsx
```diff
@@ -62,7 +62,7 @@ export function HomePage({ mediaPlayer }: { mediaPlayer: MediaPlayer }) {
     ) => {
         if (!playlist) return;
 
-        playlist.title = evt.target.value;
+        updatePlaylistTitle(playlist, evt.target.value);
```
gdorsi, line 65: The React Compiler doesn't like the Jazz way of updating data:

Screenshot 2024-10-14 at 21 47 29

Fixed by moving all the updates into `actions`

aeplay (approved):
Excellent work, thank you!

PR #484, opened 2024-10-04
test: add basic e2e tests for the chat app

gdorsi:


examples/chat/src/ui.tsx
```diff
@@ -31,7 +31,7 @@ export function BubbleContainer(props: {
   fromMe: boolean | undefined;
 }) {
   const align = props.fromMe ? "items-end" : "items-start";
-  return <div className={`${align} flex flex-col m-2`}>{props.children}</div>;
+  return <div className={`${align} flex flex-col m-2`} role="row">{props.children}</div>;
```
gdorsi, line 36: Added some basic roles to match the a11y patterns implemented on Whatsapp web.

Initially meant to simplify testing but I've preferred to use text selectors in the end.

examples/chat/tests/chat.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    // This breaks with (Error: Error reading from peer storage, handling msg)
+    // await chatPage.logout();
+    // await loginPage.loginAs(mario);
+
+    // await chatPage.expectLoaded();
+
+    // await page.goto(roomURL);
+
+    // await chatPage.expectMessageRow(message1ByMario);
+    // await chatPage.expectMessageRow(message2ByLuigi);
```
gdorsi, line 49: I've not been able to pinpoint the reason of the crash.

The permission structure is quite plain, and the message owners are managed correctly.

Tried with a third user, doesn't work.
After cleaning up the browser storage, creating a new user and joining the room I'm able to see the messages, so it's something related to the browser persistence.

Screenshot 2024-10-04 at 19 12 45

gdorsi: Loading another chat room also works.

So something within the shared chat room gets corrupted and isn't stored correctly on IndexedDB

examples/chat/tests/pages/ChatPage.ts
```diff
@@ -0,0 +1,33 @@
+import { Page, Locator, expect } from "@playwright/test";
+
+export class ChatPage {
+  readonly page: Page;
+  readonly messageInput: Locator;
+  readonly logoutButton: Locator;
+
+  constructor(page: Page) {
+    this.page = page;
+    this.messageInput = page.getByPlaceholder("Type a message and press Enter");
```
bensleveritt: I think ideally we should label the field rather than using the placeholder (at a guess aria-label would be most appropriate here)

gdorsi: Thanks for the suggestion!

Done with a sr-only label.

examples/chat/tests/pages/LoginPage.ts
```diff
@@ -0,0 +1,40 @@
...
+export class LoginPage {
+    readonly page: Page;
+    readonly usernameInput: Locator;
+    readonly signupButton: Locator;
+
+    constructor(page: Page) {
+        this.page = page;
+        this.usernameInput = page.getByRole("textbox");
+        this.signupButton = page.getByRole("button", {
+            name: "Sign up as new account",
```
aeplay:
```suggestion
            name: "Sign up",
```

PR #492, opened 2024-10-07
fix(co.json): allow interface types as generic argument

### LINKED ISSUE #470: co.json: Allow interface types from generics
The `co.json` type factory accepts a generic to let the Jazz users define the expected shape.

We want to make it possible to pass as argument only types that can be serialized, which means that anything that includes functions shouldn't be accepted as valid type.

[The current implementation](https://github.com/gardencmp/jazz/blob/main/packages/jazz-tools/src/implementation/schema.ts#L21) doesn't accept valid interface types, leading to some confusion when trying to use `co.json` correctly.

We want to fix the generic type in order to work in the following way:

```ts
type ValidJSON = {
    prop: string;
    prop2: number;
}

// This is valid ✅
co.json<ValidJSON>();

type NotValidJSON = {
    date: Date;
}

// This isn't valid ✅
co.json<NotValidJSON>();

type NotValidJSON2 = {
    fn: () => void;
}

// This isn't valid ✅
co.json<NotValidJSON2>();

interface ValidInterface {
    prop: string;
    prop2: number;
}

// This is flagged as error but should be valid ❌
co.json<ValidInterface>();
```

gdorsi:
fixes #470 

The main goal fo this PR is to enable support for interface type on `co.json`

This solution comes with some tradeoffs:
- We can only accept interface types at a fixed depth because we can't rely on recursion (Typescript flag the recursion as potentially infinite)
- With this hack anything that is not a number, null, string or Array is transformed to an object. We are checking against empty object types to discard invalid types such as Function or RegExp. This may lead to some unexpected results.

Thanks to @milkcask and @aeplay for the brainstorming!


packages/cojson/src/jsonValue.ts
```diff
@@ -4,3 +4,20 @@ export type JsonAtom = string | number | boolean | null;
...
+/**
+ * Since we are forcing Typescript to elaborate the indexes from the given type passing
+ * non-object values to CoJsonObjectWithIndex will return empty o
+ * E.g. 
+ *   CoJsonObjectWithIndex<() => void> --> {}
+ *   CoJsonObjectWithIndex<RegExp> --> {}
+ * 
+ * Applying the ExcludeEmpty type here to make sure we don't accept functions or non-serializable values
+ */
+export type CoJsonObjectWithIndex<T> = ExcludeEmpty<{ [K in keyof T & string]: JsonValue | undefined }>;
```
gdorsi: Tried with `export type CoJsonObjectWithIndex<T> = ExcludeEmpty<{ [K in keyof T]: CoJsonValue<T[K]> | undefined }>;` but Typescript got angry:

Screenshot 2024-10-07 at 16 17 19

PR #497, opened 2024-10-07
Add co.Date and update test

### LINKED ISSUE #467: Create co.date shortcut

bensleveritt:
Adds Date CoValue `co.Date`, happy to discuss ergonomics (is it Date or date?)

### DIFF
--- a/packages/jazz-tools/src/implementation/schema.ts
+++ b/packages/jazz-tools/src/implementation/schema.ts
@@ -6,6 +6,14 @@ import {
     CoValueFromRaw,
 } from "../internal.js";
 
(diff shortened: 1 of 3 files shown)


aeplay (approved):
My intuition was co.date, but makes sense to use lowercase for primitives and uppercase for encoded things (and it looks like JS's native Date)

bensleveritt:
> My intuition was co.date, but makes sense to use lowercase for primitives and uppercase for encoded things (and it looks like JS's native Date)

I think that's an appropriate distinction (and I too went with Date for the native implementation)

PR #511, opened 2024-10-08
feat(transport-ws): add messages batching

### LINKED ISSUE #344: Add batching to cojson-transport-ws
For apps that load lots of little CoValues, the bottleneck is currently that each of them requires one (or several) WebSocket messages.

We should be able to try out batching in a backwards-compatible way:

* Implement batching by sending multiple sync messages as newline-separated JSON in one WebSocket message
* Test client always uses batching
* Sync server uses batching with clients that are using it
* Probably a good way to batch is to collect all messages within 1ms up until a max batch size (assumption is response time mostly scales with number of CoValues anyways)

gdorsi:
Fixes #344 

In order to increase the messages thoughput of the websocket peer we are adding a batching system for the exchanged messages.

The batching will be released first turned off and with an upgrade system, to deploy it on our server peers before enabling client peers to do batching.


packages/cojson-transport-ws/src/index.ts
```diff
@@ -5,30 +5,27 @@ import {
...
     websocket,
     role,
     expectPings = true,
-}: {
-    id: string;
-    websocket: AnyWebSocket;
-    role: Peer["role"];
-    expectPings?: boolean;
-}): Peer {
+    batchingByDefault = false, // Keeping this false until we release batching upgrade to the mesh
```
gdorsi, line 28: I think that's better to keep it disabled until the version of `createWebSocketPeer` with batching is released to the production mesh.

This way we can first upgrade the mesh to support batching and then enable batching to everyone.

Since the pets e2e tests now run with the local sync we can ensure that the server side supports client that aren't batching.

packages/cojson-transport-ws/src/index.ts
```diff
@@ -43,26 +40,20 @@ export function createWebSocketPeer({
...
-        pingTimeout && clearTimeout(pingTimeout);
-
-        if (msg?.type === "ping") {
-            const ping = msg as PingMsg;
-            g.jazzPings ||= [];
-            g.jazzPings.push({
-                received: Date.now(),
-                sent: ping.time,
-                dc: ping.dc,
-            });
```
gdorsi, line 57: This code as it seemed unused, happy to restore it if we think that could be useful.

packages/cojson-transport-ws/src/BatchedOutgoingMessages.ts
```diff
@@ -0,0 +1,36 @@
+import { SyncMessage } from "cojson";
+
+export const MAX_OUTGOING_MESSAGES_CHUNK_SIZE = 10;
```
gdorsi, line 3: I've set this up to 10 to stay below the 1Kb per batch.

The websockets are on TCP so we shouldn't exceed the window size, but honestly I don't remeber the math behind the window size calculation so I'm setting this under a safe value before doing more research.

packages/cojson-transport-ws/src/index.ts
```diff
@@ -43,26 +40,27 @@ export function createWebSocketPeer({
...
+        let messagesChunk: PingMsg[] | SyncMessage[];
+
+        try {
+            messagesChunk = (event.data as string)
+                .split("\n")
+                .map((msg) => JSON.parse(msg));
+        } catch (e) {
+            console.error("Error while parsing incoming message", e);
+            return;
+        }
```
gdorsi, line 52: Sometimes the server reiceves a `undefined:1` message.

Not sure where this comes, but without this try/catch it causes the server to restart.

e2e/BinaryCoStream/src/jazz.tsx
```diff
@@ -14,10 +13,19 @@ const Jazz = createJazzReactApp();
...
+  const [auth, state] = useDemoAuth();
+
+  const signedUp = useRef(false);
+
+  useEffect(() => {
+    if (state.state === "ready" && !signedUp.current) {
+      state.signUp('Mister X');
+      signedUp.current = true;
+    }
+  }, [state.state])
```
gdorsi, line 25: Had to revert to the DemoAuth here because with the `ephemeralCredentialsAuth` the app wasn't working.

The E2E tests on the CI were working even before this change, so I'm not sure if it was something related to my local environment.

packages/cojson-transport-ws/src/index.ts
aeplay: We'll replace it with a better latency measurement system in the future

packages/cojson-transport-ws/src/index.ts
aeplay: Perfect. Thanks for thinking about how to roll this out

packages/cojson-transport-ws/src/BatchedOutgoingMessages.ts
aeplay: This seems small and it would be better to do the limit roughly by (serialised) data size. I'd suggest limiting it to 1MB initially (if a single message is bigger, that's fine)

Please may you change it to work like that?

packages/cojson-transport-ws/src/index.ts
aeplay: interesting, probably a broken client? Or you receive it even in local testing?

aeplay (changes requested):

e2e/BinaryCoStream/src/jazz.tsx
aeplay: Will need to investigate this, maybe the account doesn't sync properly after creation with `ephemeralCredentialsAuth`

packages/cojson-transport-ws/src/index.ts
gdorsi: Yes, both in local and in the production demo.

packages/cojson-transport-ws/src/BatchedOutgoingMessages.ts
gdorsi: > This seems small and it would be better to do the limit roughly by (serialised) data size.

Good idea!

> I'd suggest limiting it to 1MB initially

WebSockets are working on top of TCP so if we go over the [computed window size](https://networklessons.com/cisco/ccie-routing-switching-written/tcp-window-size-scaling) it will take more than one roundtrip to send the data, slowing down the exchange instead of speeding it up.

The window size depends on the connection quality, but I would say that 50Kb or even 25Kb would be a safer boundary.

My goal with a chunk of 10 was to send around 1Kb of data 😅

aeplay: Interesting - the current max chunk size for binary data / transactions is 100kb, wondering what you think of that regarding TCP window size. But for now let's do 25kb of batched message limit - if an individual message is bigger, it's fine

gdorsi: Done!

aeplay (approved):

PR #513, opened 2024-10-09
test(pets): run tests using local sync

gdorsi:
Trying to switch to use the local sync for the pets tests because:
- We test that the `jazz-run sync` works
- Hopefully should fix the flakyiness of the pets tests by removing the delays of the production mesh

### DIFF
--- a/examples/pets/src/2_main.tsx
+++ b/examples/pets/src/2_main.tsx
@@ -3,16 +3,22 @@ import ReactDOM from "react-dom/client";
 import { Link, RouterProvider, createHashRouter } from "react-router-dom";
 import "./index.css";
 
-import {
-    createJazzReactApp,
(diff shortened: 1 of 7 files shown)


examples/pets/src/2_main.tsx
```diff
@@ -34,7 +40,7 @@ function JazzAndAuth({ children }: { children: React.ReactNode }) {
         <>
             <Jazz.Provider
                 auth={auth}
-                peer="wss://mesh.jazz.tools/?key=pets-example-jazz@gcmp.io"
+                peer={peer}
```
gdorsi, line 43: Noticed that the `peer` param isn't anymore handled by jazz.

Is that something we removed to simplify the integration with rn?

PR #516, opened 2024-10-09
Make feature copy more succinct

aeplay:


trishalim:
@aeplay Great!! 

For this part, the old one was slightly better
> All data you load or create is persisted locally, so your users can work offline. Even on sketchy connections, local changes are continuously synced out.

Can you change to

> All data you load or create is persisted locally, so your users can work offline. When you’re back online, the local changes are synced to the server.

aeplay:
Unsure about the "server" part here, but merging it for now.

PR #525, opened 2024-10-10
Lead homepage with features

aeplay:
First iteration of @trishalim's homepage restructuring ideas.

I think it would be good to mention all features in the very beginning, because it most quickly displays value. But the feature descriptions need to be super short.

This let's us get rid of the second sentence of the intro paragraph.

Haven't made the icons cute and blue yet...

Also need to change the descriptions a bit still to keep them short but not loose important stuff from their respective second sentences.


trishalim:
The hero is now too text-heavy, but at the same time, lots of info lost here.

> All data you load or create is persisted locally, so your users can work offline.

^ gives me the impression that the data is only stored locally

> Build your app around mutable local state attached to an account.

^ doesn't explain how this results to the cross-device sync

> Create & handle structured data and binary streams/blobs the same way.

??

> Get instant updates throughout your UI every time you locally mutate data.

^ this is true for any app. The difference with Jazz is that it syncs to the server

> Groups of accounts are first class entities that you can create anywhere.

^ doesn't tell me anything

I don't think this is the first iteration. We can put 3 items there, 4 would would be pushing it. @aeplay

trishalim:
@aeplay If you're trying to tackle the issue of how the feature section is overwhelming, I think this is a good first iteration:
image
4 in one row was too squished

PR #538, opened 2024-10-11
Redesign landing page hero

trishalim:


trishalim:
@aeplay Ideally, this wouldn't be an image. For now I have 2 images for light and dark, and mobile is too squished. This is what I can finish today though. Up to you if you wanna merge or not. I think it's still an improvement.

aeplay:
> @aeplay Ideally, this wouldn't be an image. For now I have 2 images for light and dark, and mobile is too squished. This is what I can finish today though. Up to you if you wanna merge or not. I think it's still an improvement.

Agree that it shouldn't be an image but yeah it's a good iteration, let's go with it

PR #540, opened 2024-10-11
fix(pets): improve invitation flow

gdorsi:


examples/pets/src/2_main.tsx
```diff
@@ -112,6 +107,16 @@ export default function App() {
...
 }
 
+function AcceptInvite() {
+    const navigate = useNavigate();
+    useAcceptInvite({
+        invitedObjectSchema: PetPost,
+        onAccept: (petPostID) => navigate("/pet/" + petPostID),
+    });
+
+    return  <p>Accepting invite...</p>
```
gdorsi, line 117: By moving `useAcceptInvite` here we trigger it only when the invitation link is active.

Before we required a reload because the hash navigation made by pasting the invite link in the URL wasn't triggering a real page refresh and `useAcceptInvite` wasn't fired until the full page reload.

examples/pets/tests/createAndShare.spec.ts
```diff
@@ -27,6 +29,8 @@ test("create a new post and share", async ({ page }) => {
 
     const invitation = await postPage.getShareLink();
 
+    await sleep(1000);
```
gdorsi, line 32: I've reduced the image size and added a sleep to see if the problem was that the "creator" peer didn't have the time to upload the data.

gdorsi:
Screenshot 2024-10-12 at 12 55 17

Passed without any failure. Mission complete 🚀 

Any other test run was failing at least once: https://github.com/gardencmp/jazz/actions/runs/11295383002/job/31417963947

aeplay:
Thanks for figuring this out!

PR #543, opened 2024-10-14
Fix typo in browserLockSession

bensleveritt:


aeplay:
Good to see you digging around

PR #578, opened 2024-10-16
Support for Vue

pax-k:
Implements
- packages/jazz-vue
- examples/chat-vue


aeplay (changes requested):
Looks good to me, please may you add a changeset so it triggers a release PR?

packages/jazz-vue/src/createJazzVueApp.ts
```diff
@@ -0,0 +1,342 @@
...
+        setup(props, { slots }) {
+            const ctx = ref<
+                BrowserContext<Acc> | BrowserGuestContext | undefined
+            >(undefined);
+
+            provide(JazzContextSymbol, ctx);
+
+            onMounted(async () => {
+                try {
+                    console.log("props.auth", props.auth);
```
gdorsi: **question:** do we want to keep this log?

packages/jazz-vue/src/createJazzVueApp.ts
```diff
@@ -0,0 +1,342 @@
...
+                                  peer: props.peer,
+                                  storage: props.storage,
+                              },
+                    );
+
+                    ctx.value = {
+                        ...context,
+                        logOut: () => {
+                            context.logOut();
+                            ctx.value = undefined;
```
gdorsi: **observation:** Not sure if we need to create a new context on logout

@aeplay How do we handle the login -> logout -> login flow?

gdorsi (approved):
Awesome work!

Just left one question regarding the logout and context management.

packages/jazz-vue/src/createJazzVueApp.ts
pax-k: Fixed

aeplay (approved):

PR #580, opened 2024-10-16
Set meta title, description, and image

trishalim:


trishalim:
The preview for open graph is not working on vercel, but I see nothing wrong with the tags.

I used this image
social-image

But here are some other options
social-image (2)
social-image (1)

aeplay (approved):

aeplay:
Thank you!

PR #584, opened 2024-10-16
Convert hero illustration to code instead of image

trishalim:
- fix syntax errors in code sample
- implement mobile view
- move highlight to another line in step 2


bensleveritt:
Looks good to me, I was going to suggest too

homepage/homepage/components/home/Hero.tsx
```diff
@@ -1,5 +1,69 @@
...
+    fileName,
+    title,
+}: {
+    children: React.ReactNode;
+    className?: string;
+    fileName?: string;
+    title?: string;
+}) {
+    return (
+        <div
```
bensleveritt, line 51: We can swap this to `<article>` to be more semantic

bensleveritt (approved):

aeplay:
The highlighted lines feel a bit random to me, but I don't know which ones would be better so merging for now. Thanks for implementing it in code!

PR #587, opened 2024-10-17
Add links to live demo in readme

trishalim:


bensleveritt:
I think we should link from the `Live Preview`, we don't need to print the URL onscreen

trishalim:
@bensleveritt I think that would be great to do from [here](https://jazz.tools/docs/resources/examples). A link to the live demo and another link to the code.

PR #593, opened 2024-10-17
Add gcmp project

trishalim:
This a copy of gcmp.io excluding the newsletter form.

The font size and card styling are different because I'm just using what's in the design system.


aeplay:
Merging blindly so I can set up a Vercel preview

PR #596, opened 2024-10-17
chore: refactor CoValueState to manage the availability logic

gdorsi:
To gain a deeper understanding of the loading flow I've started a refactoring on `CoValueState` trying to centralize all the logic around the availability state management of the CoValues.

### DIFF
--- a/packages/cojson/src/coValueState.ts
+++ b/packages/cojson/src/coValueState.ts
@@ -1,0 +1,107 @@
+import { CoValueCore } from "./coValueCore.js";
+import { PeerID } from "./sync.js";
+
+function createResolvablePromise<T>() {
+    let resolve!: (value: T) => void;
+
+    const promise = new Promise<T>((res) => {
+        resolve = res;
+    });
+
+    return { promise, resolve };
+}
+
+class CoValueUnknownState {
+    type = "unknown" as const;
+    private peers: Map<PeerID, ReturnType<typeof createResolvablePromise<"available" | "unavailable">>>;
+    private resolve: (value: "available" | "unavailable") => void;
+
+    ready: Promise<"available" | "unavailable">;
+
+    constructor(peersIds: Iterable<PeerID>) {
+        this.peers = new Map();
+
+        for (const peerId of peersIds) {
+            this.peers.set(peerId, createResolvablePromise<"available" | "unavailable">());
+        }
+
+        const { resolve, promise } = createResolvablePromise<"available" | "unavailable">();
+
+        this.ready = promise;
+        this.resolve = resolve;
+    }
+
+    update(peerId: PeerID, value: "available" | "unavailable") {
+        const entry = this.peers.get(peerId);
+
+        if (entry) {
+            entry.resolve(value);
+        }
+
+        if (value === "available") {
+            this.resolve("available");
+            return;
+        }
+
+        this.peers.delete(peerId);
+
+        // If none of the peers have the coValue, we resolve to unavailable
+        if (this.peers.size === 0) {
+            this.resolve("unavailable");
+        }
+    }
+
+    // Wait for a specific peer to have a known state
+    waitForPeer(peerId: PeerID) {
+        const entry = this.peers.get(peerId);
+
+        if (!entry) {
+            return Promise.resolve();
+        }
+
+        return entry.promise;
+    }
+}
+
+class CoValueAvailableState {
+    type = "available" as const;
+
+    constructor(public coValue: CoValueCore) { }
+}
+
+type CoValueStateAction = {
+    type: "not-found";
+    peerId: PeerID;
+} | {
+    type: "found";
+    peerId: PeerID;
+    coValue: CoValueCore;
+};
+
+export class CoValueState {
+    constructor(public state: CoValueUnknownState | CoValueAvailableState) { }
+
+    static Unknown(peersToWaitFor: Set<PeerID>) {
+        return new CoValueState(new CoValueUnknownState(peersToWaitFor));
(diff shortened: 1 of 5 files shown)


packages/cojson/src/localNode.ts
```diff
@@ -302,10 +300,21 @@ export class LocalNode {
...
+
+        const updatedEntry = this.coValues[id];
+
+        if (updatedEntry?.state.type === "available") {
+            return Promise.resolve(updatedEntry.state.coValue);
+        }
+
+        delete this.coValues[id];
+
+        return "unavailable";
```
gdorsi, line 315: Not sure on what we should do when a CoValue is not available.

It could be the case of a CoValue that have not synced yet by another user but we've got already an invitation.

I think we should store all the unavailable coValues somewhere and periodically retry to load them (e.g. every 30s)

packages/cojson/src/sync.ts
```diff
@@ -131,53 +132,19 @@ export class SyncManager {
...
 
-            const coValueEntry = this.local.coValues[id];
-            if (coValueEntry?.state !== "loading") {
-                continue;
-            }
-            const firstStateEntry = coValueEntry.firstPeerState[peer.id];
-            if (firstStateEntry?.type !== "waiting") {
-                throw new Error("Expected firstPeerState to be waiting " + id);
+            if (coValueEntry?.state.type === "unknown") {
+                await coValueEntry.state.waitForPeer(peer.id);
```
gdorsi, line 146: Most of the logic that was here has been moved inside CoValueState

packages/cojson/src/sync.ts
```diff
@@ -514,39 +478,25 @@ export class SyncManager {
...
-                )
-            ) {
-                entry.resolve("unavailable");
+
+            if (!availableOnPeer) {
+                entry.dispatch({
+                    type: "not-found",
+                    peerId: peer.id,
+                });
             }
```
gdorsi, line 489: I'm not updating the peer state to "available" because:
- notifying the listeners by resolving the peer state promise breaks the sync
- the coValue is thruly available in the peer if we have got it's headers

packages/cojson/src/sync.ts
```diff
@@ -566,35 +514,27 @@ export class SyncManager {
...
-                onProgress: entry.onProgress,
-            };
+            coValue = new CoValueCore(msg.header, this.local);
 
-            this.local.coValues[msg.id] = entry;
+            entry.dispatch({
+                type: "found",
+                coValue,
+                peerId: peer.id,
+            });
```
gdorsi, line 533: Most of the logic here have been moved inside the dispatch

packages/cojson/src/sync.ts
```diff
@@ -648,15 +588,13 @@ export class SyncManager {
...
-            );
-
-            entry.onProgress?.(ourTotalnTxs / theirTotalnTxs);
+            // const theirTotalnTxs = Object.values(
+            //     peer.optimisticKnownStates[msg.id]?.sessions || {},
+            // ).reduce((sum, nTxs) => sum + nTxs, 0);
+            // const ourTotalnTxs = [...coValue.sessionLogs.values()].reduce(
+            //     (sum, session) => sum + session.transactions.length,
+            //     0,
+            // );
```
gdorsi, line 597: Removed the `onProgress` hook because it was unused. (happy to restore it if I'm wrong)

Keeping the code commented because we could reuse it to improve the load flow and resolve only when all the known sessions have been loaded.

packages/cojson/src/sync.ts
```diff
@@ -680,10 +618,6 @@ export class SyncManager {
             );
         }
 
-        if (resolveAfterDone) {
-            resolveAfterDone(coValue);
-        }
```
gdorsi, line 685: Moved the resolve up.

Promises are resolved on the next tick of the event loop anyway so, since there are no await in the middle, it shouldn't matter where we resolve the covalue state.

packages/cojson/src/coValueState.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    return { promise, resolve };
+}
+
+class CoValueUnknownState {
+    type = "unknown" as const;
+    private peers: Map<PeerID, ReturnType<typeof createResolvablePromise<"available" | "unavailable">>>;
+    private resolve: (value: "available" | "unavailable") => void;
+
+    ready: Promise<"available" | "unavailable">;
```
gdorsi, line 19: Renamed the states:
 'loading' -> 'unknown'
 'loaded' -> 'available'
 
Reasoning about the names helped me in the understanding process.
Happy to revert the change if the previous ones have a better fit in this context.

aeplay: This is better, what was perviously "loaded" isn't actually loaded, it's just "we've received the first content for it" - so "available" is much better

In the future we should have some kind of "loaded" state meaning "up-to-date as per all peer states"

packages/cojson/src/localNode.ts
aeplay: I'd say exponential retry starting with 1s up to a max period of 30s - please create another ticket for that

packages/cojson/src/sync.ts
aeplay: Wonderful, this helps untangling things

aeplay (approved):
Great improvement!

packages/cojson/src/localNode.ts
gdorsi: https://github.com/gardencmp/jazz/issues/598

PR #597, opened 2024-10-17
feat: single threaded-server benchmarks

### LINKED ISSUE #301: Research task: Create a microbenchmark setup to test the efficiency of WebSockets vs HTTP2/3 + SSE
In order to get an idea how best to proceed with #233, it would be good to have ballpark numbers of the performance characteristics of WebSockets vs HTTP requests + Server-Sent Events for our needs.

## Setup

We can get this data - completely decoupled from the internals of Jazz - by creating some synthetic microbenchmarks.

- We need a client & server setup that can handle WebSockets, HTTP 2 & 3 and Server-Sent Events.
- The client should be in TypeScript, using only Browser-native APIs (WebSocket, fetch, EventSource)
- For servers, I would like to evaluate
  - A) a plain Node.JS server (or higher-level wrapper of that)
  - B) [uWebSockets.js](https://github.com/uNetworking/uWebSockets.js), for both web sockets and HTTP3 requests
  - C) plain Node.JS server behind a Caddy proxy that handles HTTP3 & SSL
  - all of them should be single threaded, on a single node

### Original data use that needs to be simulated

Currently, Jazz uses WebSockets to sync [CoValue](https://jazz.tools/docs#intro-to-covalues) state between the client and syncing/persistence server.

The communication typically consists of three scenarios:

1. Loading a CoValue
- The client doesn't have a CoValue's edit history or only has it partially and requests the rest from the sync server.
- The sync server responds with the missing parts of the CoValue edit history

2. Creating a CoValue
- The client creates a new CoValue locally and wants to upload it completely to the sync server

3. Subscribing to and mutating a CoValue
- A client is interested in realtime changes to a CoValue and subscribes to it
- The server sends individual changes to the CoValue to the client as they happen
- The client sends individual changes to the CoValue to the server when the user modifies the CoValue locally

### Websockets vs Requests & SSE

Currently, 1., 2. and 3. happen over WebSockets, with 1 package per request/response/incoming update

For using Requests and SSE instead, we would use Requests & Responses for 1. and 2., while for 3. we listen to incoming updates with Server-Sent Events and publish outgoing updates as a Request with no expected Response.

### Simulation spec

There are roughly two classes of CoValues: structured CoValues (thousands of <50 byte edits) and binary-data CoValues (few edits that are each 100kB).

Since we are only interested in the data transmission performance, we can model the scenarios using packets containing random data:

1. Loading a CoValue
- structured: A 100 byte request from the client and a 5kB response from the server
- binary: A 100 byte request from the client and a 50MB response from the server, streamed in 100kB chunks

2. Creating a CoValue
- structured: A 10kB request from the client and a 10 byte response from the server
- binary: A 50MB request from the client, streamed in 100kB chunks and a 10 byte response from the server

3. Subscribing to and mutating a CoValue
- structured: 50 byte incoming SSE messages/WebSocket packets, mutations are 50 byte outgoing messages as a request/WebSocket packet
  - assume one client creating a mutation that is published to 10 other clients
- binary: 100kB incoming SSE messages/WebSocket packets, mutations are 100kB outgoing messages as a request/WebSocket packet

No extra HTTP headers should be set (other than what browser set by default, and these should be minimised if possible)

### Target metrics

The main variables we are interested in are

1. Loading a CoValue
- a) How many CoValues can be "simulation loaded" at once on a client? Do we get head-of-line blocking effects? (you can do structured/binary separately)
- b) How many clients can "simulation load" a CoValue at the same time per server?

2. Creating a CoValue
- c) How many CoValues can be "simulation created" at once on a client? Do we get head-of-line blocking effects? (you can do structured/binary separately)
- d) How many clients can "simulation create" a CoValue at the same time per server?

3. Subscribing to and mutating a CoValue
- e) How many updates per second can we push from a client?
- f) How many updates per second can a server handle (receive and broadcast to interested clients)?
- g) What data throughput can we achieve for binary CoValue updates?
- h) What's the latency like

### Variables

It would be good to get results for the metrics above assuming

#### Different network conditions

- I Ideal network conditions
- II 4G speeds
- III 3G speeds
- IV connections with high packet loss, including so bad that we need to reconnect WebSockets

#### Different protocols

- WS: WebSockets only (with reconnect on timeout)
- H2: HTTP2 + SSE
- H3: HTTP3 + SSE

You don't need to actually deploy a server anywhere if you can simulate these conditions locally, just make sure to note down your hardware specs and use exactly one thread/core for the server

## Dimensions summary

So in total we have the following dimensions:
- Server tech: A, B and C
- Target metrics: a), b), c), d), e), f), g), h)
- Network condition: I, II, III, IV
- Protocols: WS, H2, H3

## Deliverable

- Create a completely new project in a new folder in this monorepo called 'experiments'
- Create an independent pnpm workspace in there
- Use whatever test-harness / benchmarking libraries / node setups / browser automation you think are best
- Make sure the setup is replicable / can be run by anyone
- Post results from your machine in a README in the same folder

I realise this spec is a lot, so feel free to ask lots of clarifying questions before & after accepting the task!

ayewo:
This PR is intended to close #301. 

## 1.1. Overview
Out of the total of 9 different web server + protocol combos, only 6 web server + protocol combos were feasible.

Below is a snapsot of web server + protocol combos that were built:

| # | Server      | Layer 7 | Layer 6 | Layer 4 | Built | Status |
|----|------------|--------|------------|--------|--------|--------|
| A1 | Node.js    | WebSocket (WS) |  TLSv1.3 (Optional) | TCP | ✔️ | **Complete** |
| A2 |     | HTTP/1.1 + Server-Sent Events (SSE) |  TLSv1.3 (Optional) | TCP | ✔️ | **Complete** |
| A3 |     | HTTP/2 + SSE |  TLSv1.3 (Optional) | TCP | ✔️ | **Complete** |
| A4 |     | HTTP/3 + SSE |  TLSv1.3 (Mandatory) | UDP (QUIC) | ❌ | **N/A** |
| B1 | uWebSockets.js | WebSocket |  TLSv1.3 (Optional) | TCP | ✔️ | **Complete** |
| B2 |     | HTTP/1 + SSE |  TLSv1.3 (Optional) | TCP | ✔️ | **Complete** |
| B3 |     | HTTP/2 + SSE |  TLSv1.3 (Optional) | TCP | ❌ | **N/A** |
| B4 |     | HTTP/3 + SSE |  TLSv1.3 (Mandatory) | UDP (QUIC) | ❌ | **N/A** |
| C1 | Node.js + Caddy    | HTTP/3 + SSE |  TLSv1.3 (Mandatory) | UDP (QUIC) | ✔️ | **Complete** | 

## 1.2. Server + Protocol Combos
As can be seen in the table above, 6 different web servers + protocol combos were built for the benchmarks:
1. Node.js (WS) code-named `node-ws`;
2. Node.js (HTTP/1.1 + SSE) code-named `node-http1`;
3. Node.js (HTTP/2 + SSE) code-named `node-http2`;
4. uWebSockets.js (WS) code-named `uws-ws`;
5. uWebSockets.js (HTTP/1.1 + SSE) code-named `uws-http1`;
6. Caddy + Node.js (HTTP/3 + SSE) code-named `caddy-http3`.

## 1.3. Server Memory Limit
On the development machine I used, which has 16GB of RAM installed, the memory limit for the `node` process defaults to 4GB. 

For the benchmarks proper, each web server is started with a higher memory limit of 8GB. This is set in `package.json` via `--max-old-space-size=8192` for all 6 web servers, but if running on a machine with higher installed memory, you can raise it as needed.

## 1.4. Server Test Data 
I used [faker.js](https://v7.fakerjs.dev/) to generate test data served by each web server. 

To keep things fast, the generated test data are not stored in a database. Instead, everything is served from memory, with the exception of binary files used to simulate binary CoValue downloads and uploads. These are stored on disk at: `public/downloads` and `public/uploads`.

Each web server listens on https://localhost:3000/ and serves a default `index` page from the file `public/client/{ws|http}/index.html`, depending on whether the server's transport is based on websockets or plain HTTP + SSE. 

The default `index` page hosts all JavaScript code that will be driven by [`playwright`](https://playwright.dev/) and [`artillery`](https://www.artillery.io/) load tests.

The default page includes a minified version of `faker.js` which is also used to create test data in the browser. The only downside is this [caveat](https://v7.fakerjs.dev/guide/usage.html#browser) from the `faker.js` documentation:
> Using the browser is great for experimenting 👍. However, due to all of the strings Faker uses to generate fake data, Faker is a large package. It's > 5 MiB minified. Please avoid deploying the full Faker in your web app.

## 1.5. Server Network Conditions
With respect to simulating various network conditions (I, II, III, IV), I explored:

OS-level throttling using third-party tools like:
- [Network Link Conditioner for MacOS](https://apple.stackexchange.com/questions/24066/how-to-simulate-slow-internet-connections-on-the-mac)
- [NetLimiter for Windows](https://www.netlimiter.com/)

Or doing the throttling at the browser-level using:
- [Chrome DevTools](https://developer.chrome.com/docs/devtools/network#throttle) 

I eventually came across Shopify's [`toxiproxy`](https://github.com/Shopify/toxiproxy) which has first-class support for programmatic use and would have been perfect for our needs. Unfortunately, `toxiproxy` was designed to [proxy TCP traffic](https://github.com/Shopify/toxiproxy?tab=readme-ov-file#2-populating-toxiproxy), meaning it would work for TCP-based protocols like HTTP/1.1 and HTTP/2, but not for HTTP/3, which is based on UDP (QUIC), so I had to abandon the idea.

I was able to achieve the desired OS-level programmatic throttling using a combination of macOS' `pfctl` (packet filter control) and `dnctl` (dummynet control) CLI tools. Note that some programmatic invocations of `pfctl` and `dnctl` by scripts used in the benchmarks will require sudo privileges.

The only downside is that the benchmarks would need to be adapted to use `netem` (network emulator) and `tc` (traffic control), if there's a need to execute it on a server-class machine running Linux.

## 1.6. Summary
This is a high-level summary of the different components that make up the microbenchmark.

Please refer to the included [README](https://github.com/garden-co/jazz/blob/main/experiments/servers/README.md) for additional details.

/claim #301


aeplay:
Closing for now since we can access the results & code in this PR but don't really need it on main day-to-day. Thanks again for the hard work @ayewo !

PR #601, opened 2024-10-18
Group Inheritance

aeplay:
`cojson` (core)
- [X] Add parent/child relationship to group keys
- [X] Roles get inherited and enforced correctly for write-access
  - [X] Implementation
  - [X] Test (one level deep)
  - [X] Test (more than one level deep)
- [X] Ensure child group read keys can be revealed to parent group read keys and looked up correctly
  - [X] Implementation
  - [X] Test (one level deep)
  - [x] Test (more than one level deep)
- [x] Ensure key rotation in child groups is done when parent group key is rotated
  - [x] Implementation
  - [x] Test (one level deep)
  - [x] Test (more than one level deep)

`jazz-tools`
- [x] High level API
  - [x] Implementation
  - [x] Simple tests to make sure low-level extension / rotation is called correctly


aeplay:
Sorry for the noise due to the changeset pre-release. As you can see the tests were passing before the version update and the merge conflicts should also go away once we exit pre-release mode

packages/cojson-storage-indexeddb/src/index.ts
```diff
@@ -692,9 +692,19 @@ function getDependedOnCoValues(
...
+                      ? [
+                          key.replace(
+                            "parent_",
+                            "",
+                          ) as CojsonInternalTypes.RawCoID,
+                        ]
+                      : []
+                  : [],
               );
           }),
```
gdorsi: **suggestion:** how do you feel about moving this entire thernary branch in a separate module?

Just to give a name on what this is doing and cover it more easily with tests.

packages/cojson/src/coValueCore.ts
```diff
@@ -950,9 +990,15 @@ export class CoValueCore {
...
-          .filter((k): k is RawAccountID => k.startsWith("co_"))
+      ? [
+          ...expectGroup(this.getCurrentContent())
+            .keys()
+            .filter((k): k is RawAccountID => k.startsWith("co_")),
+          ...expectGroup(this.getCurrentContent())
+            .keys()
+            .filter((k) => k.startsWith("parent_"))
+            .map((k) => k.replace("parent_", "") as RawCoID),
+        ]
```
gdorsi: **suggestion:** this can be probably abstracted away to be reused also in the storage peers.

Thinking about this kind of function:
```ts
function getGroupDependentKeys(keys: string[]) {
   const groupDependentKeys: RawCoID[] = [];
   
   for (const key of keys) {
      if (key.startsWith("parent_")) {
         groupDependentKeys.push(key.replace("parent_", "")  as RawCoID);
      } else if (key.startsWith("co_")) {
         groupDependentKeys.push(key as RawCoID);
      }
   }
   
   return groupDependentKeys;
}
```

packages/cojson/src/coValues/group.ts
```diff
@@ -61,12 +64,68 @@ export class RawGroup<
...
+      const roleInParent = parentGroup.roleOfInternal(accountID);
+
+      if (
+        roleInParent &&
+        roleInParent.role !== "revoked" &&
+        isMorePermissiveAndShouldInherit(roleInParent.role, roleInfo?.role)
+      ) {
+        roleInfo = { role: roleInParent.role, via: parentGroup.id };
+      }
+    }
```
gdorsi, line 107: **suggestion:** Could you add some comments to document what this function is doing?

I'm not able to understand completely what's happening here.

packages/cojson/src/coValues/group.ts
```diff
@@ -61,12 +64,68 @@ export class RawGroup<
...
+
+    return roleInfo;
+  }
+
+  getParentGroups(): RawGroup[] {
+    return (
+      this.keys().filter((key) =>
+        key.startsWith("parent_"),
+      ) as `parent_${CoID<RawGroup>}`[]
```
gdorsi: **suggestion:** we should abstract these checks with type guards

It would reduce the need of doing casting.

packages/cojson/src/coValues/group.ts
```diff
@@ -203,7 +262,75 @@ export class RawGroup<
...
+            secret: parentReadKeySecret,
+          },
+          toEncrypt: newReadKey,
+        }).encrypted,
+        "trusting",
+      );
+    }
+
+    for (const child of this.getChildGroups()) {
+      console.log("Rotating child", child.id);
```
gdorsi: **question:** are these console.log meant to stay?

packages/cojson/src/coValues/group.ts
```diff
@@ -203,7 +262,75 @@ export class RawGroup<
       "trusting",
     );
 
+    console.log("Setting", `readKey`, "to", newReadKey.id, "in", this.id);
+
     this.set("readKey", newReadKey.id, "trusting");
+
+    for (const parent of this.getParentGroups()) {
```
gdorsi: **suggestion:** could you document your intent here via code comments?

packages/cojson/src/coValues/group.ts
```diff
@@ -347,6 +474,34 @@ export class RawGroup<
   }
 }
 
+function isMorePermissiveAndShouldInherit(
+  roleInParent: Role,
+  roleInChild: Exclude<Role, "revoked"> | undefined,
+) {
```
gdorsi, line 538: **praise:** thanks for the very descriptive function name!

It makes the function content way easier to understand.

gdorsi (approved):
Looks good!

Left some comments for code documentation and organization.

packages/jazz-tools/src/tests/groupsAndAccounts.test.ts
```diff
@@ -85,3 +86,95 @@ describe("Custom accounts and groups", async () => {
...
+      group._raw.core.getDependedOnCoValuesUncached(),
+      parentGroup.id,
+    );
+
+    console.log(
+      (group._raw.core.getCurrentContent() as RawGroup)
+        .keys()
+        .filter((k) => k.startsWith("parent_"))
+        .map((k) => k.replace("parent_", "")),
+    );
```
gdorsi: **question:** would it be ok to remove these console logs?

They make the test results harder to read.

packages/cojson/src/coValues/group.ts
aeplay: To give accounts read access to a group, we reveal the group's `readKey` to the account's (public, asymmetric) key.

To give parent groups read access to a child group, we now similarly reveal the child group's `readKey` to the parent group's (secret, symmetric) `readKey`

And finally, when we rotate our `readKey` (because someone got kicked out), we also need to (recursively) rotate the `readKey`s of all child groups (so they are kicked out there as well)

packages/cojson/src/coValues/group.ts
```diff
@@ -61,12 +64,68 @@ export class RawGroup<
...
+  getChildGroups(): RawGroup[] {
+    return (
+      this.keys().filter((key) =>
+        key.startsWith("child_"),
+      ) as `child_${CoID<RawGroup>}`[]
+    ).map((childKey) => {
+      const child = this.core.node.expectCoValueLoaded(
+        childKey.slice("child_".length) as CoID<RawGroup>,
+        "Expected child group to be loaded",
+      );
```
gdorsi, line 159: @aeplay 

How can we be sure that the child groups have been loaded?
They are not sent as dependencies

packages/cojson/src/coValues/group.ts
gdorsi: > To give parent groups read access to a child group, we now similarly reveal the child group's readKey to the parent group's (secret, symmetric) readKey

Wait, why a parent group should have read access to a child group?

aeplay: Because the roles should be inherited - anyone who is a reader in the parent should also be a reader in the child. By exposing the child key to the parent key we can efficiently share it with everyone who is a reader in the parent

aeplay:
Consider the loading of child groups for key rotation approved

Issue #307, opened 2024-08-14
Make sure CoValues are fully synced in the background even if they stop being used actively


aeplay:
Right now, CoValues are synced while they are actively used in a LocalNode.

But there are many cases, especially with BinaryCoStreams, where they get created and then not immediately used (particularly the highest resolution of ImageDefinitions, for example)

While local storage is often fast enough to store them in their entirety, they might be forgotten in the local node before they have finished syncing - or the user goes offline in the meantime.

The same applies not just for newly created CoValues, but also for CoValues with lots of new edits, which are then not looked at anymore.

The problem is that fixing this requires maintaining and *persisting sync state* in the storage peers as well. So right now this means 3 implementations of persisting this extra metadata:

* IndexedDB storage peer
* SQLite storage peer
* Filesystem-based storage peer (abstract interface of which OPFS is an implementation)

This will likely further complicate state handling within CoValueCores and the communication to storage peers (technically the whole sync protocol needs to be adopted to communicate 'sync status to third peers') - but supporting this is vitally important to ensure that local data loss (particularly with IndexedDB) is not fatal - so we'll take the extra complexity hit

It might be worth introducing a new peer role "storage" that is like "server" but implies interest in the additional "is it synced" metadata

gdorsi:
So if I understand correctly a repro of this bug could be:
- The user stores a file using BinaryCoStream
- The user logs-out before the sync with the server is complete
- The user logs-in again

**Expected:** The sync should resume even if the file isn't currently "loaded" in the current view
**Actual result:** The sync doesn't resume until the BinaryCoStream is consumed somewhere

To fix this issue we want to persist the sync state in the fastest peers we have (storage).

The sync state we are talking about here is the [PeerState](https://github.com/gardencmp/jazz/blob/main/packages/cojson/src/PeerState.ts#L12) of the non-storage peers.

Am I correct?

gdorsi:
So if I understand correctly a repro of this bug could be:

* The user stores a file using BinaryCoStream
* The user logs-out before the sync with the server is complete
* The user logs-in again

**Expected:** The sync should resume even if the file isn't currently "loaded" in the current view

**Actual result:** The sync doesn't resume until the BinaryCoStream is consumed somewhere

To fix this issue we want to persist the sync state in the fastest peers we have (storage).

[...]

PR #606, opened 2024-10-19
feat: make the PeerKnownState subscribable

gdorsi:
Part of #307 

The idea is to make possible to subscribe to the PeerKnownState updates in order to sent them to the storage peers.

So when building the Jazz context we could create a "WebSocket PeerKnownState" -> "Storage Peer" stream to continuously persist the sync state of the co values.

### DIFF
--- a/packages/cojson/src/PeerState.ts
+++ b/packages/cojson/src/PeerState.ts
@@ -1,16 +1,94 @@
-import { RawCoID } from "./ids.js";
+import { RawCoID, SessionID } from "./ids.js";
 import { CO_VALUE_PRIORITY } from "./priority.js";
 import {
     PriorityBasedMessageQueue,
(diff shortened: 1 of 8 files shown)


packages/cojson/src/PeerState.ts
```diff
@@ -1,16 +1,140 @@
...
+            const knownState = new PeerCoValueKnownState(emptyKnownState(id));
+            this.coValues.set(id, knownState);
+
+            this.trigger({
+                type: "SET_AS_EMPTY",
+                id,
+            });
+        
+            return knownState;
+        }
```
gdorsi: Will refactor this flow because it isn't great.

gdorsi:
Closing to do a full refactoring to collect the knownState (not only the optmisiticKnownState)

PR #612, opened 2024-10-21
Move code samples from hero section to its own section

trishalim:


trishalim:
@aeplay wdyt? 

Now the CoValues explanation is duplicated but we can remove or move it to docs.

I also want to reframe the chat example as a "get started quickly by cloning this example app, or see other examples" and move it to the bottom right before the CTA.

aeplay:
This is great!!

I almost want it before the "hard things are easy" section - but I think it's better after. I've been meaning to ask you to maybe make the typography and diagrams in "hard things are easy" smaller, it seems a bit huge compared to the more elegant other elements you've introduced since

And I love the on-topic testimonial - only not sure about the stars. They look great and are a nice visual anchor, but it makes it look too much like a review to me... wondering what else would work? A line version of the starry-eyed emoji? Too cheesy?

trishalim:
@aeplay Like this 🤩🤩🤩🤩🤩? Yes it's too much. What about this?
image

trishalim:
@aeplay I added an extra step to connect with jazz cloud. Is that too much? 🤔

aeplay:
@trishalim no, that's great, it reduces the amount of unexplained magic.

Also that new styling is 🥵

Love the quote symbol for the quote as well, and the subtle highlighted bit

Two nit-picks:
- I don't think we have to explain the email-as-api-key part here
- The numbers don't look aligned with the first line of the sentence to me

trishalim:
@aeplay Thanks! Addressed your feedback

PR #616, opened 2024-10-22
poc: resume the partial syncs on app start

gdorsi:


packages/cojson/src/localNode.ts
```diff
@@ -240,6 +311,10 @@ export class LocalNode {
                 );
             }
 
+            for (const peer of serverPeers) {
+                void node.syncManager.requestSyncStateHydration(peer.id);
+            }
```
gdorsi: We should define two new possible states for the `localNode`: "bootstrapping" and "ready"

The sync state hydration should start when we are ready so we won't need a timeout inside of it.

PR #620, opened 2024-10-22
test: add an E2E test to test the sync resumability

gdorsi:
Added a repro test for #307

### DIFF
--- a/e2e/CoValues/tests/ResumeSyncState.spec.ts
+++ b/e2e/CoValues/tests/ResumeSyncState.spec.ts
@@ -1,0 +1,47 @@
+import { test, expect } from "@playwright/test";
+import { setTimeout } from "node:timers/promises";
+
+test.describe("ResumeSyncState", () => {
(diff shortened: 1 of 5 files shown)


e2e/CoValues/tests/ResumeSyncState.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    // Create a new incognito instance and try to load the coValue
+    const newUserPage = await (await browser.newContext()).newPage();
+    await newUserPage.goto(`/resume-sync?signUp=Luigi&id=${id}`);
+
+    await expect(newUserPage.getByTestId("id")).toBeInViewport();
+
+    // The initial user should have synced the value even if the coValue was not loaded
+    // when the user is back online
+    await expect(newUserPage.getByRole("textbox", { name: "Value" })).toHaveValue("Mammamia!");
```
gdorsi: As expected this test fails in the last step because the last synced value is "Let's go!"

PR #622, opened 2024-10-23
Rewrite hero copy

trishalim:
Rewrite the hero copy to communicate: "local first is the best new way to build apps in general and we're the best framework for it."

### DIFF
--- a/homepage/homepage/components/home/Hero.tsx
+++ b/homepage/homepage/components/home/Hero.tsx
@@ -1,4 +1,5 @@
 import { LucideIcon } from "lucide-react";
+import Link from "next/link";
 
 export function Hero({
     features,
@@ -6,7 +7,7 @@ export function Hero({
     features: { title: string; icon: LucideIcon }[];
 }) {
     return (
-        <div className="container grid gap-x-8 gap-y-10 py-12 lg:py-24 lg:gap-0 lg:grid-cols-3">
+        <div className="container grid gap-x-8 gap-y-10 py-12 md:py-16 lg:py-24 lg:gap-0 lg:grid-cols-3">
             <div className="flex flex-col justify-center gap-4 lg:col-span-3 lg:gap-8">
                 <p className="uppercase text-blue tracking-wide text-sm font-medium dark:text-stone-400">
                     Local-first development toolkit
@@ -20,7 +21,7 @@ export function Hero({
                         Jazz is a framework for building local-first apps
                         &mdash; an architecture that lets apps like Figma and Linear play in a league of their own.
                     </p>
-                    <p>Open source. Easily self-host or use Jazz Cloud.</p>
+                    <p>Open source. Self-host or use <Link href="/cloud">Jazz Cloud</Link> for zero-config magic.</p>
                 </div>
 
                 <div className="grid grid-cols-2 gap-2 max-w-3xl sm:grid-cols-4 sm:gap-4">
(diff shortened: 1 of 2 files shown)


trishalim:
@aeplay 
Using this as a guide
image

Bad alternatives
- Framework for building local-first apps

Better
- Make better apps, with less code
- Ship higher quality products, with less code
- Better UX and better DX 

Action
- Make better apps, with less code

Objections

- It’s too new. Does it really have all the features I need?
- Can’t self-host, need to rely on the platform (but there's no dashboard)
- Backend developers express skepticism towards Jazz due to its ability to replace traditional stacks, which they perceive as "too much magic." This skepticism arises because Jazz replaces, like, 10 plus tech choices and things you need to deploy, which can be seen as taking away the backend developers' role of integrating various technologies. This creates a challenge, as developers inherently mistrust solutions that seem to do everything for them, potentially leading to a feeling of **losing control over the technical architecture**.

I'll explain every line below
image

> Local-first product development toolkit

This part is called a kicker which is a short line of text that sits above a headline and adds context. This can reduce the need to shove the product category into the headline and reduce cognitive load.

> Ship top-tier apps, with less code.

Bold claim and benefit

> Jazz is a framework for building local-first apps

This answers "what is jazz?"

> an architecture embraced by industry leaders today like Linear and Figma

Supports "top-tier" from the headline
Creates interest on "local-first"
We have to be careful not to say that Linear/Figma are using Jazz though 😅

> Fully open source

Moved this down so that the first sentence focuses on what jazz is and local-first

> Host with Jazz Cloud or self-host.

This is the best argument I have for backend devs' skepticism. Anything else?

After the hero, we can add a section that talks about local-first, what it is, why it's the new standard. Something like "The bar for software is higher than ever before. Your MVP (a different word) just doesn't cut it anymore. This is where local-first comes in." and then we can list the benefits of local-first for the users and also the DX. This will help us shorten the features section too.

aeplay:
Love the inspiration and detailed step-by-step execution @trishalim . The only problem is you’ve made me have ideas of my own. Thinking about how to best synthesise them and argue them with as much rigour as you have

aeplay:
Whatever, I think we can just iterate on yours

> kicker

great concept and great execution, works as promised 

> Ship top-tier apps, with less code.

a bit generic, but very punchy. Hard to beat in terms of meaning per word. The nice thing is that everyone likes “less code”.

> an architecture embraced by industry leaders today like Linear and Figma

This is very strong backup for our claim that I hadn’t considered before. I like it! Only the way you put it sounds a bit business-y. How about “An architecture that lets apps like Figma and Linear play in a league of their own”?

> Moved this down so that the first sentence focuses on what jazz is and local-first

That’s great! And open source and self-hosted is an important objection-response.

maybe even mention self hosting first but then qualify Jazz Cloud?

“Host it yourself or use Jazz Cloud for zero-config instant magic.”?

trishalim:
> a bit generic

I thought this too

> mention self hosting first 

Agreed! Plus you have the docs for this coming up

trishalim:
"league of their own" is a great description
Last line is a bit long 
image

How about this
image

aeplay:
@trishalim  Self-host shouldn't sound tooooo easy :D Just mentioning that you can is enough. And agree that my suggestion is too long but maybe we should mention that Jazz Cloud is free to start or something like that?

aeplay:
Or how about "Self-host or use Jazz Cloud for zero-config magic"?

trishalim:
What's better? Free or easy? 🤔
image
image

trishalim:
Let's go with zero-config

They'll figure out it's free pretty easily

aeplay:
Sounds great now, thanks! Let's 🚢 it

PR #634, opened 2024-10-28
Redesign diagrams section

trishalim:


trishalim:
The bold text is overdone. It feels like everything is emphasized (nothing is emphasized)
image

I added a description under the heading
I like that it starts with a compliment
image

..before we start to get kinda aggressive
image

In the end, I feel like I wrote too much without saying much :D
But my intention here was to find a unique voice / be memorable
image

It definitely looks better now, but unsure about the text. @aeplay thoughts?

Will try to trim it more.

aeplay:
The design is great!

I see where you're going with the tone, but I don't dig it :(
But again there's a lot of good new ideas in there.

I'll probably suggest some stuff based on it later today

trishalim:
@aeplay I will also revisit this later today

Edit: tomorrow

trishalim:
@aeplay Let's make this PR just a redesign for now, as I'm out of ideas at the moment. Also, I force pushed because it was easier to resolve conflicts this way.

aeplay:
Makes sense, thank you!

PR #638, opened 2024-10-28
feat(jazz-run): add a --flaky option to the sync command to work on the network reliability

gdorsi:


aeplay:
Something something our software is so good we need to add flags to make it worse

PR #641, opened 2024-10-29
Typography: styling, refactor defaults

trishalim:
- Move prose styling to tailwind config as much as possible. There were loads of class names there showing on the source code (annoying)
- Replace occurences of `<span className="font-medium text-stone-900 dark:text-white">` with `<Prose/>` and `<strong/>`
- Move styling for code examples into `<CodeGroup />`. When you don't use use this, it's gonna look rekt, so please use it.
- Decrease heading font sizes. This applies to prose only. We can use bigger font sizes for the landing pages. Those look too big in the docs.
- Style links

image


aeplay:
Excellent, thanks!

PR #643, opened 2024-10-29
Set default border color for dark mode

trishalim:
We keep repeating `border-stone-200 dark:border-stone-800`, so I've set the defaults through the design system tailwind config.


homepage/homepage/app/globals.css
```diff
@@ -2,78 +2,6 @@
...
-        --destructive: 0 62.8% 30.6%;
-        --destructive-foreground: 60 9.1% 97.8%;
-
-        --border: 12 6.5% 15.1%;
-        --input: 12 6.5% 15.1%;
-        --ring: 24 5.7% 82.9%;
-    }
-
-    * {
-        @apply border-border;
```
trishalim, line 69: This bit was interfering with my styles for the border color so I removed it.

Then I ran into a bunch of issues, and removing all this fixed it. We don't seem to be using it? I see it's from shadcn

Also, the naming (`border-border`) is weird

Anyway, site looks the same to me without all this

aeplay:
Thank you!

PR #646, opened 2024-10-29
Improve the jazz-run create command and make it possible to subscribe to the peer sync state

gdorsi:
In this PR I'm encapsulating all the peers optimisticKnownStates updates in a class called PeerKnownStates.

The idea is to make possible to subscribe to the `optimisticKnownState` to check the sync status of a coValue.

I've used this new "feature" to improve the create account command on jazz-run by replacing the `sleep` with a `waitForSync` function.

### DIFF
--- a/packages/cojson/src/tests/PeerState.test.ts
+++ b/packages/cojson/src/tests/PeerState.test.ts
@@ -20,17 +20,6 @@ function setup() {
 }
 
 describe("PeerState", () => {
-    test("should initialize with correct properties", () => {
-        const { peerState } = setup();
-        expect(peerState.id).toBe("test-peer");
-        expect(peerState.role).toBe("peer");
-        expect(peerState.priority).toBe(1);
-        expect(peerState.crashOnClose).toBe(false);
-        expect(peerState.closed).toBe(false);
-        expect(peerState.optimisticKnownStates).toEqual({});
-        expect(peerState.toldKnownState).toEqual(new Set());
-    });
-
     test("should push outgoing message to peer", async () => {
         const { mockPeer, peerState } = setup();
         const message: SyncMessage = { action: "load", id: "co_ztest-id", header: false, sessions: {} };
--- a/packages/jazz-run/src/accountCreate.ts
+++ b/packages/jazz-run/src/accountCreate.ts
@@ -1,0 +1,125 @@
+import { Command, Options } from "@effect/cli";
+import { Console, Effect } from "effect";
+import { createWebSocketPeer } from "cojson-transport-ws";
+import { WebSocket } from "ws";
+import {
+    Account,
+    Peer,
+    WasmCrypto,
+    createJazzContext,
+    isControlledAccount,
+} from "jazz-tools";
+import { fixedCredentialsAuth, randomSessionProvider } from "jazz-tools";
+import { CoValueCore } from "cojson";
+
+const name = Options.text("name").pipe(Options.withAlias("n"));
+const peer = Options.text("peer")
+    .pipe(Options.withAlias("p"))
+    .pipe(Options.withDefault("wss://cloud.jazz.tools"));
+
+const accountCreate = Command.make(
+    "create",
+    { name, peer },
+    ({ name, peer: peerAddr }) => {
+        return Effect.gen(function* () {
+            const crypto = yield* Effect.promise(() => WasmCrypto.create());
+
+            const peer = createWebSocketPeer({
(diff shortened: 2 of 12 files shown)


packages/cojson/src/tests/PeerState.test.ts
```diff
@@ -20,17 +20,6 @@ function setup() {
...
-    test("should initialize with correct properties", () => {
-        const { peerState } = setup();
-        expect(peerState.id).toBe("test-peer");
-        expect(peerState.role).toBe("peer");
-        expect(peerState.priority).toBe(1);
-        expect(peerState.crashOnClose).toBe(false);
-        expect(peerState.closed).toBe(false);
-        expect(peerState.optimisticKnownStates).toEqual({});
-        expect(peerState.toldKnownState).toEqual(new Set());
-    });
```
gdorsi, line 32: Removed this test because it wasn't actually useful

packages/jazz-run/src/accountCreate.ts
```diff
@@ -0,0 +1,125 @@
...
+} from "jazz-tools";
+import { fixedCredentialsAuth, randomSessionProvider } from "jazz-tools";
+import { CoValueCore } from "cojson";
+
+const name = Options.text("name").pipe(Options.withAlias("n"));
+const peer = Options.text("peer")
+    .pipe(Options.withAlias("p"))
+    .pipe(Options.withDefault("wss://cloud.jazz.tools"));
+
+const accountCreate = Command.make(
```
gdorsi, line 20: Most of the code here has been extracted from `./index.ts`

packages/jazz-run/src/accountCreate.ts
```diff
@@ -0,0 +1,125 @@
...
+                syncManager.syncCoValue(accountCoValue),
+            );
+            yield* Effect.promise(() =>
+                syncManager.syncCoValue(accountProfileCoValue),
+            );
+
+            yield* Effect.promise(() => Promise.all([
+                waitForSync(account, peer, accountCoValue),
+                waitForSync(account, peer, accountProfileCoValue),
+            ]));
```
gdorsi, line 58: This is the actual change.

Before this was done with an `Effect.sleep(2000)`

packages/jazz-run/src/accountCreate.ts
```diff
@@ -0,0 +1,125 @@
...
+const accountBase = Command.make("account");
+
+export const account = accountBase.pipe(Command.withSubcommands([accountCreate]));
+
+function waitForSync(account: Account, peer: Peer, coValue: CoValueCore) {
+    const syncManager = account._raw.core.node.syncManager;
+    const peerState = syncManager.peers[peer.id];
+
+    return new Promise((resolve) => {
+        const unsubscribe = peerState?.optimisticKnownStates.subscribe((id, peerKnownState) => {
```
gdorsi, line 96: @aeplay 

Tried to make a non-optimistic known state but it seems to me that the current exchanged messages aren't enough to collect a fully up-to-date known state.

We will need that to fully proceed with #307 so I think that would be good to start a discussion around this issue.

packages/jazz-run/src/accountCreate.ts
```diff
@@ -0,0 +1,125 @@
...
+            yield* Effect.promise(() =>
+                syncManager.syncCoValue(accountProfileCoValue),
+            );
+
+            yield* Effect.promise(() => Promise.all([
+                waitForSync(account, peer, accountCoValue),
+                waitForSync(account, peer, accountProfileCoValue),
+            ]));
+
+            // Spawn a second peer to double check that the account is fully synced
```
gdorsi, line 60: @aeplay 

Is this comment correct?

aeplay:
✨

aeplay:
changeset pls (even though it doesn't functionally change jazz-run account create - but it is more than a temporary fix now)

gdorsi:
Changeset added

packages/jazz-run/src/accountCreate.ts
aeplay: yes

aeplay (changes requested):
It's not a big deal, but for the sake of clarity in the per-package changelogs, please may you delete the changeset and then create two change sets, one for cojson (subscribe to peer state) and one for jazz-run (improve account create). Thank you!

Everything else LGTM

gdorsi:
@aeplay changeset splitted!

PR #650, opened 2024-10-30
Redesign Jazz Cloud section

trishalim:
- move Jazz Cloud right under the features section because it's one of the features
- make it more prominent
- also added "You can rely on us, or self-host" but commented this out until we have the docs for self-hosting

Some more room for improvement:
- differentiate Jazz Cloud as a separate product. This can be through color, or a logo.
- illustration on the right?


aeplay:
- I like that it is more prominent and it is the right place to mention it
- but it feels too disruptive to the flow of the overall page

What do you think about including it as a "super-wide special tile" in the features grid (spanning 3 cols, having gradient bg?)

- eventually I would love some cool mesh-covering-the-globe or mesh-in-the-sky type abstract illustration to use in-line and on /cloud

trishalim:
@aeplay I can try merging it into the features section. First, I need to break out the feature section before adding to it because it's already huge. Will get back to this.

trishalim:
Working on this in https://github.com/gardencmp/jazz/pull/671

PR #651, opened 2024-10-30
feat: introduce the storage peer role

gdorsi:
Introducing the "storage" peer role that is going to be used as part of #307 for the initial discovery of which peer is elegible to store the other peers known states.

This kind of flow is implemented in the POC #616 and can be summarized this way:
- During the app lifecycle the localNode will subscribe to the known states of all the server peers
- An extra trigger will be added inside [actuallySyncCoValue](https://github.com/gardencmp/jazz/pull/616/files#diff-c90e0211d2243ec0555557cabc377441df15b3b8ca4b76caa07882039f29da68R745) to track the synced state of each CoValue/Peer
- All these known states plus a synced flag are going to be stored inside the storage peers
- On the next app start the localNode will query the storage peers for all the CoValues that are not fully synced and try to resume the sync for them.

What has yet to be figured out is:
- We can't rely on the optimisticKnownState to determine if a CoValue is fully synced. Most of the reason is that optimisticKnownState is updated when an update is pushed to the peers but our WebSocket peer has an internal buffer that might be not empty when the user logs out.
  - A good start could be to try to track the "sent" messages and update a parallel known state (non-optimistic) that would be more reliable to track the actual sync state. Once added this "ack" flow we improve it by adding a real "ack" system in our WebSocket peers if we see the need.
- The peers don't have an unique identifier that can be used for storage. We need to find a stable way to identify them over the sessions.

### DIFF
--- a/.changeset/swift-fireants-type.md
+++ b/.changeset/swift-fireants-type.md
@@ -1,0 +1,7 @@
+---
+"cojson-storage-indexeddb": patch
+"cojson-storage-sqlite": patch
+"cojson": patch
+---
+
+Introduce "storage" peer role
--- a/packages/cojson-storage-indexeddb/src/index.ts
+++ b/packages/cojson-storage-indexeddb/src/index.ts
@@ -94,7 +94,7 @@ export class IDBStorage {
             "storage",
             {
                 peer1role: "client",
-                peer2role: "server",
+                peer2role: "storage",
                 trace,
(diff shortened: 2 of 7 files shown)


aeplay:
Looks good, thank you!

aeplay:
Added a changes to make sure packages are mutually compatible when we do a release

Issue #562, opened 2024-10-14
Improve the testing experience for the React components/hooks


gdorsi:
We should make it easy to write unit tests for React and provide fixture data for the Jazz state.<br><br>Example on how testing with React should work:

```typescript
// Creates the account and the context
const { me, context } = await createJazzTestAccount(MusicaAccount, {});
const playlist = Playlist.create({ title: '2024 Music Hits!' }, me);

render(<MyComponent playlistId={playlist.id} />, { me, context });

act(() => {
  playlist.title = '2025 Music Hits!';
});

expect(screen.getByText('2025 Music Hits!')).toBeInTheDocument();
```

<br>Custom renderer

```
const customRender = (
  ui: React.ReactElement,
  options?: Omit<RenderOptions, 'queries'> & ExtendedOptions,
) => {
  const wrapper = <Jazz.Provider context={options.context} user={options.me}>
          {children}
        </Jazz.Provider>;

  return render(ui, { ...options, wrapper });
};
```

PR #653, opened 2024-10-30
fix: ensure that the react provider doesn't run twice on dev

gdorsi:
Over time while testing my changes on our examples I've found some weird behavoir with the auth that made our apps lock in the loading state.

Everything was working well on the E2E tests, which made me even more confused.

I've finally found some time to debug the issue and noticed that the problem started with the double execution of `useEffect` in dev mode.

In this PR I'm setting up a quick fix to the issue by blocking the double execution of the useEffect inside the Provider.

We will probably need to redesign the API to not handle this kind of intialization inside of a `useEffect` and let the users control the creation of the jazz context to give them more flexibility on how to handle the setup of the auth and the app in general.

Looking at the `useEffect` dependencies it seems too easy to get in the situation where multiple contexts are created in a single session.

I'm thinking about something like this:

```ts
const jazz = createJazzBrowserContext({ AccountSchema, auth: new PasskeyAuth(), peer, storage });

function App() {
  const authState = usePasskeyAuth(jazz);
  
  return <>
       <JazzProvider value={jazz}><MyAppComponents /></JazzProvider>
       <MyAuthUI auth={authState} />
   </>
}
```

We can plan the redesing along with #562 so we introduce the breaking changes in the React API only once.

### DIFF
--- a/.changeset/wicked-icons-repeat.md
+++ b/.changeset/wicked-icons-repeat.md
@@ -1,0 +1,5 @@
+---
+"jazz-react": patch
+---
+
+Temporary fix to prevent double creation of Jazz context in React StrictMode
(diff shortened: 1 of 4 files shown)


bensleveritt:
This is good, HMR is killing my app every save at the moment

PR #676, opened 2024-11-04
Store clerk authentication locally

### LINKED ISSUE #673: Make Clerk auth work offline even when token expires
This means storing the logged-in Jazz account credentials in localStorage on successful login, similar to DemoAuth and PasskeyAuth and only doing the full Clerk login flow when no logged-in account is present in localStorage<br><br>This also means that we should encourage people to use the logout function provided by the Jazz context / Auth Method since that lets us clear localStorage and then log out of Clerk<br><br>(Needed for SA4)

bensleveritt:
Introduce LocallyStoredCredentials for storing and validating local storage values

### DIFF
--- a/packages/jazz-browser-auth-clerk/src/index.ts
+++ b/packages/jazz-browser-auth-clerk/src/index.ts
@@ -41,6 +41,7 @@ export class BrowserClerkAuth implements AuthMethod {
                 return {
                     type: "existing",
                     credentials,
+                    saveCredentials: async () => {}, // No need to save credentials when recovering from local storage
                     onSuccess: () => {},
                     onError: (error: string | Error) => {
                         this.driver.onError(error);
@@ -69,6 +70,18 @@ export class BrowserClerkAuth implements AuthMethod {
                             storedCredentials.jazzAccountID as ID<Account>,
                         secret: storedCredentials.jazzAccountSecret as AgentSecret,
                     },
+                    saveCredentials: async (credentials: {
+                        accountID: ID<Account>;
+                        secret: AgentSecret;
+                    }) => {
+                        localStorage.setItem(
+                            localStorageKey,
+                            JSON.stringify({
+                                accountID: credentials.accountID,
+                                secret: credentials.secret,
+                            }),
+                        );
+                    },
                     onSuccess: () => {},
                     onError: (error: string | Error) => {
                         this.driver.onError(error);
@@ -91,6 +104,13 @@ export class BrowserClerkAuth implements AuthMethod {
                         accountID: ID<Account>;
                         secret: AgentSecret;
                     }) => {
+                        localStorage.setItem(
+                            localStorageKey,
+                            JSON.stringify({
+                                accountID: credentials.accountID,
+                                secret: credentials.secret,
+                            }),
+                        );
                         await this.clerkClient.user?.update({
                             unsafeMetadata: {
                                 jazzAccountID: credentials.accountID,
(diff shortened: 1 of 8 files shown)


packages/jazz-browser-auth-clerk/src/index.ts
```diff
@@ -24,6 +31,33 @@ export class BrowserClerkAuth implements AuthMethod {
 
     async start(): Promise<AuthResult> {
         if (this.clerkClient.user) {
+            // Check local storage for credentials
```
aeplay: I haven't tried this, but in case the Clerk cookie expires / we reload, do we even have a user in the clerkClient? I.e. do we need to check local storage even above this `if`?

bensleveritt: Fair point, I'll test that scenario

packages/jazz-tools/src/implementation/createContext.ts
```diff
@@ -15,6 +15,10 @@ export type AuthResult =
     | {
           type: "existing";
           credentials: { accountID: ID<Account>; secret: AgentSecret };
+          saveCredentials?: (credentials: {
```
bensleveritt: Seems like a good hook for saving to localStorage, but it's not on the FixedCreds... putting in as an optional

aeplay (changes requested):
Thank you! Please create a patch changeset for the small jazz-tools change and the jazz-browser-auth-clerk changes

packages/jazz-browser-auth-clerk/src/tests/BrowserClerkAuth.test.ts
```diff
@@ -0,0 +1,117 @@
...
+            });
+
+            await expect(auth.start()).rejects.toThrow("Not signed in");
+        });
+
+        it("should save credentials to localStorage", async () => {
+            const auth = new BrowserClerkAuth(mockDriver, mockClerkClient);
+            const result = await auth.start();
+            result.saveCredentials?.({
+                accountID: "test-account-id" as ID<Account>,
```
gdorsi: **note:** the linter is breaking the build because of this error (referring to the Github annotations on line 101)

gdorsi (commented):
LGTM!

packages/jazz-browser-auth-clerk/src/tests/BrowserClerkAuth.test.ts
```diff
@@ -98,10 +98,12 @@ describe("BrowserClerkAuth", () => {
         it("should save credentials to localStorage", async () => {
             const auth = new BrowserClerkAuth(mockDriver, mockClerkClient);
             const result = await auth.start();
-            result.saveCredentials?.({
-                accountID: "test-account-id" as ID<Account>,
-                secret: "test-secret" as AgentSecret,
-            });
+            if (result.saveCredentials) {
+                await result.saveCredentials?.({
```
gdorsi: **nit:** the `?.` is now unnecessary

packages/jazz-browser-auth-clerk/src/tests/BrowserClerkAuth.test.ts
bensleveritt: Good spot, fixed

packages/jazz-browser-auth-clerk/src/tests/BrowserClerkAuth.test.ts
bensleveritt: Fixed

bensleveritt:
> Thank you! Please create a patch changeset for the small jazz-tools change and the jazz-browser-auth-clerk changes

Done!

aeplay (approved):

aeplay (changes requested):
Super nit picky, but please may you instead create two separate change sets, one for jazz-tools just describing the change you did there, and one for the clerk packages?

Because the messages and up in the changelogs for each package and the clerk stuff is higher-level than jazz-tools

packages/jazz-browser-auth-clerk/src/index.ts
bensleveritt: Fixed

aeplay (approved):

PR #698, opened 2024-11-05
Allow to define multiple storage options

### LINKED ISSUE #693: Expose ability to setup multiple storage peers
`indexedDB" | "singleTabOPFS" | ["indexedDB", "singleTabOPFS"]`

tobiaslins:
Resolves #693


packages/jazz-browser/src/index.ts
```diff
@@ -64,38 +68,53 @@ export async function createJazzBrowserContext<Acc extends Account>(
...
+    if (useSingleTabOPFS) {
+        peersToLoadFrom.push(
+            await LSMStorage.asPeer({
+                fs: new OPFSFilesystem(crypto),
+                // trace: true,
+            }),
+        );
+    }
+
+    if (useIndexedDB) {
```
gdorsi: This should be `useIndexedDB || !useSingleTabOPFS` because (for now) IndexedDB is our default when no storage option is specified.

**Optional:** It would be nice to extract the `StorageConfig` parsing logic in a function and add some tests on it. An LLM should do a good job on generating the tests for this kind of functions.

gdorsi: You can run the test suite by executing `pnpm test` from the project root.

To run the test on a subset of files you can add the file pattern like this `pnpm test parseStorageConfig.test.ts`

gdorsi:
Thanks a mill for the contribution!

packages/jazz-browser/src/index.ts
tobiaslins: great points! - added tests and extracted into a function

gdorsi (approved):
Perfect! 🚀

aeplay:
Thank you very much! Did the merge with the giant formatting PR and added a changeset (so it triggers a release & changelogs PR)

PR #699, opened 2024-11-05
Add more information to example apps page

trishalim:
I did not include Vue because we don't have docs for it. I think it's not "done" until there are docs.


aeplay:
Looking great already 😍

gdorsi:

trishalim:
I don't think this needs to be under `/docs`. It can be its own page (without the side nav) when I link it from the main nav.

PR #707, opened 2024-11-06
Fix coList header docs

bensleveritt:


aeplay:
Thanks for spotting this!

PR #709, opened 2024-11-06
Add some initial coFeed and FileStream docs

bensleveritt:
Adding coFeed and FileStream docs

### DIFF
--- a/packages/jazz-tools/src/coValues/interfaces.ts
+++ b/packages/jazz-tools/src/coValues/interfaces.ts
@@ -60,7 +60,7 @@ export function isCoValueClass<V extends CoValue>(
 
 /**
  * IDs are unique identifiers for `CoValue`s.
- * Can be used as a type argument to refer to a specific `CoValue` type.
+ * Can be used with a type argument to refer to a specific `CoValue` type.
  *
  * @example
  *
--- a/packages/jazz-tools/src/coValues/interfaces.ts
+++ b/packages/jazz-tools/src/coValues/interfaces.ts
@@ -58,7 +58,18 @@ export function isCoValueClass<V extends CoValue>(
   return typeof value === "function" && value.fromRaw !== undefined;
 }
 
-/** @category CoValues */
+/**
+ * IDs are unique identifiers for `CoValue`s.
+ * Can be used as a type argument to refer to a specific `CoValue` type.
+ *
+ * @example
+ *
+ * ```ts
+ * type AccountID = ID<Account>;
+ * ```
+ *
+ * @category CoValues
+ */
 export type ID<T> = CojsonInternalTypes.RawCoID & IDMarker<T>;
 
 type IDMarker<out T> = { __type(_: never): T };
--- a/packages/jazz-tools/src/coValues/coFeed.ts
+++ b/packages/jazz-tools/src/coValues/coFeed.ts
@@ -123,6 +123,10 @@ export class CoFeed<Item = any> extends CoValueBase implements CoValue {
    */
   [key: ID<Account>]: CoFeedEntry<Item>;
 
+  /**
+   * The current account's view of this `CoFeed`
+   * @category Content
+   */
   get byMe(): CoFeedEntry<Item> | undefined {
     if (this._loadedAs._type === "Account") {
       return this[this._loadedAs.id];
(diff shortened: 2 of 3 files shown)


bensleveritt:
@aeplay I've taken some liberties, but I figure writing some sub-par docs will prompt improvements. Please correct anything I've done, or let me know where I've got it too wrong and I'll correct it

gdorsi:

packages/jazz-tools/src/coValues/coStream.ts
```diff
@@ -50,44 +50,97 @@ export type SingleCoStreamEntry<Item> = {
...
 };
 
-/** @category CoValues */
+/**
+ * CoStreams are collaborative logs of data.
+ *
+ * @categoryDescription Content
+ * They are similar to `CoList`s, but with a few key differences:
+ * - They are append-only
+ * - They have a per-session view
```
aeplay:
```suggestion
 * - They consist of several internal append-only logs, one per account session (tab, device, app instance, etc.)
 * - They expose those as a per-account aggregated view (default) or a precise per-session view
```

packages/jazz-tools/src/coValues/coStream.ts
```diff
@@ -50,44 +50,97 @@ export type SingleCoStreamEntry<Item> = {
...
   static _schema: any;
+  /** @internal */
   get _schema(): {
     [ItemsSym]: SchemaFor<Item>;
   } {
     return (this.constructor as typeof CoStream)._schema;
   }
 
+  /**
+   * The per-account view of this `CoStream`
```
aeplay: Adding an example here would be good to make clear how you can
- directly index the CoFeed by account ID
- use something like Object.entries on it

packages/jazz-tools/src/coValues/coStream.ts
```diff
@@ -138,6 +195,10 @@ export class CoStream<Item = any> extends CoValueBase implements CoValue {
     return instance;
   }
 
+  /**
+   * Push items to this `CoStream`
```
aeplay: Maybe explain here how this will append it's to the current session's log?

packages/jazz-tools/src/coValues/coStream.ts
```diff
@@ -227,15 +300,23 @@ export class CoStream<Item = any> extends CoValueBase implements CoValue {
     return subscribeToCoValue<S, Depth>(this, id, as, depth, listener);
   }
 
-  /** @category Subscription & Loading */
+  /**
+   * Ensure a `CoStream` is loaded
```
aeplay: Make clearer here that this returns a new instance of the same CoFeed that's deeply loaded as per the spec, or undefined if it can't be loaded that deeply

packages/jazz-tools/src/coValues/interfaces.ts
```diff
@@ -58,7 +58,18 @@ export function isCoValueClass<V extends CoValue>(
   return typeof value === "function" && value.fromRaw !== undefined;
 }
 
-/** @category CoValues */
+/**
+ * IDs are unique identifiers for `CoValue`s.
+ * Can be used as a type argument to refer to a specific `CoValue` type.
```
aeplay:
```suggestion
 * Can be used with a type argument to refer to a specific `CoValue` type.
```

aeplay (changes requested):
Thanks @bensleveritt! Added a couple comments.

In addition, please may you adapt this to use the new names `CoFeed` and `FileStream`?

packages/jazz-tools/src/coValues/coStream.ts
bensleveritt: Addressed now

packages/jazz-tools/src/coValues/coStream.ts
bensleveritt: Added explanation and example

packages/jazz-tools/src/coValues/coStream.ts
bensleveritt: Added some examples

.changeset/fifty-gifts-wash.md
```diff
@@ -0,0 +1,5 @@
+---
```
aeplay: Why is this changeset in this PR?

aeplay (changes requested):

.changeset/fifty-gifts-wash.md
bensleveritt: Ah, that seems to have been @gdorsi merging from master to keep it in sync. I think I can remove the commit and it'll **just work**™️

aeplay (approved):

PR #725, opened 2024-11-08
Immediately ack new content before syncing it upstream

aeplay:


packages/cojson/src/tests/sync.test.ts
```diff
@@ -1656,6 +1652,10 @@ describe("SyncManager - knownStates vs optimisticKnownStates", () => {
       return push.call(jazzCloudConnectionAsPeer.outgoing, msg);
     });
 
+    map.set("key2", "value2", "trusting");
```
aeplay, line 1655: Had to move the new-content-creating bit after the mock, or it wouldn't work.

I'm sleepy so let me know if this still makes sense @gdorsi

gdorsi (approved):

packages/cojson/src/sync.ts
```diff
@@ -667,6 +665,8 @@ export class SyncManager {
         console.error("Error sending known state", e);
       });
     }
+
+    await this.syncCoValue(coValue);
```
aeplay, line 669: We do send a correction/ack message before syncing to give an immediate response to the peers that are waiting for confirmation that a value is fully synced

PR #735, opened 2024-11-11
Add user prop to demo auth to skip login on demos

trishalim:
This won't affect any existing demos.

If you pass a user prop, you won't be able to log out anymore but hopefully we just won't need that. If we do then I'll fix it.

### DIFF
--- a/packages/jazz-react/src/auth/DemoAuth.tsx
+++ b/packages/jazz-react/src/auth/DemoAuth.tsx
@@ -69,15 +69,28 @@ export function useDemoAuth({
 export const DemoAuthBasicUI = ({
   appName,
   state,
+  user,
 }: {
   appName: string;
   state: DemoAuthState;
+  user?: string;
 }) => {
   const [username, setUsername] = useState<string>("");
   const darkMode =
     typeof window !== "undefined"
       ? window.matchMedia("(prefers-color-scheme: dark)").matches
       : false;
+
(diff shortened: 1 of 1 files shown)


aeplay:
Looks like a good way to do it

aeplay:
Oops sorry forgot to ask for a change set - @trishalim can you please add one on the same branch and open a new PR? Thank you!

packages/jazz-react/src/auth/DemoAuth.tsx
```diff
@@ -69,16 +69,29 @@ export function useDemoAuth({
...
     typeof window !== "undefined"
       ? window.matchMedia("(prefers-color-scheme: dark)").matches
       : false;
 
+  if (user && state.state === "ready") {
+    if (state.existingUsers.includes(user)) {
+      state.logInAs(user);
+      return <></>;
+    } else {
+      state.signUp(user);
```
gdorsi, line 89: Side effects should not run in render, as React can render components multiple times.

We could fix the issue using a `useEffect`:
```ts
const isAutoLogin = Boolean(user && state.state === "ready");

useEffect(() => {
  if (!isAutoLogin) return;
  
  if (state.existingUsers.includes(user)) {
      state.logInAs(user);
    } else {
      state.signUp(user);
    }
}, [isAutoLogin]);

if (isAutoLogin) return <></>;
```

PR #739, opened 2024-11-11
Move auto login check to useEffect

trishalim:


packages/jazz-react/src/auth/DemoAuth.tsx
```diff
@@ -81,16 +81,19 @@ export const DemoAuthBasicUI = ({
       ? window.matchMedia("(prefers-color-scheme: dark)").matches
       : false;
 
-  if (user && state.state === "ready") {
+  const isAutoLogin = !!(user && state.state === "ready");
```
trishalim, line 84: @gdorsi The `Boolean()` gave me TS errors that don't make sense to me
image

gdorsi: Ah, it's because Typescript doesn't do type narrowing when using `Boolean`.

gdorsi (approved):
Looks good to me!

aeplay:
Thanks for the advice @gdorsi and for the updated implementation @trishalim !

PR #741, opened 2024-11-12
Onboarding example

marinoska:
This PR introduces the JAZZ Onboarding example application, which demonstrates a simple onboarding flow.

- As an HR manager, a user initiates a new employee profile in the onboarding flow.
- HR managers can create and delete employees during the onboarding process until completion.
- Only admins have the permission to complete the final onboarding confirmation.
- Invited coworkers becomes writers and participate in each onboarding step but the final confirmation.
- Only the admin can finalise the third step by performing the final confirmation, which locks the profile from further changes.

Screenshot 2024-11-12 at 21 14 26

https://www.loom.com/share/b2a01237db094dff9678e4f98c840950?sid=294e1d3c-649b-4bdc-9bd7-a5d85a740e82


examples/onboarding/src/pages/NewEmployee.tsx
```diff
@@ -0,0 +1,88 @@
...
+export function NewEmployee({
+  employeeListCoId,
+}: {
+  employeeListCoId: ID<EmployeeCoList>;
+}) {
+  const navigate = useNavigate();
+  const { me } = useAccount();
+  const writerGroup = Group.create({ owner: me });
+  const readerGroup = Group.create({ owner: me });
+  readerGroup.addMember("everyone", "reader");
```
gdorsi: **nit:** having these "Group.create" here means that we do create these groups at each render.

Since we use them only inside `createEmployee` I would suggest to move the groups cration inside there.

gdorsi (approved):
LGTM!

Very well poslished and thanks for the tests!

examples/onboarding/src/pages/NewEmployee.tsx
marinoska: Well spotted! fixed that.

aeplay (approved):
The implementation looks really good!

PR #743, opened 2024-11-12
fix: fixes the react provider intialization when multiple storage options are provided

gdorsi:


packages/jazz-react/src/index.tsx
```diff
@@ -50,72 +50,75 @@ export function createJazzReactApp<Acc extends Account>({
...
 
-      return () => {
-        void promise.then((context) => context.done());
-      };
-    }, [AccountSchema, auth, peer, storage, sessionCount]);
+        return () => {
+          void promise.then((context) => context.done());
+        };
+      },
+      [AccountSchema, auth, peer, sessionCount].concat(storage as any),
```
gdorsi, line 120: The core change is here, the rest is formatting.

gdorsi: Used `concat` because it already handles the difference between a single value and an array

aeplay (approved):

PR #744, opened 2024-11-12
fix: dispatch more updates to knownState

gdorsi:
Reviewing the code to look for the next steps I've noticed that some of the update that shouldn't be optimistic only weren't send to known states.

So I've added an API to dispatch to both states and increased the coverage of the known states to all the "non-optimistic" updates.

### DIFF
--- a/packages/cojson/src/sync.ts
+++ b/packages/cojson/src/sync.ts
@@ -400,7 +400,7 @@ export class SyncManager {
   }
 
   async handleLoad(msg: LoadMessage, peer: PeerState) {
-    peer.optimisticKnownStates.dispatch({
+    peer.dispatchToKnownStates({
       type: "SET",
       id: msg.id,
       value: knownStateIn(msg),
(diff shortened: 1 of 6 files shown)


packages/cojson/src/sync.ts
```diff
@@ -454,7 +454,7 @@ export class SyncManager {
       const loaded = await entry.state.ready;
 
       if (loaded === "unavailable") {
-        peer.optimisticKnownStates.dispatch({
+        peer.dispatchToKnownStates({
           type: "SET",
           id: msg.id,
           value: knownStateIn(msg),
```
gdorsi, line 460: The load related message should not be considered optimistic

packages/cojson/src/sync.ts
```diff
@@ -688,7 +682,7 @@ export class SyncManager {
   }
 
   async handleCorrection(msg: KnownStateMessage, peer: PeerState) {
-    peer.optimisticKnownStates.dispatch({
+    peer.dispatchToKnownStates({
```
gdorsi, line 685: Same goes for the "new-content" and correction messages

aeplay (approved):

aeplay:
Changeset pls

PR #748, opened 2024-11-13
Show two chat demos side by side

trishalim:
- show 2 iframes for chat demo
- move chat demo right after covalues section
- remove code from chat app
- skip login by passing a username to DemoAuth
- getting 2 iframes to be open on the same chat room was tricky, so I hid the log out button so we don't have to handle those situations

To test, please run locally, and run 2 instances of the chat app.


examples/chat/src/app.tsx
```diff
@@ -15,6 +24,15 @@ export function App() {
...
     router.navigate("/#/chat/" + chat.id);
+
+    if (window.parent) {
+      waitForUpload(chat.id, me).then(() => {
+        window.parent.postMessage(
+          { type: "chat-load", id: "/chat/" + chat.id },
+          "*",
+        );
+      });
+    }
```
trishalim: @aeplay @gdorsi all this code for the homepage chat demo is gonna get confusing for those referencing the chat app as a guide. What can I do, make another chat app specifically for the homepage demo?

gdorsi: What about extracting this logic away in a separate module to make it less visible?

Would that work?

Better to keep one example if possible.

trishalim: Alright I managed to reduce it to one line

aeplay (changes requested):
Hey, thanks for this, this is a lot more inviting to use!

- I pushed a commit with a better approach to hiding the logout button for the demo, but I did appreciate the pragmatism of the CSS equivalent of duct-tape
- please could you move the testimonial above the chat example? For me it belongs to the CoValue section
- please could you introduce an interactive section (maybe where the testimonial used to be?) that encourages people to try connecting with an extra tab/device by showing the chat URL once we have it, both as copyable input and QR code? (in the external device you would get the normal login flow and could chat with Alice and Bob "as yourself")

trishalim:
@aeplay 
How can I chat with Alice and Bob on a third window? Do we need a third instance of the chat app?

aeplay:
> @aeplay How can I chat with Alice and Bob on a third window? Do we need a third instance of the chat app?

Yes @trishalim  I guess we need a third instance, one for Alice, one for Bob and then the "normal one" at chat.jazz.tools to open elsewhere

aeplay:
Beautiful ✨

PR #750, opened 2024-11-13
Redesign desktop nav, use shadcn and radix

trishalim:
I ran into an issue with radix with positioning the dropdown
https://github.com/shadcn-ui/ui/issues/1588
https://github.com/shadcn-ui/ui/issues/94

The best I could do was left align it

Moving your cursor on the space between the nav and the dropdown causes the dropdown to close, so I removed the space in between
image

I'd prefer to use [headless](https://github.com/gardencmp/jazz/pull/713) but they open the dropdown on click, not hover.


bensleveritt:
Looks good to me

bensleveritt (approved):

PR #752, opened 2024-11-13
fix: reuse the crypto instance when creating the browser context

gdorsi:
When using both the `IndexedDB` and the OPFS storage peers we were getting:

```
index.esm.js:248 Uncaught (in promise) RangeError: WebAssembly.instantiate(): Out of memory: Cannot allocate Wasm memory for new instance
    at index.esm.js:248:46
    at Generator.next (<anonymous>)
    at fulfilled (index.esm.js:27:58)
```

Reusing the crypto instance fixes the issue.

### DIFF
--- a/packages/cojson-storage-indexeddb/src/index.ts
+++ b/packages/cojson-storage-indexeddb/src/index.ts
@@ -91,7 +91,7 @@ export class IDBStorage {
   ): Promise<Peer> {
     const [localNodeAsPeer, storageAsPeer] = cojsonInternals.connectedPeers(
       localNodeName,
-      "storage",
+      "indexedDB",
       {
         peer1role: "client",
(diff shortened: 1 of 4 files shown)


packages/cojson-storage-indexeddb/src/index.ts
```diff
@@ -91,7 +91,7 @@ export class IDBStorage {
   ): Promise<Peer> {
     const [localNodeAsPeer, storageAsPeer] = cojsonInternals.connectedPeers(
       localNodeName,
-      "storage",
+      "indexedDB",
```
gdorsi, line 94: Added this change to have a different name between this peer and the OPFS one.

aeplay (approved):

PR #757, opened 2024-11-14
Rename CoStream to CoFeed

marinoska:


gdorsi (commented):
Would it be possible to export `CoStream` as an alias?

Otherwise this would be technically a breaking change.

marinoska:
> Would it be possible to export `CoStream` as an alias?
> 
> Otherwise this would be technically a breaking change.

Why not, really?

marinoska:
> Would it be possible to export `CoStream` as an alias?
> 
> Otherwise this would be technically a breaking change.

For sure

marinoska:
@aeplay What about BinaryCoStream?

gdorsi:
> @aeplay What about BinaryCoStream?

Maybe we should rename them into "FileStream"?

aeplay (approved):
Perfect, thanks for the graceful deprecation / backwards-compatibility!

aeplay:
> @aeplay What about BinaryCoStream?

I'd say BinaryStream - as it can also be used for ongoing livestreams, not just finite files (cc @marinoska  @gdorsi)

gdorsi:
> > @aeplay What about BinaryCoStream?
> 
> I'd say BinaryStream - as it can also be used for ongoing livestreams, not just finite files (cc @marinoska @gdorsi)

The most common usage would be related to files and I think that using the "File" prefix helps with discoverabilty (one should be able to find and use it with just the TS declarations without looking at the docs).

For livestreams we can do an alias or another view type. I think that would give the feeling of a more solid support from our side.

marinoska:
> > > @aeplay What about BinaryCoStream?
> > 
> > 
> > I'd say BinaryStream - as it can also be used for ongoing livestreams, not just finite files (cc @marinoska @gdorsi)
> 
> The most common usage would be related to files and I think that using the "File" prefix helps with discoverabilty (one should be able to find and use it with just the TS declarations without looking at the docs).
> 
> For livestreams we can do an alias or another view type. I think that would give the feeling of a more solid support from our side.

@aeplay wdyt?

aeplay:
@marinoska @gdorsi OK, let's go with FileStream for now and invent a new view when we actually have high-level conveniences for live streaming

PR #759, opened 2024-11-14
feat: retry unavailable coValues

gdorsi:
With this PR we introduce a retry system when clients are trying to load CoValues and the server responds with an "unavailable" state.

This should help on the share interactions where the reference is shared before a full upload the CoValue.


packages/cojson-transport-ws/src/index.ts
```diff
@@ -18,6 +18,7 @@ export type CreateWebSocketPeerOpts = {
   role: Peer["role"];
   expectPings?: boolean;
   batchingByDefault?: boolean;
+  retryUnavailableCoValues: boolean;
```
gdorsi: Added a flag to handpick the peers to retry.

This is mostly for two reasons:
1. We don't need to retry on storage peers. If a value is unavailable there we can only hope to get it from the server peers
2. I didn't want to introduce the retry on the sync servers.

We could also use a role-based logic (e.g. only "server" peers") but preferred to go for a more explicit route to be sure of where we do retries.

packages/cojson/src/coValueState.ts
```diff
@@ -1,53 +1,50 @@
...
-
-  const promise = new Promise<T>((res) => {
-    resolve = res;
-  });
+export const CO_VALUE_LOADING_MAX_RETRIES = 5;
 
-  return { promise, resolve };
+export class CoValueUnknownState {
+  type = "unknown" as const;
 }
```
gdorsi, line 10: Moved all the loading management from here into a new `CoValueLoadingState`

packages/cojson/src/coValueState.ts
```diff
@@ -1,53 +1,50 @@
...
-    ReturnType<typeof createResolvablePromise<"available" | "unavailable">>
-  >;
-  private resolve: (value: "available" | "unavailable") => void;
+    ReturnType<typeof createResolvablePromise<void>>
+  >();
+  private resolve: (value: CoValueCore | "unavailable") => void;
 
-  ready: Promise<"available" | "unavailable">;
+  result: Promise<CoValueCore | "unavailable">;
+  resolution: undefined | CoValueCore | "unavailable";
```
gdorsi, line 21: Added this `resolution` attribute to be able to read the resolved state syncronously

packages/cojson/src/coValueState.ts
```diff
@@ -71,12 +69,16 @@ class CoValueUnknownState {
...
-class CoValueAvailableState {
+export class CoValueAvailableState {
   type = "available" as const;
 
   constructor(public coValue: CoValueCore) {}
 }
 
+export class CoValueUnavailableState {
+  type = "unavailable" as const;
+}
```
gdorsi, line 80: Added this state to mark a CoValue as unavailable.

Used in the load function to see if we already tried to load it and failed.
In that case we try to load the CoValue again. (cc @marinoska this might solve some of the "need to reload" issues)

packages/cojson/src/coValueState.ts
```diff
@@ -89,29 +91,166 @@ type CoValueStateAction =
...
+    return new CoValueState(id, new CoValueLoadingState(peersIds));
   }
 
   static Available(coValue: CoValueCore) {
-    return new CoValueState(new CoValueAvailableState(coValue));
+    return new CoValueState(coValue.id, new CoValueAvailableState(coValue));
   }
 
-  dispatch(action: CoValueStateAction) {
+  async loadFromPeers(peers: PeerState[]) {
```
gdorsi, line 172: Moved `loadFromPeers` from the SyncManager to here.

This way all the loading state management is self-contained inside `CoValueState` and it is easier to cover with tests.

packages/cojson/src/coValueState.ts
```diff
@@ -89,29 +91,174 @@ type CoValueStateAction =
...
+    await peer.pushOutgoingMessage({
+      action: "load",
+      id: coValueEntry.id,
+      header: false,
+      sessions: {},
+    });
+
+    if (coValueEntry.state.type === "loading") {
+      await coValueEntry.state.waitForPeer(peer.id);
+    }
```
gdorsi, line 276: Once the coValueEntry enters into an "available" state we can add to the "load" messages the header/sessions we have in the coValue.

We should also give a look at `subscribeToIncludingDependencies` because we might need to subscribe to the dependencies.

Could maybe fix part of #760  (cc @marinoska)

packages/cojson/src/coValueState.ts
```diff
@@ -89,29 +91,174 @@ type CoValueStateAction =
...
+}
+
+function getPeersWithoutErrors(peers: PeerState[], coValueId: RawCoID) {
+  return peers.filter((p) => {
+    if (p.erroredCoValues.has(coValueId)) {
+      console.error(
+        `Skipping load on errored coValue ${coValueId} from peer ${p.id}`,
+      );
+      return false;
+    }
```
gdorsi, line 324: Added this helper to keep the warning from the sync manager.

e2e/BinaryCoStream/src/DownloaderPeer.tsx
```diff
@@ -1,17 +1,14 @@
...
-    uploadedFileId,
-    me,
-    Boolean,
-    {},
-  );
+  const uploadedFile = await UploadedFile.load(uploadedFileId, me, {});
+
+  if (!uploadedFile) {
+    throw new Error("Uploaded file not found");
+  }
```
gdorsi, line 11: When testing locally with the "offline then resume" test I'm getting some consistent failures.

It doesn't happen on CI but need to investigate more nonethless.

packages/cojson/src/sync.ts
```diff
@@ -511,7 +480,10 @@ export class SyncManager {
...
     }
 
-    if (entry.state.type === "unknown") {
+    // When we receive a known state from a peer and the header is "false"
+    // it means that the peer doesn't know anything about this coValue
+    // So we can say that the coValue has been not found on this peer
+    if (entry.state.type !== "available") {
       const availableOnPeer = peer.optimisticKnownStates.get(msg.id)?.header;
 
       if (!availableOnPeer) {
```
gdorsi, line 472: If this message is lost we could end up infinitely waiting when loading a coValue.

This problem makes me think about @marinoska suggestion about having "an ack of the ack".
I think `known` messages need to have the insurance to be delivered because if we miss some of them CoValues might get stuck.

gdorsi: It's actually the same with the "load" messages. 

The only ones we can afford to lose are the "content" messages.

e2e/BinaryCoStream/src/DownloaderPeer.tsx
gdorsi: Found the issue and fixed: https://github.com/gardencmp/jazz/pull/759/commits/bb126009c89c0f332f497e8393cff7f7764afcb3

packages/cojson/src/sync.ts
```diff
@@ -408,52 +388,32 @@ export class SyncManager {
     let entry = this.local.coValues[msg.id];
 
     if (!entry) {
-      // console.log(`Loading ${msg.id} from all peers except ${peer.id}`);
+      this.local.coValues[msg.id] = entry = CoValueState.Unknown(msg.id);
+    }
```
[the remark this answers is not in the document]
gdorsi: Follow-up PR https://github.com/gardencmp/jazz/pull/775

packages/cojson/src/CoValuesStore.ts
```diff
@@ -0,0 +1,38 @@
...
+    return entry;
+  }
+
+  setAsAvailable(id: RawCoID, coValue: CoValueCore) {
+    const entry = this.get(id);
+    entry.dispatch({
+      type: "available",
+      coValue,
+    });
+  }
```
gdorsi, line 25: Introduced this store to enforce two things in the CoValues state management:
- That reads always return an entry defaulting to an Unknown state
- Once set an entry is always the same. We never override o delete entries (we will delete entries once we implement the garbage collection system)

packages/cojson/src/sync.ts
```diff
@@ -474,31 +422,37 @@ export class SyncManager {
...
+        if (
+          dependencyEntry.state.type === "available" ||
+          dependencyEntry.state.type === "loading"
+        ) {
+          this.local.loadCoValueCore(msg.id, peer.id).catch((e) => {
+            console.error(
+              `Error loading coValue ${msg.id} to create loading state, as dependency of ${msg.asDependencyOf}`,
+              e,
+            );
+          });
```
gdorsi, line 454: @aeplay 

Not sure that the new state checks I've added are reflecting the previous intention here.

.vscode/settings.json
bensleveritt, line 1: Can we move this into a new PR? I agree that we probably want to avoid committing VS Code settings, but I don't think this relates to this change, and I think there might be some scope for discussion

gdorsi: https://github.com/gardencmp/jazz/pull/784

bensleveritt: Grazie 🙏

packages/cojson-transport-ws/src/index.ts
aeplay: Please make it rule-based to be only for server peers. This is exactly the behaviour we want (the edge sync servers should also retry the core sync server).

You're correct that it doesn't make sense on the storage peers.

aeplay (changes requested):

packages/cojson/src/sync.ts
```diff
@@ -511,36 +464,34 @@ export class SyncManager {
       }
     }
 
-    if (entry.state.type === "unknown") {
+    // When we receive a known state from a peer and the header is "false"
+    // it means that the peer doesn't know anything about this coValue
+    // So we can say that the coValue has been not found on this peer
+    if (entry.state.type !== "available") {
       const availableOnPeer = peer.optimisticKnownStates.get(msg.id)?.header;
```
marinoska, line 471: This line is familiar to me as the failing invitation flow consists of two steps of handling two messages in a row - the known one without a header and the content one with a header. I thought the lacking header in the known was a "feature" but now I see that actually the header is expected for the value well known on a peer. 
Not sure how but this PR solves the lack of the header in the known msg in that case.

gdorsi: > I thought the lacking header in the known was a "feature" but now I see that actually the header is expected for the value well known on a peer.

The header is a boolean value that tells us if the other peer do have information about the header.
If it's false in this point it means that the coValue is unavailable on the other peer.

In the case of the invitation flow, if we are getting `header: false` here, it means that the coValues related to the invitation haven't been uploaded yet to the sync server.

With this new logic we are going to get an "unavailable" here and the CoValueState is going to retry the load.

This doesn't solve our problem with invitations in general because the coValues might not be uploaded for a range of reasons, but should fix the most common issue related to the timing of sharing.

marinoska: The peer responded with both known and content messages bundled in one single response and as I remember the known msg had header: false , whereas content msg was complete. And that situation isn't right as I understand it now because the covalue was actually known and the header had to be provided.
I'll try to reproduce it now

gdorsi: Yeah, definitely a bug!

marinoska: And I can't reproduce the invitation issue anymore. I'll be keeping it in mind though

gdorsi: Ok, looking again at the code, I should have fixed that bug.

When handling a new content [we flag the coValue as available](https://github.com/gardencmp/jazz/blob/5724735a9539fb25a6fbb5ea386ede1c62a161b1/packages/cojson/src/sync.ts#L493) if it wasn't before.

This may happen after that the same peer flagged the coValue as unavailable.

I've noticed this possible flow and covered it with a test [here](https://github.com/gardencmp/jazz/blob/5724735a9539fb25a6fbb5ea386ede1c62a161b1/packages/cojson/src/tests/coValueState.test.ts#L260).

The new CoValueState is designed to have an explicit "unavailable" state and manage the transition from "unavailable" to "available" correctly.

Also, the returned value (from getValue) must be always consistent with the current internal state.

aeplay: Perfect, this is an edge case I was worried about! Thanks for ensuring it's handled correctly.

aeplay (approved):

PR #765, opened 2024-11-15
Move examples page outside of docs

trishalim:


aeplay (changes requested):
Thanks for moving it out!

This is kind of offtopic but please can we make the example file in the preview for the music app "take-five.mp3"? More Jazz themed and probably public domain i.e. not implying piracy :D

PR #768, opened 2024-11-15
Enhance onboarding test - add page context

marinoska:
Looks neater and I prefer the context's approach, but it won't solve the flaky problem as the app fails at the invitation reception stage which is endlessly loading every other time when I use the app myself

### DIFF
--- a/examples/onboarding/playwright.config.ts
+++ b/examples/onboarding/playwright.config.ts
@@ -20,7 +20,7 @@ export default defineConfig({
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     /* Base URL to use in actions like `await page.goto('/')`. */
-    baseURL: "http://localhost:5173/?peer=ws://localhost:1234",
+    baseURL: "http://localhost:5173/",
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
     trace: "on-first-retry",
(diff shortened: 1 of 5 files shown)


gdorsi:
>  as the app fails at the invitation reception stage which is endlessly loading every other time when I use the app myself

That's because the app shows an alert which stops the main thread blocking the sync messages required to upload the invitation.

Changing the `alert` with a [toast](https://github.com/gardencmp/jazz/blob/main/examples/music-player/src/3_HomePage.tsx#L69) should fix the issue (or make it less likely to happen).

marinoska:
> > as the app fails at the invitation reception stage which is endlessly loading every other time when I use the app myself
> 
> That's because the app shows an alert which stops the main thread blocking the sync messages required to upload the invitation.
> 
> Changing the `alert` with a [toast](https://github.com/gardencmp/jazz/blob/main/examples/music-player/src/3_HomePage.tsx#L69) should fix the issue (or make it less likely to happen).

Invitation is accepted  in the other window not the one with the alert. I can see the problem when I manually close the alert popup and insert the link in another new window, then probably need to reload it a couple of times.

gdorsi:
@marinoska 

Pushed a fix.
When testing locally I was getting in the test console `Failed to accept invite Error: Trying to accept invite: Group/owned value unavailable from all peers`

This happens when the Group on the other side haven't been uploaded yet.

Added a fix with by adding a timeout after the invite creation but the test is still failing on CI on the first run (succedes on the second retry)

Hopefully #759 would bring a little more stability on this side.

gdorsi (approved):
LGTM

**praise:** I like ALOT how you set up the test code.

Very clean and easy to understand.

gdorsi:
Tested [here](https://github.com/gardencmp/jazz/actions/runs/11870682042/job/33082380686) if the retry system would help but no luck.

We need to do a deeper investigation on the problem.

marinoska:
> LGTM
> 
> **praise:** I like ALOT how you set up the test code.
> 
> Very clean and easy to understand.

 Thank you!

PR #780, opened 2024-11-18
fix(SyncManager): add flag to delete peers on close

gdorsi:
Keeping the PeerState of the client websocket peers that are connecting to our sync servers is causing their memory to grow over time.

Since keeping the PeerState is something we need only in client apps and between the sync servers in our cloud I've added a flag to be able to selectively delete the peer states that we don't need to keep.

### DIFF
--- a/.changeset/modern-zebras-invent.md
+++ b/.changeset/modern-zebras-invent.md
@@ -1,0 +1,5 @@
+---
+"jazz-run": patch
+---
+
+Enable the deletePeerStateOnClose on the Websocket peers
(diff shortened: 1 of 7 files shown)


aeplay (approved):
I'm wondering if this should just be default on for "client" peers but I'm unsure so for now it's good to have a flag we can explicitly enable on the sync server.

In the future we might want a "delete after timeout" in case the client is flakily reconnecting, but that may be solved by allowing peers to be internally temporarily disconnected

PR #783, opened 2024-11-18
Tweak 'Docs coming soon' copy

bensleveritt:


trishalim (approved):

aeplay:
@bensleveritt the tone is a bit too corporate for my taste, but it's better than what we had before, so I'm merging it

PR #788, opened 2024-11-18
Add accounts & migration docs

bensleveritt:


homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx
```diff
@@ -0,0 +1,68 @@
+import { CodeGroup } from "@/components/forMdx";
+
+# Accounts & Migrations
```
Comment (trishalim) [inline homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx]: @bensleveritt can we use sentence case please? Right now we have a mix of sentence case and title case but I'll fix it

homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+To create a custom account type, extend the `Account` class and register it with the Jazz runtime.
+
+<CodeGroup>
+```ts
+import { Account, CoMap } from "jazz-tools"; 
+
+class MyAccountRoot extends CoMap {
+  privateString = co.string; // Readable by the account owner
+}
```
Comment (trishalim) [inline homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx:21]: What makes MyAccountRoot private? Is it the `{ owner: this }` on migration? If so, then we could also make that public right?

homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx
```diff
@@ -0,0 +1,70 @@
...
+// React
+const Jazz = createJazzReactApp({
+  AccountSchema: MyAccount,
+});
+
+// Vue
+const Jazz = createJazzVueApp({
+  AccountSchema: MyAccount,
+});
+```
```
Comment (trishalim) [inline homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx:62]: Also add react native here?

And put in a separate CodeGroup so that they can be copied separately with the copy button

Not related to this PR:
We'll have a lot of instances like this where the code is different for each framework.. We can do a global toggle maybe

Comment (bensleveritt) [reply homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx:62]: Will put React Native example in, and separate CodeGroups 

Yeah a language-toggleable group would be a good component to make.

homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx
Comment (aeplay) [reply homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx:21]: Correct on both points, but the convention is that profile is public stuff, root is private

homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx
```diff
@@ -0,0 +1,70 @@
+import { CodeGroup } from "@/components/forMdx";
+
+# Accounts & migrations
+
+Accounts are a fundamental building block for user identity and data organization in Jazz applications, providing a structured way to manage both public and private user data while maintaining type-safety and access control.
+
+Account data and metadata are publicly visible to other users.
+
+For private data, we suggest nesting privately owned CoValues within the account.
```
Comment (aeplay) [inline homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx]:
```suggestion
For private data, we suggest nesting privately owned CoValues within the account root.
```

homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx
```diff
@@ -0,0 +1,70 @@
...
+```ts
+import { Account, CoMap } from "jazz-tools"; 
+
+class MyAccountRoot extends CoMap {
+  privateString = co.string; // Readable by the account owner
+}
+
+class MyAccount extends Account {
+  root = co.ref(MyAccountRoot);
+  publicString = co.string; // Readable by everyone
```
Comment (aeplay) [inline homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx]: We discourage putting stuff into the account directly, public metadata should go in the profile

aeplay (changes requested):
This is a good start and infinitely better than what we have. I will likely extend on it soon - but in the meantime the comments I made are the minimal changes necessary to ensure it's correct advice

homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+class MyProfile extends Profile {
+  publicString = co.string; // Readable by everyone
+}
+
+class MyAccount extends Account {
+  root = co.ref(MyAccountRoot);
+  profile = co.ref(MyProfile);
+}
+```
```
Comment (trishalim) [inline homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx:31]: In this code snippet (without the migration yet), root and profile would both be public right?

homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+Accounts can be migrated to add new data or update schemas. The `migrate()` method runs automatically when an account is created and each time a user logs in.
+
+You can use migrations to initialize the account root and set up any other required CoValues.
+
+
+<CodeGroup>
+```ts
+class MyAccount extends Account {
+  async migrate() {
```
Comment (trishalim) [inline homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx:75]: missing `creationProps` here?

homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx
Comment (trishalim) [reply homepage/homepage/app/docs/[...slug]/schemas/accounts.mdx:31]: If you only read this part, I think it gets confusing. I suggest to write only about the Profile. 

In the Migrations section, we can talk about adding root, and how that's where private data should go.

trishalim:
This is useful and already 80% of the way there. May I continue with this? @bensleveritt @aeplay Unless of course you want to @bensleveritt, I don't wanna be a task thief 😆

PR #794, opened 2024-11-18
PassphraseAuth - Move saving credentials into `saveCredentials`

bensleveritt:
Moves saving credentials into `saveCredentials`, to match other implementations


aeplay:
Great, thx!

PR #798, opened 2024-11-18
SchemaUnion.Of() implementation

pax-k:


.changeset/warm-readers-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Implemented SchemaUnion.Of()
```
gdorsi: I'd like us to try to be more descriptive on the changelogs.
I know it's not easy, but if we work toghether should be more doable.

```suggestion
Added SchemaUnion, a new API that let's you assign multiple possible schemas to a given coValue and selectively apply them at runtime.
```

packages/jazz-tools/src/coValues/schemaUnion.ts
```diff
@@ -0,0 +1,105 @@
...
+ * const widget = await WidgetUnion.load(id, me, {});
+ * if (widget instanceof ButtonWidget) {
+ *   console.log(widget.label);
+ * } else if (widget instanceof SliderWidget) {
+ *   console.log(widget.min, widget.max);
+ * }
+ * ```
+ *
+ * @category CoValues
+ */
```
gdorsi, line 56: **praise:** great docs!

gdorsi (approved):
LGTM

Thanks for the extensive docs!

.changeset/warm-readers-invite.md
aeplay: Well, if we're being precise, I'd suggest

```suggestion
Added SchemaUnion.Of(), a new API that lets you express unions of other Schema definitions (often different subclasses of a parent Schema class) that are correctly narrowed in TypeScript and correctly instantiated at runtime based on a discriminating field in the raw loaded CoValue.
```

packages/jazz-tools/src/coValues/schemaUnion.ts
```diff
@@ -0,0 +1,105 @@
...
+ *   switch (raw.get("type")) {
+ *     case "button": return ButtonWidget;
+ *     case "slider": return SliderWidget;
+ *     default: throw new Error("Unknown widget type");
+ *   }
+ * });
+ * ```
+ *
+ * @categoryDescription Loading
+ * When loading a SchemaUnion, the correct subclass will be instantiated based on the discriminator.
```
aeplay, line 43:
```suggestion
 * When loading a SchemaUnion, the correct subclass will be instantiated based on the discriminator.
 * You can narrow the returned instance to a subclass by using `instanceof` like so:
```

aeplay (changes requested):
Great work! Added a couple suggestions to make the changelog/docs as helpful as possible

PR #805, opened 2024-11-19
Website search functionality with Pagefind

pax-k:
- integrated website search functionality using [Pagefind](https://pagefind.app/) and [cmd+k](https://github.com/pacocoursey/cmdk)
- search widget is triggered either by pressing `cmd+k` anywhere on the website, either by clicking "Quick search" displayed in Docs and API Reference
- using `react-singleton-hook` for sharing a singleton hook between multiple components (instead of using context providers)
- tweaked Pagefind indexing by adding custom weights to, or ignoring, specific HTML blocks
- frontmatter support and parsing using `remarkMdxFrontmatter`
- improved SEO titles for some static and dynamic routes

How to run:
```
cd homepage/homepage
pnpm i
pnpm run build
pnpm run start
```

Note: running `pnpm run dev` erases the contents of `.next` directory each time, including built `pagefind/` indexes, so `pnpm run build` has to be used instead


trishalim:
Looks great! 🥳 Thanks a lot @pax-k 

I think it would make more sense to search only in docs, api ref, examples. I can see it also searches from the home page and Jazz Cloud page. 

Some more things I can work on to improve on this (after this PR is merged)
- add a button to open search from mobile
- light mode styling for the modal

@bensleveritt Also consider this when thinking about the structure for API Reference. Here I am searching for `CoMap` but it looks like it's returning the wrong result at first glance. It's not, it's just returning the jazz-tools page where `CoMap` is. But if `CoMap` had its own page then these results would be better. Should we do that or improve the search results instead? I think it's low priority though, but something to keep in mind.
image

bensleveritt:
Yes, I agree not finding the CoMap docs first would be frustrating. To fix it specifically I think we can just weigh the Docs more than the API (https://pagefind.app/docs/weighting/). That said SEO would love more prominent mentions of CoValues, so maybe a bit of that too?

bensleveritt:
Sorry, looking into it more that weighting is for content weight. There might be a way to do page weights, but needs more investigation. In the mean time we'll need headings with the key terms we want to rank up (SEO++)

aeplay:
Closed in favour of #2378

PR #814, opened 2024-11-20
Add minimal Clerk example

bensleveritt:
Just adds a minimal example of Clerk authentication


aeplay (approved):
Nice minimal app that still uses a Jazz hook 👍

PR #827, opened 2024-11-20
fix(coValueState): once a coValue is loaded send the known state to the rest of the peers

gdorsi:


packages/cojson/src/coValueState.ts
```diff
@@ -264,12 +264,19 @@ async function loadCoValueFromPeers(
...
+        action: "load",
+        id: coValueEntry.id,
+        header: false,
+        sessions: {},
+      });
+    } else if (coValueEntry.state.type === "available") {
+      await peer.pushOutgoingMessage({
+        action: "load",
+        ...coValueEntry.state.coValue.knownState(),
+      });
```
gdorsi, line 278: The dependencies should subscribe automatically to the dependencies

PR #842, opened 2024-11-21
feat: pressing cmd+j in a jazz app exports the jazz account to inspector

pax-k:
Usage in a Jazz app:
```
import { listenForCmdJ } from "jazz-tools";
const cleanup = listenForCmdJ();
```

On cmd+J press, redirects the user to `https://inspector.jazz.tools/import/${id}/${base64Secret}`, which in turn knows how to handle the import.

`cmd+J` listener registered for `examples/password-manager`


packages/jazz-tools/src/utils/export-account-inspector.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @category Other
+ **/
+export function listenForCmdJ(localStorageKey?: string) {
+  if (typeof window === "undefined") return;
+
+  const cb = (e: any) => {
+    if (e.metaKey && e.key === "j") {
+      if (
+        confirm(
+          "Are you sure you want to export your account to the inspector?",
```
aeplay:
```suggestion
          "Are you sure you want to inspect your account using inspector.jazz.tools? This lets anyone with the secret inspector URL read your data and impersonate you.",
```

aeplay (changes requested):
Left a suggestion for making the confirmation consequences clearer.

Furthermore

- Please may you put the secret into the hash part of the URL for the inspector? That way it never gets shared with the server hosting the inspector.

- Could you always enable the shortcut if the app's url contains "allowJazzInspector"? That way we can instruct users to enable it even in prod if the devs didn't

pax-k:
> Left a suggestion for making the confirmation consequences clearer.
> 
> Furthermore
> 
> * Please may you put the secret into the hash part of the URL for the inspector? That way it never gets shared with the server hosting the inspector.
> * Could you always enable the shortcut if the app's url contains "allowJazzInspector"? That way we can instruct users to enable it even in prod if the devs didn't

Done.

If a user opens an URL like `http://localhost:5173/#/vault?allowJazzInspector`, then pressing `cmd+j` will show the confirmation and will redirect the user.

aeplay (changes requested):
Thank you!

Sorry, last requests:

- please can you move the shortcut logic / function to jazz-browser? It doesn't really make sense in jazz-tools
- please add a changeset that describes the feature well for the changelog!

packages/jazz-browser/src/utils/export-account-inspector.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Automatically sets up the Cmd+J listener if 'allowJazzInspector' is present in the URL
+ * @returns A cleanup function if the listener was set up, undefined otherwise
+ */
+export function setupInspector() {
+  if (typeof window === "undefined") return;
+
+  const url = new URL(window.location.href);
+  console.log(url.hash);
```
gdorsi:
```suggestion
```

packages/jazz-browser/src/utils/export-account-inspector.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (url.hash.includes("allowJazzInspector")) {
```
gdorsi, line 50: **suggestion:** what about adding `|| process.env.NODE_ENV !== 'production'` to enable this on dev?

gdorsi, line 51: **suggestion:** what about adding `|| process.env.NODE_ENV !== 'production'` to enable this on dev?

gdorsi, line 50:
```suggestion
  if (url.hash.includes("allowJazzInspector") || process.env.NODE_ENV === "development") {
```

**suggestion:** what about enabling the shortcut by default on dev?

gdorsi: cc @anselm

PR #850, opened 2024-11-22
fix: restore offline support and faster load when storage is active

gdorsi:
After some testing I've noticed that Jazz was generally faster with a local sync server even when the data was fully stored locally.

After some more debugging I've found where the extra waiting was and that we weren't operating correctly when the sync server is down.

This PR fixes the issue

### DIFF
--- a/packages/cojson/src/coValueState.ts
+++ b/packages/cojson/src/coValueState.ts
@@ -4,6 +4,7 @@ import { RawCoID } from "./ids.js";
 import { PeerID } from "./sync.js";
 
 export const CO_VALUE_LOADING_MAX_RETRIES = 5;
+export const CO_VALUE_LOADING_TIMEOUT = 5000;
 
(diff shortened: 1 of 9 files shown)


packages/cojson/src/coValueState.ts
```diff
@@ -264,13 +260,27 @@ async function loadCoValueFromPeers(
...
+        });
+      }
+    }, 3000);
+
     if (coValueEntry.state.type === "available") {
-      await peer.pushOutgoingMessage({
+      void peer.pushOutgoingMessage({
         action: "load",
         ...coValueEntry.state.coValue.knownState(),
       });
```
gdorsi, line 281: We don't need to wait for the message to be delivered here.

This way when the coValue becomes available because it's cached we don't wait for the websocket peer to consume the messages queue before moving forward.

PR #856, opened 2024-11-22
API Reference: typography and spacing adjustments

### LINKED ISSUE #830: Add padding to the scrollTo target in API docs
We're losing a line or two to the header when scrolled to.

### LINKED ISSUE #832: API sections should be more differentiated from the content
It's hard to parse the sections from the annotation blocks in the api docs, we should bump the significance of the section headers

trishalim:
fixes #830, fixes #832


bensleveritt (approved):
Nice, thanks

PR #867, opened 2024-11-22
Make coming soon section 4-col

trishalim:


homepage/homepage/components/home/ComingSoonSection.tsx
```diff
@@ -15,39 +15,45 @@ export function ComingSoonSection() {
     <div>
       <SectionHeader title="More features coming soon" />
 
-      <GappedGrid>
-        <GridCard>
-          <H3>Cursors & carets</H3>
-          <P className="text-lg">Ready-made spatial presence.</P>
+      <GappedGrid cols={4}>
+        <Card className="p-4">
+          <h3 className="text-stone-900 font-medium md:text-base dark:text-stone-100 mb-2">
```
bensleveritt: Should we capture this heading style into a component? If only to keep track of what styles we're using

bensleveritt (approved):
LGTM

homepage/homepage/components/home/ComingSoonSection.tsx
trishalim: agreed

trishalim: I placed in inside a component `FeatureCard`. I don't want to change our `H3` component. Styling based on headings works for article content, but feels very limiting on a landing page.

PR #873, opened 2024-11-25
Refactor IDB to have a better control over the flow and reduce amount of messages

marinoska:
**Refactored IDB package**
- Split into idbClient, idbNode, syncManager
- Removed promise hell

**Bugs**
- Fixed issue with duplicating dependencies during deeper recursive iterations, a test added

**Optimised**
- Removed replying to peers known messages

**Covered with tests**
- All branches as on the schema below
Screenshot 2024-11-29 at 09 15 57


packages/cojson-storage-indexeddb/src/index.ts
```diff
@@ -1,733 +1 @@
...
-            const dependedOnCoValues = getDependedOnCoValues(
-              coValueRow,
-              newContentPieces,
-              theirKnown,
-            );
-
-            return SyncPromise.all(
-              dependedOnCoValues.map((dependedOnCoValue) =>
-                this.sendNewContentAfter(
-                  {
```
marinoska, line 370: Here we could send the same coValue a few times while looping through the the dependencies recursively

packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -0,0 +1,234 @@
...
+          transactions: IDBObjectStore;
+          signatureAfter: IDBObjectStore;
+        };
+      }) => {
+        const request = handler(stores);
+        request.onerror = () => {
+          console.error("Error in request", request.error);
+          this.currentTx = undefined;
+          reject(request.error);
+          // TODO: recover pending requests in new tx
```
gdorsi: **suggestion:** would it be possible to provide more context for this TODO?

packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -0,0 +1,234 @@
...
+            next({ stores });
+          } else {
+            if (this.currentTx === txEntry) {
+              this.currentTx = undefined;
+            }
+          }
+        };
+      };
+
+      if (!txEntry || performance.now() - txEntry.startedAt > 20) {
```
gdorsi, line 111: **suggestion:** would it be possible to document why we are doing the performance.now check?

packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -0,0 +1,234 @@
...
+    );
+  }
+
+  async addCoValue(
+    msg: CojsonInternalTypes.NewContentMessage,
+  ): Promise<number> {
+    return (await this.makeRequest<IDBValidKey>(({ coValues }) =>
+      coValues.put({
+        id: msg.id,
+        header: msg.header!,
```
gdorsi, line 189: suggestion: we can cast msg.header to Boolean so we can be sure of the final type

packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  addSignatureAfter({
+    sessionRowID,
+    idx,
+    signature,
+  }: { sessionRowID: number; idx: number; signature: Signature }) {
+    return this.makeRequest(({ signatureAfter }) =>
+      signatureAfter.put({
+        ses: sessionRowID,
+        // TODO: newLastIdx is a misnomer, it's actually more like nextIdx or length
```
gdorsi: **question:** where is newLastIdx?

packages/cojson-storage-indexeddb/src/syncManager.ts
```diff
@@ -0,0 +1,333 @@
...
+    theirKnown,
+    ourKnown,
+    newContentPieces,
+  }: {
+    sessionRow: StoredSessionRow;
+    theirKnown: CojsonInternalTypes.CoValueKnownState;
+    ourKnown: CojsonInternalTypes.CoValueKnownState;
+    newContentPieces: CojsonInternalTypes.NewContentMessage[];
+  }) {
+    ourKnown.sessions[sessionRow.sessionID] = sessionRow.lastIdx;
```
gdorsi: **suggestion:** could you add comments on the steps inside this function?

Can't understand it only by reading the code

packages/cojson-storage-indexeddb/src/syncManager.ts
```diff
@@ -0,0 +1,333 @@
...
+
+      const contentMessages = contentMessageMap[coId as RawCoID] || [];
+      contentMessages.forEach((msg) => this.sendStateMessage(msg));
+    });
+  }
+
+  private async collectCoValueData(
+    coValueKnownState: CojsonInternalTypes.CoValueKnownState,
+    knownMessageMap: Record<RawCoID, KnownStateMessage> = {},
+    contentMessageMap: Record<RawCoID, NewContentMessage[]> = {},
```
gdorsi: **question:** why are we using two maps instead of lists here?

From what I see a list would be better suited for the job.

Something like
```ts
type ListEntry = { id: RawCoID, knownMessage: KnownStateMessage, contentMessages: NewContentMessage[] };
```

New entries would always be added at the top.
If we need to dedupe we can use a support map.

packages/cojson-storage-indexeddb/src/syncManager.ts
```diff
@@ -0,0 +1,333 @@
...
+    await Promise.all(
+      allCoValueSessions.map((sessionRow) =>
+        this.handleSessionUpdate({
+          sessionRow,
+          theirKnown: coValueKnownState,
+          ourKnown: newCoValueKnownState,
+          newContentPieces: contentMessages,
+        }),
+      ),
+    );
```
gdorsi, line 150: **suggestion:** could you describe in a comment what's happening here?

packages/cojson-storage-indexeddb/src/syncManager.ts
```diff
@@ -0,0 +1,333 @@
...
+
+  handleLoad(msg: CojsonInternalTypes.LoadMessage) {
+    return this.sendNewContent(msg);
+  }
+
+  async handleContent(
+    msg: CojsonInternalTypes.NewContentMessage,
+  ): Promise<void | unknown> {
+    const coValueRow = await this.idbClient.getCoValue(msg.id);
+    if (!msg.header && !coValueRow) {
```
gdorsi: **suggestion:** could you describe which scenario are we handling here?

packages/cojson-storage-indexeddb/src/syncManager.ts
```diff
@@ -0,0 +1,333 @@
...
+            nextIdx + i,
+            newTransaction,
+          );
+        }),
+      ),
+    );
+  }
+
+  handleKnown(_msg: KnownStateMessage) {
+    // No need to process known messages from the local node as IDB storage can't be updated by another peer
```
gdorsi: **praise:** thanks for the context!

gdorsi (commented):
Looks great!
Thanks for this refactoring!

gdorsi:
Pets tests are failing, could you check if it's a real regression?

packages/cojson-storage-indexeddb/src/idbClient.ts
marinoska: This is a leftover, I didn't rewrite it, just reorganised mostly. I wasn't dare to remove it, probably :D But I'll do then 👍

packages/cojson-storage-indexeddb/src/idbClient.ts
marinoska: Honestly I didn't change anything in that function as it's out of our focus. Just left it as it is as it works and IDB is not our priority. Wdyt?

packages/cojson-storage-indexeddb/src/idbClient.ts
marinoska: A leftover as well. I did no change to makeRequest fn. Will remove it

packages/cojson-storage-indexeddb/src/syncManager.ts
marinoska: Like I mentioned I didn't rewrite all the code. If we want to make it more readable I need to work on the whole thing more.  No problem. And it'd be better than to write a comment after every other line. This is not the only place that is not clear from the first sight. So I can leave a comment or make just the code better

packages/cojson-storage-indexeddb/src/syncManager.ts
marinoska: It can be smth like this, I can change it

packages/cojson-storage-indexeddb/src/idbClient.ts
aeplay: What this implements is "transaction batching" which generally helps with IndexedDB performance. I'm not sure if we're doing it the best way but it would be good to at least comment that this is what we're doing.

packages/cojson-storage-indexeddb/src/syncManager.ts
aeplay: It's not precisely correct: two tabs can write to the same IndexedDB. But we don't intend to use IndexedDB itself as a synchronisation mechanism, so we can ignore the known messages. @marinoska please may you change the comment accordingly?

aeplay (changes requested):
Just a couple small requests.

Monumental work! Love the precise tests and how you set them up!

packages/cojson-storage-indexeddb/src/syncManager.ts
marinoska: When we were sent some content without a header for the unknown value. Then it sends back a correction.
I thought this piece is quite self-descriptive but can you see smth wrong in there? I can add a comment of course

gdorsi: Now that you explain it is more obvious, but won't hurt to add a comment to connect the missing header with the need of a correction.

packages/cojson-storage-indexeddb/src/idbClient.ts
marinoska: Added a comment

packages/cojson-storage-indexeddb/src/idbClient.ts
marinoska: Header is of type CoValueHeader|undefined and we save it as it is (no boolean).
This code l.182 had to narrow types as I thought 

```
   if (!msg.header) {
      throw new Error("Header is required, coId: " + msg.id);
    }
```
    
But TS still says  it could be undefined so I had to make it explicit `msg.header!`

packages/cojson-storage-indexeddb/src/syncManager.ts
marinoska: Made it a bit more readable and added a comment

marinoska: Took out the l.141 from `this.handleSessionUpdate` to make it neater

packages/cojson-storage-indexeddb/src/syncManager.ts
```diff
@@ -0,0 +1,339 @@
...
+      this.sendStateMessage(knownMessage);
+
+      contentMessages?.length &&
+        contentMessages.forEach((msg) => this.sendStateMessage(msg));
+    });
+  }
+
+  private async collectCoValueData(
+    peerKnownState: CojsonInternalTypes.CoValueKnownState,
+    messageMap: OutputMessageMap = {},
```
marinoska, line 100: Replaced knownMap and contentMap with a single messageMap which looks nicer.

packages/cojson-storage-indexeddb/src/syncManager.ts
```diff
@@ -0,0 +1,339 @@
...
+
+  async handleContent(
+    msg: CojsonInternalTypes.NewContentMessage,
+  ): Promise<void | unknown> {
+    const coValueRow = await this.idbClient.getCoValue(msg.id);
+
+    // We have no info about coValue header
+    const invalidAssumptionOnHeaderPresence = !msg.header && !coValueRow;
+
+    if (invalidAssumptionOnHeaderPresence) {
```
marinoska, line 196: Made this piece more self-descriptive and added a comment

marinoska:
@gdorsi I removed leftover Todos, added some commentaries and made a few places a bit readable, I hope - my comments inline.

packages/cojson-storage-indexeddb/src/syncManager.ts
```diff
@@ -0,0 +1,339 @@
...
+            nextIdx + i,
+            newTransaction,
+          );
+        }),
+      ),
+    );
+  }
+
+  handleKnown(_msg: KnownStateMessage) {
+    // We don't intend to use IndexedDB itself as a synchronisation mechanism, so we can ignore the known messages
```
marinoska, line 324: @aeplay the comment you suggested

PR #877, opened 2024-11-25
Bring CoPlainText up to current

bensleveritt:
Adds `CoPlainText` and `CoRichText` CoValues to Jazz-tools, finishing off work from @aeplay in preparation for supporting rich text editors.

`CoPlainText` extends `CoList` adding support for common string operations (insert, remove), while `CoRichText` is a `CoMap` that contains a text string (`CoPlainText`), and a list of marks indicating rich text markup ranges.


packages/cojson/src/coValues/account.ts
```diff
@@ -103,6 +104,36 @@ export class RawControlledAccount<Meta extends AccountMeta = AccountMeta>
...
+        ...this.core.crypto.createdNowUnique(),
+      })
+      .getCurrentContent() as T;
+
+    if (init) {
+      text.insertAfter(0, init, initPrivacy);
+    }
+
+    return text;
+  }
```
gdorsi: The create methods are defined on `group` 

Since account extends group we don't need to add this method here because it should be inherited

packages/cojson/src/coValues/coPlainText.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+type PlaintextIdxMapping = {
+  opIDbeforeIdx: OpID[];
+  opIDafterIdx: OpID[];
+  idxAfterOpID: { [opID: StringifiedOpID]: number };
+  idxBeforeOpID: { [opID: StringifiedOpID]: number };
+};
+
+const segmenter = new Intl.Segmenter("en", { granularity: "grapheme" });
```
gdorsi: [The baseline for this API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) is only 2024 and it isn't supported by React Native.

We should move this evaluation inside `RawCoPlainText`, so the polyfill/fallback will be required only when using `RawCoPlainText`

packages/cojson/src/coValues/coPlainText.ts
```diff
@@ -0,0 +1,136 @@
...
+
+export class RawCoPlainText<
+  Meta extends JsonObject | null = JsonObject | null,
+> extends RawCoList<string, Meta> {
```
gdorsi, line 25: IMHO we shouldn't rely on inheritance here.

`RawCoList` is already quite complex, extending it makes `RawCoPlainText` practically unmaintainable and harder to use.

What if we move the shared bits into a separate class and reuse it across the two primitives?

gdorsi (changes requested):
Added some changes requests.

Also, since the new API on `jazz-tools` aren't working yet, what about moving them on a separate PR so we can merge only the changes to `cojson`?

packages/cojson/src/coValues/coPlainText.ts
aeplay: As far as I remember, RawCoPlainText uses most of the functionality of RawCoList. RawCoList just _is_ complex. And RawCoPlainText really is a list of characters/graphemes. That's why I originally just extended it.

If there is a common core that can be extracted out to make everything cleaner, I'm happy for that - but in the interest of moving this PR forward I'd suggest doing it in a separate PR.

examples/richtext/job-template.nomad
```diff
@@ -0,0 +1,56 @@
+job "chat$BRANCH_SUFFIX" {
```
aeplay: We don't need this file anymore, examples are now deployed to Vercel

aeplay:
> since the new API on `jazz-tools` aren't working yet

What do you mean by this @gdorsi?

examples/richtext/job-template.nomad
bensleveritt: Removed

packages/cojson/src/coValues/account.ts
bensleveritt: Removed

gdorsi:
> > since the new API on `jazz-tools` aren't working yet
> 
> What do you mean by this @gdorsi?

From the PR description:
> For now, we should consider CoRichText to be experimental, this PR only seeks to add support to the sync server, while rich text editor support is still being developed. 

To me this means that only the changes inside `cojson` are functional, while we are still figuring out how to develop the `jazz-tools` API to best integrate with the rich text editors.

Since the goal is to release the cojson changes so simplify the testing process, I'm proposing to merge only the cojson part

bensleveritt:
> > > since the new API on `jazz-tools` aren't working yet
> > 
> > 
> > What do you mean by this @gdorsi?
> 
> From the PR description:
> 
> > For now, we should consider CoRichText to be experimental, this PR only seeks to add support to the sync server, while rich text editor support is still being developed.
> 
> To me this means that only the changes inside `cojson` are functional, while we are still figuring out how to develop the `jazz-tools` API to best integrate with the rich text editors.
> 
> Since the goal is to release the cojson changes so simplify the testing process, I'm proposing to merge only the cojson part

Yeah, I can split the PR into two parts to allow the cojson part to be merged before the prosemirror support

packages/cojson/src/coValues/coPlainText.ts
bensleveritt: Moved into class

packages/cojson/src/coValues/coPlainText.ts
bensleveritt: Created a follow up task https://linear.app/garden-computing/issue/JAZZ-629/extract-common-core-from-rawcolist-and-rawcoplaintext

packages/cojson/src/coValues/coPlainText.ts
```diff
@@ -0,0 +1,145 @@
...
+    NonNullable<typeof this._cachedEntries>,
+    PlaintextIdxMapping
+  >;
+
+  constructor(core: CoValueCore) {
+    super(core);
+    this._cachedMapping = new WeakMap();
+    if (!Intl.Segmenter) {
+      throw new Error(
+        "Intl.Segmenter is not supported for this compilation target. Use a polyfill to get coPlainText support in Jazz. (eg. https://formatjs.github.io/docs/polyfills/intl-segmenter/)",
```
gdorsi:
```suggestion
        "Intl.Segmenter is not supported. Use a polyfill to get coPlainText support in Jazz. (eg. https://formatjs.github.io/docs/polyfills/intl-segmenter/)",
```

packages/cojson/src/tests/coPlainText.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+test("should throw on creation if Intl.Segmenter is not available", () => {
+  vi.stubGlobal("Intl", {
+    Segmenter: undefined,
+  });
+
+  const node = new LocalNode(...randomAnonymousAccountAndSessionID(), Crypto);
+  const group = node.createGroup();
+  expect(() => group.createPlainText()).toThrow(
+    "Intl.Segmenter is not supported for this compilation target. Use a polyfill to get coPlainText support in Jazz. (eg. https://formatjs.github.io/docs/polyfills/intl-segmenter/)",
```
gdorsi:
```suggestion
    "Intl.Segmenter is not supported. Use a polyfill to get coPlainText support in Jazz. (eg. https://formatjs.github.io/docs/polyfills/intl-segmenter/)",
```

gdorsi (approved):
LGTM

PR #878, opened 2024-11-26
Fix console errors on Button component

trishalim:
There was an error because I was passing attributes to `<button>` that aren't valid


bensleveritt (approved):
Looks great, nice one

PR #879, opened 2024-11-26
Check for clerk.user before rendering the app

sventies:
The chat-clerk app hangs on load (if not yet authenticated): https://chat-clerk-demo.jazz.tools/
This is because it tries to render the App before actually being authenticated. 

Changes: 
Check if we're actually authed before rendering children (App component). This is done properly in the docs so I could simply copy from there:
https://jazz.tools/docs/authentication/auth-methods#clerk


aeplay:
Thank you very much @sventies!

PR #890, opened 2024-11-27
fix: reset observable value when unsubscribed

pax-k:


packages/jazz-react/src/index.tsx
```diff
@@ -199,20 +199,45 @@ export function createJazzReactApp<Acc extends Account>({
...
+          return () => {
+            unsubscribe();
+          };
         },
         [Schema, id, context],
       ),
-      () => observable.getCurrentValue(),
-      () => observable.getCurrentValue(),
+      () => resolvedValue,
+      () => resolvedValue,
```
gdorsi: suggestion: I think that would be better to fix the issue inside "observable" and return `undefined` when `observable.getCurrentValue()` is called when there are no subscribers.

This way we won't need to rely on additional states outside `useSyncExternalStore`

gdorsi (changes requested):
Would prefer to go on a simpler route and tweak [the observable](https://github.com/garden-co/jazz/blob/main/packages/jazz-tools/src/coValues/interfaces.ts#L252) to set currentValue to `undefined` when `unsubscribe` is called.

Should also be easier to cover with a test.

gdorsi (approved):
LGTM

Thanks!

pax-k:
Tests fail in GH Actions because of some circular imports already present in the codebase. We will fix them at a later time.

Here is the log for locally ran tests: 
```
❯ pnpm run test -t "createCoValueObservable"

> jazz-tools@0.8.34 test /Users/pax/Documents/Repos/jazz/packages/jazz-tools
> vitest --run --root ../../ --project jazz-tools "-t" "createCoValueObservable"


 RUN  v1.5.3 /Users/pax/Documents/Repos/jazz

stdout | src/tests/coMap.test.ts:40:11
TestMap schema undefined

 ↓ |jazz-tools| src/lib/cache.test.ts (3) [skipped]
 ↓ |jazz-tools| src/tests/schemaUnion.test.ts (2) [skipped]
 ↓ |jazz-tools| src/tests/schema.test.ts (12) [skipped]
 ↓ |jazz-tools| src/tests/groupsAndAccounts.test.ts (1) [skipped]
 ↓ |jazz-tools| src/tests/subscribe.test.ts (5) [skipped]
 ↓ |jazz-tools| src/tests/coList.test.ts (10) [skipped]
 ↓ |jazz-tools| src/tests/deepLoading.test.ts (3) [skipped]
 ✓ |jazz-tools| src/tests/createCoValueObservable.test.ts (3)
 ↓ |jazz-tools| src/tests/coFeed.test.ts (12) [skipped]
 ↓ |jazz-tools| src/tests/coMap.test.ts (26) [skipped]

 Test Files  1 passed | 9 skipped (10)
      Tests  3 passed | 74 skipped (77)
```

PR #900, opened 2024-11-28
Content deadlock fix with client <-> edge <-> core setup

aeplay:


marinoska (approved):

packages/cojson/src/sync.ts
```diff
@@ -403,27 +403,39 @@ export class SyncManager {
     }
 
     if (entry.state.type === "loading") {
-      const value = await entry.getCoValue();
```
marinoska, line 406: Probably just removing `await` keyword would be enough :)

aeplay: ?? it wouldn't, everything in the then() still needs to happen after the covalue is loaded. (The diff view is very confusing unfortunately)

marinoska: Oh sorry you're right!

gdorsi (approved):

gdorsi:
The test is failing also on main, so don't worry about it

PR #919, opened 2024-12-03
Refactor SQLite storage

marinoska:
- Refactor SQLite storage in order to embrace the recent changes to IDB storage
- Create `cojson-storage` package to accommodate synchronization logic common to SQLite and IDB storage
- Refactor SQLite and IDB storage in order to move SyncManager to the newly created `cojson-storage` package
- Make IDBClient and SQLiteClient implement the shared interface DBClientInterface


packages/cojson-storage/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "cojson-storage",
+  "version": "0.8.35",
```
marinoska, line 3: Not sure what should I state here for a new package

gdorsi: You made the right call to set the same version as the other packages.

I think it makes easier to manage dependencies this way.

gdorsi (approved):
LGTM

Very clean!

aeplay (changes requested):
Looks very good, thanks @marinoska !

~~Small request: can we call the shared package `cojson-storage-relational`?~~

marinoska:
> Looks very good, thanks @marinoska !
> 
> Small request: can we call the shared package `cojson-storage-relational`?

No problem! But why this name?

aeplay (approved):

PR #922, opened 2024-12-04
Explicitly reference homepage dependencies

### LINKED ISSUE #921: Explicitly reference homepage dependencies
The Homepage project is dependant on the jazz-\* packages, but `turbo` and `pnpm` are not aware. The dependencies can be modelled with `"jazz-react": "workspace:../*"` etc.

bensleveritt:
As title.

This should allow us to drop the custom build command in Vercel, and reduce build times (as only dependent packages are being built, rather than everything)


bensleveritt:
@coderabbitai configuration

aeplay (approved):

PR #927, opened 2024-12-04
Optimise large record-like CoMaps for access of latest value

aeplay:


packages/cojson/src/coValues/coMap.ts
```diff
@@ -39,44 +40,107 @@ export class RawCoMapView<
...
     this.core = core;
-    this.ops = {};
+    this.latest = {};
+    const latest = this.latest;
+    this.latestTxMadeAt = 0;
+    this.options = options;
+
+    const validTransactions = core.getValidSortedTransactions(options);
+
+    for (let i = validTransactions.length - 1; i >= 0; i--) {
```
aeplay: By default, we only collect the latest change for each key now. We iterate backwards over transactions because then we only need to set the entry for each key once (since they are sorted by madeAt already)

packages/cojson/src/coValues/coMap.ts
```diff
@@ -39,44 +40,107 @@ export class RawCoMapView<
...
+          entry.madeAt = madeAt;
+          entry.changeIdx = changeIdx;
+          entry.change = change;
+        }
+      }
+    }
+  }
+
+  private getOps() {
+    if (this.cachedOps) {
```
aeplay, line 115: For the cases where we do need the full history, we get it lazily (and cache it)

packages/cojson/src/coValues/coMap.ts
```diff
@@ -91,13 +155,17 @@ export class RawCoMapView<
...
   atTime(time: number): this {
-    const clone = Object.create(this) as this;
-    clone.id = this.id;
-    clone.type = this.type;
-    clone.core = this.core;
-    clone.ops = this.ops;
-    clone.atTimeFilter = time;
-    return clone;
+    if (time >= this.latestTxMadeAt) {
+      return this;
```
aeplay, line 165: Optimisation for atTime where atTime is larger than all transactions in this CoMap. This especially helps with Groups where we often call atTime when determining validity of transactions in owned CoValues

packages/cojson/src/coValues/coMap.ts
```diff
@@ -91,13 +153,17 @@ export class RawCoMapView<
...
+    } else {
+      const clone = new RawCoMapView(this.core, {
+        ignorePrivateTransactions:
+          this.options?.ignorePrivateTransactions ?? false,
+        atTime: time,
+      });
+      Object.setPrototypeOf(clone, this);
```
gdorsi, line 164: **question:** could you explain this line?

PR #935, opened 2024-12-04
Add depth fallback

bensleveritt:


gdorsi (approved):
Do you think that would be possible to cover these changes with some tests?

Even adding an e2e test on the svelte app would be ok

bensleveritt:
> Do you think that would be possible to cover these changes with some tests?
> 
> Even adding an e2e test on the svelte app would be ok

Yes, you're right

gdorsi (approved):
LGTM

Thanks for the tests!

PR #942, opened 2024-12-05
Add linting to cojson-storage-indexeddb

bensleveritt:


packages/cojson-storage-indexeddb/src/syncManager.ts
```diff
@@ -87,12 +87,14 @@ export class SyncManager {
 
     // reverse it to send the top level id the last in the order
     const collectedMessages = Object.values(outputMessages).reverse();
-    collectedMessages.forEach(({ knownMessage, contentMessages }) => {
+    for (const { knownMessage, contentMessages } of collectedMessages) {
```
marinoska: I haven't used a `for` loop in a while. I mostly use methods like map, filter, reduce, and forEach, which I thought was the "industry standard" by my own personal experience. 
I see the `for` loops have its pros when you need to await within a loop iteration and it's faster on large amounts. It's more talkative and the syntax is heavier imho. Do we have a reason to use this rule?

marinoska (approved):
Looks nice! Approved with a comment

packages/cojson-storage-indexeddb/src/syncManager.ts
aeplay: I think it's a style question mostly. I and the rest of the codebase typically use array methods for everything except forEach - so just for consistency's sake let's follow the rule

packages/cojson-storage-indexeddb/src/syncManager.ts
```diff
@@ -87,12 +87,14 @@ export class SyncManager {
...
-
-      contentMessages?.length &&
-        contentMessages.forEach((msg) => this.sendStateMessage(msg));
-    });
+      if (contentMessages?.length) {
+        for (const msg of contentMessages) {
+          this.sendStateMessage(msg);
+        }
+      }
+    }
```
marinoska: This one as well - two nested blocks vs zero and necessity to declare a variable another time. Of course this is about coding style which is a personal thing. Some might not like `&&` but still - `for ... of` is something I would't expect in JS a lot

packages/cojson-storage-indexeddb/src/syncManager.ts
marinoska: But who makes the rule? :) That was my question - do we have a reason to use this rule? Other way around - forcing to use `for...of` makes it not consistent discriminating one of the array methods and allowing the others imho.

aeplay: It's a default rule of Biome, our new formatter & linter.

The rule description has some arguments: https://biomejs.dev/linter/rules/no-for-each/

I would strongly trust their judgement and not deviate from defaults 🤷‍♂️

marinoska: It's arguable. I could extend their arguments to all array methods and go ES5, it’s a very unusual rule to me. But it's just a discussion, after all, I approved the PR in the first place.

bensleveritt: Yeah for me, I know it's easy to get bogged down in the dogma and personal preference, and just have to trust that 'one codebase, written as one' is best in the long run. Happy to discuss rules, but imo if Biome (or any linter for that matter) put a rule into their recommendations, they did so with good reason (and hopefully good arguments).

For me, I'm a big functional fan, but the for ... of argument is very compelling and I've seen a number of (fairly public) people switch to it from .forEach

PR #947, opened 2024-12-05
Move vue project setup to guide

trishalim:


trishalim:
This is not important

PR #950, opened 2024-12-05
feat: add the APIs to wait for a sync state

gdorsi:
Adds two new APIs to wait for the sync to be completed.

Examples:
```ts
const chat = Chat.create([], { owner: me });

await chat.waitForSync(); // Waits for the chat to be uploaded
await me.waitForAllCoValuesSync(); // Waits all the available coValues to be uploaded
```


packages/cojson/src/SyncStateManager.ts
```diff
@@ -134,8 +152,3 @@ function getIsUploadCompleted(
 
   return true;
 }
-
-function simpleMemoize<T>(fn: () => T): () => T {
```
gdorsi, line 138: Had to remove the memoization because was creating some issues in the subscribe tests

packages/cojson/src/sync.ts
```diff
@@ -117,10 +117,10 @@ export class SyncManager {
 
   constructor(local: LocalNode) {
     this.local = local;
-    this.syncStateSubscriptionManager = new SyncStateSubscriptionManager(this);
+    this.syncState = new SyncStateManager(this);
```
gdorsi, line 120: Renamed to make it easier to read and access

packages/cojson/src/sync.ts
```diff
@@ -708,35 +708,61 @@ export class SyncManager {
...
+      }, timeout);
     });
   }
 
+  async waitForSync(id: RawCoID, timeout = 30_000) {
+    const peers = this.getPeers();
+
+    return Promise.all(
+      peers
+        .filter((peer) => peer.role !== "storage")
```
gdorsi: Removed the storage peers because we are not getting an ACK for the sent content.

@marinoska @aeplay 

Maybe we could set the storage peers to not have an "optimistic" state but to consider the "optimistic" state definitive.

WDYT?

packages/jazz-run/src/createWorkerAccount.ts
```diff
@@ -45,16 +45,7 @@ export const createWorkerAccount = async ({
...
-    ]),
-    failAfter(
-      4_000,
-      "Timeout: Didn't manage to upload the account and profile",
-    ),
-  ]);
+  await account.waitForAllCoValuesSync({ timeout: 4_000 });
 
   // Spawn a second peer to double check that the account is fully synced
   const peer2 = createWebSocketPeer({
```
gdorsi: @aeplay 

We can get rid of this double now.
WDYT?

packages/cojson/src/sync.ts
gdorsi: Fixed by changing PeerState to share the same instance of "optimisticKnownState" and "knownState" when the role is Peer.

packages/cojson/src/PeerState.ts
```diff
@@ -14,7 +14,14 @@ export class PeerState {
     knownStates: PeerKnownStates | undefined,
   ) {
     this.optimisticKnownStates = knownStates?.clone() ?? new PeerKnownStates();
-    this.knownStates = knownStates?.clone() ?? new PeerKnownStates();
+
+    // We assume that exchanges with storage peers are always successful
```
aeplay, line 18: I think this is OK!

aeplay (approved):
LGTM, I think only thinking about upload is the right thing

PR #955, opened 2024-12-06
chore: add an up-down counter to track the amount of connected peers

Elfo404:
Adds `@opentelemetry/api` as a dependency and adds an UpDownCounter to track the number of connected peers by role.
This should have no effect on the browser aside from that package being included in the bundle.

### DIFF
--- a/packages/cojson/src/sync.ts
+++ b/packages/cojson/src/sync.ts
@@ -1,3 +1,4 @@
+import { ValueType, metrics } from "@opentelemetry/api";
 import { PeerState } from "./PeerState.js";
 import { SyncStateManager } from "./SyncStateManager.js";
 import { CoValueHeader, Transaction } from "./coValueCore.js";
@@ -114,6 +115,12 @@ export class SyncManager {
       | { done: Promise<void>; nRequestsThisTick: number }
       | undefined;
   } = {};
+
+  peersCounter = metrics.getMeter("cojson").createUpDownCounter("jazz.peers", {
+    description: "Amount of connected peers",
+    valueType: ValueType.INT,
+    unit: "peer",
+  });
 
   constructor(local: LocalNode) {
     this.local = local;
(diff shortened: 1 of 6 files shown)


aeplay:
Wondering if instead we should add event listeners for "peer added", "peer removed" (and whatever we want to instrument in the future) to LocalNode instead? That way we could keep the open telemetry dep only where needed.

cc @gdorsi

aeplay (changes requested):
@Elfo404 please may you remove / add to gitignore the .direnv files?

packages/cojson/src/sync.ts
```diff
@@ -115,6 +116,12 @@ export class SyncManager {
       | undefined;
   } = {};
 
+  peersCounter = metrics.getMeter("cojson").createUpDownCounter("jazz.peers", {
+    description: "Amount of connected peers",
+    valueType: ValueType.INT,
+    unit: "peer",
+  });
```
gdorsi, line 123: suggestion: what about creating a metrics adapter?

Something like:
```ts
class Metrics {
    peersCounter = metrics.getMeter("cojson").createUpDownCounter("jazz.peers", {
      description: "Amount of connected peers",
      valueType: ValueType.INT,
      unit: "peer",
    });
    
    trackAddPeer(peerRole: Peer['role']) {
       this.peersCounter.add(1, { role: peer.role });
    }
    
    trackRemovePeer(peerRole: Peer['role']) {
       this.peersCounter.add(-1, { role: peer.role });
    }
}

export const metrics = new Metrics();
```

Should make it easier to write tests and makes the tracking calls cleaner.

gdorsi (changes requested):
Could you add some tests?

packages/cojson/src/sync.ts
Elfo404: saw this a bit too late, but i think it's worth exploring. I'll experiment with it and follow up if that's ok

gdorsi (approved):

PR #965, opened 2024-12-07
perf(CoMap): optimize the atTime processing

gdorsi:


packages/cojson/src/coValues/coMap.ts
```diff
@@ -49,6 +49,9 @@ export class RawCoMapView<
   cachedOps?: {
     [Key in keyof Shape & string]?: MapOp<Key, Shape[Key]>[];
   };
+  /** @internal */
+  validSortedTransactions?: DecryptedTransaction[];
```
gdorsi, line 53: This can also be cached, guess that's also where most of the overhead comes from.

packages/cojson/src/coValues/coMap.ts
```diff
@@ -59,27 +62,58 @@ export class RawCoMapView<
...
   constructor(
     core: CoValueCore,
-    options?: { ignorePrivateTransactions: boolean; atTime?: number },
+    options?: {
+      ignorePrivateTransactions: boolean;
+      atTime?: number;
+      validSortedTransactions?: DecryptedTransaction[];
+      cachedOps?: {
+        [Key in keyof Shape & string]?: MapOp<Key, Shape[Key]>[];
+      };
```
gdorsi, line 71: Added the cached values to carry over them when cloning on "atTime"

packages/cojson/src/coValues/coMap.ts
```diff
@@ -182,10 +214,10 @@ export class RawCoMapView<
       return undefined;
     }
 
-    if (this.atTimeFilter) {
-      return this.getOps()[key]?.filter(
-        (op) => op.madeAt <= this.atTimeFilter!,
-      );
+    const atTimeFilter = this.options?.atTime;
```
gdorsi, line 222: Fixed the atTimeFilter to read from options

packages/cojson/src/coValues/coMap.ts
```diff
@@ -197,17 +229,13 @@ export class RawCoMapView<
...
-      if (!latest) {
-        return undefined;
-      }
+      const latestChange = this.latest[key];
 
-      const includeUntil = this.atTimeFilter;
-      const lastEntry = includeUntil
-        ? this.getOps()[key]?.findLast((entry) => entry.madeAt <= includeUntil)
-        : latest!;
+      if (!latestChange) {
```
gdorsi, line 239: Optimized to use `latest` since it's already processed considering the "atTime" filter

packages/cojson/src/coValues/coMap.ts
```diff
@@ -248,7 +269,7 @@ export class RawCoMapView<
       [K in keyof Shape & string]: Shape[K];
     }> = {};
 
-    for (const key of this.keys()) {
+    for (const key of Object.keys(this.latest) as (keyof Shape & string)[]) {
```
gdorsi, line 277: Using `Object.keys` instead of `this.keys` because we don't need to filter them by last operation since we use `this.get` internally

packages/cojson/src/coValues/coMap.ts
```diff
@@ -268,34 +289,21 @@ export class RawCoMapView<
...
-        by: RawAccountID | AgentID;
-        tx: TransactionID;
-        at: Date;
-        value?: Shape[K];
-      }
-    | undefined {
-    const ops = this.timeFilteredOps(key);
-    if (!ops || ops.length <= n) {
+  nthEditAt<K extends keyof Shape & string>(key: K, n: number) {
+    const ops = this.getOps()[key];
```
gdorsi, line 298: Switched to use `this.getOps` to skip the filtering operation.

We can check the "atTimeFilter" here

packages/cojson/src/coValues/coMap.ts
```diff
@@ -310,10 +318,13 @@ export class RawCoMapView<
...
     const ops = this.timeFilteredOps(key);
-    if (!ops || ops.length === 0) {
+    const lastEntry = ops?.[ops.length - 1];
+
+    if (!lastEntry) {
       return undefined;
     }
-    return this.nthEditAt(key, ops.length - 1);
+
+    return operationToEditEntry(lastEntry);
```
gdorsi, line 332: Avoided to call `nthEditAt` since we already have the entry that we want to return.

This can be further optimized by removing `timeFilteredOps` but didn't want to complicate this function too much.

PR #975, opened 2024-12-09
chore: Add message queue size metric

Elfo404:
Adds an updown counter to track the size of the message queue for each priority.

### DIFF
--- a/packages/cojson/package.json
+++ b/packages/cojson/package.json
@@ -30,11 +30,13 @@
     "@noble/ciphers": "^0.1.3",
     "@noble/curves": "^1.3.0",
     "@noble/hashes": "^1.4.0",
-    "@opentelemetry/api": "^1.9.0",
     "@scure/base": "^1.1.1",
     "hash-wasm": "^4.9.0",
     "neverthrow": "^7.0.1",
     "queueueue": "^4.1.2"
+  },
+  "peerDependencies": {
+    "@opentelemetry/api": "^1.0.0"
   },
   "scripts": {
     "dev": "tsc --watch --sourceMap --outDir dist/web -p tsconfig.web.json",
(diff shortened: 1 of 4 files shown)


packages/cojson/package.json
```diff
@@ -35,6 +35,9 @@
     "neverthrow": "^7.0.1",
     "queueueue": "^4.1.2"
   },
+  "peerDependencies": {
+    "@opentelemetry/api": "^1.0.0"
+  },
```
gdorsi, line 40: Not sure about having this as peer dependency.

If we are importing it and the user doesn't install open telemetry doesn't it break the build?

gdorsi (approved):
LGTM

Unsure about having the package as peerDependency

packages/cojson/package.json
Elfo404: Having it as a peer dependency ensures that the users installs the version they need/want and avoid inconsistencies as only one version is going to be installed - in general all instrumented libraries mark it as a peer dep. Some of them - like nextjs - make it also optional by also shipping a pre bundled version that is used if the user does not install another one explicitly - i guess to remove the friction when using a package manager that does not install peer deps automatically.

Elfo404: yep. and some other mark it as a dependency, like https://github.com/vercel/ai/blob/311b17306f9aada804622e319270094347be004c/packages/ai/package.json

gdorsi: If you feel confident that works it's ok for me.

Was just worried by the fact that peerDependencies haven't been correctly implemented by some package managers over time and that someone might incur in some issues.

But this might be something related to the past, honestly I don't know how the situation is nowadays.

aeplay: let's see if it still breaks people's installs

aeplay (approved):

PR #976, opened 2024-12-09
New API draft

aeplay:


packages/jazz-tools-2/src/schema.ts
```diff
@@ -0,0 +1,199 @@
...
+  status = Person;
+}
+
+const profile = await Profile.load("co_skjdfhi743");
+if (profile.status.$loaded) {
+  profile.status.name;
+}
+
+class ListOfProfiles extends CoList {
+  item = Profile;
```
tobiaslins, line 174: what if its `$item` or something? otherwise it may feel like a custom attribute again?

PR #979, opened 2024-12-09
Send image in chat example

trishalim:


examples/chat/src/chatScreen.tsx
```diff
@@ -1,21 +1,49 @@
...
-  return chat ? (
+  if (!chat)
+    return (
+      <div className="flex-1 flex justify-center items-center">Loading...</div>
+    );
+
+  const sendImage = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (!me?.profile) return;
+
+    const file = event.currentTarget.files?.[0];
```
gdorsi, line 32: Maybe we should check the file size before uploading, just to be sure that the next time we go on HN people won't upload GB of data.

e.g.
```ts
if (file?.size > 10_000_000) return; // Drop any upload bigger than 10mb
```

examples/chat/src/chatScreen.tsx
```diff
@@ -1,21 +1,49 @@
...
+
+    if (file) {
+      createImage(file, { owner: me.profile._owner }).then((image) => {
```
gdorsi: Not sure why this works (guess that the profile is shared to everyone by default) but would be cleaner to link the image to the chat group:

```suggestion
      createImage(file, { owner: chat._owner }).then((image) => {
```

homepage/homepage/components/home/ChatDemoSection.tsx
```diff
@@ -133,7 +133,7 @@ export function ChatDemoSection() {
         title="See it for yourself"
         slogan={
           <>
-            A chat app in 174 lines of client-side code.{" "}
+            A chat app with image upload in 300+ lines of client-side code.{" "}
```
gdorsi: Maybe would be better to say `~300 lines`?

300+ means that the lines could be way more

gdorsi (approved):
LGTM

Love how it looks!

examples/chat/src/chatScreen.tsx
trishalim: What are the permissions of `me.profile._owner`? I am guessing it's a Group where `me` is the admin and everyone else is a reader?

homepage/homepage/components/home/ChatDemoSection.tsx
trishalim: yeah you're right

examples/chat/src/chatScreen.tsx
trishalim: Good idea, thanks!

I set the limit to 5MB. I think that's plenty for an image.

examples/chat/src/chatScreen.tsx
trishalim: I think I should change the owner of Message to `me.profile._owner` so that, if we add Edit and Delete, then only the sender can perform those.

bensleveritt (approved):
Looks good to me. Only quibble would be to make the add image button look more like a button (background on hover or something)

gdorsi:
There are some conflicts, after resolving them let's merge!

PR #988, opened 2024-12-10
chore: adds message queue size metrics tests

Elfo404:
Follow up to https://github.com/garden-co/jazz/pull/975

adds tests to ensure the correct metrics are preoduced with the correct values.

i figured `should pull messages in priority order` was a good fit to add assertions related to this, but i'm happy to move them to a different tests if we prefer.

### DIFF
--- a/packages/cojson/package.json
+++ b/packages/cojson/package.json
@@ -21,6 +21,7 @@
   "license": "MIT",
   "version": "0.8.37",
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


gdorsi (approved):
LGTM 

Thanks for the tests

Elfo404:
@gdorsi eeeeh, it's failing :D fixing it

PR #990, opened 2024-12-10
Chore: Bump node version to 22

bensleveritt:


packages/cojson-storage-sqlite/package.json
```diff
@@ -6,13 +6,13 @@
   "types": "src/index.ts",
   "license": "MIT",
   "dependencies": {
-    "better-sqlite3": "^8.5.2",
+    "better-sqlite3": "^11.7.0",
```
gdorsi, line 9: This means that we are dropping support for node v21: https://github.com/WiseLibs/better-sqlite3/releases/tag/v11.0.0

gdorsi (approved):
LGTM

Thanks!

packages/cojson-storage-sqlite/package.json
bensleveritt: Good point. Is that a concern, or should we try to stay under?

gdorsi (approved):
Many thanks!

Issue #114, opened 2023-10-04
Inbox abstraction


aeplay:
As more users adopt server workers, they find the concept of "workers subscribing to a CoValue and performing side-effects on updates" intuitive.

But what remains unsolved is how to notify workers of "new CoValues" they should care about.

For this purpose, we need a CoValue that can act like an "Inbox" - a single value that can be shared widely, even hardcoded by ID, that producers of work for the worker can push "request-like" or "invite-notfication-like" items into.

The CoValue that is closest to this is a CoStream. It already ensures that each participant (session) can only push items into their own session log.

However, all participants are still able to read the items of *all other participants*.

To meaningfully achieve Inbox-like semantics, we want "senders" to only be able to see their own pushed items in the Inbox, while the "receiver" can see all items by everyone.

This requires as a must the introduction of a new writer-only role (see #112 for more details & implications).

Ideally, the sync server should also make sure that senders will only be able to sync their own sessions in an inbox and not uselessly sync all the other sessions they can't read. To support this, we need a very basic form of authenticated sync, where a peer proves controlling a specific account and the sync server takes that into consideration when syncing a CoValue in a group with writer-only members.

PR #991, opened 2024-12-10
feat: add writeOnly role on groups

gdorsi:
This PR adds a new "writeOnly" role intended to give users access only to their own changes.

This new role is intended as building block for #114 enabling a new patterns for workers subscriptions.

### DIFF
--- a/packages/cojson/src/coValues/group.ts
+++ b/packages/cojson/src/coValues/group.ts
@@ -364,7 +364,7 @@ export class RawGroup<
     return keys;
   }
 
-  getCurrentKeyReadId() {
+  getCurrentReadKeyId() {
     if (this.myRole() === "writeOnly") {
       const accountId = this.core.node.account.id;
 
--- a/packages/cojson/src/coValues/group.ts
+++ b/packages/cojson/src/coValues/group.ts
@@ -13,7 +13,7 @@ import {
   isParentGroupReference,
 } from "../ids.js";
 import { JsonObject } from "../jsonValue.js";
-import { Role } from "../permissions.js";
+import { AccountRole, Role } from "../permissions.js";
 import { expectGroup } from "../typeUtils/expectGroup.js";
 import {
   ControlledAccountOrAgent,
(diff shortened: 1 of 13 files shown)


packages/cojson/src/coValues/group.ts
```diff
@@ -236,44 +238,168 @@ export class RawGroup<
...
+      const role = this.get(key);
+      return (
+        role === "admin" ||
+        role === "writer" ||
+        role === "reader" ||
+        role === "adminInvite" ||
+        role === "writerInvite" ||
+        role === "readerInvite"
+      );
+    });
```
gdorsi, line 397: Added the invite roles here to support [this use case](https://github.com/garden-co/jazz/blob/46df5b7efc2497073f1d6c5f014ba702601952ee/packages/cojson/src/tests/group.test.ts#L280).

Before when rotating keys we weren't revealing the new keys to invite roles making the invites invalid.

packages/cojson/src/coValues/group.ts
```diff
@@ -558,13 +719,20 @@ function isMorePermissiveAndShouldInherit(
...
   if (roleInParent === "writer") {
-    return !roleInChild || roleInChild === "reader";
+    return (
+      !roleInChild || roleInChild === "reader" || roleInChild === "writeOnly"
+    );
   }
 
   if (roleInParent === "reader") {
     return !roleInChild;
   }
```
gdorsi, line 729: Configured the `reader` role to not be inherited by writerOnly members.

Considering them incompatible and kind-of at the same level.

packages/cojson/src/coValues/group.ts
```diff
@@ -236,44 +238,168 @@ export class RawGroup<
...
+    const memberKey = typeof account === "string" ? account : account.id;
+    const agent =
+      typeof account === "string"
+        ? account
+        : account.currentAgentID()._unsafeUnwrap({ withStackTrace: true });
+
+    /**
+     * WriteOnly members can only see their own changes.
+     *
+     * We don't want to reveal the readKey to them so we create a new one and reveal it
```
aeplay:
```suggestion
     * We don't want to reveal the readKey to them so we create a new one specifically for them and also reveal it to everyone else with a reader or higher-capability role (but crucially not to other writer-only members)
```

packages/cojson/src/coValues/group.ts
aeplay: correct

packages/cojson/src/permissions.ts
```diff
@@ -254,14 +259,31 @@ function determineValidTransactionsForGroup(
...
+        memberState[transactor] !== "readerInvite" &&
+        memberState[transactor] !== "writeOnlyInvite"
       ) {
         console.warn("Only admins can reveal keys");
         continue;
       }
 
-      // TODO: check validity of agents who the key is revealed to?
+      /**
+       * We don't want to give the ability to invite members to override
```
aeplay, line 270: thanks again for spotting this

aeplay (approved):

PR #993, opened 2024-12-10
Add form example for creating and updating

trishalim:
This is a simple form example that shows you how to make a form for creating and editing a `CoValue`, 
called `BubbleTeaOrder`, with fields of different types such
as single-select, multi-select, date, text, and boolean.

To create a new `BubbleTeaOrder`, we create an empty order. Because `BubbleTeaOrder` has some
required fields, we can't create an empty `BubbleTeaOrder` directly. 

Instead, we create a `DraftBubbleTeaOrder`,
which has the same structure as `BubbleTeaOrder`, but with all fields set to `optional`.

When the user is ready to submit the order, we treat `DraftBubbleTeaOrder` as a "real order" by
converting it into a `BubbleTeaOrder`.


examples/form/src/CreateOrder.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  const newOrder = BubbleTeaOrder.create(
+    {
+      baseTea: BubbleTeaBaseTeaTypes[0],
+      addOns: ListOfBubbleTeaAddOns.create([], { owner: me.profile._owner }),
+      deliveryDate: new Date(),
+      withMilk: false,
+    },
+    { owner: me?.profile._owner },
+  );
```
gdorsi: It's not great to call create in the component body, this way a new order gets created on each render.

The React-friendly way to do this is:
```ts
const [newOrder, setNewOrder] = useState<BubbleTeaOrder | undefined>();

const publicGroup = me?.profile._owner;
useEffect(() => {
  if (!publicGroup) return;
  
   const newOrder = BubbleTeaOrder.create({
      baseTea: BubbleTeaBaseTeaTypes[0],
      addOns: ListOfBubbleTeaAddOns.create([], { owner: me.profile._owner }),
      deliveryDate: new Date(),
      withMilk: false,
    },
    { owner: publicGroup },
  );
  
  setNewOrder(newOrder);
}, [publicGroup])

if (!newOrder) return null;
```

gdorsi: @aeplay This could become something we provide as hook:

```ts
const newOrder = useCoStateCreate((me) => {
   const publicGroup = me.profile._owner;
   
   return BubbleTeaOrder.create({
      baseTea: BubbleTeaBaseTeaTypes[0],
      addOns: ListOfBubbleTeaAddOns.create([], { owner: me.profile._owner }),
      deliveryDate: new Date(),
      withMilk: false,
    },
    { owner: publicGroup }
    );
})
```

This way we remove the ugly waiting for the account loading.

gdorsi (commented):
I like the pattern, but it's not always possible to provide the default values for required fields.

For that situations, what about declaring a Draft entity?

e.g.
```ts
class DraftBubbleTeaOrder extends CoMap {
  baseTea = co.optional.literal(...BubbleTeaBaseTeaTypes);
  // other fields
}
```

We could even make a "Draft.Of" API that marks all the fields of a CoMap as optional.

examples/form/src/CreateOrder.tsx
aeplay: Good idea to introduce a hook that helps with the state handling.

In general, the hard part with the create + edit form is when you have fields that are required on a created CoValue, but obviously need to be optional _while creating the CoValue_. I think in this example we conveniently avoid this, but it would be good to introduce one required field that doesn't have an obvious "empty" state, so we are forced to find a solution to this and can offer a helpful pattern @trishalim @gdorsi.

edit: this will be a good solution https://github.com/garden-co/jazz/pull/993#pullrequestreview-2494995237

trishalim:
Update @aeplay @gdorsi 

In Profile, I added `draft = co.ref(DraftBubbleTeaOrder);` and initialized it on `migrate`.

When creating an order, I use `draft`. When user changes this without submitting the form, the changes are kept there and they can come back to it anytime. On submit, the draft gets pushed into `profile.orders` and becomes a real order. The `profile.draft` is reset into an empty `DraftBubbleTeaOrder`.

aeplay:
> Update @aeplay @gdorsi
> 
> In Profile, I added `draft = co.ref(DraftBubbleTeaOrder);` and initialized it on `migrate`.
> 
> When creating an order, I use `draft`. When user changes this without submitting the form, the changes are kept there and they can come back to it anytime. On submit, the draft gets pushed into `profile.orders` and becomes a real order. The `profile.draft` is reset into an empty `DraftBubbleTeaOrder`.

That's a cool pattern!!

examples/form/src/CreateOrder.tsx
```diff
@@ -0,0 +1,65 @@
...
+} from "./schema.ts";
+
+export function CreateOrder() {
+  const { me } = useAccount({ profile: { draft: {}, orders: [] } });
+  const router = useIframeHashRouter();
+
+  if (!me?.profile) return;
+
+  const onSave = (draft: DraftBubbleTeaOrder) => {
+    // turn the draft into a real order
```
aeplay, line 29: Can we either
- do some actual validation here to check all fields are present?
- leave a comment that in general this _would be_ a good place to do validation?

aeplay (changes requested):

gdorsi (approved):
Awesome job!

examples/form/src/CreateOrder.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const { me } = useAccount({ profile: { draft: {}, orders: [] } });
+  const router = useIframeHashRouter();
+
+  if (!me?.profile) return;
+
+  const onSave = (draft: DraftBubbleTeaOrder) => {
+    // validate if the draft is a valid order
+    const validation = draft.validate();
+    if (validation?.error) {
+      alert(validation.error);
```
aeplay: Please may you extend this to proper field errors displayed on the form. I know it means you need to figure out some state tracking for it, but I think it would be worth having a full-blown "form with validation errors" example

aeplay (changes requested):

aeplay (approved):

aeplay (approved):

PR #994, opened 2024-12-10
fix(jazz-nodejs): wait for sync before closing and fix the reconnection

gdorsi:
Changes:
- Wait for the all coValues to be uploaded before closing the worker
- Improve the reconnection flow by reacting to the "close" handler and removing the setInterval
- Export startSyncServer and createWorkerAccount on jazz-tools to make possible to use them via code

### DIFF
--- a/packages/jazz-nodejs/src/index.ts
+++ b/packages/jazz-nodejs/src/index.ts
@@ -55,6 +55,8 @@ export async function startWorker<Acc extends Account>({
   node = context.account._raw.core.node;
 
   async function done() {
+    await context.account.waitForAllCoValuesSync();
+
     wsPeer.done();
     context.done();
   }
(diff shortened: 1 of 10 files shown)


packages/jazz-nodejs/src/index.ts
```diff
@@ -41,31 +40,66 @@ export async function startWorker<Acc extends Account>({
...
     }),
     AccountSchema,
     // TODO: locked sessions similar to browser
     sessionProvider: randomSessionProvider,
-    peersToLoadFrom: [wsPeer],
+    peersToLoadFrom: [wsPeer.peer],
     crypto: await WasmCrypto.create(),
   });
 
-  setInterval(async () => {
```
gdorsi, line 56: This was preventing the scripts from ending because it keeps a timer active even when the work is "done"

bensleveritt (approved):
Looks good to me! 👍

PR #996, opened 2024-12-10
Move to use turbo build

bensleveritt:
To run the homepage build more effectively, we should use `turbo build` rather than `pnpm build` as there are a number of dependencies that can be cached by turbo, which aren't when run with pnpm.

Adds generating code samples to turbo's scripts.


gdorsi (approved):
Neat, thanks!

PR #1009, opened 2024-12-12
Proof-of-concepts for hosting Y.js, automerge, Loro on top of cojson

aeplay:


packages/auto-jazz/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "auto-jazz",
```
gdorsi, line 2: Auto-jazz might be misleading, what about naming this `jazz-automerge`?

gdorsi (commented):
Tested locally and found some issues on the build.

Pushed a fix.

Also noticed that all the messages are considered as sent from the same user.
Is that something you forgot to implement or is it as intended?

Issue #763, opened 2024-11-15
Implement ShardedPeer


aeplay:
For edge nodes of a sync server cluster, simple load balancing is enough to distribute load across cores & machines. But so far, these edge nodes had to rely on a single core server (and single CPU core) as their upstream. And that core server had to process updates to *all* CoValues, making it a bottleneck.

In order to parallelise the core server and start introducing redundancy, we need to **shard** it by CoValue ID.

To achieve this, we need to implement a `ShardedPeer` which consists of multiple other `Peer` definitions ("sub-peers"). It should multiplex outgoing messages to the correct peer by using [Rendezvous Hashing](https://en.wikipedia.org/wiki/Rendezvous_hashing) - picking the sub-peer with the smallest value of `hash(coValueID, peerName)`.

It should support dynamic changes to the existence/availability of sub-peers.

Incoming messages from all sub-peers should be coalesced to give the appearance of `ShardedPeer` being a single peer.

Note that a shard becoming unavailable will cause CoValue unavailability issues, because we only sync each CoValue with a single shard in this initial naive setup. In this case, the known-state handling of the `ShardedPeer` as implemented in the `SyncManager` also becomes slightly wrong for affected CoValues, because it reflects the state of the now-unavailable shard.

This is however already a marked improvement to the single-core-server situation and a first step to implement smarter behaviour in the future (such as picking two or three peers for each CoValue ID, which Rendezvous Hashing lets us gracefully upgrade to later)

Setting up the shard peers correctly and making sure the shard peers know about each other, too, is left as an exercise to the cluster deployer. This is important, because CoValue dependencies (such as value-owned-by-group, account-member-of-group, or group-extended-by-group) need to be resolved correctly, and depended-on CoValues might live on another shard.

gdorsi:
Done

PR #1010, opened 2024-12-12
feat: Inbox, a new experimental API to simplfy the communication with Workers

gdorsi:
We're introducing a new experimental API to simplify the inital handshake between accounts (on both Workers and Users)

## The problem

We want to introduce a new API to let users model different flows for connecting accounts.

Currently the only "built-in" flow is invite based:
- The owner creates the CoValue and create an invite
- The other account consumes the invite to get access to the CoValue

While this works most of the time it can be too much manual for some scenarios, such as communication with Workers.

One example is the communication with the Workers, where we want to use the Inbox to make the following flow possible:
- Let's say we have a worker that generates a PDF of a given WebPage
- The Worker Account ID is known beforehand and passed as ENV variable
- An user account creates a PDFRequest CoValue and sends it to the Worker via Inbox
- The Worker gets the PDFRequest, starts the processing and assign the result to one of the PDFRequest props
- The user account subscribed to PDFRequest and gets the result through it once ready

In this flow the Inbox provides an easy-to-use API to send a CoValue to any account by knowing only their AccountID.
It also automatically tracks the processed values and manages failures with a configurable retry system.

## Code examples

This is the example code to implement the communication to a Worker that generates PDFs of a given website URL:

schema.ts
```ts
export class PDFRequest extends CoMap {
  url = co.string;
  error = co.optional.string;
}
```

worker.ts
```ts
const { worker, experimental: { inbox }, done } = await startWorker();

// The value is marked as processed as soon as the callback promise is resolved 
// We pass the senderAccountID so the Worker can more easily add the sender as
// group member of the result. 
// A much simpler solution can also be to reuse the value group but may not be fitted to every use case
inbox.subscribe(PDFRequest, async (value, senderAccountID)  => { 
  // We are going to introduce a Task API to better handle this async workflows
  return generatePDF(value, senderAccountID);
})
```

On the app side:
```ts
function PDFDownloader(props) {
   const { me } = useAccount();
   const { sendMessage } = useInboxSender(import.meta.env.VITE_WORKER_ID);
   
   function handleSubmit(url: string) {
     const group = Group.create({ owner: me });
     const request = PDFRequest({ url }, { owner: group });
     const fileId = await sendMessage(request); // The inbox owner is automatically added as Writer
     
     downloadFileStream(fileId);
    }
}
```

## Design choices

The general thinking done when designing this API can be summarized in these points:
- To connect the two accounts need only their "AccountID"
- Permissions for this kind of communication should be easy to implement
- The messages sent should be only visible to the reiceiver, for that we have introduced the "writeOnly" role
- Generally, an account should send a CoValue to the other account and the rest of the communication should be handled by updating that CoValue.
- Managing failures is HARD and requires a much more complex API. Anselm is designing a Task/Workflow API that's going to handle this kind of stuff more in-depth.
- Marking this a expermiental to let us play with the API in various apps and improve it further before releasing to the general public.

Along the way we found also that the Inbox can also be used for communication between users.

With the Inbox developers can implement an approval based flow, where the accounts that want to get access to a given resource can ask for access through the Inbox and owners can approve it.
We think this flow can be quite useful if used with the [Organization pattern](https://github.com/garden-co/jazz/pull/1045) where the users need to do it only once.

For that we have added a `useInboxListener` hook to subscribe to the current account Inbox messages but requires further exploration and it's outside of the main goals of this MVP.

### Scenarios

> What if the task fails

While the heavy lift should be done with a system designed to better handle failures (our next Task API or any other workflow system available in the wild) things may still be fail.

For that we provide:
- A "failed" stream that devs can use to check the failed CoValues
- A configurable retry system, which retries to run the callback for X times (default is 3) before giving up

> What if the user wants pass different possible schemas through the Inbox?

While we suggest to use different workers for most of the situations where this might be required it's possible to handle this scenario using a [SchemaUnion](https://jazz.tools/api-reference/jazz-tools#SchemaUnion):
```ts
const UnionOfMessageTypes = SchemaUnion.Of<CoMap>(value => {
   switch (value.get('type')) {
      case 'A':
         return MessageTypeA;
      case 'B':
         return MessageTypeB;
   }
})

inbox.subscribe(UnionOfMessageTypes, (value) => {
   if (value instanceof MessageTypeA) return processMessageTypeA(value);
   return  processMessageTypeB(value);
})
```

It would be very nice to support an API like the following one:
```ts
inbox.subscribe(MessageTypeA, (value) => {
})

inbox.subscribe(MessageTypeB, (value) => {
})
```

This would be very ergonomic especially on the App side, but we don't have the required level of introspection now and that's currently out of scope for now.

> What if the user subscribes multiple times on the same inbox

On the Worker side only one subscribe will work simply because we don't see the need for that there.

When using `useInboxListener` the update is pushed to every subscriber.

## Outside this MVP

There are some other stuff to figure out:
- What's our limit with the current message stream?
- How's going to be possible to distribute work across workers? (might steal the sharding approach from #763 or go for something like Raft)
- Explore the user-to-user communication use case and iterate the API around our findings there.

### DIFF
--- a/packages/jazz-nodejs/src/index.ts
+++ b/packages/jazz-nodejs/src/index.ts
@@ -3,25 +3,31 @@ import {
   Account,
   AccountClass,
   ID,
+  Inbox,
   createJazzContext,
   fixedCredentialsAuth,
   randomSessionProvider,
 } from "jazz-tools";
-import { WebSocket } from "ws";
 import { webSocketWithReconnection } from "./webSocketWithReconnection.js";
 
-/** @category Context Creation */
-export async function startWorker<Acc extends Account>({
-  accountID = process.env.JAZZ_WORKER_ACCOUNT,
-  accountSecret = process.env.JAZZ_WORKER_SECRET,
-  syncServer = "wss://cloud.jazz.tools",
-  AccountSchema = Account as unknown as AccountClass<Acc>,
-}: {
+type WorkerOptions<Acc extends Account> = {
   accountID?: string;
   accountSecret?: string;
   syncServer?: string;
   AccountSchema?: AccountClass<Acc>;
-}): Promise<{ worker: Acc; done: () => Promise<void> }> {
+};
+
+/** @category Context Creation */
+export async function startWorker<Acc extends Account>(
+  options: WorkerOptions<Acc>,
+) {
+  const {
+    accountID = process.env.JAZZ_WORKER_ACCOUNT,
+    accountSecret = process.env.JAZZ_WORKER_SECRET,
+    syncServer = "wss://cloud.jazz.tools",
+    AccountSchema = Account as unknown as AccountClass<Acc>,
+  } = options;
+
   let node: LocalNode | undefined = undefined;
   const wsPeer = webSocketWithReconnection(syncServer, (peer) => {
     node?.syncManager.addPeer(peer);
@@ -52,7 +58,14 @@ export async function startWorker<Acc extends Account>({
     crypto: await WasmCrypto.create(),
   });
 
-  node = context.account._raw.core.node;
+  const account = context.account as Acc;
+  node = account._raw.core.node;
+
+  if (!account._refs.profile?.id) {
+    throw new Error("Account has no profile");
+  }
+
+  const inbox = await Inbox.load(account);
 
   async function done() {
     await context.account.waitForAllCoValuesSync();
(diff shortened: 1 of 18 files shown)


packages/jazz-nodejs/src/index.ts
```diff
@@ -61,5 +74,15 @@ export async function startWorker<Acc extends Account>({
     context.done();
   }
 
-  return { worker: context.account as Acc, done };
+  const inboxPublicApi = {
+    subscribe: inbox.subscribe.bind(inbox) as Inbox["subscribe"],
+  };
```
gdorsi, line 79: Reducing the API surface to limit the possible operations on Inbox only to subscribe

packages/jazz-nodejs/src/index.ts
```diff
@@ -61,5 +74,15 @@ export async function startWorker<Acc extends Account>({
...
+
+  return {
+    worker: context.account as Acc,
+    experimental: {
+      inbox: inboxPublicApi,
+    },
```
gdorsi, line 85: Wrapping the API into an "experimental" object to make it clear that it's not stable yet

PR #1018, opened 2024-12-13
Jazz 0.12.0 - A clearer syntax for deep loading

aeplay:
TODO:
- [x] Improve the loading of unauthorized optional values (return null instead of an empty value)
- [x] ~~Collect the list of errors when encountering unavailable or unauthorized values~~ (out of scope)
- [x] Test for deepLoading required refs that are missing
- [ ] Shall we include a redesign of the useCoState API?
- [x] Upgrade docs
- [ ] Deep loading docs


packages/jazz-svelte/src/lib/jazz.svelte.ts
bensleveritt, line 1: Looks good to me 👍

examples/passkey/vite.config.ts.timestamp-1734345846936-31300bb333a5.mjs
```diff
@@ -0,0 +1,8 @@
...
+import react from "file:///Users/anselm/jazz/jazz/node_modules/@vitejs/plugin-react/dist/index.mjs";
+var vite_config_default = defineConfig({
+  plugins: [react()],
+});
+export { vite_config_default as default };
+//# sourceMappingURL=data:application/json;base64,ewogICJ2ZXJzaW9uIjogMywKICAic291cmNlcyI6IFsidml0ZS5jb25maWcudHMiXSwKICAic291cmNlc0NvbnRlbnQiOiBbImNvbnN0IF9fdml0ZV9pbmplY3RlZF9vcmlnaW5hbF9kaXJuYW1lID0gXCIvVXNlcnMvYW5zZWxtL2phenovamF6ei9leGFtcGxlcy9wYXNza2V5XCI7Y29uc3QgX192aXRlX2luamVjdGVkX29yaWdpbmFsX2ZpbGVuYW1lID0gXCIvVXNlcnMvYW5zZWxtL2phenovamF6ei9leGFtcGxlcy9wYXNza2V5L3ZpdGUuY29uZmlnLnRzXCI7Y29uc3QgX192aXRlX2luamVjdGVkX29yaWdpbmFsX2ltcG9ydF9tZXRhX3VybCA9IFwiZmlsZTovLy9Vc2Vycy9hbnNlbG0vamF6ei9qYXp6L2V4YW1wbGVzL3Bhc3NrZXkvdml0ZS5jb25maWcudHNcIjtpbXBvcnQgcmVhY3QgZnJvbSBcIkB2aXRlanMvcGx1Z2luLXJlYWN0XCI7XG5pbXBvcnQgeyBkZWZpbmVDb25maWcgfSBmcm9tIFwidml0ZVwiO1xuXG4vLyBodHRwczovL3ZpdGUuZGV2L2NvbmZpZy9cbmV4cG9ydCBkZWZhdWx0IGRlZmluZUNvbmZpZyh7XG4gIHBsdWdpbnM6IFtyZWFjdCgpXSxcbn0pO1xuIl0sCiAgIm1hcHBpbmdzIjogIjtBQUEwUyxPQUFPLFdBQVc7QUFDNVQsU0FBUyxvQkFBb0I7QUFHN0IsSUFBTyxzQkFBUSxhQUFhO0FBQUEsRUFDMUIsU0FBUyxDQUFDLE1BQU0sQ0FBQztBQUNuQixDQUFDOyIsCiAgIm5hbWVzIjogW10KfQo=
```
gdorsi: We don't need this one, right?

packages/jazz-tools/src/coValues/deepLoading.ts
```diff
@@ -82,58 +75,69 @@ export function fulfillsDepth(depth: any, value: CoValue): boolean {
...
+              | boolean
           : V extends {
                 _type: "CoStream";
                 byMe: CoFeedEntry<infer Item> | undefined;
               }
             ?
-                | [
-                    DepthsIn<
+                | {
+                    $each: RefsToResolve<
```
gdorsi, line 268: What about adding a typedoc comment on this property?

If we can make the IDE show a brief documentation on `$each` I think would be a huge help for people to get comfortable with the Resolve API.

.changeset/fast-beans-decide.md
```diff
@@ -0,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Implement new API for deep loading
```
gdorsi, line 5: Guess this one should be a minor as it's a breaking change

packages/jazz-tools/src/coValues/deepLoading.ts
aeplay: Unfortunately doesn't get passed out of this complex type :(

packages/jazz-react/src/index.tsx
```diff
@@ -126,12 +126,12 @@ export function createJazzReactApp<Acc extends Account>({
...
   function useAccount(): { me: Acc; logOut: () => void };
-  function useAccount<D extends DepthsIn<Acc>>(
-    depth: D,
-  ): { me: DeeplyLoaded<Acc, D> | undefined; logOut: () => void };
-  function useAccount<D extends DepthsIn<Acc>>(
-    depth?: D,
-  ): { me: Acc | DeeplyLoaded<Acc, D> | undefined; logOut: () => void } {
+  function useAccount<const O extends { resolve?: RefsToResolve<Acc> }>(
+    options: O,
+  ): { me: Resolved<Acc, O> | undefined; logOut: () => void };
```
gdorsi: This kind of typing makes ok to add new options like:

Screenshot 2024-12-17 at 16 17 56

Do you think it would be ok to have just:

```ts
useAccount<const R extends RefsToResolve<Acc>>(
    options: { resolve?: R },
  ): { me: Resolved<Acc, R> | undefined; logOut: () => void };
```

gdorsi: Also noticed that with this change it would be possible to add docs on the resolve option:
Screenshot 2024-12-17 at 16 23 34

gdorsi: Noticed two other issues:
1. Additional properties added inside to resolve aren't marked as invalid
1. Errors on the value side are quite cryptic

We probably can't do much to make errors easier to read, but we should definitely mark additional properties as errors.

See:
Screenshot 2024-12-17 at 16 42 13

Screenshot 2024-12-17 at 16 38 31

gdorsi:
Closing in favor of https://github.com/garden-co/jazz/pull/1642 to have a PR more up in the list and with a branch name that reflects the version upgrade.

PR #1025, opened 2024-12-14
Pagefind modal and button styling improvements

trishalim:


trishalim:
This is really, really old now 🗑️

Issue #1030, opened 2024-12-14
Wait for all incoming content (potentially streamed) from storage peers before attempting sync from server peers


aeplay:
Currently, sync prioritisation seems to be behaving correctly when all content to be loaded fits into a single content message:

* load from storage with empty known state
* receive all content from storage
* load from server with already complete known state
* receive nothing redundant from server

However for large CoValues, where we stream content in multiple content messages, something like the following happens

* load from storage with empty known state
* receive content piece 1 from storage
* load from server with known state "after piece 1"
* receive content piece 2 from storage
* receive content piece 2 from server
* …etc

This means that especially where it is most costly (large CoValues) we redundantly load everything except the first piece of content from the sync server *even though it is available locally.*

PR #1032, opened 2024-12-14
Send content from storage peers in one message instead of streaming it

### LINKED ISSUE #1031: Send content from storage peers in one message instead of streaming it
Peers in cojson stream content of large CoValues in pieces, with intermediate signatures. This enables fairer use of the bandwidth, less head-of-line blocking and incremental validation of incoming content.

Storage peers currently implement the same piecemeal sending of content, even though they

* have all content loaded already at once from the underlying storage
* have non-fallible, high bandwidth connection to the localNode (aka passing values in memory)
* sending multiple pieces just incurs a higher overhead of signature validation

Crucially, there is a bug when a storage peer and a server peer are present when loading more than one content piece: [JAZZ-593](https://linear.app/garden-computing/issue/JAZZ-593/wait-for-all-incoming-content-potentially-streamed-from-storage-peers)<br><br>This deserves a proper fix, but the correct behaviour will be easiest to express after [JAZZ-569](https://linear.app/garden-computing/issue/JAZZ-569/re-architect-localnode-and-covaluecore-to-use-render-passes) 

In the meantime we can circumvent the issue by *just not doing any streaming from storage peers at all* - leading to correct behaviour and no redundantly shared state

aeplay:
Fixes #1031 
Circumvents #1030

### DIFF
--- a/packages/cojson-storage/src/syncUtils.ts
+++ b/packages/cojson-storage/src/syncUtils.ts
@@ -5,30 +5,19 @@ import {
   Stringified,
   cojsonInternals,
 } from "cojson";
-import {
-  SignatureAfterRow,
-  StoredCoValueRow,
-  StoredSessionRow,
-  TransactionRow,
-} from "./types.js";
+import { StoredCoValueRow, StoredSessionRow, TransactionRow } from "./types.js";
 
 export function collectNewTxs({
   newTxsInSession,
   newContentMessages,
   sessionRow,
-  signaturesAndIdxs,
-  peerKnownState,
   firstNewTxIdx,
 }: {
   newTxsInSession: TransactionRow[];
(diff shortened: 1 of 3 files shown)


packages/cojson-storage/src/syncUtils.ts
```diff
@@ -46,20 +35,7 @@ export function collectNewTxs({
...
-      signaturesAndIdxs.shift();
-      newContentMessages.push({
-        action: "content",
-        id: peerKnownState.id,
-        new: {},
-        priority: cojsonInternals.getPriorityFromHeader(undefined),
-      });
-    } else if (idx === firstNewTxIdx + newTxsInSession.length - 1) {
-      sessionEntry.lastSignature = sessionRow.lastSignature;
-    }
```
marinoska, line 61: I think the exact piece you removed was something I din't fully understand :D

Issue #714, opened 2024-11-07
New API version with less magic


aeplay:
[...]

Issue #565, opened 2024-10-14
Discovery: immutable updates and refs


gdorsi:
Let's say we have the following 2 entity graph:

```
Playlist['Best songs'] -> MusicTrack['Friday!']
```

<br>When we update the MusicTrack, since the updates in Jazz are immutable, we are going to have two versions of the graph

```
v0 Playlist['Best songs'] -> MusicTrack['Friday!']
```

```
v1 Playlist['Best songs'] -> MusicTrack['Monday!']
```

<br>The issue is that if currently I try to resolve MusicTrack from v0 I'm going to get `'Monday!'` and not `'Friday!'`<br><br>The updates to be really immutable should generate a full snapshot of the resolved graph.<br><br>**Why this can be a problem?**

Immutability is one [of the rules of React](https://react.dev/reference/rules#components-and-hooks-must-be-pure) and breaking it might case unexpected behavoirs when using advanced features like [useTransition](https://react.dev/reference/react/useTransition#usetransition) or [useDeferredValue](https://react.dev/reference/react/useDeferredValue#usage).

React might render multiple versions of the same component "concurrently" and if one version is updated unexpectedly (e.g like in the above example) the UI might experience some visual bugs.

gdorsi:
Going to be part of the new Schema & value APIS

PR #1034, opened 2024-12-14
perf: incremental computation on transactions

gdorsi:
Implements incremental transactions computation on CoMap and CoStream top optimize the updates management.

### DIFF
--- a/packages/cojson/src/coValues/coMap.ts
+++ b/packages/cojson/src/coValues/coMap.ts
@@ -76,19 +76,10 @@ export class RawCoMapView<
     this.latest = {};
     this.knownTransactions = {};
 
-    this.processLatestTransactions();
+    this.processNewTransactions();
   }
 
-  private getNextValidTransactions() {
-    const nextValidTransactions = this.core.getValidTransactions({
-      ignorePrivateTransactions: this.ignorePrivateTransactions,
-      knownTransactions: this.knownTransactions,
-    });
-
-    return nextValidTransactions;
-  }
-
-  private processLatestTransactions() {
+  processNewTransactions() {
     if (this.isTimeTravelEntity()) {
       throw new Error("Cannot process transactions on a time travel entity");
     }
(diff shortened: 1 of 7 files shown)


packages/cojson/src/coValues/coMap.ts
```diff
@@ -374,6 +365,10 @@ export class RawCoMap<
     value: Shape[K],
     privacy: "private" | "trusting" = "private",
   ): void {
+    if (this.isTimeTravelEntity()) {
+      throw new Error("Cannot set value on a time travel entity");
+    }
```
gdorsi, line 370: Added these checks because it would be too weird/complex to add transactions on a time traveled entity and I think it's fair to consider it a "frozen" entity.

packages/cojson/src/coValues/coMap.ts
```diff
@@ -232,36 +188,64 @@ export class RawCoMapView<
...
   }
 
+  getRaw<K extends keyof Shape & string>(key: K) {
+    let latestChange = this.latest[key];
+
+    if (latestChange === undefined) {
+      const entries = this.ops[key];
+
+      // Time travel values are lazily computed
+      if (entries && !(key in this.latest)) {
```
gdorsi, line 213: This was actually my main goal.

With the `writeOnly` role and the Inbox pattern we will probably have groups with a huge number of keys, so I wanted to make time travel in that setting as fast as possible

packages/jazz-tools/src/tests/subscribe.test.ts
```diff
@@ -286,7 +286,6 @@ describe("subscribeToCoValue", () => {
     const lastValue = updateFn.mock.lastCall[0];
     expect(lastValue).not.toBe(initialValue);
     expect(lastValue.name).toBe("Me and Luigi");
-    expect(initialValue.name).toBe("General");
```
gdorsi, line 289: Keeping the content ref breaks this check, because initialValue content is the same as lastValue.

This can be an issue only with React Concurrent, but we have this problem anyway (see #565)

This issue can be solved once we land #714

aeplay (approved):

PR #1039, opened 2024-12-16
[Draft] Protocol simplification proposal

marinoska:
This is to close. Replaced by https://github.com/garden-co/jazz/pull/1164


packages/cojson/src/sync/LoadService.ts
```diff
@@ -0,0 +1,66 @@
+import { CoValueCore } from "../coValueCore.js";
+import { CO_VALUE_LOADING_TIMEOUT, CoValueEntry } from "../coValueEntry.js";
+import { LocalNode } from "../exports.js";
+import { PeerEntry, getPeersWithoutErrors } from "../peer/index.js";
+import { emptyKnownState } from "./types.js";
+
+export class LoadService {
```
aeplay, line 7: Why is this a class? Can't loadCoValue just be a free-standing function?

marinoska: Fair point. Will answer within https://github.com/garden-co/jazz/pull/1164. I'm closing this one.

PR #1045, opened 2024-12-16
Add Organization example app

trishalim:
- [x] share data between users through by sharing an Organization
- [x] add project in Organization
- [x] invite users through invite links
- [x] docs
- [ ] invite users by adding their account ID (I'll do this when Inbox is out)

image


examples/organization/src/AcceptInvitePage.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  const onAccept = (organizationId: ID<Organization>) => {
+    if (me?.root?.organizations) {
+      Organization.load(organizationId, me, []).then((organization) => {
+        if (organization) {
+          // avoid duplicates
+          const ids = me.root.organizations.map(
+            (organization) => organization?.id,
+          );
+          if (ids.includes(organizationId)) return;
```
gdorsi, line 18: @aeplay We definitely need CoSet

This kind of checks here doesn't ensure that the organization is added only once in the list.

I actually also have a use-case also for "CoOrderedSet" on the music app to manage the playlist tracks.
The user likely wants the tracks to be ordered and don't want duplicates.

gdorsi (approved):
LGTM, great job!

PR #1047, opened 2024-12-18
create-jazz-app

pax-k:


trishalim:
I tested all options (except for RN) with pnpm and it works great!

PR #1054, opened 2024-12-19
Add account root design pattern docs

trishalim:


trishalim:
Closing this because @aeplay will work on it

PR #1069, opened 2024-12-20
Add form design pattern to docs

trishalim:
[Preview here](https://jazz-homepage-git-jazz-598-add-form-design-pattern-to-docs-gcmp.vercel.app/docs/react/design-patterns/form)

### DIFF
--- a/homepage/homepage/app/docs/[framework]/[...slug]/design-patterns/form.mdx
+++ b/homepage/homepage/app/docs/[framework]/[...slug]/design-patterns/form.mdx
@@ -2,67 +2,35 @@ import { CodeGroup } from "@/components/forMdx";
 
 # Creating and updating CoValues in a form
 
-Normally, when creating a form, we create a state to store the value of each field.
-Then we update that state as the user makes changes, like this:
+Normally we implement forms using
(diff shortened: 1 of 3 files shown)


homepage/homepage/app/docs/[framework]/[...slug]/design-patterns/form.mdx
```diff
@@ -0,0 +1,369 @@
...
+</CodeGroup>
+
+Then, each Account will have a `DraftBubbleTeaOrder`.
+
+<CodeGroup>
+```tsx
+// schema.ts
+export class AccountRoot extends CoMap {
+  draft = co.ref(DraftBubbleTeaOrder);
+}
```
Comment (gdorsi) [inline homepage/homepage/app/docs/[framework]/[...slug]/design-patterns/form.mdx]: The part related to the AccountRoot is a nice plus but not a core requirement of the draft pattern.

Users can use the draft without having the user "remember" about the last edits.

I would suggest to move this part on the bottom and focus on the "core" of the pattern which is create a "Draft" object for the creation form.

Then explain that we can link the draft to the AccountRoot to have persistence of the unsaved edits.

gdorsi (approved):
Awesome job!

PR #1078, opened 2024-12-21
fix(demo-auth): fix the user credentials storage when the username contains non aplhanumeric characters

gdorsi:
Fixes the user credentials storage when the username contains non aplhanumeric characters


pax-k:
LGTM!

gdorsi:
@pax-k 

Rewritten the fix to serialize the username using `btoa` and handle the migration from the old keys.
I've also added a test suite for DemoAuthMethod

PR #1079, opened 2024-12-21
Fix chat app dark mode

trishalim:


gdorsi:
Works great on dark mode!

Screenshot 2024-12-23 at 12 39 45

When testing I've noticed that the reiceved messages text color doesn't have enough contrast in the light mode:
Screenshot 2024-12-23 at 12 39 16

Reproduced also in the prod version so this is not a regression

examples/chat/src/main.tsx
```diff
@@ -25,9 +26,12 @@ function JazzAndAuth({ children }: { children: React.ReactNode }) {
...
-      <App />
-    </JazzAndAuth>
-  </StrictMode>,
+  <ThemeProvider>
+    <StrictMode>
+      <JazzAndAuth>
+        <App />
+      </JazzAndAuth>
+    </StrictMode>
+    ,
```
gdorsi: This extra comma is going to be rendered in the app

trishalim:
@gdorsi Thanks for testing! Fixed now.
image

PR #1092, opened 2024-12-27
feat: migrate to a request to join pattern

gdorsi:


aeplay (commented):
Would be great to add this as a third permission pattern to docs (direct add, invite link, request to join)

bensleveritt:
When someone requests to join, where does the invitation go?

PR #1103, opened 2024-12-30
feat: simplify the .create calls by accepting Account | Group as second param

gdorsi:
Making the CoValues creation less verbose by accepting the owner as second param:

```ts
      const orders = ListOfBubbleTeaOrders.create([], account);
      const draft = DraftBubbleTeaOrder.create(
        {
          addOns: ListOfBubbleTeaAddOns.create([], account),
        },
        account,
      );

      this.root = AccountRoot.create({ draft, orders }, account);
```

### DIFF
--- a/.changeset/loud-llamas-compare.md
+++ b/.changeset/loud-llamas-compare.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Simplify the .create calls by accepting directly "Account | Group" as second param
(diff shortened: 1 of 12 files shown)


aeplay (approved):
The original idea was to teach people about ownership just through the syntax alone, but it does get tedious - and the fact that it's a required parameter means you need to confront what it's for anyways. Thanks!

PR #1104, opened 2024-12-30
feat: top-level imports for Jazz react hooks

gdorsi:
Introducing top-level imports for Jazz react hooks:
- [The new way to setup Jazz](https://jazz-homepage-git-jazz-361-make-jazz-react-hooks-global-gcmp.vercel.app/docs/react/project-setup?_vercel_share=SJjWtKwQft6D0LxLSESncews97ZAWs19)
- Jazz hooks can now be imported from `jazz-react` e.g. `import { useAccount, useCoState } from "jazz-react";`

Using tantstack-router a example the type-sefety for custom Account schema is achieved by having this snippet in the App root:

```ts
declare module "jazz-react" {
    interface Register {
        Account: MyAppAccount;
    }
}
```

These changes are not backward compatible and the previous way of setting up Jazz with React.

Other changes:
- Moved the platform-agnostic code in a `jazz-react-core` package to share code between React DOM and React Native
- Added a `jazz-react/testing` module to simplify the writing of unit tests for components / hooks that use Jazz
- Added tests to cover the hooks logic
- Removed `useInboxListener` experimental hook. Haven't found a good use-case for it so removing it until the need comes up.
- Optimized useCoState to render only once if the value is available in-memory
- Updated the React docs


examples/chat/src/app.tsx
```diff
@@ -34,3 +38,13 @@ export function App() {
...
+
+createRoot(document.getElementById("root")!).render(
+  <ThemeProvider>
+    <StrictMode>
+      <JazzAndAuth>
+        <App />
+      </JazzAndAuth>
+    </StrictMode>
+  </ThemeProvider>,
+);
```
gdorsi, line 50: Had to move this from main because Vite was resolving `app.tsx` as main entry.

Before this was working because main was imported here

packages/jazz-tools/src/coValues/interfaces.ts
```diff
@@ -191,45 +190,59 @@ export function subscribeToCoValue<V extends CoValue, Depth>(
   id: ID<V>,
   as: Account | AnonymousJazzAgent,
   depth: Depth & DepthsIn<V>,
-  listener: (value: DeeplyLoaded<V, Depth>) => void,
+  listener: (value: DeeplyLoaded<V, Depth>, unsubscribe: () => void) => void,
   onUnavailable?: () => void,
+  syncResolution?: boolean,
```
gdorsi, line 195: Added the `syncResolution` option to use it to optimize the React and other frameworks subscriptions.

Didn't enable the syncResolution by default because it would have been a tricky breaking change

examples/book-shelf/README.md
```diff
@@ -38,4 +38,4 @@ If you have feedback, let us know on [Discord](https://discord.gg/utDMjHYg42) or
 
 By default, the example app uses [Jazz Cloud](https://jazz.tools/cloud) (`wss://cloud.jazz.tools`) - so cross-device use, invites and collaboration should just work.
 
-You can also run a local sync server by running `npx jazz-run sync` and adding the query param `?sync=ws://localhost:4200` to the URL of the example app (for example: `http://localhost:5173/?peer=ws://localhost:4200`), or by setting the `sync` parameter of the `<Jazz.Provider>` provider component in [./src/components/JazzAndAuth.tsx](./src/components/JazzAndAuth.tsx).
+You can also run a local sync server by running `npx jazz-run sync` and adding the query param `?sync=ws://localhost:4200` to the URL of the example app (for example: `http://localhost:5173/?peer=ws://localhost:4200`), or by setting the `sync` parameter of the `<JazzProvider>` provider component in [./src/components/JazzAndAuth.tsx](./src/components/JazzAndAuth.tsx).
```
trishalim, line 41: The readme for the other examples also need this change

homepage/homepage/app/docs/[framework]/[...slug]/project-setup/react.mdx
```diff
@@ -43,28 +15,32 @@ Currently, the recommended pattern to set up a React app with Jazz is to create
...
                 >
                     {children}
-                </Jazz.Provider>
+                </JazzProvider>
                 <PasskeyAuthBasicUI state={passKeyState} />
             </>
         );
     }
+
+    // Register the Account schema for type safety
```
Comment (aeplay) [inline homepage/homepage/app/docs/[framework]/[...slug]/project-setup/react.mdx]: Let's be more specific here

```suggestion
    // Register the Account schema so `useAccount` returns our custom `MyAppAccount`
```

Comment (aeplay) [reply homepage/homepage/app/docs/[framework]/[...slug]/project-setup/react.mdx]: And here: https://github.com/garden-co/jazz/pull/1104/files#diff-7546d93c0714a65185a6409209e5412cdd7ab1b64006bd2cf86fb5bdd4448e47R47

We need to change the sentence: remove "extracted and exported" and just call them Jazz hooks or sth like that

aeplay (approved):
Looks great, just a couple doc suggestions!

PR #1112, opened 2025-01-01
Create react starter app for create-jazz-app

trishalim:
This is now the first option for a starter when running create-jazz-app.
I've kept the chat app as a second option.

image

Todo

- [x] fix TS error
- [x] add changeset
- [ ] update docs


packages/create-jazz-app/src/config.ts
```diff
@@ -75,7 +75,11 @@ export const frameworkToAuthExamples: Partial<
   Record<FrameworkAuthPair, { name: string; repo: string | undefined }>
 > = {
   "react-demo-auth": {
-    name: "React + Jazz + Demo Auth",
+    name: "React + Jazz + Demo Auth + Tailwind",
+    repo: "garden-co/jazz/starters/react-demo-auth-tailwind",
+  },
+  "react-demo-auth-tailwind-chat": {
```
trishalim: @pax-k can you please help me with the TS errors here? Because the naming doesn't comply with the types.

gdorsi: Looks like the type wants the name to end with `auth`

gdorsi: Maybe it would be better to just remove the chat, having one preset should be enough

trishalim: Yeah, I can do that.

gdorsi:
@trishalim Is this PR ready?

gdorsi (approved):

aeplay (commented):
@trishalim @gdorsi would it make sense and be worth adding just a little bit of complexity to also add a private field to edit, so we set people up with a private account root as well?

This is often a source of confusion and having both profile and root set up would make the distinction clear.

trishalim:
@gdorsi It is ready but I don't want to merge it until Andrei is back on Monday in case I break `create-jazz-app` :D

@aeplay That makes sense!

PR #1115, opened 2025-01-01
React guide fixes

trishalim:
This was clickable but it doesn't work, and it looked like a bug. So I made it interactive.
image

Fix indents


gdorsi (approved):
Thanks for the fixes!

There are conflicts to fix

PR #1120, opened 2025-01-02
Add version history example app

trishalim:
image


packages/jazz-tools/src/coValues/coMap.ts
```diff
@@ -43,6 +43,7 @@ type CoMapEdit<V> = {
   ref?: RefIfCoValue<V>;
   by?: Account;
   madeAt: Date;
+  key?: keyof V;
```
trishalim: Is this correct?
Getting a type error
image

aeplay: you want `CoKeys<V>` https://github.com/garden-co/jazz/blob/1ca1d15639c8ab51e3ee84490c98015c76759640/packages/jazz-tools/src/coValues/coMap.ts#L570

But it seems very strange that the keys you're getting here sound like methods on Array?!

trishalim: That doesn't work
image

But it's a string here, so I will just use string.
https://github.com/garden-co/jazz/blob/1ca1d15639c8ab51e3ee84490c98015c76759640/packages/jazz-tools/src/coValues/coMap.ts#L161

aeplay (approved):

PR #1127, opened 2025-01-03
[react native] Fix permissions of messages and fix chat loading

tobiaslins:
- Fixes a bug where that prevented loading existing chats (from browser etc)
- Fixes permissions of messages created in a chat (same group as chat)


gdorsi (approved):
Thanks for the fix!

PR #1129, opened 2025-01-03
Add upgrade guides section to docs

trishalim:
Preview's not showing correctly for some reason (not the first time this happened) so here's a screenshot
image
Don't merge this until we have content


gdorsi:
Merging into next to have a starting point for the migration guides

PR #1131, opened 2025-01-06
Add new Briscola card game example

Elfo404:
Adds an example app implementing the [Briscola](https://en.wikipedia.org/wiki/Briscola) game using jazz, featuring an Inbox and a worker.

I think it's kinda ready for review, even though i can probably spend some more time making it nicer, and better handling some things


gdorsi:
Closing because we made a more minimal example for workers: https://github.com/garden-co/jazz/tree/main/examples/jazz-paper-scissors

PR #1139, opened 2025-01-07
feat(jazz-vue): simplify app setup and make composables available as top-level imports

gdorsi:
Introducing top-level imports for Jazz react hooks:
- [The new way to setup Jazz](https://jazz-homepage-git-top-level-imports-vue-gcmp.vercel.app/docs/vue/project-setup?_vercel_share=ZUhP5TYlAaibnKgIZmoEpFvp2pmEiF2N)
- Jazz composables can now be imported from `jazz-vue` e.g. `import { useAccount, useCoState } from "jazz-vue";`

Using tantstack-router a example the type-sefety for custom Account schema is achieved by having this snippet in the App root:

```ts
declare module "jazz-vue" {
    interface Register {
        Account: MyAppAccount;
    }
}
```

These changes are not backward compatible and the previous way of setting up Jazz with React.

Other changes:
- Added a `jazz-vue/testing` module to simplify the writing of unit tests for components / hooks that use Jazz
- Added tests to cover the hooks logic
- Optimized useCoState to render only once if the value is available in-memory
- Updated the Vue docs


homepage/homepage/app/docs/[framework]/[...slug]/groups/sharing.mdx
```diff
@@ -82,48 +82,3 @@ useAcceptInvite({
...
-<ContentByFramework framework="svelte">
-`useAcceptInvite` is exported from your Jazz app.
-
-<CodeGroup>
-```ts
-const Jazz = createJazzApp();
-export const { useAcceptInvite } = Jazz;
-```
-</CodeGroup>
-</ContentByFramework>
```
Comment (gdorsi) [inline homepage/homepage/app/docs/[framework]/[...slug]/groups/sharing.mdx:127]: Removed since with the top level import there is no need to give instructions on how to import the hook

lefthook.yml
```diff
@@ -1,6 +1,6 @@
 pre-commit:
   commands:
     check:
-      glob: "*.{js,ts,cjs,mjs,d.cts,d.mts,jsx,tsx,json,jsonc}"
+      glob: "*.{js,ts,cjs,mjs,d.cts,d.mts,jsx,tsx,json,jsonc,vue,svelte}"
```
gdorsi, line 4: Vue and Svelte files weren't auto formatted on commit

packages/jazz-vue/vite.config.ts
```diff
@@ -1,27 +1,27 @@
...
 import { defineConfig } from "vite";
 import dts from "vite-plugin-dts";
 
 export default defineConfig({
-  // @ts-expect-error types
-  plugins: [vue(), dts({ include: ["src/**/*.ts"], outDir: "dist" })],
+  plugins: [
+    vue(),
+    dts({ include: ["src/**/*.ts"], outDir: "dist" }),
+    depsExternal(),
```
gdorsi, line 11: Added this plugin to automatically mark all the dependencies as external files

PR #1151, opened 2025-01-08
feat(svelte): syncronously resolve local values in useCoState

gdorsi:


packages/jazz-svelte/src/lib/jazz.svelte.ts
```diff
@@ -159,10 +159,8 @@ export function useCoState<V extends CoValue, D extends DepthsIn<V> = []>(
 
   // Create state and a stable observable
   let state = $state.raw<DeeplyLoaded<V, D> | undefined>(undefined);
-  const observable = $state.raw(createCoValueObservable());
```
gdorsi, line 162: Removing this fixes an infinte loop issue I've encountered when moving to the sync resolution.

It's also non-required in this context.

bensleveritt (approved):
Thanks, looks good to me

PR #1163, opened 2025-01-08
ProseMirror integration + RichText example

bensleveritt:
Adds ProseMirror integration libs, as well as an example app that uses them.

- `jazz-browser-prosemirror` - The core of the ProseMirror integration with Jazz, interfacing CoRichText and ProseMirror transactions
- `jazz-react-prosemirror` - React wrapper around ProseMirror's EditorView that gives an abstraction for just interacting with a rich text editor and the Document CoMap handles the rest.
- `jazz-example-richtext` - A React app that uses `jazz-react-prosemirror` for collaborative rich text editing

CleanShot 2025-01-14 at 16 06 17


packages/jazz-browser-prosemirror/src/schema.ts
```diff
@@ -0,0 +1,7 @@
+import { CoRichText } from "jazz-tools";
+
+/**
+ * Extends CoRichText to represent a collaborative rich text document.
+ * Used as the underlying data structure for the ProseMirror editor.
+ */
+export class Document extends CoRichText {}
```
aeplay: Sorry, I'm not sure why I introduced this alias in my original branch - we don't need it and everything would be clearer if we just dealt with CoRichText everywhere. Please may you get rid of it @bensleveritt ?

aeplay (changes requested):

packages/jazz-browser-prosemirror/src/schema.ts
bensleveritt: Removed!

packages/jazz-tools/src/coValues/coPlainText.ts
```diff
@@ -114,6 +119,18 @@ export class CoPlainText extends String implements CoValue {
     return new this({ fromRaw: raw });
   }
 
+  applyDiff(other: string) {
```
bensleveritt, line 122: APPLY DIFF HERE.

aeplay:
Closing in favour of #1861 & #1849

PR #1164, opened 2025-01-08
Implement new message sync for local node

marinoska:
In this PR: Remake of protocol messaging system.

**Four Atomic Operations:**
PULL and PUSH: represent initial (requesting) states.
DATA and ACK: represent terminal (responding) states.
Terminal operations designate a state of full sync and can't produce any sync message

**General Assumption:**
All peers are assumed to have complete, fully synchronized data. When a new piece of data is generated, it is broadcast based on this assumption.

**Important changes.**
Peer known states are not stored.
No additional assumptions are made about peers' states.

**Flow 1: PULL**
A node initiates a flow with a PULL action.
It receives the DATA msg in response, considered satisfying the initial request.
When a DATA is received from the peer, the flow is terminated.
If the peer is not aware of the coValue the peer sends back the empty DATA msg with {known: false} as a terminal state.
If the peer has no new states to respond with it sends back the empty DATA msg as a terminal state.

**Flow 2: PUSH**
A node initiates a flow with a PUSH action when a new transaction created or during the initial loading.
It receives either an ACK or PULL in response, both considered satisfying the initial request.
When an ACK or PULL is received from the peer, the flow is terminated.
If the peer lacks some transactions in between and needs correction it initiates a new PULL flow, otherwise, sends ACK as a terminal state

**Main behaviour.**
When receiving a PULL or PUSH request, the node will always attempt to subscribe if not already subscribed.
When new data is received, the node broadcasts a PUSH request to all peers.

Screenshot 2025-01-08 at 19 25 32

I explained the changes this PR introduces in the following recording
https://www.loom.com/share/40744499b9a547568fc84c3a3635efb4?sid=e78603ba-1a6a-4b01-b73d-30123adbf4c0

- Renamed: `CoValueState` to `CoValueEntry`
- Renamed: `PeerState` to `PeerEntry`
- New folders `peer` and `sync` in the cojson for peer logic and sync messages handlers
- `PeerOperations.ts` - The PeerOperations class centralizes the sending logic for the atomic synchronization operations:
pull, push, ack, and data, implementing the protocol
- `sync.ts` reorganised in order to delegate sync logic to appropriate sync services.
- Removed peer states and `syncStateSubscriptionManager`
- Dependencies are requested on demand by sync message handlers via `DependencyService.ts`. Unknown dependencies loaded one by one before adding new incoming data.
- Removed await from `syncManager.handleSyncMessage`. Added queueing per message action type per coValue id - in other words - processing messages one by one in a separate queue for the pair of coValue.id and msg.action in `AbstractMessageHandler.handle`
- Added `uploadState` in `CoValueEntry` (formerly, `CoValueState`) for managing uploading to peers
- Added list of unaware peers in `CoValueEntry.uploadState.unawarePeersIds` (to sync to as soon as the coValue appears on local node)

https://www.loom.com/share/5302d2ef958244d0ba9f8b8bfed1854d?sid=009dcc22-bc9b-49d0-98bf-21c5053fe57f


packages/cojson/src/sync/LoadService.ts
```diff
@@ -0,0 +1,66 @@
+import { CoValueCore } from "../coValueCore.js";
+import { CO_VALUE_LOADING_TIMEOUT, CoValueEntry } from "../coValueEntry.js";
+import { LocalNode } from "../exports.js";
+import { PeerEntry, getPeersWithoutErrors } from "../peer/index.js";
+import { emptyKnownState } from "./types.js";
+
+export class LoadService {
+  constructor() {}
+
```
marinoska: #1039 [aeplay](https://github.com/aeplay) asked: Why is this a class? Can't loadCoValue just be a free-standing function?

gdorsi (commented):
Skimmed thorugh the PR and watched the explaination video, I like the solution!

To help move this forward, I have a few suggestions:
- Could you add some comments in the PR to explain your design choices and the things that still needs to be unwrapped? While the video explanation is excellent, having written context would facilitate deeper discussions around the implementation
- Would it be possible to add more tests? For example to cover the protocol compat layer and any issue you have encountered that wasn't covered by tests
- Do you have already a rollout strategy in mind? Shall we do some brainstorming about that?

marinoska:
> * Could you add some comments in the PR to explain your design choices and the things that still needs to be unwrapped? While the video explanation is excellent, having written context would facilitate deeper discussions around the implementation

Sure! Will add more comments.

> Would it be possible to add more tests? For example to cover the protocol compat layer and any issue you have encountered that wasn't covered by tests

Will be fixing and adding new tests during next week.

> Do you have already a rollout strategy in mind? Shall we do some brainstorming about that?

We'd better to talk through that a bit, I might not be aware of all the nuances.

Looking into music app now - it's loaded but not re-rendered for some reason...

packages/cojson/src/localNode.ts
```diff
@@ -246,50 +329,46 @@ export class LocalNode {
...
+    returnCore?: false,
+  ): Promise<"unavailable" | T>;
+  async load<T extends RawCoValue>(
+    id: RawCoID,
+    returnCore: true,
+  ): Promise<"unavailable" | CoValueCore>;
+  async load<T extends RawCoValue>(
+    id: RawCoID,
+    returnCore: boolean = false,
+  ): Promise<"unavailable" | CoValueCore | T> {
```
gdorsi: Can we keep the two methods separated?

Having the same method returning two completely different things can be confusing

packages/cojson/src/sync/index.ts
```diff
@@ -0,0 +1,7 @@
+export * from "./types.js";
+export * from "./DataResponseHandler.js";
+export * from "./PushRequestHandler.js";
+export * from "./PullRequestHandler.js";
+export * from "./AckResponseHandler.js";
+export * from "./SyncService.js";
+export * from "./LoadService.js";
```
gdorsi: The usage of barrel modules could potentially create some challenges with circular dependencies and bundling. Would it be possible to switch to direct imports instead?

packages/cojson/src/sync/SyncService.ts
```diff
@@ -0,0 +1,69 @@
+import { CoValuesStore } from "../CoValuesStore.js";
+import { CoValueEntry } from "../coValueEntry.js";
+import { LocalNode } from "../exports.js";
```
gdorsi: We should never import from the `exports` module, it will probably lead to bundling issue due the circular issues.

packages/cojson/src/sync/PullRequestHandler.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * "Pull" request must be followed by "data" message response according to the protocol:
+ * - Sends new content if it exists.
+ * - Sends an empty data message otherwise.
+ * - Sends an empty data message with `{ known: false }` in the message if the `coValue` is unknown by local node.
+ *
+ * Handler initiates a new "pull" requests to load the coValue from peers if it is not known by the node.
+ */
+export class PullRequestHandler extends AbstractMessageHandler {
```
gdorsi, line 21: Could you describe how the MessageHandlers works and where they are used?

Looking around the PR can't find the consumer.

packages/cojson/src/utils/parallelQueueRunner.ts
```diff
@@ -0,0 +1,59 @@
+import { RawCoID } from "../ids.js";
+type DeferredFn = () => Promise<unknown>;
+
+export class ParallelQueueRunner {
```
gdorsi, line 4: Could you help me understand how this works and what problem it's solving?

packages/cojson-storage/src/syncManager.ts
```diff
@@ -175,26 +170,71 @@ export class SyncManager {
     return messageMap;
   }
 
-  handleLoad(msg: CojsonInternalTypes.LoadMessage) {
+  handlePull(msg: CojsonInternalTypes.PullMessage) {
```
gdorsi, line 181: I think we should document which kind of requests the handle* methods are handling, to give some easy-to-access context around the push/pull/data verbs

packages/cojson/src/localNode.ts
marinoska: There is a single place in the codebase `packages/jazz-tools/src/implementation/subscriptionScope.ts:89` which uses returned `CoValueCore` value and all the rest consume returned `RawCoValue`. Maybe I better refactor that single consuming place in order and simplify the `load` method ?
The initial naming seemed a little misleading as `load` and `loadCore` sounded as loading different things but they ware literally `loadCoValueAndReturnCore` and `loadCoValueAndReturnRaw` as their main responsibility was the same - to load the coValue.

packages/cojson/src/sync/index.ts
marinoska: > Would it be possible to switch to direct imports instead?

No problem at all.

packages/cojson/src/sync/PullRequestHandler.ts
marinoska: They are used by `SyncManager` which delegates message processing to a particular handler. 

```ts
  handleSyncMessage(msg: SyncMessage, peer: PeerEntry) {
    const entry = this.local.coValuesStore.get(msg.id);

    let handler: MessageHandlerInterface;
    switch (msg.action) {
      case "data":
        handler = this.dataResponseHandler;
        break;
      case "push":
        handler = this.pushRequestHandler;
        break;
      case "pull":
        handler = this.pullRequestHandler;
        break;
      case "ack":
        handler = this.ackResponseHandler;
        break;
      default:
        throw new Error(
          `Unknown message type ${(msg as unknown as { action: "string" }).action}`,
        );
    }
    return handler.handle({ msg, peer, entry });
  }
```

packages/cojson/src/utils/parallelQueueRunner.ts
marinoska: Used by messages handlers to queue up message processing per coValue so that the messages with the same action type for the same coValue are processed one by one in a separate queue but all such queues run separately from each other.
This is why I removed `await` in `processSyncMessage` where the `handleSyncMessage` is invoked. Everything drops in its own queue.

This way when we receive 10 "data" messages for the same covalue (imagine a stream for a picture) they'll be processed one by one in a single queue. But other message action types also for different covalues will go separately.

I think I need to move `ParallelQueueRunner` under the `sync` folder. Thought a lot about of naming, so pls let me know if you have a better name in mind.

```ts
export abstract class AbstractMessageHandler
  implements MessageHandlerInterface
{
  private readonly queuesRunner = new ParallelQueueRunner();

  handle({ msg, peer, entry }: MessageHandlerInput) {
// here the handle puts the callback to process the msg for particular `msg.id` at the end of the prior ones
    this.queuesRunner.defferPer(msg.id, () =>
      this.routeMessage({ msg, peer, entry }),
    );
  }

```

packages/cojson-storage/src/syncManager.ts
marinoska: This is a storage and handling is quite straightforward  in there. Here is the `handleSyncMessage` in the  storage sync manager.
```ts
  async handleSyncMessage(msg: SyncMessage) {
    switch (msg.action) {
      case "pull":
        return this.handlePull(msg);
      case "data":
        return this.handleData(msg);
      case "push":
        return this.handlePush(msg);
      case "known":
        return this.handleKnown(msg);
    }
  }
  ```
  I think we need to add the general protocol messaging description in notion, don't you? Also I'd improve diagrams/schemas

File: packages/cojson-storage-sqlite/src/transformers.ts [hunk 8]
```diff
@@ -0,0 +1,83 @@
...
+  SyncMessage,
+  unknownDataMessage,
+} from "cojson";
+import CoValueContent = CojsonInternalTypes.CoValueContent;
+import { SQLiteNode } from "./sqliteNode.js";
+
+export const transformOutgoingMessageToPeer = (
+  msg: SyncMessage,
+): SyncMessage[] => {
+  if (SQLiteNode.USE_PROTOCOL2) {
```
marinoska, line 13: @gdorsi this is set in `true` inside `jazz-run` to use new protocol locally. So it'll work as before on remote server as default set to `false`

packages/cojson/src/utils/parallelQueueRunner.ts
gdorsi: > Used by messages handlers to queue up message processing per coValue so that the messages with the same action type for the same coValue are processed one by one in a separate queue but all such queues run separately from each other.

> This way when we receive 10 "data" messages for the same covalue (imagine a stream for a picture) they'll be processed one by one in a single queue. But other message action types also for different covalues will go separately.

Could you help me to understand which advantages we have following this approach?

I'm concerned about the complexty related to ensuring execution order across different message types (push/pull/data) and coValues which is an hard rule for us.

packages/cojson-storage/src/syncManager.ts
gdorsi: I prefer have the documentation co-located on the code when possible.

A brief description of the verbs would suffice in this case, and help occasional contributors to better understand this core part of our codebase.

packages/cojson-storage-sqlite/src/transformers.ts
gdorsi: Not sure if we will ever need to use this transform.

The sqlite storage will be updated along with cojson so the will both speak PROTOCOL_2

Am I missing something?

gdorsi: Also, mind that we have another sqlite adapter for React Native that needs the protocol upgrade.

packages/cojson/src/coValueCore.ts
```diff
@@ -1084,6 +1094,55 @@ export class CoValueCore {
   }) {
     return this.node.syncManager.waitForSync(this.id, options?.timeout);
   }
+
+  addNewContent(content: CoValueContent) {
```
marinoska, line 1098: Moved here from `sync.ts handleNewContent`

packages/cojson/src/localNode.ts
gdorsi: > The initial naming seemed a little misleading as load and loadCore sounded as loading different things 

This was intentional as they are two different things.

The "content" is a view that wraps the CoValueCore and getting it is slower than accessing the core.

The core is the raw value that we use as a unit for the sync process.

The sync process should never access the content (for performance reasons) ouside of the group values that we use to check the permissions.

Having two different names it's meant to make the intent more explicit and avoid loading the content where we shouldn't.

It's ok to `loadCoValueCore` have a cryptic name, we may just add some comments on where it's meant to be used and the difference with `load`.

packages/cojson/src/utils/parallelQueueRunner.ts
marinoska: > I'm concerned about the complexty related to ensuring execution order across different message types (push/pull/data) and coValues which is an hard rule for us.

The code below (from `depencencyService.ts: 63` would ensure the dependencies are loaded in the correct order and this is the only place responsible for downloading dependencies). In this sense the complexity is reduced to this short piece of code and there is only one method responsible for the dependencies loading. 

```ts
    for await (const dependency of unknownDependencies) {
      await this.loadService.loadCoValue(dependency);
    }
```

This wouldn't even work if we had a general `await` in processMessages: 
`await syncManager.handlesyncMessage`.  

As we parallelise the message processing - the execution context won't be suspended by any message awaiting for something. So you can think of message processing independently - one task will be awaiting freely for loadCoValue or whatever else, while others processing incoming messages.
The benefit is that the only parallelism in the code is in the top level `void syncManager.handlesyncMessage()` which would reduce general complexity of all nested code which could be syncronous further on so more readable and controllable. And next time you shouldn't need to think where not to put await  to not delay the incoming messages queue processing

packages/jazz-tools/src/coValues/inbox.ts
```diff
@@ -159,7 +159,7 @@ export class Inbox {
             const id = item.value;
 
             node
-              .load(id)
+              .load<InboxMessage<I, O>>(id)
```
gdorsi: You shouldn't need to add this type because it should be inferred from id.

If you need it's because probably the API signature changed in a way that the infer doesn't work anymore and this is a regression.

packages/cojson/src/peer/transformers.ts
```diff
@@ -0,0 +1,81 @@
+import { unknownDataMessage } from "../exports.js";
+import { SessionID } from "../ids.js";
+
+import { CoValueContent, SyncMessage } from "../sync/types.js";
+
+export const transformOutgoingMessageToPeer = (
+  msg: SyncMessage,
+  id: string,
+): SyncMessage[] => {
+  if (!id.includes("cloud")) {
```
marinoska, line 10: @gdorsi This is how I define if  the message needs to be converted to the older messaging system and I don't know if this simple check would be enough and I don't like such a hardcode. Is there another kind of remote server like jazz.cloud named differently?

packages/jazz-tools/src/coValues/inbox.ts
marinoska: This is because the load vs loadCore thing. I'll bring it back but still consider the `loadCore` name misleading :D

packages/cojson-storage-sqlite/src/transformers.ts
marinoska: I updated the storages, also the FileSystem.ts (recently, to make music-player work:D )

> Also, mind that we have another sqlite adapter for React Native that needs the protocol upgrade.

Good I took out the storages's `SyncManager` into shared `cojson-storage` package previously and which I had to tweak along and it's going to work for all kind of storages inheriting 'DBClientInterface'.

Didn't I miss anything?

packages/cojson/src/localNode.ts
```diff
@@ -66,6 +84,72 @@ export class LocalNode {
...
+                ? v.slice(0, 20) + "..."
+                : v,
+          )}`,
+          { cause: e },
+        );
+      }
+    }
+  }
+
+  async addPeer(peerData: Peer) {
```
marinoska, line 119: Moved addPeer into LocalNode as this is the node's responsibility to be aware of its own peers.

packages/cojson-storage-sqlite/src/transformers.ts
marinoska: > Not sure if we will ever need to use this transform.
The sqlite storage will be updated along with cojson so the will both speak PROTOCOL_2

As I understand the sync server uses SQLiteStorage and speaks PROTOCOL_1 with SQLiteStorage. That's why PROTOCOL_2 is false by default but `jazz-run` sets it in true

packages/cojson/src/coValueCore.ts
```diff
@@ -1118,3 +1177,85 @@ export type TryAddTransactionsError =
...
+): e is TryAddTransactionsException {
+  return (
+    e &&
+    e.type === "TryAddTransactionsError" &&
+    typeof e.message === "string" &&
+    e.error !== undefined
+  );
+}
+
+export function getDependedOnFromContent(msg: Required<CoValueContent>) {
```
marinoska, line 1197: Retrieving dependencies from message data (header/sessions etc) - copy/pasted from storage which duplicates the code. I haven't found a good place to share it from. But I'd not import directly it from the cojson-storage

marinoska:
Just to set some context - I was following the principle of separation of concerns, and most of the decisions were based on that. Generally, I wanted to start from decoupling the actors inside cojson as they intertwined tightly referring to each other which appeared to be highly untrivial task so I rolled back some of my changes and concentrated specifically on making independent at least parts of the sync process. 
This way the `sync` folder appeared, which contains low-level services to be instantiated and used by and within the SyncManager which provides the api for the outer consuming code,  like local node.

packages/cojson/src/coValueEntry.ts
```diff
@@ -100,29 +100,115 @@ type CoValueStateType =
   | CoValueAvailableState
   | CoValueUnavailableState;
 
-export class CoValueState {
+class UploadState {
+  protected peers = new Map<
```
marinoska, line 104: UploadState is to be used by coValueEntry to keep track of uploading status to a certain peer. Used via coValueEntry by methods like waitForUploadIntoPeer

packages/cojson/src/coValueEntry.ts
```diff
@@ -255,64 +354,16 @@ export class CoValueState {
           currentState.markAsUnavailable(action.peerId);
         }
 
+        this.uploadState.unawarePeers.add(action.peerId);
```
marinoska, line 357: We save a "not-found-in" peer id  inside the coValueEntry uploadState temporary - to be able to send the coValue to the peer later in case it's found in the others - the third scenario when we take responsibility to upload the coValue intentionally, apart from initial sync and push new data txs

packages/cojson/src/sync/index.ts
marinoska: Fixed.

packages/cojson/src/sync/LoadService.ts
marinoska: It's a good question. There are 6 dependencies instantiated by sync manager and located inside `./sync` to which it delegates all messaging logic distributed between push/data/ack/pull handlers and a couple of sync and load services. Finally, the load service's become something very simple but its purpose hasn't changed and it's intended to be a sync manager's dependency encapsulated by it. Since we use classes and all services I mentioned above are used uniformly I'd leave it to be a class. As well as they all could be functions either, like in a different js style but I'd go for consistency.
Also I hope this codebase is going to evolve and let's see what it'll become.

packages/cojson-storage-sqlite/src/transformers.ts
gdorsi: > Good I took out the storages's SyncManager into shared cojson-storage

Ah, that's great!

marinoska:
@gdorsi Let's talk about the rollout after we finish going through the PR and find ourselves on the same page then see how the parts of the system are interconnected. I lack the holistic picture of what kind of servers and apps are running currently and I might miss anything.

packages/cojson-storage-sqlite/src/transformers.ts
gdorsi: > As I understand the sync server uses SQLiteStorage and speaks PROTOCOL_1 with SQLiteStorage

The storage is upgraded along with the sync server, so they should always be on the same version.

packages/cojson/src/localNode.ts
```diff
@@ -66,6 +84,72 @@ export class LocalNode {
...
+      }
+      if (msg === "PingTimeout") {
+        console.error("Ping timeout from peer", peer.id);
+        return;
+      }
+      try {
+        if (TRACE_SYNC_MESSAGES) {
+          console.log("🔵 ===>>> Received from", peer.id, msg);
+        }
+        this.syncManager.handleSyncMessage(
```
marinoska, line 100: Important change - removed `await` from the `syncManager.handleSyncMessage` call so each message is processed independently. Asynchronicity is taken on the top level like in a regular API controller. But the internal calls can be all synchronous now (using await)

packages/cojson-storage-sqlite/src/transformers.ts
marinoska: > > As I understand the sync server uses SQLiteStorage and speaks PROTOCOL_1 with SQLiteStorage
> 
> The storage is upgraded along with the sync server, so they should always be on the same version.

What package contains the remote server code?

packages/cojson/src/localNode.ts
marinoska: Done

packages/cojson/src/sync/SyncService.ts
marinoska: Done

packages/cojson-storage/src/syncManager.ts
marinoska: Added

packages/jazz-tools/src/coValues/inbox.ts
marinoska: Done

packages/cojson/src/localNode.ts
```diff
@@ -42,6 +54,13 @@ const { localNode } = useJazz();
 ```
 */
 export class LocalNode {
+  static peers = new Peers();
```
marinoska: This shouldn't be `static` and I'm looking for how to leave `peers` it inside LocalNode but not as a static. The problem was the more properties are on the local node instance the harder to maintain the current copy-into-into-new-node-with-account strategy.
This is why the peers were located on the sync manager instance

PR #1173, opened 2025-01-09
Improve coRichText support

bensleveritt:
Adds more tests, and improves coPlainText and coRichText in preparation for the richtext editor integrations.


packages/jazz-tools/src/tests/coRichText.test.ts
```diff
@@ -34,24 +48,887 @@ describe("Simple CoRichText operations", async () => {
...
+        // Add mark covering "hello"
+        text.insertMark(0, 4, Marks.Strong, { tag: "strong" });
+
+        const marks = text.resolveMarks();
+        expect(marks).toHaveLength(1);
+        expect(marks[0]).toMatchObject({
+          startAfter: 0,
+          startBefore: 1,
+          endAfter: 3,
+          endBefore: 5,
```
bensleveritt, line 66: I think this should be 4 right @aeplay ? There are a few other instances of endBefore being a little off, so I'm wondering if the resolution is correct there. I've accounted for it in the client library, just to marry up the expectation, but I'll need to fix the source lib if that's wrong

packages/cojson/src/coValues/coPlainText.ts
```diff
@@ -88,33 +83,22 @@ export class RawCoPlainText<
...
+      }
+    } else {
+      // For other insertions, use append after the specified index
+      // We append in forward order to maintain the text order
+      let after = idx - 1;
+      for (const grapheme of graphemes) {
+        this.append(grapheme, after, privacy);
+        after++; // Move the insertion point forward for each grapheme
       }
     }
```
gdorsi, line 101: We can try to optimize this using `appendItems` which creates a single transaction.

It should also handle the "reverse" part, so it might help on simplifying the code here

packages/jazz-tools/src/tests/coPlainText.test.ts
```diff
@@ -1,33 +1,161 @@
...
+  describe("Loading and availability", () => {
+    test("can load text across peers", async () => {
+      const { me, text } = await initNodeAndText();
+      const id = text.id;
+
+      // Set up peer connections
+      const [initialAsPeer, secondPeer] = connectedPeers("initial", "second", {
+        peer1role: "server",
+        peer2role: "client",
+      });
```
gdorsi, line 94: You can use `linkAccounts` (from src/testing) to simplfy this:

```
    const spiderman = await Account.create({
      creationProps: { name: "Spiderman" },
      crypto: Crypto,
    });
    
    linkAccounts(me, spiderman);
```

gdorsi (approved):
LGTM

Haven't tested locally, but code wise looks good.

Do we need a changeset?

packages/jazz-tools/src/tests/coPlainText.test.ts
bensleveritt: I'll put the fixes into https://github.com/garden-co/jazz/pull/1163

packages/cojson/src/coValues/coPlainText.ts
bensleveritt: Yes, that makes sense, I'll give that a go

bensleveritt: Actually it doesn't port over, but I think that's a defect in `appendItems` I'll see if I can fix it

PR #1176, opened 2025-01-09
feat: zero-setup auth, anonymous user upgrade and localOnly sync

gdorsi:
Step 1 of the Auth refactoring.

Changes highlights:
- Made the auth prop optional on JazzProvider. When omitted the user is logged as anonymous.
- Made usePasskeyAuth and useClerkAuth usable inside JazzProvider. When done they reuse the existing anonymous user credentials.
- Centralized the browser secrets storage management
- Added localOnly prop to disable the cloud sync

This PR will be merged against a feature branch because it's not complete yet:
- Looking for simplifying the auth flow and improve the error management
- Support Passphrase auth inside JazzProvider
- Deprecate or remove DemoAuth
- Store somewhere the anonymous user credentials when the Auth provider discards them
- Better backward compatibility. Thinking about generating a new keypair when upgrading so we are not binded to the local credentials limits (e.g. PassphraseAuth can't be used to store the Anonymous user credentials)
- Once we agree on the format, will need to migrate React Native, Svelte & Vue to the new auth


examples/music-player/src/3_HomePage.tsx
```diff
@@ -66,6 +71,8 @@ export function HomePage({ mediaPlayer }: { mediaPlayer: MediaPlayer }) {
     });
   };
 
+  const isAnonymousUser = useIsAnonymousUser();
```
gdorsi, line 74: Will add a `useIsLocalOnly` to let devs change the logic based on the sync model.

It also kind-of explains better why share is disabled for anonymous users

examples/music-player/src/components/AuthModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const [username, setUsername] = useState("");
+  const [isSignUp, setIsSignUp] = useState(true);
+
+  const { me } = useAccount();
+
+  const [, authState] = usePasskeyAuth({
+    appName: "Jazz Music Player",
+    onAnonymousUserUpgrade: ({ username, isSignUp }) => {
+      if (isSignUp) {
+        me.profile!.name = username;
```
gdorsi, line 30: Since profile is always available, shall we fix the type to reflect that? @aeplay

packages/jazz-browser/src/auth/AnonymousAuth.ts
```diff
@@ -1,43 +1,37 @@
...
+    public driver: BrowserAnonymousAuth.Driver,
   ) {}
 
   /**
    * @returns A `JazzAuth` object
    */
-  async start() {
-    const existingUser = localStorage[STORAGE_KEY];
+  async start(crypto: CryptoProvider) {
+    AuthSecretStorage.migrate();
```
gdorsi, line 28: Added a `migrate` call to move the auth systems to use the same credentials format

packages/jazz-browser/src/auth/AnonymousAuth.ts
```diff
@@ -51,19 +45,22 @@ export class BrowserOnboardingAuth implements AuthMethod {
         logOut,
       } satisfies AuthResult;
     } else {
+      const secretSeed = crypto.newRandomSecretSeed();
```
gdorsi, line 48: Generating a secret seed to make this compatible with Passkey and Passphrase

packages/jazz-react/src/provider.tsx
```diff
@@ -1,117 +1,204 @@
...
   storage?: BaseBrowserContextOptions["storage"];
   AccountSchema?: AccountClass<Acc>;
 };
 
+export type JazzProviderProps<Acc extends Account = RegisteredAccount> = {
+  children: React.ReactNode;
+  localOnly?: boolean | "anonymous";
+} & Omit<JazzContextManagerProps<Acc>, "localOnly">;
+
+class JazzContextManager<Acc extends Account = RegisteredAccount> {
```
gdorsi, line 32: Refactored to use `useSyncExternalStorage` and use a "propsChanged" check to avoid the double invocation on dev

packages/jazz-browser/src/auth/AuthSecretStorage.ts
```diff
@@ -0,0 +1,96 @@
+import { AgentSecret } from "cojson";
+import { Account, ID } from "jazz-tools";
+
+const STORAGE_KEY = "jazz-logged-in-secret";
```
gdorsi, line 4: Created a secret manager to centralize the session management.

PR #1180, opened 2025-01-10
Fix and test react starter app

trishalim:
Added green highlights. The starter app was missing the account schema prop.
image

Added a test


gdorsi (approved):
Thanks a mill!

PR #1191, opened 2025-01-12
Use example apps as starter with create-jazz-app

trishalim:
Added an example param to create-jazz-app to get started with examples quickly

Added the following to examples README

```
npm create-jazz-app@latest --example chat --project-name chat
```

Updated create-jazz-app to accept partial options (project name only, starter only, etc).


aeplay (approved):
Looks great!

PR #1213, opened 2025-01-17
perf: skip coValue content creation on access

gdorsi:
The coValue content was always created when creating a coValue with a group.

This is an unnecessary operation on the sync servers that we can avoid to reduce the memory footprint and the CPU cost of creating a coValue.

This PR fixes that by centralizing the notify method and ensuring that the content is created only when there are listeners (which should be never the case on sync servers)

### DIFF
--- a/packages/cojson/src/coValueCore.ts
+++ b/packages/cojson/src/coValueCore.ts
@@ -126,10 +126,7 @@ export class CoValueCore {
         .expectCoValueLoaded(header.ruleset.group)
         .subscribe((_groupUpdate) => {
           this._cachedContent = undefined;
-          const newContent = this.getCurrentContent();
-          for (const listener of this.listeners) {
-            listener(newContent);
-          }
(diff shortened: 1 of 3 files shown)


packages/cojson/src/coValueCore.ts
```diff
@@ -126,10 +126,7 @@ export class CoValueCore {
         .expectCoValueLoaded(header.ruleset.group)
         .subscribe((_groupUpdate) => {
           this._cachedContent = undefined;
-          const newContent = this.getCurrentContent();
-          for (const listener of this.listeners) {
-            listener(newContent);
-          }
+          this.notifyUpdate("immediate");
         });
```
gdorsi, line 130: We have already a test on this logic to ensure that I'm not intorducing a regression.

Checking if we can optimize this block by resetting the cache only when required (updates on the readKey and getting the read access)

aeplay (approved):

PR #1215, opened 2025-01-17
fix: improve error management on initial auth, fixed an infinite loop when migration fails

gdorsi:


packages/jazz-tools/src/implementation/createContext.ts
```diff
@@ -128,129 +128,106 @@ export async function createJazzContext<Acc extends Account>(
 ): Promise<JazzContext<Acc>>;
 export async function createJazzContext<Acc extends Account>(
   options: ContextParamsWithAuth<Acc> | BaseContextParams,
-): Promise<JazzContext<Acc>> {
-  // eslint-disable-next-line no-constant-condition
-  while (true) {
```
gdorsi, line 133: Removed the `while (true) {` and the inner try/catch blocks

PR #1216, opened 2025-01-18
fix: phone theme drives light/dark mode in Demo Auth

boorad:
this is huge


bensleveritt:
Thanks for your contribution! It looks good to me, so I'll merge in!

PR #1245, opened 2025-01-22
Extract friendly user name for profile creation when authenticating using Clerk with a regular email address

pax-k:
```
creationProps: {
            name:
              this.clerkClient.user.fullName ||
              this.clerkClient.user.firstName ||
              this.clerkClient.user.username ||
              this.clerkClient.user.primaryEmailAddress?.emailAddress?.split(
                "@",
              )[0] ||
              this.clerkClient.user.id,
            other: {
              email: this.clerkClient.user.primaryEmailAddress?.emailAddress,
            },
          }
```

- friendly name extracted from `this.clerkClient.user.primaryEmailAddress?.emailAddress?.split("@")[0]`
- email passed to `creationProps`
- clear stored clerk credentials

### DIFF
--- a/packages/jazz-browser-auth-clerk/src/index.ts
+++ b/packages/jazz-browser-auth-clerk/src/index.ts
@@ -1,25 +1,8 @@
+import type { Clerk } from "@clerk/clerk-js";
 import { AgentSecret } from "cojson";
 import { Account, AuthMethod, AuthResult, Credentials, ID } from "jazz-tools";
 
 const localStorageKey = "jazz-clerk-auth";
-
-export type MinimalClerkClient = {
-  user:
-    | {
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        unsafeMetadata: Record<string, any>;
-        fullName: string | null;
-        username: string | null;
-        id: string;
-        update: (args: {
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          unsafeMetadata: Record<string, any>;
-        }) => Promise<unknown>;
-      }
-    | null
-    | undefined;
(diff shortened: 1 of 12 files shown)


packages/jazz-browser-auth-clerk/src/index.ts
```diff
@@ -1,26 +1,9 @@
+import type { Clerk } from "@clerk/clerk-js";
```
gdorsi: If we depend on the type this is not a devDependency but an actual dependency.

I guess the `MinimalClerkClient` type was inlined with the intention of avoiding that.

Could you revert the change and update the local type?

gdorsi (changes requested):

packages/jazz-browser-auth-clerk/src/index.ts
```diff
@@ -96,7 +79,11 @@ export class BrowserClerkAuth implements AuthMethod {
           creationProps: {
             name:
               this.clerkClient.user.fullName ||
+              this.clerkClient.user.firstName ||
               this.clerkClient.user.username ||
+              this.clerkClient.user.primaryEmailAddress?.emailAddress?.split(
+                "@",
+              )[0] ||
```
gdorsi, line 125: Could you cover this change with a test?

packages/jazz-browser-auth-clerk/src/index.ts
```diff
@@ -96,7 +79,11 @@ export class BrowserClerkAuth implements AuthMethod {
...
               this.clerkClient.user.id,
           },
```
trishalim, line 130: @pax-k While you're at it, can you please also store the emailAdress?

PR #1270, opened 2025-01-24
feat: improve the clerk integration

gdorsi:


aeplay (approved):
I think this is an OK approach! Thanks!

PR #1279, opened 2025-01-25
feat: e2e-rn-cli

boorad:


boorad:
we can re-open if we ever want a non-Expo e2e app

PR #1280, opened 2025-01-26
Update README.md

doingandlearning:
The current jazz-vue package describes itself as bindings for React. 

Update to fix this.


gdorsi (approved):
Thanks for the fix!

PR #1291, opened 2025-01-27
feat: make auth state platform agnostic

gdorsi:
In order to simplify the maintenance of our auth logic I'm moving some parts to jazz-tools.

This should increase the amount of code shared between our platforms (Browser and React Native).

Changes:
- Refactor AuthSecretStorage to use a KVStore instead of localStorage
- Made AuthSecretStorage a class, to manage the listeners/auth state locally in a context
- Moved DemoAuth and PassphraseAuth inside jazz-tools so they can be reused on ReactNative
- Moved the auth logic down to jazz-tools
- Added tests


packages/jazz-browser/src/BrowserContextManager.ts
```diff
@@ -21,6 +27,15 @@ export type JazzContextManagerProps<Acc extends Account> = {
 export class JazzBrowserContextManager<
   Acc extends Account,
 > extends JazzContextManager<Acc, JazzContextManagerProps<Acc>> {
+  getKvStore() {
+    if (typeof window === "undefined") {
+      // To handle running in SSR
+      return new InMemoryKVStore();
```
gdorsi, line 31: This should make it easier to run Jazz in a SSR context

packages/jazz-browser/src/createBrowserContext.ts
```diff
@@ -32,15 +29,13 @@ setupInspector();
...
   node: LocalNode;
-  toggleNetwork: (enabled: boolean) => void;
   logOut: () => void;
   done: () => void;
 };
 
 export type BrowserGuestContext = {
   guest: AnonymousJazzAgent;
   node: LocalNode;
-  toggleNetwork: (enabled: boolean) => void;
```
gdorsi, line 43: Removed this API from the public context.

Now the anonymous localOnly switch is handled internally and if the user wants to manually control "localOnly" we re-create the context.

Done this to keep the complexity low, if we see that manual control over localOnly becomes a common pattern we can improve it.

packages/jazz-browser/src/createBrowserContext.ts
```diff
@@ -163,50 +153,51 @@ export async function createJazzBrowserContext<Acc extends Account>(
...
+      if (isAuthenticated) {
+        toggleNetwork(true);
+      } else {
+        toggleNetwork(false);
+      }
+    }
+
+    unsubscribeAuthUpdate = authSecretStorage.onUpdate(handleAuthUpdate);
+    handleAuthUpdate(authSecretStorage.isAuthenticated);
+  }
```
gdorsi, line 158: Moved the auth based localOnly toggle inside here to make it easier to maintain.

packages/jazz-react-core/src/auth/DemoAuth.tsx
```diff
@@ -0,0 +1,43 @@
...
+ * `useDemoAuth` is a hook that provides a `JazzAuth` object for demo authentication.
+ *
+ *
+ * ```ts
+ * const { state, logIn, signUp, existingUsers } = useDemoAuth();
+ * ```
+ *
+ * @category Auth Providers
+ */
+export function useDemoAuth() {
```
gdorsi, line 16: This is now on `jazz-react-core` so we can use the same hook on both RN and web

packages/jazz-react-core/src/auth/PassphraseAuth.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * `usePassphraseAuth` hook provides a `JazzAuth` object for passphrase authentication.
+ *
+ * @example
+ * ```ts
+ * const [auth, state] = usePassphraseAuth({ appName, appHostname, wordlist });
+ * ```
+ *
+ * @category Auth Providers
+ */
+export function usePassphraseAuth({
```
gdorsi, line 16: This is now on jazz-react-core so we can use the same hook on both RN and web

packages/jazz-react/src/auth/PassphraseAuth.tsx
```diff
@@ -1,56 +1,31 @@
...
-  } as const;
-}
+import { usePassphraseAuth } from "jazz-react-core";
+import { useState } from "react";
 
 export const PassphraseAuthBasicUI = (
   props: ReturnType<typeof usePassphraseAuth>,
 ) => {
-  const { logIn, signUp, generateRandomPassphrase } = props;
+  const { logIn, signUp } = props;
```
gdorsi, line 7: Thinking about removing this BasicUI and build an example.

I think it would be more effective on showing how best use this Auth method and now that auth is not mandatory anymore it's ok to not have a one-click setup.

packages/jazz-svelte/src/lib/tests/Provider.svelte.test.ts
```diff
@@ -1,120 +0,0 @@
-import { describe, it, expect, vi, beforeEach } from 'vitest';
```
gdorsi, line 1: Going to cover this kind of components with a browser-based test, otherwise it's very hard to maintain.

packages/jazz-tools/src/auth/AuthSecretStorage.ts
```diff
@@ -0,0 +1,109 @@
...
+};
+
+export class AuthSecretStorage {
+  private listeners: Set<(isAuthenticated: boolean) => void>;
+  public isAuthenticated: boolean;
+
+  constructor() {
+    this.listeners = new Set();
+    this.isAuthenticated = false;
+  }
```
gdorsi, line 23: Made this a class to handle the `isAuthenticated` state here and not make it global.

The change was driven by the switch to the async API that made reading from storage not possible in the init phase of the UI components.

PR #1294, opened 2025-01-29
feat: update web examples to new auth system

boorad:
Change the example web apps to use new auth system.

> I'm not sure I got all of the `localOnly="anonymous"` and removing `DemoAuthBasicUI` correct...

After this commit, `pnpm build` was able to complete successfully.

### DIFF
--- a/examples/password-manager/src/2_main.tsx
+++ b/examples/password-manager/src/2_main.tsx
@@ -6,21 +6,18 @@ import App from "./5_App.tsx";
 import "./index.css";
 
 function JazzAndAuth({ children }: { children: React.ReactNode }) {
-  const [auth, state] = usePasskeyAuth({
+  const auth = usePasskeyAuth({
     appName: "Jazz Password Manager",
   });
 
   return (
-    <>
-      <JazzProvider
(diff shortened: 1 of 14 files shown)


examples/password-manager/src/2_main.tsx
```diff
@@ -6,21 +6,18 @@ import App from "./5_App.tsx";
...
-        {children}
-      </JazzProvider>
-      <PasskeyAuthBasicUI state={state} />
-    </>
+    <JazzProvider
+      AccountSchema={PasswordManagerAccount}
+      peer="wss://cloud.jazz.tools/?key=password-manager-example-jazz@garden.co"
+    >
+      <PasskeyAuthBasicUI {...auth} />
+      {children}
```
gdorsi, line 19: To keep the code simple I've added a requirement for the auth hooks to run inside the provider.

Since the builtin UI components are kind-of blocking I think we can just mount the hook inside there to simplify the setup and migrations.

Going to push a fix!

gdorsi (approved):
Thanks!

This helped me to find some improvement areas  on the new APIs!

PR #1296, opened 2025-01-29
feat: improve builtin auth UIs

gdorsi:
Mostly fixed the broken tests and polished the example apps to better leverage the new APIs.

The most notable change is related to the builtin auth UIs:
- Moved the auth hooks inside the UI components. This way the migration will become easier because devs will be less likely driven on setting the auth hook outside the JazzProvider. It's also a better pattern to suggest when the devs need an auth wall.
- Made the passphrase auth more aligned to a "realistic" use case (see  https://jazz-todo-git-authv2-example-apps-gcmp.vercel.app/#/)
- Added a passphrase minimal example
- Added passphrase & passkey hooks on Vue

Thinking about refactoring the auth minimal examples to use a custom UI to better show how to use the auth hooks.


aeplay (approved):
Makes it clear how much simpler the new flow is!

PR #1298, opened 2025-01-29
Change isMe to indicate whether an account is the currently active account

Elfo404:
Also adds `isLocalNodeOwner` to replace the previous behavior to check if an account is the local node owner.


aeplay (approved):
Looks good to me but please make it patch change - anything relying on the old broken behaviour is considered a bug

PR #1301, opened 2025-01-29
Jazz 679 Move wasm-crypto code into rust lib

bensleveritt:
Ports much of the external WASMCrypto dependancies to our own crypto rust lib, allowing treeshaking and reducing the crossing of the WASM boundary.

Particularly `@hazae41/berith` (x25519 and ed25519) and `hash-wasm` (blake3)


bensleveritt:
Closing in favour of creating a new repo

packages/cojson/src/crypto/WasmCrypto.ts
```diff
@@ -54,93 +66,83 @@ export class WasmCrypto extends CryptoProvider<Uint8Array> {
...
-      .to_bytes()
-      .copyAndDispose();
+    const signingKey = ed25519_signing_key_from_bytes(
+      base58.decode(secret.substring("signerSecret_z".length)),
+    );
+    const signature = ed25519_signing_key_sign(
+      signingKey,
+      textEncoder.encode(stableStringify(message)),
+    );
     return `signature_z${base58.encode(signature)}`;
```
gdorsi, line 119: Can we make this a single sign call to WASM? @bensleveritt 

Same goes for the verify, the less roundtrips we do to WASM memory the better.

I mean putting `ed25519_signing_key_from_bytes`, `ed25519_signing_key_sign` and `base58.encode` behind a single sign call:

```ts
const signature = jazz_ed25519_sign(secret.substring("signerSecret_z".length), stableStringify(message))

return `signature_z${signature}`
```

bensleveritt: Yeah definitely. I'll just do it in the next PR. I hesitated as the base58 lib isn't wasm, but the fewer boundary movements would indeed be better

PR #1302, opened 2025-01-29
fix: handle websocket errors and add onSuccess callback

gdorsi:
Adding an error handler to make the WebSocket restart correctly when the server connection goes down on workers.

Added an onSuccess callback for future use on managing the client reconnections (part of AuthV2)


packages/cojson-transport-ws/src/tests/syncServer.ts
```diff
@@ -0,0 +1,89 @@
+import { createServer } from "http";
+import { ControlledAgent, LocalNode, WasmCrypto } from "cojson";
+import { WebSocket, WebSocketServer } from "ws";
+import { createWebSocketPeer } from "../index";
+
+export const startSyncServer = async (port?: number) => {
```
gdorsi, line 6: Added a test sync server to write the integration tests.

aeplay (approved):
😍

PR #1304, opened 2025-01-30
Add framework and API key parameter to create jazz app

trishalim:


packages/create-jazz-app/src/index.ts
```diff
@@ -147,6 +155,50 @@ async function scaffoldProject({
...
+        "src/routes/+layout.svelte",
+      ];
+
+      for (const file of filesToSearch) {
+        const filePath = `${projectName}/${file}`;
+        if (fs.existsSync(filePath)) {
+          let content = fs.readFileSync(filePath, "utf8");
+          // Replace any Jazz API key pattern
+          const keyPattern = /key=([^&\s"']+)/g;
+          const updatedContent = content.replace(keyPattern, `key=${apiKey}`);
```
gdorsi: @trishalim 

I think this approach is too flaky, because nothing guarantees that our future changes don't break this replacement.

What about having a `src/apiKey.ts` file in each example/starter where we dump our key?

By default it can be:
```ts
export const API_KEY = "you@example.com"
```

So we can just dump the whole file without need of pattern matching or file lists.

trishalim: Agreed

gdorsi (approved):
Thanks for the cleanup!

Would it be possible to target the `authv2` branch?

We can ship this into the canary release and ensure that it's compatible with the next version of the example apps.

trishalim:
@gdorsi I changed the base to `authv2` and moved the api keys into `src/apiKey.ts`

aeplay (approved):
Very nice!

PR #1307, opened 2025-01-30
Jazz 679 Use jazz-crypto-rs

bensleveritt:
Most of the fun is over [here](https://github.com/garden-co/jazz-crypto-rs/pull/1)

Ports the functions of WASMCrypto to our own `jazz-crypto-rs` lib to reduce WASM boundary crossing, and dependencies on WASM libs.


gdorsi (approved):
Well done!

PR #1308, opened 2025-01-30
feat: API to handle the authenticate on top of an anonymous user

gdorsi:
This PR introduces a `onAnonymousUserDiscarded` handler to pass the data to the logged user before the data on the anonymous user gets lost.

Got the idea of adding a `transferOwnership` API on Account, which would make the things way easier because it doesn't require a good permissions design/understanding.

Tried that road, but it isn't a quick win so I've decided to set for a dumb API that's hard to use but gets the job done.

Howerver once works it is wonderful:

https://github.com/user-attachments/assets/81fff4ef-43d2-4397-b30f-3630d557a7af


aeplay (approved):
LGTM!

PR #1313, opened 2025-02-01
feat: add Maestro e2e tests for RN

boorad:
This adds [Maestro](https://maestro.mobile.dev/) end-to-end tests that click through an app and assert what is on the screen of a simulator.

To test Jazz functionality, it exercises the `chat-rn` example application through login, logout, starting a chat, sending a message, and getting an invite code.

The Github Actions workflow builds the app with `xcodebuild` in a `release` variant so we don't have to run the Metro bundler.  This may add some time to the execution, but simplifies the flow, as we don't have to deal with Expo startup screens and grabbing Metro logs for diagnosing errors, etc.  This can be changed in the future if the runtimes are too long.

This change also consolidated some of the source code dependency install and build steps into a composite action that can be shared across workflows.


.github/workflows/build-examples.yaml
```diff
@@ -23,41 +23,13 @@ jobs:
...
-        key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
-        restore-keys: |
-          ${{ runner.os }}-pnpm-store-
-
-
-    - name: Install dependencies
-      run: pnpm install --frozen-lockfile
-
+    - name: Setup Source Code
+      uses: ./.github/actions/source-code/
```
gdorsi, line 32: Nice touch, thanks!

examples/chat-rn/test/e2e/erase_text.yml
```diff
@@ -0,0 +1,15 @@
+appId: com.jazz.chatrn
+---
+- copyTextFrom:
+   id: ${id}
+- repeat:
+    times: 4
```
gdorsi, line 10: Why 4 times?

.github/workflows/e2e-rn-test.yml
```diff
@@ -0,0 +1,86 @@
+name: End-to-End Tests for React Native
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  push:
+    branches:
+      - 'main'
```
gdorsi: We could filter the action to only when the RN examples/packages are changed

gdorsi (commented):
Great job!

examples/chat-rn/test/e2e/erase_text.yml
boorad: Just to be sure we got all the existing text.  I had a `while {condition true}` for a bit, and that ran forever once.

The underlying bug is noted in Maestro docs.  Initial click into a textinput is in the middle, and if the text is longer than 50%, only that left half is erased.

.github/workflows/e2e-rn-test.yml
boorad: That won't test changes that occur (far more often) in the core Jazz code, tho :(

examples/chat-rn/test/e2e/erase_text.yml
gdorsi: Thanks for the explaination!

Could you add a little comment on the code? So we won't forget the why.

.github/workflows/e2e-rn-test.yml
gdorsi: I know that's not ideal, but we can't run a 30m pipeline on each PR.

But while we would skip the action on PRs, filtering by RN examples/packages ensures that the test runs in the "Version Packages" PR when one of their dependency changes.

This should keep us safe from releasing broken packages for React Native, which is good enough for now.

boorad: yeah, fine by me.  I realize 30 mins was a no-go.  That's why I was saying maybe merges into main (before releases).  But this sounds like it will catch things quickly enough to get back to the original author while it's fresh in their mind.

examples/chat-rn/test/e2e/erase_text.yml
boorad: done

gdorsi (approved):
LGTM 🚀

PR #1319, opened 2025-02-04
feat: make ensureLoaded throw when the resolved value is undefined

gdorsi:


packages/jazz-tools/src/coValues/interfaces.ts
```diff
@@ -196,16 +196,22 @@ export function loadCoValue<V extends CoValue, Depth>(
...
     existing.id,
     existing._loadedAs,
     depth,
   );
+
+  if (!response) {
+    throw new Error("Failed to deeply load CoValue " + existing.id);
+  }
+
+  return response;
```
gdorsi, line 214: @aeplay 

Wondering if we should expose a `queryResolved` method that returns undefined/null when not found

PR #1320, opened 2025-02-04
feat: upgrade vitest to v3 and vite to v6

gdorsi:
- Upgrading vitest to be able to use the browser mode for the integration tests.
- Dropped node 14 polyfill for globalThis.crypto to remove a Vite 6 warning.
- Migrated the indexed db tests on browser mode


packages/cojson/src/crypto/WasmCrypto.ts
```diff
@@ -38,23 +38,9 @@ export class WasmCrypto extends CryptoProvider<Uint8Array> {
...
-      createBLAKE3(),
-      initBundledOnce(),
-      new Promise<void>((resolve) => {
-        if ("crypto" in globalThis) {
-          resolve();
-        } else {
-          return import(/*webpackIgnore: true*/ "node:crypto").then(
-            ({ webcrypto }) => {
-              // eslint-disable-next-line @typescript-eslint/no-explicit-any
-              (globalThis as any).crypto = webcrypto;
```
bensleveritt, line 51: Sweet, I did wonder if we could remove this

bensleveritt:
Are there any speed implications of running unit tests in browsers? Also, does this impact the e2e tests?

gdorsi:
> Are there any speed implications of running unit tests in browsers? Also, does this impact the e2e tests?

The ones running on the browser are generally slower, but not by a significant margin.

Generally I think we need this kind of tests to cover the edge cases on the integration between storage & network in the browsers.

We need something realistic for that, so I think it's ok to take a complexity hit for that.

PR #1324, opened 2025-02-05
test: cover browser sync scenarios with tests

gdorsi:
Changes:
- Add browser integration tests to check if everything still works when the connection plays pranks
- return a Promise to wait for the credentials clear when calling logOut


packages/jazz-browser/src/storageOptions.ts
```diff
@@ -17,7 +17,7 @@ export function getStorageOptions(storage?: StorageConfig): {
     !storage ||
     (Array.isArray(storage) && storage.includes("indexedDB")) ||
     storage === "indexedDB" ||
-    !useSingleTabOPFS;
+    (!useSingleTabOPFS && storage !== "disabled");
```
gdorsi: Added the disabled option to storage to be able to test the sync without a shared storage between accounts.

packages/jazz-browser/src/createBrowserContext.ts
```diff
@@ -139,7 +139,7 @@ export async function createJazzBrowserGuestContext(
       context.done();
     },
     logOut: () => {
-      context.logOut();
+      return context.logOut();
```
gdorsi, line 142: The tests showed up that the logOut wasn't waiting for the credentials to be deleted...

PR #1337, opened 2025-02-10
feat: add role mapping to group extension

gdorsi:
Adding role mapping to group extension.

When setting a mapped role all the members with inheritable roles from the parent (admin, writer and reader) are added to the child group with the mapped role.

The mapped role always wins over the parent role and follows the same "most-permissive" rules as the inherited roles when calculating the member role on the child group.


homepage/homepage/app/(docs)/docs/[framework]/[...slug]/groups/inheritance.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+When a group extends another group, members of the parent group will become automatically part of the child group.
+
+## Basic Usage
+
+Here's how to extend a group:
+
+<CodeGroup>
+```typescript
+const playlist = Group.create();
```
Comment (aeplay) [inline homepage/homepage/app/(docs)/docs/[framework]/[...slug]/groups/inheritance.mdx]:
```suggestion
const playlistGroup = Group.create();
```

Can we call all of them ..Group please? And maybe show how the groups are used as an owner on something?

I know it's a bit tedious but as-is it's confusing

homepage/homepage/app/(docs)/docs/[framework]/[...slug]/groups/inheritance.mdx
```diff
@@ -0,0 +1,161 @@
...
+track.extend(playlist);
+```
+</CodeGroup>
+
+When you extend a group:
+- Members of the parent group get access to the child group
+- Their roles are inherited (with some exceptions, see [below](#role-inheritance-rules))
+- Removing a member from the parent group also removes their access to child groups
+
+## Setting the extension role
```
Comment (aeplay) [inline homepage/homepage/app/(docs)/docs/[framework]/[...slug]/groups/inheritance.mdx]:
```suggestion
## Inheriting members but overriding their role
```

homepage/homepage/app/(docs)/docs/[framework]/[...slug]/groups/inheritance.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+You can set the extension role by passing a second argument to `extend`:
```
Comment (aeplay) [inline homepage/homepage/app/(docs)/docs/[framework]/[...slug]/groups/inheritance.mdx]:
```suggestion
In some cases you might want to inherit all members from a parent group but override/flatten their roles to the same specific role in the child group. You can do so by passing an "override role" as a second argument to `extend`:
```

homepage/homepage/app/(docs)/docs/[framework]/[...slug]/groups/inheritance.mdx
```diff
@@ -0,0 +1,161 @@
...
+const organization = Group.create();
+
+const billing = Group.create();
+
+// This way the members of the organization can only read the billing data
+billing.extend(organization, "reader"); 
+```
+</CodeGroup>
+
+The extension role always overrides the role of the accounts in the parent group:
```
Comment (aeplay) [inline homepage/homepage/app/(docs)/docs/[framework]/[...slug]/groups/inheritance.mdx]: Now that we explain this in the intro we can combine the preceding and the following listing and delete this sentence

homepage/homepage/app/(docs)/docs/[framework]/[...slug]/groups/inheritance.mdx
```diff
@@ -0,0 +1,161 @@
...
+// Remove member from parent
+await parentGroup.removeMember(bob);
+
+// Bob loses access to both parent and child groups
+```
+</CodeGroup>
+
+## Role Inheritance Rules
+
+If the account is already a member of the child group, it will get the highest possible role:
```
Comment (aeplay) [inline homepage/homepage/app/(docs)/docs/[framework]/[...slug]/groups/inheritance.mdx]:
```suggestion
If the account is already a member of the child group, it will get the more permissive role:
```

homepage/homepage/app/(docs)/docs/[framework]/[...slug]/upgrade/0-10-0.mdx
```diff
@@ -355,3 +356,22 @@ const { worker } = await startWorker({
 ```
 </CodeGroup> 
 </div>
+
+<h3 id="group-inheritance">Group inheritance with role mapping</h3>
+<div>
+  You can set the extension role by passing a second argument to `extend`.
```
Comment (aeplay) [inline homepage/homepage/app/(docs)/docs/[framework]/[...slug]/upgrade/0-10-0.mdx]:
```suggestion
  You can override the inherited role by passing a second argument to `extend`.
```

aeplay (changes requested):

homepage/homepage/app/(docs)/docs/[framework]/[...slug]/groups/inheritance.mdx
Comment (gdorsi) [reply homepage/homepage/app/(docs)/docs/[framework]/[...slug]/groups/inheritance.mdx]: Changed this sentence and the example into:

```
The "override role" works in both directions:

<CodeGroup>
  ```typescript
  const parentGroup = Group.create();
  parentGroup.addMember(bob, "reader");
  parentGroup.addMember(alice, "admin");
  
  const childGroup = Group.create();
  childGroup.extend(parentGroup, "writer");
  
  // Bob and Alice are now writers in the child group
 /```
</CodeGroup>
```

Comment (gdorsi) [reply homepage/homepage/app/(docs)/docs/[framework]/[...slug]/groups/inheritance.mdx]: I might look redundant, but I want to make this behavoir more clear and preserve the billing example which is the main driver of the whole feature.

aeplay (approved):

PR #1338, opened 2025-02-10
fix: disallow extra props in the resolve type

gdorsi:


packages/jazz-tools/src/coValues/deepLoading.ts
```diff
@@ -145,7 +149,7 @@ export type DeeplyLoaded<
   CurrentDepth extends number[] = [],
 > = DepthLimit extends CurrentDepth["length"]
   ? V
-  : Depth extends true | undefined
+  : Depth extends boolean | undefined // Checking against boolean instead of true because the inference from RefsToResolveStrict transforms true into boolean
```
gdorsi, line 152: Applying any kind of `extends` check on the generic R makes Typescript change the inference of `true` to boolean.

Screenshot 2025-02-10 at 19 53 27

Had to change this here to avoid getting broken types from Resolved:
Screenshot 2025-02-10 at 19 06 05

Since the type becomes { $each: boolean } the if we don't stop the `false` types here the `DeeplyLoaded` type would continue the recursion until reaching the limit of 5.

gdorsi:
Merging to continue wokring on the feature branch, feedback will be handled in a separate PR

PR #1350, opened 2025-02-11
Fix: docs in main nav doesn't get highlighted when active

trishalim:


gdorsi (commented):
Screenshot 2025-02-11 at 11 40 44

Checked the homepage preview, the issue is still there

trishalim:
@gdorsi Dunno what's up with that, because it's working for me locally.
image

homepage/design-system/src/app/components/organisms/Footer.tsx
```diff
@@ -100,7 +101,7 @@ function FooterLink({
       className={clsx(
         "py-0.5 px-0 text-sm",
         className,
-        path === href
+        isActive(href)
```
gdorsi, line 104: @trishalim where does this `href` prop comes from?

Could it be that the problem is that it contains the full origin in production? (e.g. `https://jazz.tools/docs/react`)

trishalim: [It works now](https://jazz-homepage-git-jazz-701-docs-in-main-nav-doesnt-b2c7f7-gcmp.vercel.app/docs/react)
Just needed to do a redeploy on vercel :D

PR #1353, opened 2025-02-11
Mark vue and svelte as experimental

trishalim:


gdorsi (approved):
Looks good! Thanks!

gdorsi:
@trishalim 

Kind-of realized it after merging, but the menu width is not aligned with the dropdown trigger:

Screenshot 2025-02-11 at 16 04 10

While aestetically good I think it would be better for UX having the menu as the same size of the button, to require only vertical mouse movement to select an option

PR #1359, opened 2025-02-12
feat: remove the unstable_enablePackageExports requirement from RN

gdorsi:
To remove the unstable_enablePackageExports requirement from RN I've modified cojson and jazz-tools builds to output a single dist and exported the Crypto providers directly from their module.


.github/workflows/pre-release.yml
```diff
@@ -19,4 +19,90 @@ jobs:
         run: pnpm turbo build --filter="./packages/*"
 
       - name: Pre publish
-        run: pnpm exec pkg-pr-new publish "./packages/*"
+        run: pnpm exec pkg-pr-new publish --json output.json --comment=off "./packages/*"
+
+      - name: Post or update comment
+        uses: actions/github-script@v6
+        with:
```
gdorsi, line 26: Improved the pre-release comment to be able to use the pre-release version more easily.

Now it is enough to set something like:
```
"jazz-react": "https://pkg.pr.new/garden-co/jazz/jazz-react@4bc5bd0a7bf227774265528e61cc6e7031a80add",
"jazz-tools": "https://pkg.pr.new/garden-co/jazz/jazz-tools@4bc5bd0a7bf227774265528e61cc6e7031a80add"
```

In the package.json deps because the URL format is the same of the internal deps.

aeplay (approved):

PR #1366, opened 2025-02-12
fix: correctly setup the metro config on React Native templates

gdorsi:


gdorsi (commented):
Tested locally:

Screenshot 2025-02-12 at 19 31 07

PR #1372, opened 2025-02-13
Jazz 705 rename e2e test scripts

bensleveritt:
A general scattergun of test fixes that generally make turbo and vitest work faster, particularly from the root level.

- *Importantly*: `turbo test` now runs all unit/integration tests in any project that has them, and not `e2e` (which are distracting and slow 🥱 ).
- `tests/browser-integration` runs headless. Definitely. For sure.
- Turbo tests now run like CI, ie. not test watching.


vite.config.ts
```diff
@@ -4,7 +4,7 @@ import { defineConfig } from "vitest/config";
 export default defineConfig({
   root: "./",
   test: {
-    workspace: ["packages/*", "tests/browser-integration"],
+    workspace: ["packages/*"],
```
gdorsi: This is going to exclude the integration tests from the CI.

I also like that the integration tests runs when I type `pnpm test` on the root, because they help alot on catching regressions.

Why did you remove that?

bensleveritt: As I understand it Vite treats the contents of the folder as a src package and that breaks Turbo's test dependency resolution (running `turbo test` in the root doesn't pick up the dependencies before running certain package's tests which if you haven't built recently will cause mystery build errors, though it's just the build dependencies not resolving).

Vitest doesn't need it to be there to run the tests (`pnpm test` in the package and `turbo test` at the root still run the tests). And they run in CI because we use `turbo test` from root.

bensleveritt: Welp, we don't use `turbo test` in the unit-test workflow... Investigating.

bensleveritt: Reverting for now

PR #1376, opened 2025-02-13
Remove redundant passkey auth starter in create-jazz-app

trishalim:
image

- Since the react starter app (previously called Demo Auth) has been changed to use anon auth + passkey, I am renaming it to passkey in create-jazz-app and removing the old passkey starter app.
- Changed "Choose a starter" to "Choose an authentication method"


gdorsi (approved):
Looks good

PR #1379, opened 2025-02-14
Jazz 409 passkeyauth provide fallback api for moving between vendors

### LINKED ISSUE #581: PasskeyAuth: provide fallback API for moving between vendors
While Passkeys provide robust security, they present limitations for users needing to access their accounts across multiple devices, as they are bound to specific OS/browser environments.

We should provide a secure fallback API that devs can use to provide users a way to authenticate to another device without using Passkeys.

This solution can be similar to the Passphrase login but without the wordlist requirement.

Ideally it should be something that can be transmitted by URL so it would be possible to authenticate using QRCodes

bensleveritt:
This is an early pass at supporting users who have passkey authentication on a device, but want another device to be authenticated.

Still trying to work out the scope of what this PR will be, as I think there are lots of ways we can use it.

Use cases include: 
- mobile passkey -> desktop device (ie. authenticate on Windows when Passkey is on iOS.
- desktop passkey -> mobile (sign up on desktop, access from mobile)
- desktop -> desktop (moving from passkey to a different auth)
- "temporary" access (probably mobile -> desktop) (logging into a public machine with mobile passkey)
- migrating passkey vendors (out of scope?)

I'm particularly interested in the idea of moving the authentication from one device to another, using a CoMap to act as the actual authentication intermediary, but that also could be out of scope here.

fixes #581


packages/jazz-tools/src/auth/SecretURLAuth.ts
```diff
@@ -0,0 +1,165 @@
...
+ * @returns The authentication URL.
+ */
+export function createAuthURL(secret: string, expiresAtParam?: number) {
+  const expiresAt = expiresAtParam ?? Date.now() + 15 * 60 * 1000;
+  const payload = JSON.stringify({ s: secret, t: expiresAt });
+  const encoded = bytesToBase64url(new TextEncoder().encode(payload)).replace(
+    /=/g,
+    "",
+  );
+  return `${window.location.origin}/auth#${encoded}`;
```
gdorsi: Since we want this to be used on mobile (ergo React Native) we can't depend directly on window and the origin must be injected.

gdorsi (commented):
I like the idea of the CoMap based exchange.

We have two devices, Bob's desktop and Bob's tablet.

Bob uses passkey on desktop and wants to login on tablet, but they have different vendors (Windows and Mac).

So we are going to implement the following flow to auth the tablet:
1. on desktop we generate an empty CoMap that will be used to transfer the credentials
2. the device generates a writeOnly invite to that CoMap and shows it via QRCode. (we are using writeOnly because this way the other device can write in the CoMap but can't read it's content)
3. the tablet scans the QRCode and accepts the invite with a temporary agent. We want to use a temporary agent to reveal the credentials on to a throwaway account and don't risk that they are exposed to any persistent account.
4. then, if everything happens within the expiration period, the desktop upgrades the new writeOnly member to become a writer (so they can see the CoMap content) and write the account credentials in the CoMap
5. then, the tablet gets the credentials and uses them to authenticate

This way the expiration time is checked by the main device and can't be tricked.

WDYT @bensleveritt @aeplay

aeplay:
I like this, nice and simple, uses existing mechanisms

packages/jazz-tools/src/auth/SecretURLAuth.ts
jmsv: ✅ 2ab276b23f0b40f8656c9da45e160be7f8866655

jmsv:
thinking of this as a first-pass for now – some follow-up stuff:

- #1662
- maybe supporting this flow in reverse: logged-in device scans the QR code, not the other way around (like Discord, Whatsapp etc.) – [discussion](https://discord.com/channels/1139617727565271160/1349319303886929995/1350039114161852487)
- might be worth rethinking the API – as-of this PR it uses the invites flow (`createInviteLink` and `useAcceptInvite`) but this might be fiddly if the app's `/invite` page requires the user to be logged-in (also [#1662](https://github.com/garden-co/jazz/issues/1662) means the logged-in account wouldn't be accepting the invite)

jmsv:
closing to restart with [these changes](https://github.com/garden-co/jazz/pull/1379#issuecomment-2733161493) in mind

PR #1382, opened 2025-02-14
fix: Add back ability to use different crypto providers for RN

boorad:
`CryptoProvider` is reinstated as an optional attribute for `<JazzProvider>` and associated places.

`RNQuickCrypto` is exported from `jazz-react-native` - this doesn't follow the pattern of others, as they are all located in `cojson/crypto`.

Also `chat-rn` example app is converted to use the `RNQuickCrypto` as its `CryptoProvider`, which will increase testing on the experimental module.   Quick Crypto version is bumped to latest everywhere.

Finally, docs are updated to show the option is available for RN apps.


homepage/homepage/app/(docs)/docs/[framework]/[...slug]/project-setup/react-native.mdx
```diff
@@ -205,7 +205,15 @@ Wrap your app components with the `JazzProvider:
...
-You can optionally pass a custom `kvStore` and `AccountSchema` to `createJazzRNApp()`, otherwise, it defaults to `ExpoSecureStoreAdapter` and `Account`.
+You can optionally pass a few custom attributes to `<JazzProvider>`:
+- `kvStore`
+  - `ExpoSecureStoreAdapter` (default)
+  - example: `MMKVStore` - roll your own, using MMKV
+- `AccountSchema`
+  - `Account` (default)
+- `CryptoProvider`
+  - `PureJSCrypto` (default)
+  - `RNQuickCrypto` - C++ accelerated crypto provider
```
Comment (gdorsi) [inline homepage/homepage/app/(docs)/docs/[framework]/[...slug]/project-setup/react-native.mdx:216]: Should this be C++?

Comment (boorad) [reply homepage/homepage/app/(docs)/docs/[framework]/[...slug]/project-setup/react-native.mdx:216]: What do you mean?  Is that library written in C++?  Yes it is a RN native module that uses Nitro Modules to bridge between JS & C++

Comment (gdorsi) [reply homepage/homepage/app/(docs)/docs/[framework]/[...slug]/project-setup/react-native.mdx:216]: Reviewing from mobile made it look like a typo 😅 

Sorry!

PR #1386, opened 2025-02-15
tony/fix-contributing-md

antoniel:
## Context
When cloning the repository, I encountered errors running tests immediately after installing dependencies. I concluded that workspace dependencies were not fully exposed, particularly those for the homepage. It appears that a pnpm configuration prevents the extraction of homepage dependencies during the initial installation.

## Update CONTRIBUTING.md
- Add an explicit step to install homepage dependencies.
- Include package build instructions before running tests.


gdorsi (approved):
Thanks!

PR #1393, opened 2025-02-17
feat(llms-txt): add the music player example code

gdorsi:
Over the weekend I tried to create a v0 app asking to use Jazz via the llms-txt.

The results weren't great, so I've tried adding a full example code and the results were way better.

Used the music player because it uses FileStream, it's based on React and has the structure I wanted (schema+actions+optional auth)


bensleveritt:
Maybe try your test again, as I've just fixed llms-full.txt in #1396

trishalim (approved):
LGTM

PR #1395, opened 2025-02-17
Allow selection of external security keys when creating passkey

3f765d5e-ac23-4170-97a5-68b8feee155b:
Deleting the line removes the restriction of only being able to select on-device passkey creation.

This will allow using an external security key, for example a yubikey, when creating an account.

### DIFF
--- a/packages/jazz-browser/src/auth/PasskeyAuth.ts
+++ b/packages/jazz-browser/src/auth/PasskeyAuth.ts
@@ -130,7 +130,6 @@ export class BrowserPasskeyAuth {
           },
           pubKeyCredParams: [{ alg: -7, type: "public-key" }],
           authenticatorSelection: {
-            authenticatorAttachment: "platform",
             requireResidentKey: true,
             residentKey: "required",
           },


packages/jazz-browser/src/auth/PasskeyAuth.ts
```diff
@@ -130,7 +130,6 @@ export class BrowserPasskeyAuth {
           },
           pubKeyCredParams: [{ alg: -7, type: "public-key" }],
           authenticatorSelection: {
-            authenticatorAttachment: "platform",
```
gdorsi, line 133: Tested on [the MusicPlayer example](https://music-demo-git-fork-3f765d5e-ac23-4170-97a5-68b8fee-606337-gcmp.vercel.app/?_vercel_share=L9OFtyv6qg7gEeYy9PLUX70VOAw8jF0r) using a YubiKey but the experience on Mac isn't great:
- Firefox handles the auth with YubiKey correctly
- Chrome comes up with a "Your device can't be used on this site" error message
- Safari gets stuck on waiting for the physical approval on the Yubikey

Maybe we need to tweak other parameters in order to make it work better.

@3f765d5e-ac23-4170-97a5-68b8feee155b How is your experience on Linux?

3f765d5e-ac23-4170-97a5-68b8feee155b: > How is your experience on Linux?

Works as expected in FF and Chrome. Asks to unlock security key and then I'm able to create an account.

> Maybe we need to tweak other parameters in order to make it work better.

Maybe worth looking how it works on github and vercel to see what's possible in terms of user experience.

One possible solution I see here is leave the code as is (only on-device passkeys) and if we end up in the catch block try again but only allow external security keys. If both fail the device does not support passkeys.

I cannot test the behavior on a device that has both ways of creating passkeys. The above solution would favor on-device passkey creation which could be a problem if the user wants to use an external key

> Chrome comes up with a "Your device can't be used on this site" error message

Just guessing but maybe adding additional algorithms to `pubKeyCredParams` would help. Per developer.mozilla.org for the best coverage -7, -8 and -257 should be specified

aeplay: @gdorsi just to clarify: we should be able to use any kind of pubKeyCredParams - this is not something we're relying on. All we need is to put/get the Jazz credential payload in publicKey.user.id

gdorsi: Applied the changes on a different PR, working on a fork is tricky: https://github.com/garden-co/jazz/pull/1446/files

Got it working with Safari, Chrome and Firefox on MacOS.

I have two Yubikeys (a 7 years old Yubikey 4 and a 6 years old Yubikey 5 NFC)

Got it working only on the Yubikey 5 NFC, but I think it's good enough 😄

gdorsi:
Just merged https://github.com/garden-co/jazz/pull/1446 which supersedes this PR.

Thanks for the contribution, the suggestions and the research on the topic!

PR #1403, opened 2025-02-17
 fix: fixes clerk auth flow

Elfo404:


packages/jazz-react-core/src/hooks.ts
```diff
@@ -57,6 +57,7 @@ export function useCoState<V extends CoValue, D>(
   >(
     React.useCallback(
       (callback) => {
+        observable.reset();
```
gdorsi: We can also remove this reset, as in theory we do this already with the sync resolution

packages/jazz-tools/src/implementation/ContextManager.ts
```diff
@@ -44,6 +44,7 @@ export class JazzContextManager<
   protected props: P | undefined;
   protected authSecretStorage = new AuthSecretStorage();
   protected authenticating = false;
+  protected prevContextCreation?: Promise<void>;
```
gdorsi: We can remove this.

Out of order context creation needs a more complete solution

packages/jazz-tools/src/tests/ContextManager.test.ts
```diff
@@ -1,5 +1,5 @@
 import { WasmCrypto } from "cojson/crypto/WasmCrypto";
-import { beforeEach, describe, expect, test, vi } from "vitest";
+import { assert, beforeEach, describe, expect, test, vi } from "vitest";
```
gdorsi: nit: unused import

gdorsi (approved):

PR #1418, opened 2025-02-18
fix: improve the rollback on failure when handling new content in storage

gdorsi:
We spotted some issues when mixing multiple browser tabs and data intensive applications.

Added a E2E test to cover that case and improved the transactions management so when something bad happens the storage is left in a consistent state.


packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -10,30 +10,25 @@ import type {
...
+    transactions: IDBObjectStore;
+    signatureAfter: IDBObjectStore;
+  };
+  running: boolean;
+  startedAt: number;
+  isControlledTransaction?: boolean;
+  pendingRequests: ((txEntry: TxEntry) => void)[];
+};
+
 export class IDBClient implements DBClientInterface {
```
gdorsi, line 28: Need to add a test to verify what happens if the same content message is sent twice

packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -220,7 +249,15 @@ export class IDBClient implements DBClientInterface {
     );
   }
 
-  async unitOfWork(operationsCallback: () => unknown[]) {
-    return Promise.all(operationsCallback());
+  async transaction(operationsCallback: () => unknown) {
```
gdorsi, line 176: Simplified the signature here to make it easier to use when the intent is to have a transaction with rollback

packages/cojson-storage/src/syncManager.ts
```diff
@@ -217,8 +217,8 @@ export class SyncManager {
 
     let invalidAssumptions = false;
 
-    await this.dbClient.unitOfWork(() =>
-      (Object.keys(msg.new) as SessionID[]).map((sessionID) => {
+    for (const sessionID of Object.keys(msg.new) as SessionID[]) {
+      await this.dbClient.transaction(async () => {
```
gdorsi, line 212: We probably lose some perf here, but we gain more resilience against errors.

This way only the errored sessions are reverted.

tests/e2e/src/pages/ConcurrentChanges.tsx
```diff
@@ -0,0 +1,76 @@
...
+async function count(counter: Counter) {
+  if (!counter.byMe) return;
+
+  let value = counter.byMe.value?.value ?? 0;
+
+  while (value <= 300) {
+    if (value === 200) {
+      counter.push({ value: value++ });
+      counter.push({ value: value++ });
+      window.location.reload();
```
gdorsi, line 70: Managed to reproduce an InvalidSignature error by reloading the page in the middle of the counting

packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -220,7 +249,15 @@ export class IDBClient implements DBClientInterface {
...
+    const txEntry = this.startTransaction();
+    try {
+      await operationsCallback();
+      this.closeTransaction(txEntry);
+    } catch (error) {
+      txEntry.tx.abort();
+      this.closeTransaction(txEntry);
+      throw error;
+    }
```
gdorsi, line 261: I mean to improve this code, the main idea was to bulk operations in the same transactions only when declared.

packages/cojson-storage/src/syncManager.ts
```diff
@@ -217,9 +208,13 @@ export class SyncManager {
...
 
-    await this.dbClient.unitOfWork(() =>
-      (Object.keys(msg.new) as SessionID[]).map((sessionID) => {
-        const sessionRow = allOurSessions[sessionID];
+    for (const sessionID of Object.keys(msg.new) as SessionID[]) {
+      await this.dbClient.transaction(async () => {
+        const sessionRow = await this.dbClient.getSingleCoValueSession(
+          storedCoValueRowID,
+          sessionID,
+        );
```
gdorsi, line 216: Had to move the Session get inside the transaction to fix completely the InvalidSignature error

tests/e2e/tests/ConcurrentChanges.test.ts
```diff
@@ -0,0 +1,35 @@
...
+      }
+    });
+
+    await page.reload();
+
+    await expect(page.getByTestId("done")).toBeVisible();
+
+    expect(
+      errorLogs.find((log) => log.includes("InvalidSignature")),
+    ).toBeUndefined();
```
gdorsi, line 33: This test fails the 100% of the times on main.

It's easier to reproduce this issue with Playwright because all the pages are 100% active during the test.
With a real browser the non-focused pages are throttled so it's way harder to reproduce concurrency issues.

packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -8,143 +8,92 @@ import type {
...
+    this.autoBatchingTransaction = tx;
+
+    return tx.handleRequest<T>(handler);
   }
 
   async getCoValue(coValueId: RawCoID): Promise<StoredCoValueRow | undefined> {
-    return this.makeRequest<StoredCoValueRow | undefined>(({ coValues }) =>
-      coValues.index("coValuesById").get(coValueId),
+    return this.makeRequest<StoredCoValueRow | undefined>((tx) =>
+      tx.getObjectStore("coValues").index("coValuesById").get(coValueId),
```
gdorsi, line 43: Small refactoring to access the objectStore only when needed.

Not a big difference but made the code on the transaction management more lean

aeplay (approved):

PR #1431, opened 2025-02-19
fix: isAuthenticated out-of-sync with the account state

gdorsi:
While working on some internal apps we noticed that during login and logout the account and isAuthenticated states are out-of-sync for a couple of renders.

Debugging the code we found that:
- The emit of AuthSecretStorage was triggered with a different timing compared to the account context updates
- The useCoState wasn't unsubscribing correctly from the account coValues

We fixed both issues here but haven't addressed the unsubscribe issue outside of useCoState and going to do that ASAP in another PR.


packages/cojson/src/tests/permissions.test.ts
```diff
@@ -2029,9 +2030,13 @@ test("Can give write permissions to 'everyone' (high-level)", async () => {
       .getCurrentContent(),
   );
 
+  connectTwoPeers(group.core.node, childContent2.core.node, "server", "server");
+
+  // Ensure that the group is available to newAccount
+  await group.core.waitForSync();
```
gdorsi, line 2036: Added because this test was flaky

packages/jazz-react-core/src/hooks.ts
```diff
@@ -26,16 +34,69 @@ export function useJazzContext<Acc extends Account>() {
...
+    getCurrentValue() {
+      return ref.current.getCurrentValue();
+    },
+    getCurrentObservable() {
+      return ref.current;
+    },
+    reset() {
+      ref.current = createCoValueObservable<V, D>({
+        syncResolution: true,
+      });
```
gdorsi, line 97: As a workaround for the issues related to the unsubscribe I've added a reset.

Going to fix the unsubscribe ASAP, but I think it's still better to fully isolate the subscribe actions with different observables.

packages/jazz-react-core/src/tests/useAccount.test.ts
```diff
@@ -50,4 +55,125 @@ describe("useAccount", () => {
...
+        {
+          "accountIndex": 0,
+          "isAuthenticated": true,
+        },
+        {
+          "accountIndex": 1,
+          "isAuthenticated": false,
+        },
+      ]
+    `);
```
gdorsi, line 114: Before we had an additional 0/false entry in the middle:

```json
   {
          "accountIndex": 0,
          "isAuthenticated": false,
        },
```

packages/jazz-tools/src/implementation/ContextManager.ts
```diff
@@ -43,6 +43,10 @@ export class JazzContextManager<
   protected context: PlatformSpecificContext<Acc> | undefined;
   protected props: P | undefined;
   protected authSecretStorage = new AuthSecretStorage();
+  protected authSecretStorageWithNotify = Object.assign(
+    Object.create(this.authSecretStorage),
+    { notify: true },
+  );
```
gdorsi, line 49: Decided to change the auth notify behavoir to be decoupled when used inside the context flow and to be automatic when used outside (e.g.  in the auth providers).

In the context flow we trigger an update at the same time when we trigger the context update, to align the auth state update timing with the account value updates.

packages/jazz-tools/src/auth/AuthSecretStorage.ts
```diff
@@ -16,6 +16,7 @@ export type AuthSetPayload = {
 export class AuthSecretStorage {
   private listeners: Set<(isAuthenticated: boolean) => void>;
   public isAuthenticated: boolean;
+  notify = false;
```
Elfo404, line 19: does this ever change?

Elfo404: lol never mind, it's literally in the next file 🤦

packages/jazz-react-core/src/hooks.ts
```diff
@@ -118,14 +218,44 @@ export function createUseAccountHooks<Acc extends Account>() {
...
+              .getCurrentObservable()
+              .subscribe(
+                Schema,
+                agent.id,
+                agent,
+                depth ?? ([] as D),
+                callback,
+                () => {
+                  callback();
+                },
```
Elfo404: nit
```suggestion
                callback,
```

packages/jazz-tools/src/tests/ContextManager.test.ts
```diff
@@ -26,6 +26,7 @@ import {
   setupJazzTestSync,
 } from "../testing";
 
+// @ts-ignore Typescript in VSCode doesn't like top level await
```
Elfo404, line 29:
```suggestion
// @ts-expect-error Typescript in VSCode doesn't like top level await
```
so the day vs code is happy we notice, this disable altogether

packages/jazz-react-core/src/hooks.ts
```diff
@@ -87,17 +152,52 @@ export function createUseAccountHooks<Acc extends Account>() {
...
```
Elfo404: nit
```suggestion
                callback,
```

packages/jazz-react-core/src/hooks.ts
```diff
@@ -59,14 +116,22 @@ export function useCoState<V extends CoValue, D>(
...
-        return observable.subscribe(Schema, id, agent, depth, callback, () => {
-          console.log("unavailable");
-          callback();
+          observable.reset();
+
+          return observable
+            .getCurrentObservable()
+            .subscribe(Schema, id, agent, depth, callback, () => {
+              callback();
+            });
```
Elfo404: nit
```suggestion
            .subscribe(Schema, id, agent, depth, callback, callback);
```

Elfo404 (approved):
makes sense 👍

packages/jazz-tools/src/tests/ContextManager.test.ts
gdorsi: The problem is that tsc doesn't consider it an error so `@ts-expect-error` makes the typecheck fail

Elfo404: ah good point, nvm then

PR #1446, opened 2025-02-20
feat: support external authenticators on passkey authentication

gdorsi:
Tested with a Yubikey on Safari, Chrome and Firefox on MacOS.

I have two Yubikeys (a 7 years old Yubikey 4 and a 6 years old Yubikey 5 NFC)

Got it working only on the Yubikey 5 NFC, but I think it's good enough 😄

### DIFF
--- a/.changeset/poor-bats-applaud.md
+++ b/.changeset/poor-bats-applaud.md
@@ -1,0 +1,5 @@
+---
+"jazz-browser": patch
+---
+
+Support external authenticators on passkey auth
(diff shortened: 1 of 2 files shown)


aeplay (approved):

3f765d5e-ac23-4170-97a5-68b8feee155b:
asskey 😏

3f765d5e-ac23-4170-97a5-68b8feee155b:
I think setting the attachment to `cross-platform` breaks on-device passkey creation? On my phone I only get the option to use a USB key or using a different device via bluetooth

3f765d5e-ac23-4170-97a5-68b8feee155b:
Plus I get a privacy warning in firefox https://support.mozilla.org/en-US/kb/privacy-web-authentication because some additional information about the security key is requested

PR #1449, opened 2025-02-20
add anchor to headings in mdx components

Elfo404:
didn't add it for h1 to avoid linking the fist header of a page, which doesn't make much sense probably


trishalim:
Thanks, this is helpful!

I tried to match the # with the heading font size. If that works for you, feel free to merge

image

PR #1459, opened 2025-02-21
feat: migrate to declaration files

gdorsi:
By linking the type sources we make our users do type-validation on our typescript sources, which may lead to errors in case of incompatible typescript versions or configuration.

With this PR we migrate over declaration files, which are the suggested solution of the Typescript team but come with the following drawbacks:
-  ~~Click to reference will now link to the d.ts files instead of the source~~ EDIT: fixed with declarationMap
- The type validation in the repo now requires the packages build, and we can't validate a package before building their dependencies

For this reasons I've pushed back this change since now, but the errors reported on the user side are getting weirder so I think that this is the best choice now.


packages/jazz-react-core/src/hooks.ts
```diff
@@ -138,139 +142,130 @@ export function useCoState<V extends CoValue, D>(
...
+  logOut: () => void;
+};
+function useAccount<A extends RegisteredAccount, D extends DepthsIn<A>>(
+  depth: D,
+): { me: DeeplyLoaded<A, D> | undefined | null; logOut: () => void };
+function useAccount<A extends RegisteredAccount, D extends DepthsIn<A>>(
+  depth?: D,
+): { me: A | DeeplyLoaded<A, D> | undefined | null; logOut: () => void } {
+  const context = useJazzContext<A>();
+  const contextManager = useJazzContextManager<A>();
```
gdorsi, line 156: Moved away from the factory pattern because the RegisteredAccount injection was broken by the `d.ts` compilation.

Now the jazz-react and jazz-react-native simply re-export this hook while registering the RegisteredAccount type.

This way the user can still do the declaration on these two packages and reach the RegisteredAccount type here

packages/jazz-tools/src/coValues/coMap.ts
```diff
@@ -151,7 +151,7 @@ export class CoMap extends CoValueBase implements CoValue {
   }
 
   /** @internal */
-  private getEditFromRaw(
+  public getEditFromRaw(
```
gdorsi, line 154: Typescript was whining about this restricted visibility but I don't have a clear explaination

packages/jazz-tools/src/coValues/coRichText.ts
```diff
@@ -615,7 +615,10 @@ export function splitNode(
  * Collection of predefined mark types for common text formatting.
  * Includes marks for headings, paragraphs, links, and text styling.
  */
-export const Marks = {
+export const Marks: Record<
+  "Heading" | "Paragraph" | "Link" | "Strong" | "Em",
+  typeof Mark
+> = {
```
gdorsi, line 621: @bensleveritt 

Is this ok for you?
Another fix for the d.ts, probably related to the circular references.

bensleveritt: Yeah LGTM 👍

PR #1461, opened 2025-02-21
feat: allow profile migrations, enforce profile ownership

Elfo404:


packages/jazz-tools/src/coValues/account.ts
```diff
@@ -259,13 +259,29 @@ export class Account extends CoValueBase implements CoValue {
...
-      const profileGroup = RegisteredSchemas["Group"].create({ owner: this });
-      profileGroup.addMember("everyone", "reader");
-      this.profile = Profile.create(
-        { name: creationProps.name },
-        { owner: profileGroup },
-      );
+    await (this.constructor as typeof Account).migrate(this, creationProps);
+
+    // if the user has not defined an account themselves, we create one
+    if (!this.profile && creationProps) {
```
gdorsi, line 265: This should be `this.profile === undefined`

If this.profile is null it means that the value is there but not loaded yet

packages/jazz-tools/src/coValues/account.ts
```diff
@@ -259,13 +259,29 @@ export class Account extends CoValueBase implements CoValue {
...
+      this.profile.name = creationProps.name;
+
+      // by convention, the profile must be owned by a publicly accessible group, in case the user
+      // created the profile themeselves and has not set the owner group to be (at least) publicly readable,
+      // we add the everyone as a reader.
+      if (
+        !this.profile._owner.members.some(
+          (m) =>
+            m.id === "everyone" && (m.role === "reader" || m.role === "writer"),
+        )
```
gdorsi, line 281: This breaks if the owner is an Account.

We do a similar thing to the inbox, enforcing a writer role on the inbox messages: https://github.com/garden-co/jazz/blob/main/packages/jazz-tools/src/coValues/inbox.ts#L68

packages/jazz-tools/src/coValues/profile.ts
```diff
@@ -1,11 +1,44 @@
...
+    this: CoValueClass<M>,
+    init: Simplify<CoMapInit<M>>,
+    options?:
+      | {
+          owner: Group;
+        }
+      | Group,
+  ) {
+    return super.create<M>(init, options);
+  }
```
gdorsi, line 43: Question: why are you adding these methods here?

examples/music-player/src/1_schema.ts
```diff
@@ -87,15 +87,15 @@ export class MusicaAccount extends Account {
    *  The account migration is run on account creation and on every log-in.
    *  You can use it to set up the account root and any other initial CoValues you need.
    */
-  migrate() {
-    if (this.root === undefined) {
+  static override migrate(me: MusicaAccount) {
```
gdorsi, line 90: question: is this override key required by Typescript?

Elfo404: no, but it will make sure that you are overriding something. typescript will throw an error if you define a method with `override` that is not actually overriding anything

packages/jazz-tools/src/coValues/profile.ts
Elfo404: This is to make typescript complain if you try to create a Profile with an account as owner

Elfo404: as per `get _owner()` it could be simply    `return super._owner as Group;, but similar reasons

packages/jazz-tools/src/coValues/account.ts
Elfo404: with the changes in profile the idea was to make it impossible to create Profiles owned by accounts. this of course means also doing some runtime check before this and in Profile.create itself

PR #1471, opened 2025-02-23
jazz-739-handle-reloading-page-while-image-upload-is-in-progress

### LINKED ISSUE #1458: Handle reloading page while image upload is in progress
Reloading the page before an image upload is done causes an error in Firefox

In Chrome, you still get the image, but you don't get the full size

In image upload example:

* handle UI for image upload progress (currently the image shown goes from small to large until it overflows the page)
* warn the user that reloading the page will halt the upload progress

bensleveritt:
- Added a Promise response to `createImages` that resolves when image is processed (not just uploaded).
- Added a beforeunload handler to notify when images are being processed.
- Added a resize interaction for large images.

CleanShot 2025-02-23 at 17 00 48


gdorsi:
I'm still getting the issue on Firefox:

https://github.com/user-attachments/assets/760aa72c-e4da-4738-8f79-5bbbaaaa030a

gdorsi:
I think that the image is still processed after we resolve the promise, and the error is returned because the task is killed before the first chunk of the higher resolution image is stored.

examples/image-upload/src/ImageUpload.tsx
```diff
@@ -1,44 +1,82 @@
...
   const { me } = useAccount();
-
+  const [isUploading, setIsUploading] = useState(false);
   const inputRef = useRef<HTMLInputElement>(null);
 
+  useEffect(() => {
+    const handleBeforeUnload = (e: BeforeUnloadEvent) => {
+      if (isUploading) {
+        e.preventDefault();
+        return "Upload in progress. Are you sure you want to leave?";
```
trishalim, line 14: This works, but it's showing the default message. That's fine, we don't really need a custom one. But this string made me think we had a custom one. So maybe just change the string to something like "true", and a comment: `return a non-empty string to trigger dialog`

bensleveritt: Depends on your browser. Older browsers will show the message, while newer ones seem to drop it entirely in favour of the default response. This accounts for both

trishalim: Makes sense

gdorsi:
Pushed some changes:
- Added a test for createImage in browser-integration
- Simplified the fix for createImage by removing the setTimeout
- Aligned the branch with main
- added sideEffects: false to the images package
- added the changeset files

Done this changes here to be able to re-export createImage on jazz-react and jazz-react-native

PR #1472, opened 2025-02-23
Next SSR example working

tobiaslins:


gdorsi:
Now we support with `experimental_enableSSR` flag! 🚀 

Thanks for the proposal that helped us to move in the right direction!

PR #1487, opened 2025-02-24
fix(PasskeyAuth): set name iff username given

### LINKED ISSUE #1433: Profile data from anon user doesn't get carried over after signup
In the react starter app, after signing up, the name disappears, but the dateOfBirth is carried over. This is because the dateOfBirth is stored in the root, and the name is in the profile.

emmyoh:
If an empty username is passed to `signUp`, the `profile.name` of the user is not updated.

Fixes: #1433

### DIFF
--- a/.changeset/six-lizards-sip.md
+++ b/.changeset/six-lizards-sip.md
@@ -3,4 +3,4 @@
 "form": minor
 ---
 
-Sets `profile.name` only if a non-empty username is passed to `signUp`
+PasskeyAuth: Sets `profile.name` only if a non-empty username is passed to `signUp`
--- a/.changeset/six-lizards-sip.md
+++ b/.changeset/six-lizards-sip.md
@@ -1,0 +1,6 @@
+---
+"jazz-browser": minor
+"form": minor
+---
+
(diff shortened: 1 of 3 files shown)


gdorsi (approved):
LGTM

Could you create a changeset?

It can be done with `pnpm changeset` on the repo root and the change should be a minor because it's breaking

emmyoh:
> LGTM
> 
> Could you create a changeset?
> 
> It can be done with `pnpm changeset` on the repo root and the change should be a minor because it's breaking

@gdorsi Sounds good! Pushed a new commit to this PR; hopefully I did it correctly :)

.changeset/six-lizards-sip.md
```diff
@@ -0,0 +1,6 @@
+---
+"jazz-browser": minor
+"form": minor
+---
+
+Sets `profile.name` only if a non-empty username is passed to `signUp`
```
gdorsi:
```suggestion
PasskeyAuth: Sets `profile.name` only if a non-empty username is passed to `signUp`
```

emmyoh:
@gdorsi I'll go ahead and merge this if you're OK with doing so.

Issue #1476, opened 2025-02-24
Improve the type inference on useCoState and all the load/subscribe APIs


gdorsi:
If we take this example:

```
const playlist = useCoState(Playlist, playlistId, {
    tracks: [],
});
```

If the type of playlistId is `ID<CoMap>` the whole inferred type will become CoMap, making it look like that the types on useCoState are broken.<br><br>Proposed solution:<br><br>We could change the id param to be `ID<CoValue>` so the type inference would pick the coValue type only from the Schema param.

PR #1489, opened 2025-02-24
fix: type inference on `useCoState`

emmyoh:
The `id` parameter should be `ID<CoValue>` so the type inference picks up the CoValue type only from the `Schema` parameter.

See: #1476

### DIFF
--- a/packages/jazz-svelte/src/lib/jazz.svelte.ts
+++ b/packages/jazz-svelte/src/lib/jazz.svelte.ts
@@ -1,6 +1,4 @@
-import {
-  consumeInviteLinkFromWindowLocation
-} from 'jazz-browser';
+import { consumeInviteLinkFromWindowLocation } from 'jazz-browser';
 import type {
   AnonymousJazzAgent,
   AuthSecretStorage,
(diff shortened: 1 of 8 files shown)


packages/jazz-svelte/src/lib/jazz.svelte.ts
emmyoh, line 1: Some of these changes came from running `pnpm format-and-link` and are purely cosmetic.

emmyoh:
Turning this into a draft PR since tests are failing. On my Linux machine `pnpm test` did not show new failures, but I'm going to double-check this on macOS which has official Playwright support.

gdorsi (commented):
Good job!

Could you add a test using `expectTypeOf` passing to useCoState something like `const value = useCoState(MyCoMap, id as ID<CoValue>, [])`

`value` should be of the type `MyCoMap`

packages/jazz-svelte/src/lib/tests/useCoState.svelte.test.ts
emmyoh, line 1: Many changes here are also cosmetic and came from the formatter.

gdorsi (approved):
Good job!

Could you add a changeset?
After that you can merge this PR directly.

emmyoh:
> Good job!
> 
> Could you add a changeset? After that you can merge this PR directly.

Perfect—thanks for your review!

PR #1506, opened 2025-02-25
Image upload example UX improvement

trishalim:


examples/image-upload/src/ImageUpload.tsx
```diff
@@ -66,35 +52,46 @@ export default function ImageUpload() {
...
-              ref={inputRef}
-              type="file"
-              accept="image/png, image/jpeg, image/gif, image/bmp"
-              onChange={onImageChange}
-              disabled={isUploading}
-            />
-            {isUploading && <span>Uploading...</span>}
-          </div>
-        )}
+  if (imagePreviewUrl) {
```
trishalim, line 69: @bensleveritt @gdorsi Not sure if it's worth doing this "uploading" state because it's super fast

examples/image-upload/src/ImageUpload.tsx
```diff
@@ -1,62 +1,48 @@
 import { createImage } from "jazz-browser-media-images";
 import { ProgressiveImg, useAccount } from "jazz-react";
-import { ImageDefinition } from "jazz-tools";
 import { ChangeEvent, useEffect, useRef, useState } from "react";
 
-function Image({ image }: { image: ImageDefinition }) {
-  const [isFullSize, setIsFullSize] = useState(false);
```
trishalim, line 7: I removed this part because the user can just click "open image in new tab"

examples/image-upload/src/ImageUpload.tsx
bensleveritt: "Uploading" now includes image processing, which isn't instant, so it might be worth keeping?

PR #1507, opened 2025-02-25
feat(PasskeyAuth): support random passphrase generation

gdorsi:
Added registerNewUser and generateRandomPassphrase methods to PasskeyAuth to make it possible to build a "reroll passphrase" kind-of UI.

`registerNewUser` accepts a random passphrase that can be generated with `generateRandomPassphrase`, creates a new account and does the authentication.

Data can be migrated from the anonymous user with the `onAnonymousAccountDiscarded`

### DIFF
--- a/packages/jazz-tools/src/auth/PassphraseAuth.ts
+++ b/packages/jazz-tools/src/auth/PassphraseAuth.ts
@@ -23,25 +23,13 @@ import { AuthSecretStorage } from "./AuthSecretStorage.js";
 export class PassphraseAuth {
   passphrase: string = "";
 
-  private crypto: CryptoProvider;
-  private authenticate: AuthenticateAccountFunction;
-  private register: RegisterAccountFunction;
-  private authSecretStorage: AuthSecretStorage;
-  public wordlist: string[];
-
   constructor(
-    crypto: CryptoProvider,
-    authenticate: AuthenticateAccountFunction,
-    register: RegisterAccountFunction,
-    authSecretStorage: AuthSecretStorage,
-    wordlist: string[],
-  ) {
-    this.crypto = crypto;
-    this.authenticate = authenticate;
-    this.register = register;
-    this.authSecretStorage = authSecretStorage;
-    this.wordlist = wordlist;
-  }
+    private crypto: CryptoProvider,
+    private authenticate: AuthenticateAccountFunction,
+    private register: RegisterAccountFunction,
+    private authSecretStorage: AuthSecretStorage,
+    public wordlist: string[],
+  ) {}
 
   logIn = async (passphrase: string) => {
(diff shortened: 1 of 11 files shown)


packages/jazz-tools/src/auth/PassphraseAuth.ts
```diff
@@ -20,12 +23,25 @@ import { AuthSecretStorage } from "./AuthSecretStorage.js";
...
+    register: RegisterAccountFunction,
+    authSecretStorage: AuthSecretStorage,
+    wordlist: string[],
+  ) {
+    this.crypto = crypto;
+    this.authenticate = authenticate;
+    this.register = register;
+    this.authSecretStorage = authSecretStorage;
+    this.wordlist = wordlist;
+  }
```
Elfo404: why not just
```ts
  constructor(
    private crypto: CryptoProvider,
    private authenticate: AuthenticateAccountFunction,
    private register: RegisterAccountFunction,
    private authSecretStorage: AuthSecretStorage,
    public wordlist: string[],
  ) {}
```
to remove some noise?

packages/jazz-tools/src/tests/PassphraseAuth.test.ts
```diff
@@ -150,3 +195,152 @@ describe("PassphraseAuth", () => {
...
+
+      assert(context && "me" in context);
+      expect(context.me.id).toBe(accountId);
+      expect(context.me.profile?.name).toBe("Test User");
+
+      expect(authSecretStorage.isAuthenticated).toBe(true);
+
+      const credentials = await authSecretStorage.get();
+      expect(credentials?.accountID).toBe(accountId);
+      expect(credentials?.provider).toBe("passphrase");
```
Elfo404: nit:
```suggestion
      const credentials = await authSecretStorage.get();
      assert(credentials);
      expect(credentials.accountID).toBe(accountId);
      expect(credentials.provider).toBe("passphrase");
```

Elfo404 (approved):
lgtm

packages/jazz-tools/src/auth/PassphraseAuth.ts
gdorsi: This is one of the Typescript features with runtime code and the new direction for Typescript now that has become part of Node.js is to start to deprecate this kind of features.

It's a very long process and it's not a solid reason to stop using it, but I've done this change just to see how bad it looks to go back and use the "manual" initializers.

I can revert the change, kind-of agree that looks quite verbose compared to before.

Elfo404: makes sense, in general i'd leave "style" changes isolated to their own PRs (and possibly added to .git-blame-ignore-revs), but not a biggie. 🐑 👽

PR #1512, opened 2025-02-25
feat!(Groups): API to unextend group

### LINKED ISSUE #1504: Add an API to unextend a group
We need an API to revert a group extension, to automatically kick-out the members inherited from a group.<br><br>Proposed solution:

* We should add an API `Group.revokeExtend`
* This API sets the parent key on the child group and the child key to the parent group as `revoked`
* Rotate the keys on the child group

emmyoh:
Closes #1504.


emmyoh:
Tests not passing on `0-11-0` as of writing, unrelated to this PR.

packages/cojson/src/coValues/group.ts
```diff
@@ -626,6 +641,34 @@ export class RawGroup<
...
+    }
+
+    // Set the parent key on the child group to `revoked`
+    this.set(`parent_${parent.id}`, "revoked", "trusting");
+
+    // Set the child key on the parent group to `revoked`
+    parent.set(`child_${this.id}`, "revoked", "trusting");
+
+    // Rotate the keys on the child group
+    this.rotateReadKey();
```
gdorsi, line 669: To rotate the key we need to ensure that all the child groups are loaded:

```ts
  async removeMember(
    account: RawAccount | ControlledAccountOrAgent | Everyone,
  ) {
    // Ensure all child groups are loaded before removing a member
    await this.loadAllChildGroups();

    this.removeMemberInternal(account);
  }
```

This is to cover a case where we are calling `revokeExtend` on a group that has childs, otherwise `rotateKey` might not iterate all the childs.

packages/cojson/src/tests/permissions.test.ts
```diff
@@ -2913,6 +2914,70 @@ test("revoking access on a parent group doesn't block access to the child group
...
+
+  const aliceMapAfterUnextend = expectMap(
+    mapCore
+      .testWithDifferentAccount(alice, Crypto.newRandomSessionID(alice.id))
+      .getCurrentContent(),
+  );
+  // `alice` attempts to set `foo` to `def`, but fails
+  expect(() => aliceMapAfterUnextend.set("foo", "def", "private")).toThrow(
+    "Can't make transaction without read key secret",
+  );
```
gdorsi, line 2964: This should not throw but fail silently

packages/cojson/src/tests/permissions.test.ts
```diff
@@ -2913,6 +2914,70 @@ test("revoking access on a parent group doesn't block access to the child group
   expect(mapOnNode2.get("foo")).toEqual("baz");
 });
 
+test("revoking access to parent group", async () => {
```
gdorsi: We also need tests to verify that:
- After revokeExtend the account loses read access over the new transactions (key rotation)
- When revoking the extend of a granparent the related members lose read access on the grandchilds (key rotation on childs)

gdorsi (approved):
Good job!

PR #1517, opened 2025-02-26
feat(deepLoading): return undefined when an optional field is not accessible

gdorsi:


packages/jazz-tools/src/coValues/deepLoading.ts
```diff
@@ -56,8 +53,12 @@ export function fulfillsDepth(depth: any, value: CoValue): FulfillsDepthResult {
...
         if (map._raw.get(key) !== undefined) {
           if (!item) {
-            result = "unfulfilled";
-            continue;
+            if (hasReadAccess(map, key)) {
+              result = "unfulfilled";
+              continue;
+            } else {
+              return "unauthorized";
+            }
```
gdorsi, line 61: Basically I've changed the logic to check for the permissions when the item is `null`

This way we can know if the item is null because it's loading or because it's not accessible.

PR #1522, opened 2025-02-26
Fix and test cases for reported issue with optional fields

### LINKED ISSUE #1496: Allow optional fields in types passed to `co.json` (see `CoJsonValue`)

### LINKED ISSUE #1601: JSON type regression
Minimal repro from @gdorsi :

```
it("should consider string generics valid", async () => {
	class InvalidFunctionMap<T extends string> extends CoMap {
		data = co.json<T>();
	}
});
```

4rknova:


packages/jazz-tools/src/tests/schema.test.ts
```diff
@@ -163,6 +163,35 @@ describe("co.json TypeScript validation", () => {
...
+   ** See: https://github.com/garden-co/jazz/issues/1496
+   */
+  it("should apply the same validation to optional json [JAZZ-1496]", async () => {
+    interface ValidInterface0 {
+      value: string;
+    }
+    interface ValidInterface1 {
+      value: string | undefined;
+    }
+    interface InvalidInterface {
```
gdorsi: nit: this interface is technically valid

What about renaming the interface to `InterfaceWithOptionalTypes`?

packages/jazz-tools/src/tests/schema.test.ts
```diff
@@ -163,6 +163,35 @@ describe("co.json TypeScript validation", () => {
...
+      value: string | undefined;
+    }
+    interface InterfaceWithOptionalTypes {
+      value?: string;
+    }
+
+    class MapWithOptionalJSON extends CoMap {
+      data1 = co.optional.json<ValidInterface0>();
+      data2 = co.optional.json<ValidInterface1>();
+      data3 = co.optional.json<InterfaceWithOptionalTypes>();
```
gdorsi, line 193: Noticed that the CI wasn't typechecking the tests on jazz-tools.

Updated the tsconfig and now it reports that there is an indeed an issue on this type:
```
Type 'InterfaceWithOptionalTypes' does not satisfy the constraint 'CoJsonValue<InterfaceWithOptionalTypes>'.
  Type 'InterfaceWithOptionalTypes' is not assignable to type '{ value: CoJsonValue1L<string | undefined> | undefined; }'.
    Property 'value' is optional in type 'InterfaceWithOptionalTypes' but required in type '{ value: CoJsonValue1L<string | undefined> | undefined; }'
```

PR #1523, opened 2025-02-26
Maintain docs side nav scroll state on navigation

trishalim:
Scroll state was resetting on navigation because of how I was rendering the layout. I've moved it to `layout.tsx`. Then I needed to store the table of contents items in a context so that I can set the state from within the child.


homepage/homepage/app/(docs)/docs/[framework]/[...slug]/page.tsx
```diff
@@ -26,19 +23,17 @@ export default async function Page({
     const tocItems = (tableOfContents as Toc)?.[0]?.children;
 
     return (
-      <DocsLayout toc={tocItems} nav={<DocNav />}>
-        <Prose className={bodyClassName}>
-          <Content />
-        </Prose>
-      </DocsLayout>
+      <>
+        <TocItemsSetter items={tocItems} />
```
Comment (trishalim) [inline homepage/homepage/app/(docs)/docs/[framework]/[...slug]/page.tsx:27]: Weird but dunno what else to do

PR #1528, opened 2025-02-27
feat(Group.members): use the global AccountSchema and remove everyone from the returned values

gdorsi:


packages/jazz-tools/src/coValues/group.ts
```diff
@@ -152,43 +147,44 @@ export class Group extends CoValueBase implements CoValue {
...
-        role,
-        ref,
-        get account() {
-          return accessRef();
-        },
-      });
-    }
+      const accessRef = () => ref.accessFrom(this, "members." + accountID);
 
-    const everyoneRole = this._raw.roleOf("everyone");
```
gdorsi, line 183: Removed everyone from the returned values to return a list of values of the same type making the usage of this API way easier.

For the cases where the dev wants to check for everyone access esplicitely they can do:

```ts
if (group.getRoleOf("everyone") {
  // Do stuff
}
```

It doesn't come to my mind a common use case for that so I think the extra code can be justified.

examples/organization/src/components/OrganizationMembers.tsx
```diff
@@ -10,26 +9,13 @@ export function OrganizationMembers({
...
     <>
       {group.members.map((member) => (
-        <Member
-          key={member.id}
-          accountId={member.id as ID<Account>}
-          role={member.role}
-        />
+        <div key={member.id} className="px-4 py-5 sm:px-6">
+          <strong className="font-medium">
+            {member.account.profile?.name}
```
gdorsi, line 14: Note how `account` is now non-nullable

packages/jazz-tools/src/coValues/group.ts
aeplay: Hmm I disagree about this one. Group.members would often be used to render the roles on a Group or something owned by it. Even though "everyone" makes the return type more awkward, it forces you to deal with and expose in the UI that case

aeplay (changes requested):

packages/jazz-tools/src/coValues/group.ts
gdorsi: > Even though "everyone" makes the return type more awkward, it forces you to deal with and expose in the UI that case

That's not true, it only makes the UI code less readable without forcing you to do that (see https://github.com/garden-co/jazz/blob/main/examples/organization/src/components/OrganizationMembers.tsx#L15)

Having something this weird in the APIs that's hard to understand only leads devs to circumventing the type.

aeplay: OK - but this means we should instead make "everyone" behave more like an Account in the long run so we can return it as a member.

In the meantime the docs should shout out to check for "everyone"s role in Groups when displaying roles to the user

aeplay (approved):

PR #1535, opened 2025-02-27
test(groupAndAccounts): enable sync and add members test

gdorsi:


packages/jazz-tools/src/coValues/group.ts
```diff
@@ -180,7 +180,10 @@ export class Group extends CoValueBase implements CoValue {
         );
         const accessRef = () => ref.accessFrom(this, "members." + accountID);
 
-        ref.syncLoad();
+        if (!ref.syncLoad()) {
+          console.warn("Account not loaded", accountID);
```
gdorsi, line 184: Should never happen outside of the test environment

packages/jazz-tools/src/tests/groupsAndAccounts.test.ts
```diff
@@ -447,3 +457,95 @@ describe("Account permissions", () => {
     expect(nonMember.canAdmin(testObject)).toBe(false);
   });
 });
+
+describe("Group.members", () => {
```
gdorsi, line 461: Totally forgot to add these tests 🤦 

Thanks for the reminder @emmyoh

PR #1545, opened 2025-02-28
fix: useCoState stale data

pax-k:


packages/jazz-react-core/src/hooks.ts
```diff
@@ -135,7 +135,9 @@ export function useCoState<V extends CoValue, D>(
     () => observable.getCurrentValue(),
   );
 
-  return value;
+  return React.useMemo(() => {
+    return id ? value : undefined;
```
gdorsi: We can fix this also by adding an `observable.reset();` inside the id check:
```ts
if (!id) {
  observable.reset();
  return () => {};
}
```

pnpm-lock.yaml
```diff
@@ -1507,6 +1507,8 @@ importers:
         specifier: ^8.14.2
         version: 8.18.0
 
+  packages/coverage: {}
+
```
gdorsi: This change should be reverted

gdorsi:
Fixed in another PR

PR #1547, opened 2025-02-28
feat: Schema & Subscription based on Zod-like schema

### LINKED ISSUE #182: Allow eliding owner and new NestedClass if owner is the same
Before

```ts
  var pipeline = new Pipeline({
    pipelineId: "test",
    environments: new ListOfEnvironments([
      new Environment({
        name: "L3",
        status: "RUNNING",
        executionId: "execId",
      },{ owner })
    ],{ owner }),
  }, { owner });
```

After

```ts
  var pipeline = new Pipeline({
    pipelineId: "test",
    environments: [
      {
        name: "L3",
        status: "RUNNING",
        executionId: "execId",
      }
    ],
  }, { owner });
```

Thanks @markmelville

### LINKED ISSUE #1283: Move to zod-like syntax for declaring schemas
Prerequisites:
- https://github.com/garden-co/jazz/issues/2079

gdorsi:
fixes #182
fixes #1283


packages/jazz-react-core/src/useCoState2.ts
```diff
@@ -0,0 +1,145 @@
+import {
+  AnonymousJazzAgent,
```
gdorsi: Nothing particular here, it's the same as before.

packages/jazz-tools/src/schema/coMap.ts
```diff
@@ -0,0 +1,364 @@
...
+} from "../internal.js";
+import { CoValuePrototype } from "./interfaces.js";
+import { CoValueResolutionNode, ensureCoValueLoaded } from "./subscribe.js";
+export type Simplify<A> = {
+  [K in keyof A]: A[K];
+} extends infer B
+  ? B
+  : never;
+
+export class CoMapSchema {
```
gdorsi: Decided to separate Schema and Instance for two resons:
1. It will be easier to adopt another Schema system
2. Has made the instance management easier to implement

This approach has broken all the Resolve types, so I've just turned off most of the type validations.

packages/jazz-tools/src/schema/coMap.ts
```diff
@@ -0,0 +1,364 @@
...
+  declare $schema: M;
+  declare $id: ID<M>;
+  declare $resolutionNode: CoValueResolutionNode<M> | undefined;
+  refs = new Map<string, CoMapInstance<any>>();
+  private $lastUpdateTx: number;
+
+  constructor(
+    schema: M,
+    raw: RawCoMap,
+    resolutionNode?: CoValueResolutionNode<M>,
```
gdorsi: Passing the "ResolutionNode" to be able to increase the scope of the subscription.

More below.

packages/jazz-tools/src/schema/coMap.ts
```diff
@@ -0,0 +1,364 @@
...
+      });
+    }
+
+    if (refs) {
+      for (const [key, value] of refs.entries()) {
+        if (value) {
+          instance._fillRef(key as any, value);
+        }
+      }
+    }
```
gdorsi: The refs are passed as a different argument because:
- They are copied either from a previous CoMapInstance or from the CoValueResolutionNode
- Before that the valued is fully loaded in the subscribe we use `_fillRef` to mutate the instance (it's the only case of mutation)

packages/jazz-tools/src/schema/coMap.ts
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+  $set<K extends CoKeys<M>>(key: K, value: M[K]) {
+    setValue(this.$raw, this.$schema, key, value as JsonValue);
+  }
+
+  $updated(refs?: Map<string, CoMapInstance<any> | undefined>) {
+    if (this.$lastUpdateTx === this.$raw.totalProcessedTransactions && !refs) {
+      return this;
+    }
```
gdorsi: This is one of the key APIs.

It preserves the current instance unless:
- Some new transactions are coming from the core value
- new refs are provided

It keeps the resolution state consistent and ensures immutable updates.

packages/jazz-tools/src/schema/coMap.ts
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+  $request<const R extends RefsToResolve<M>>(
+    this: CoMapInstance<M>,
+    options: { resolve: RefsToResolveStrict<M, R> },
+  ) {
+    this.$resolutionNode?.request(options.resolve);
+
+    return this;
+  }
```
gdorsi: This API is meant to be used in place of the autoload, making the "lazy-load" more explicit.

For now it accepts a resolve param which is merged with the current resolve request.
In the future I will do some research to provide an API that's both predictable and type-safe.

packages/jazz-tools/src/schema/coMap.ts
```diff
@@ -0,0 +1,364 @@
...
+
+  static fromRaw<M extends CoMapSchema>(
+    schema: M,
+    raw: RawCoMap,
+    refs?: Map<string, CoMapInstance<any> | undefined>,
+    resolutionNode?: CoValueResolutionNode<M>,
+  ) {
+    const instance = Object.create(
+      new CoMapInstance(schema, raw, resolutionNode),
+    );
```
gdorsi: The CoMap values are assigned to the instance and everything else is part of the prototype.

This gives makes the CoMapInstnace easy to debug (the console log shows only the properties) and works great with JSON.stringify (see tests below)

packages/jazz-tools/src/schema/subscribe.ts
```diff
@@ -0,0 +1,290 @@
...
+    resolve = res;
+    reject = rej;
+  });
+
+  return { promise, resolve, reject };
+}
+
+type ResolvablePromise<T> = ReturnType<typeof createResolvablePromise<T>>;
+
+class Subscription {
```
gdorsi: This is an abstraction to make it look like we can subscribe to values before the load.

It's an hack that should disappear after the LocalNode improvements

packages/jazz-tools/src/schema/subscribe.ts
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  unsubscribe() {
+    if (this.unsubscribed) return;
+    this.unsubscribed = true;
+    this._unsubscribe();
+  }
+}
+
+export class CoValueResolutionNode<S extends CoMapSchema> {
```
gdorsi: This is use to build the resolution tree.

It tracks all the updates on the value and the linked childs and triggers the $update functions accordingly.

Issue #1548, opened 2025-02-28
Logout fails after `signUp`


emmyoh:
# To replicate
1. Open https://passkey-demo.jazz.tools/ or https://passphrase-demo.jazz.tools/
2. Sign-up for a new account
3. Attempt to logout

# What's known
- [The node's `syncManager.peers` shows `indexedDb` and Jazz Cloud after login, but only `indexedDb` after sign-up](https://gist.github.com/emmyoh/cb822a1a2d02fdcc3cf44811e82d5b0c)
- Guido identified that the change to `isAuthenticated` isn't communicated [here](https://github.com/garden-co/jazz/blob/3185a20777df515ad84ad126598d793fefcb6820/packages/jazz-tools/src/implementation/ContextManager.ts#L90) because of [this problematic check](https://github.com/garden-co/jazz/blob/3185a20777df515ad84ad126598d793fefcb6820/packages/jazz-tools/src/auth/AuthSecretStorage.ts#L134)
- Since `authSecretStorage` is [only updated externally](https://github.com/garden-co/jazz/blob/3185a20777df515ad84ad126598d793fefcb6820/packages/jazz-tools/src/implementation/ContextManager.ts#L46), `isAuthenticated` remains `false` after account registration

# Possible solution
Set `isAuthenticated` to `true` after registration

PR #1549, opened 2025-02-28
fix: fixes invalid authentication state when logging out after signUp

gdorsi:


packages/jazz-tools/src/auth/AuthSecretStorage.ts
```diff
@@ -16,7 +16,6 @@ export type AuthSetPayload = {
 export class AuthSecretStorage {
   private listeners: Set<(isAuthenticated: boolean) => void>;
   public isAuthenticated: boolean;
-  notify = false;
```
gdorsi, line 19: Replaced internal notify state with methods with specific intent.

packages/jazz-tools/src/implementation/ContextManager.ts
```diff
@@ -43,10 +43,6 @@ export class JazzContextManager<
   protected context: PlatformSpecificContext<Acc> | undefined;
   protected props: P | undefined;
   protected authSecretStorage = new AuthSecretStorage();
-  protected authSecretStorageWithNotify = Object.assign(
-    Object.create(this.authSecretStorage),
-    { notify: true },
-  );
```
gdorsi, line 49: This was the source of the bug.

Wanted to have a different notify the state when the AuthSecretStorage is used outside but did create also two different isAuthenticated states.

This was made to ensure that `isAuthenticated` is always in sync with the current account state.

We [already have tests](https://github.com/garden-co/jazz/blob/main/packages/jazz-react-core/src/tests/useAccount.test.ts#L59) to ensure that I'm not breaking that with this PR.

emmyoh (approved):

emmyoh:
Closes #1548

PR #1551, opened 2025-02-28
feat(examples): multi-auth example

emmyoh:
Example demonstrating multiple authentication methods in one Jazz app.
Derived from [the minimal Jazz & React example](https://stackblitz.com/edit/jazz-minimal-react).


bensleveritt:
I think it's okay to just copy the other auth examples, ie. put nothing in the app, just say 'Welcome, Name'

bensleveritt:
Just run the formatter on ./examples/counter/src/index.css too.

PR #1580, opened 2025-03-04
feat: jazz cloud auth & better auth integration

### LINKED ISSUE #2099: Cloud Authentication: E2E tests
https://github.com/garden-co/jazz/tree/emil-cloud-auth/examples/betterauth/tests

### LINKED ISSUE #2100: Cloud Authentication: Documentation
https://vercel.live/open-feedback/jazz-homepage-git-emil-cloud-auth-gcmp.vercel.app?via=pr-comment-feedback-link

emmyoh:
Supporting [Better Auth](https://www.better-auth.com/) servers for authentication would enable logging in email & password, various OAuth providers, magic links, and more.


examples/cloudauth/src/components/atoms/Icon.tsx
bensleveritt, line 1: @trishalim Do we want to use the Design System for our example apps, or is it better to keep copies of components?

examples/multiauth/src/apiKey.ts
bensleveritt, line 1: Good catch, thanks

examples/cloudauth/src/components/atoms/Icon.tsx
trishalim: It's better to keep a copy. It's easier for users imo.

Elfo404:
The changes to the design-system packages must be reverted, we use catalyst UI/tailwind plus for the design system,  we can't republish/package it in anything that is not an end-user product.

the components in jazz-react-auth-betterauth must also be removed, partly because of the previous point, and partly because they are not reusable anyway without our very specific tailwind config. `jazz-react-auth-betterauth` should not have any UI specific code. if we want to author example components as part of an example app we can do so in another PR. let's keep this focused on shipping working auth packages.

it's ok if `examples/betterauth` doesn't have any style for now or doesn't build after reverting the deisgn-system changes, do not worry about making it pretty, me and @trishalim can figure it out without blocking this PR.

emmyoh:
> The changes to the design-system packages must be reverted, we use catalyst UI/tailwind plus for the design system, we can't republish/package it in anything that is not an end-user product.
> 
> the components in jazz-react-auth-betterauth must also be removed, partly because of the previous point, and partly because they are not reusable anyway without our very specific tailwind config. `jazz-react-auth-betterauth` should not have any UI specific code. if we want to author example components as part of an example app we can do so in another PR. let's keep this focused on shipping working auth packages.
> 
> it's ok if `examples/betterauth` doesn't have any style for now or doesn't build after reverting the deisgn-system changes, do not worry about making it pretty, me and @trishalim can figure it out without blocking this PR.

Sounds good. So, I'll:
- Move the components out of `jazz-react-auth-betterauth` and into the example app
  - I'll open another PR putting unstyled components into their own package after we've merged this
- Undo anything relating to `@garden-co/design-system`
- Remove usage of the design system package from the example app

packages/jazz-react-auth-betterauth/src/contexts/Auth.tsx
```diff
@@ -0,0 +1,124 @@
...
+      if (data?.user.encryptedCredentials) {
+        useUpdateUser();
+      } else if (data && !data.user.encryptedCredentials) {
+        auth.signIn().then(() => {
+          useUpdateUser();
+        });
+      }
+      if (onSessionChange) onSessionChange();
+    });
+  }, [user, account, auth.state, isAuthenticated]);
```
Elfo404: i'm not quite sure i'm understanding the intended flow here. could you explain how the login/signup flow works?

Elfo404: so, if i'm understanding correctly:

on login:
- user sends email/pw
- a session is created on the client (here betterauth says we are logged in, jazz not yet)
- from the session - on the client - we get the encrypted account secret, send it to  the auth server
- the server returns the decrypted account credentials to the client
- the client signs in "with jazz" (at this point both better auth and jazz say we are logged in)

when the user signs in for the first time, the difference is that we encrypt & sends the local anonymous credentials to the server.

if so - do we need to expose the user object/session from better auth or should we hide that and let `useAccount` from jazz be the source of truth for the end user? (ie for the user name/email).

also, i've noticed the logging out using the better-auth auth client doesn't log out from jazz as well, and it's left to the user to implement this. is this intended?

emmyoh: Correct!

The Better Auth session should still be exposed via the authentication context, as it's useful if you're wanting to implement UI components that need Better Auth data (eg, a settings menu that list your account sessions). The state of the Better Auth account needs to be kept in the context anyway (for the `useEffect` to work correctly), so I figured it'd make sense to expose for that reason.

Regarding logging out, I could always add some two-line convenience function that logs you out of both if you think that's necessary, but AFAIK there's no way to programmatically guarantee logging out of both; someone could always write code that logs out of one but not the other. This bothered me too, but I didn't have too much time to reflect on the best way to address this (other than expecting app developers to not write invalid code). I'd appreciate it if you shared any solution you had in mind while I continue preparing this PR for merging (don't have enough mental capacity ATM).

packages/jazz-react-auth-betterauth/src/types/image.tsx
```diff
@@ -0,0 +1,19 @@
+import type { ComponentType } from "react";
```
Elfo404: image/link/router should not be needed here and can be removed also from the provider

packages/jazz-react-auth-betterauth/src/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  useJazzContext,
+} from "jazz-react";
+import { useEffect, useMemo } from "react";
+
+// biome-ignore lint/correctness/useImportExtensions: <explanation>
+export * from "./contexts/Auth";
+// biome-ignore lint/correctness/useImportExtensions: <explanation>
+export * from "./types/auth";
+// biome-ignore lint/correctness/useImportExtensions: <explanation>
+export * from "./lib/social";
```
Elfo404: does this work?
```suggestion
export * from "./contexts/Auth.js";
export * from "./types/auth.js";
export * from "./lib/social.js";
```

packages/jazz-react-auth-betterauth/src/types/image.tsx
emmyoh: Yep! Good catch. Working on it :)

packages/jazz-react-auth-betterauth/src/index.tsx
emmyoh: It didn't earlier. Let me try in a moment. I know the lint name suggests that it's about imports but I got the same lint error, which led me to ignoring it on those lines.

packages/jazz-react-auth-betterauth/src/types/image.tsx
emmyoh: What're your thoughts on keeping this so people can copy-paste the component files from the `betterauth` example app? If I modify those components to use Next.js-specific code, it'd require modifying each component after copy-pasting (as opposed to just passing things to the provider).

packages/jazz-react-auth-betterauth/src/index.tsx
emmyoh: Doesn't work for me :/

packages/jazz-react-auth-betterauth/src/types/image.tsx
Elfo404: i'm not getting why this package would have any of them tbh - it doesn't render any link/image, nor does perform any routing - which is handled completely in user-land. am i missing something?

emmyoh: There's no reason for this package itself to keep handling them; I see what you're saying. I was referring to the components in the `betterauth` example being able to be copy-pasted into a developer's new app. If we keep this, while not itself used in `jazz-react-auth-betterauth`, it'd be a convenient way to share the same `Link` or `Image` components across the copy-pasted authentication components. I see either removing it (and using the Next.js components directly in the example app) or keeping it (for ease of copy-pasting) to both have benefits, so wanted to ask. I'll remove it

packages/jazz-betterauth-server-plugin/src/index.ts
```diff
@@ -0,0 +1,271 @@
...
+                    },
+                  ],
+                  update: {
+                    encryptedCredentials: encryptedCredentials,
+                    salt: salt,
+                  },
+                });
+              }
+            }
+            symmetricSecret = ctx.body.secret;
```
gdorsi: **request:**  We should add some tests to cover this logic, it's indeed quite complex.

Cases that could lead to some serious data-loss:
- The key rotation fails in the middle of the operation
- A user signs up while we are rotating keys

Possible error situations:
- a user authenticate while we are rotating keys

I think these cases should be covered wit tests to consider this API usable, otherwise it's better not to provide the feature of rotating keys while the server is online.

To make it easier to cover this logic with tests we could extract it in a standalone module.

packages/jazz-auth-betterauth/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * Called when the authentication session changes.
+   * @param session The authentication session.
+   */
+  onUserChange = async (session?: AuthClient<T>["$Infer"]["Session"]) => {
+    if (!session || (session && !session["user"])) return;
+    const isAuthenticated = this.authSecretStorage.isAuthenticated;
+    if (!isAuthenticated) return;
+  };
```
gdorsi: Is this dead code?

examples/betterauth/tests/authFlow.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { randomBytes } from "node:crypto";
+import { test } from "@playwright/test";
+import { HomePage } from "./pages/HomePage";
+
+test("should sign up, sign in, and logout", async ({ page }) => {
```
gdorsi, line 5: How about adding these tests to the CI?

packages/jazz-auth-betterauth/src/tests/JazzBetterAuth.test.ts.disabled
```diff
@@ -0,0 +1,137 @@
+// @vitest-environment happy-dom
+
+import {
```
gdorsi: Why are these tests disabled?

packages/jazz-betterauth-server-plugin/src/crypto.ts
```diff
@@ -0,0 +1,119 @@
...
+  encoding: BufferEncoding = "hex",
+) => {
+  const passwordHash = await argon2.hash(password, {
+    salt: Buffer.from(salt, encoding),
+  });
+  const hash =
+    phc.deserialize(passwordHash).hash ??
+    Uint8Array.from(Buffer.from(passwordHash)).slice(0, 32);
+  return decryptString(ciphertext, hash, encoding);
+};
```
gdorsi: How do you feel about covering these functions with tests?

gdorsi (commented):
High-level looks good.

Maybe it would be better to comment out the key rotation APIs until we are seriously positive that the risk of data-loss there is low.

gdorsi:
The failures on CI seems to be related to the changes on the lockfile.

A possible fix would be:
```
git checkout origin/main pnpm-lock.yaml # Copy the lockfile from main
pnpm install # Update the lockfile with the new deps
```

packages/jazz-auth-betterauth/src/tests/JazzBetterAuth.test.ts.disabled
emmyoh: https://github.com/garden-co/jazz-cloud-dashboard/issues/143#issuecomment-2867684257

packages/jazz-betterauth-server-plugin/src/crypto.ts
emmyoh: Good idea. I'll get some written.

emmyoh: https://github.com/garden-co/jazz/tree/emil-cloud-auth/packages/jazz-betterauth-server-plugin/src/tests

packages/jazz-auth-betterauth/src/index.ts
emmyoh: Yep. Removed

packages/jazz-betterauth-server-plugin/src/index.ts
emmyoh: I'll rewrite most of this tomorrow, and I'll add tests too :)

Questions:
- How should I mock the DB in the tests?
- What *should* happen when a user signs up during key rotation?
  - I think the new user's credentials should be encrypted with the new secret.
- What *should* happen when a user authenticates during key rotation?
  - I think we should attempt decryption of the user's credentials with the new secret, then fall back to the old secret if that fails.
- How should we handle key rotation failing partway?
  - I think `ctx.context.adapter.update` should only be called if `[encryptedCredentials, salt]` is returned without throwing an error (ie, using `try`-`catch`)

emmyoh: I actually just removed the whole key rotation thing for now. Rotating the encryption key for the rest of the Better Auth DB isn't even supported; when we get back to this in the future (if we want it), we should either add a mechanism in Jazz Cloud to migrate data over to a new environment entirely, or we should work with the Better Auth developers to add this to their library.

packages/jazz-auth-betterauth/src/tests/JazzBetterAuth.test.ts.disabled
emmyoh: Tests deleted (E2E tests cover same functionality)

gdorsi (approved):
Let's go 🚀

PR #1589, opened 2025-03-05
Add React 19 to peer deps

3f765d5e-ac23-4170-97a5-68b8feee155b:
Hope that's enough to make the peer dep warning go away. Been running inspector with react 19 and no problems so far


gdorsi:
Thanks for the fix!

PR #1591, opened 2025-03-05
Add z-index of 1 to inspector window

3f765d5e-ac23-4170-97a5-68b8feee155b:
This is to show the inspector window at least above other elements that are relatively positioned but have no z-index applied.

Consider increasing the z-index value or make it configurable (maybe @layer could also help?)

### DIFF
--- a/packages/jazz-inspector/src/viewer/page.tsx
+++ b/packages/jazz-inspector/src/viewer/page.tsx
@@ -54,6 +54,7 @@ export function Page({
     <div
       style={{
         position: "absolute",
+        zIndex: 1,
         inset: 0,
         backgroundColor: "white",
         borderWidth: "1px",


trishalim:
It makes sense to have a higher number (maybe 100?) so that it's on top of anything else on the app

3f765d5e-ac23-4170-97a5-68b8feee155b:
> It makes sense to have a higher number (maybe 100?) so that it's on top of anything else on the app

Yes, setting it to 1 just fixes my immediate need.

Suggestion:

Set it one higher than the highest value from the default tailwind classes (which is `z-index: 50` I think). But also offer a prop to customize the z-index. And also set the same z-index for the toggle button

(Probably for a different issue: Prop to customize button position, `'top-left'`, `'bottom-right'` etc.)

trishalim:
@3f765d5e-ac23-4170-97a5-68b8feee155b 
I agree about the positioning

If the current positioning bothers you a lot right now you can also use the separate inspector app instead of the widget. 

Can you please make a changeset (patch) by running `pnpm changeset` on the jazz root dir? Thank you

3f765d5e-ac23-4170-97a5-68b8feee155b:
> Can you please make a changeset (patch) by running `pnpm changeset` on the jazz root dir? Thank you

Will do, need to set up the repo first though

PR #1612, opened 2025-03-07
create-jazz-app: replace project name param with directory name

### LINKED ISSUE #1588: Ability to specify creation path for create-jazz-app
I often already create the project folder manually and then want to run `pnpm create jazz-app .` from inside it (note the `.`)

In this case create-jazz-app should
- _not_ create a nested folder
- guess the project name

Additionally it would be nice
- in case I forget this and don't specify a path
- so I'm trying to create a project with name === current folder
- I get prompted if I want a nested folder or use the current folder directly

Note:

The "to get started" helper needs to be updated to not ask you to `cd` in case you're already in the right folder

trishalim:
`npx create-jazz-app .` creates a new jazz app under current directory, and takes current directory name as the project name to put in package.json

You can also pass a directory name `npx create-jazz-app my-app` and it create a my-app directory, and take my-app as the project name.

This makes `--project-name` redundant, so I removed that now.

fixes #1588

### DIFF
--- a/.changeset/dull-tools-run.md
+++ b/.changeset/dull-tools-run.md
@@ -1,0 +1,5 @@
+---
+"create-jazz-app": patch
+---
+
+add directory param to create-jazz-app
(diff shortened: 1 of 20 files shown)


bensleveritt (approved):
👍 Nicely done!

PR #1615, opened 2025-03-07
De-emphasise local-first on homepage

aeplay:


homepage/homepage/components/home/CollaborationFeaturesSection.tsx
```diff
@@ -10,18 +10,18 @@ const data = [
...
+    title: "Shared",
+    description: "Use Groups for precise access control.",
     codeSample: CollaborationPublic,
   },
   {
-    title: "Invite-only",
+    title: "By Invitation",
     description:
-      "Generate an invite link, and share only with people you want to collaborate with.",
+      "Works for users with no account yet.",
```
trishalim: That bit about the invite link is more interesting because this is a selling point.
"Works without an account" is the default behavior of a traditional tech stack

homepage/homepage/components/home/HowJazzWorksSection.tsx
```diff
@@ -94,7 +94,7 @@ export function HowJazzWorksSection() {
       <GappedGrid>
         <Step
           step={1}
-          description="Define your schema using Collaborative Values &mdash; your new building blocks."
+          description="Define the shape of your app’s data with Collaborative Values, your new cloud-synced building-blocks."
```
trishalim: I think "building blocks" was correct. cc @bensleveritt

PR #1590, opened 2025-03-05
Jazz 0.13 - Create dedicated packages for React Native Vanilla and Expo

boorad:
# React Native/Expo Package Restructuring

## Summary

This PR restructures our React Native and Expo implementations to provide better platform-specific support for both framework-less React Native and Expo applications. The changes include the creation of new packages, refactoring of existing ones, and reorganization of examples.

### New Package Structure

1. **New Core Package: `jazz-react-native-core`**
   - Provides shared functionality for both Expo and framework-less React Native
   - Contains platform-agnostic React Native functionality
   - Includes a generic SQLite adapter interface

2. **Refocused Package: `jazz-react-native`**
   - Now specifically targets framework-less React Native applications
   - Implements op-sqlite adapter for database storage
   - Adds MMKV store adapter for key-value storage

3. **New Package: `jazz-expo`**
   - Dedicated Expo implementation
   - Implements expo-sqlite adapter
   - Includes expo-secure-store adapter

### Storage Adapters

- **Expo SQLite adapter**: Uses `expo-sqlite` for Expo applications
- **op-sqlite adapter**: Uses `@op-engineering/op-sqlite` for React Native apps
- **MMKV store adapter**: Uses `react-native-mmkv` for key-value storage

### Example Applications

- Renamed `chat-rn` to `chat-rn-expo` to clarify its Expo affiliation
- Renamed `chat-rn-clerk` to `chat-rn-expo-clerk` to clarify its Expo affiliation
- Added new `chat-rn` example app for testing framework-less React Native
- Updated E2E test app structure (renamed from `e2e-react-native` to `e2e-rn-expo`)

### Testing

- Updated all tests to reflect the new package structure
- Fixed numerous tests for compatibility with the new architecture

### Documentation

- Updated package READMEs with clearer usage instructions
- Added better type definitions and export paths

### Compatibility

- All packages standardized on React Native version 0.76.7
- React version standardized to 18.3.1
- Updated other dependencies for consistency

## Implementation Details

- Created clear separation between platform implementations
- Ensured backward compatibility through proper exports configuration
- Added comprehensive testing for all storage adapters
- Improved VSCode debugging support with launch configurations

## Migration Guide for Existing Projects

1. **For Expo Applications**:
   - Use `jazz-expo` package
   - SQLite storage provided via `expo-sqlite`

2. **For Framework-less React Native**:
   - Use `jazz-react-native` package
   - Database storage via `op-sqlite`
   - Key-value storage via `react-native-mmkv`

[...]

PR #1618, opened 2025-03-07
cojson-storage-rn-sqlite, switch from async to sync sqlite API

### LINKED ISSUE #1544: Performance: try to use the SQLite sync API on cojson-storage-rn-sqlite
We have got reports from users that the performance of the SQLite storage on React Native is quite bad.

The idea is to try to switch to the sync API like we do on cojson-storage-sqlite and see if the performance improves

4rknova:


boorad:
@4rknova hey, so my initial thought is that this change would be short-lived.  I delete this package in #1590, and move parts of it into different packages.  For example, `getCoValue` is moved [here](https://github.com/garden-co/jazz/pull/1590/files#diff-f563eb9ad863f8479a3c13589d3981c6f6d9e678134612eddacaba32d0ac83aeR49).

Then we have to think about if this client changes depending on what sqlite flavor we are using.  Any thoughts there?

4rknova:
> @4rknova hey, so my initial thought is that this change would be short-lived. I delete this package in #1590, and move parts of it into different packages. For example, `getCoValue` is moved [here](https://github.com/garden-co/jazz/pull/1590/files#diff-f563eb9ad863f8479a3c13589d3981c6f6d9e678134612eddacaba32d0ac83aeR49).
> 
> Then we have to think about if this client changes depending on what sqlite flavor we are using. Any thoughts there?

Hey @boorad,
I agree, I'll go ahead and close this PR. Thanks!

PR #1624, opened 2025-03-10
feat: improve error logging

gdorsi:
Moving the raw errors in the { err } attribute to get better logging with the console (because the stack would be always visible) and when pino is used (because the `err` property is serialized using the "error" serializer)


Elfo404 (approved):
❤️

PR #1652, opened 2025-03-12
fix: align the profile migration in the passkey starter with the last best pratice

gdorsi:


starters/react-passkey-auth/src/schema.ts
```diff
@@ -49,5 +49,18 @@ export class JazzAccount extends Account {
         group,
       );
     }
+
+    if (this.profile === undefined) {
+      const group = Group.create();
+      group.addMember("everyone", "reader"); // The profile info are visible by everyone
```
aeplay:
```suggestion
      group.addMember("everyone", "reader"); // The profile info is visible to everyone
```

aeplay (approved):

PR #1655, opened 2025-03-13
test: fix the mobile e2e tests

gdorsi:
The latest fixes on the framework made the logout redirect correctly on the homepage, so changes on the e2e flow were needed in order to make it work.

### DIFF
--- a/examples/chat-rn/test/e2e/flow.yml
+++ b/examples/chat-rn/test/e2e/flow.yml
@@ -41,8 +41,10 @@ appId: com.jazz.chatrn
 
 # logout
 - tapOn: "Logout"
-- tapOn: "Join chat"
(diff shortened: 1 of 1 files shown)


examples/chat-rn/test/e2e/flow.yml
```diff
@@ -25,11 +25,8 @@ appId: com.jazz.chatrn
 - assertVisible: "Logout"
 
 # send a message
-- runFlow:
-    label: "Erase existing message"
-    file: erase_text.yml
-    env:
-      id: "message-input"
+- tapOn:
+    id: "message-input"
```
gdorsi, line 29: The text input is already empty, no need to clean it

boorad: I think I had restarted it once, locally, and it was not emptied.  But yeah, fresh emulator in CI, shouldn't be a problem.

PR #1717, opened 2025-03-24
fix: missing TOC on docs intro

trishalim:
There's currently no TOC on this page: https://jazz.tools/docs/react


bensleveritt:
What does the second bracket do in `[[slug]]`?

bensleveritt (approved):

trishalim:
> What does the second bracket do in `[[slug]]`?

@bensleveritt 

It means that the slug is optional https://nextjs.org/docs/app/building-your-application/routing/dynamic-routes#optional-catch-all-segments

For us, it means that `/docs/react` (no slug) is included

trishalim:
Rebased this to have the 0.12 changes

PR #1726, opened 2025-03-25
fix(inspector): CoFeeds and FileStreams are showing as "CoStream"

### LINKED ISSUE #1676: Rename CoStreams -> CoFeeds in the inspector
We currently show the internal name for CoFeeds in the inspector, which might be confusing as all of our documentation uses different terminology.

### LINKED ISSUE #1677: Label files as "FileStream" instead of "CoStream" on the inspector
Image

Image

trishalim:
fixes #1676 
fixes #1677


packages/jazz-inspector/src/viewer/use-resolve-covalue.ts
```diff
@@ -138,6 +150,12 @@ function subscribeToCoValue(
             extendedType = "record";
           }
         }
+      } else if (type === "costream") {
+        const coStream = detectCoStreamType(value as RawCoStream);
+
+        if (coStream.type === "binary") {
+          extendedType = "file";
+        }
```
trishalim, line 158: Is that right? 👀

gdorsi: Yes, that's correct

gdorsi (approved):
LGTM

Found a couple of issues unrelated to this PR.

The mime type here is incorrect:
Screenshot 2025-03-27 at 11 28 26

It's nice that we can detect PDFs, but maybe the fallback value should just be `unknown`.

The close button is hard to read in dark mode:
Screenshot 2025-03-27 at 11 29 16

PR #1734, opened 2025-03-26
perf: optimize Group.roleOf getter and made the transactions validation incremental for CoMap and CoFeed

gdorsi:
Refactored Group.roleOf to avoid any unneccessary value allocation.

Tested on the multi-cursor example.

Before:
Screenshot 2025-03-26 at 12 59 02

After the Group.roleOf optimization:
Screenshot 2025-03-26 at 12 59 12

After the transactions validation optimization:
Screenshot 2025-03-26 at 14 21 19

### DIFF
--- a/.changeset/giant-mails-sneeze.md
+++ b/.changeset/giant-mails-sneeze.md
@@ -2,4 +2,4 @@
 "cojson": patch
 ---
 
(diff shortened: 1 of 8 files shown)


bensleveritt (approved):
Works well for me!

PR #1746, opened 2025-03-28
fix(homepage): upgrade typedoc to v0.27 and ts v5.7

gdorsi:
Upgrading typedoc and typescript in the homepage package to fix the `[ERR_ASSERTION]: Missing signatures` error we are getting when running typedoc.

### DIFF
--- a/homepage/homepage/components/docs/packageDocs.tsx
+++ b/homepage/homepage/components/docs/packageDocs.tsx
@@ -38,11 +38,15 @@ export async function PackageDocs({
           <section key={category.title}>
             <h2>{category.title}</h2>
             {category.children.map((child) => (
-              <RenderPackageChild
-                child={child}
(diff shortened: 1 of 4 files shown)


homepage/homepage/components/docs/packageDocs.tsx
```diff
@@ -38,11 +38,13 @@ export async function PackageDocs({
           <section key={category.title}>
             <h2>{category.title}</h2>
             {category.children.map((child) => (
-              <RenderPackageChild
-                child={child}
-                key={child.id}
-                inPackage={packageName}
-              />
+              child.variant !== "document" && (
```
gdorsi, line 43: Ability to link external documents has been added. Turning it off for now

https://typedoc.org/documents/External_Documents.html

PR #1773, opened 2025-03-31
Cross-Device Account Transfer (formerly Magic Link Auth)

### LINKED ISSUE #581: PasskeyAuth: provide fallback API for moving between vendors

### LINKED ISSUE #1662: SecretURLAuth: Use a temporary agent to consume secret
Follow-up from #581

From [this comment](https://github.com/garden-co/jazz/pull/1379#pullrequestreview-2617232320):

> We want to use a temporary agent to reveal the credentials on to a throwaway account and don't risk that they are exposed to any persistent account.

jmsv:
- **_Logged-in device_** can show a QR code to log in a **_not-logged-in device_**, and
- **_Not-logged-in device_** can show a QR code to get logged in by a **_logged-in device_**

closes #581
closes #1662

TODO:

- [x] Docstrings on `MagicLinkAuth` functions
- [x] Changeset for `jazz-tools`, `jazz-react-core` and `jazz-react`
- [x] Tests
- [x] Documentation


packages/jazz-tools/src/auth/MagicLinkAuth.ts
```diff
@@ -0,0 +1,256 @@
...
+    { peer1role: "server", peer2role: "client" },
+  );
+  Account.getMe()._raw.core.node.syncManager.addPeer(magicLinkAuthPeer);
+
+  const { node } = await LocalNode.withNewlyCreatedAccount({
+    creationProps: { name: "Sandbox account" },
+    peersToLoadFrom: [localPeer],
+    crypto,
+  });
+  return Account.fromNode(node);
```
gdorsi: This node should be closed when the flow is completed or fails.

Can't find where this happens

packages/jazz-react-core/src/auth/MagicLinkAuth.tsx
```diff
@@ -0,0 +1,384 @@
...
+      context.authenticate,
+      authSecretStorage,
+      origin,
+      options,
+    );
+  }, [origin, options]);
+
+  useEffect(() => {
+    if (hasRunRef.current) return;
+    hasRunRef.current = true;
```
gdorsi: Does this ever become `false`?

What happens if the `url` changes?

packages/jazz-react-core/src/auth/MagicLinkAuth.tsx
```diff
@@ -0,0 +1,384 @@
...
+
+    async function handleFlow() {
```
gdorsi: Does this flow handle cancelation?

What happens if the component gets unmounted or the user cancels the auth process?

packages/jazz-react/src/auth/MagicLinkAuth.tsx
```diff
@@ -0,0 +1,40 @@
...
+export function useCreateMagicLinkAuthAsConsumer(
+  options?: Partial<MagicLinkAuthConsumerOptions>,
+) {
+  return useCreateMagicLinkAuthAsConsumerCore(window.location.origin, options);
+}
+
+export function useHandleMagicLinkAuthAsConsumer(
+  options?: Partial<MagicLinkAuthConsumerOptions>,
+) {
+  return useHandleMagicLinkAuthAsConsumerCore(
```
gdorsi: On the web we should handle `hash` navigation.

When the user paste a link with a different hash but same path the page is not reloaded.
With the current implementation the auth would not start.

packages/jazz-tools/src/auth/MagicLinkAuth/MagicLinkAuth.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    let transferCore = transfer._raw.core;
+    while (transferCore.header.ruleset.type === "ownedByGroup") {
+      transferCore = transferCore.getGroup().core;
+    }
+
+    const group = cojsonInternals.expectGroup(transferCore.getCurrentContent());
+    const inviteSecret = group.createInvite("writer");
+
+    return handlerUrl + `/${transfer.id}/${inviteSecret}`;
```
gdorsi: **suggestion:** how about throwing an error if the `handlerUrl` doesn't contain an hash?

Thinking about a way to enforce users to rely on the hash paths to reduce the expoure of the secrets

trishalim (approved):

PR #1780, opened 2025-04-02
refactor(inspector): use goober for css

trishalim:


gdorsi (approved):
Checked the inspector app and looks good

PR #1784, opened 2025-04-02
chore: replace queue size updown counter with 2 monotonic counters (push & pull)

Elfo404:
remove the previous queue size gauge and replaces it with two counters per each priority & peer role values. This will allow us to get better visibility into the amount of messages that are pushed to and pulled from the queue as opposed to only its size. We still can get the current size by subtracting them.

Also reduces the available priorities to 3, which are the ones actually being used, but i avoided changing the priority values to avoid mismatches between old versions of the clients & new version of the sync servers - or if we actually need to start using more priorities in the future

### DIFF
--- a/packages/cojson/src/PriorityBasedMessageQueue.ts
+++ b/packages/cojson/src/PriorityBasedMessageQueue.ts
@@ -1,6 +1,7 @@
 import { ValueType, metrics } from "@opentelemetry/api";
+import { PeerState } from "./PeerState.js";
 import type { CoValuePriority } from "./priority.js";
-import type { SyncMessage } from "./sync.js";
+import type { Peer, SyncMessage } from "./sync.js";
 
 function promiseWithResolvers<R>() {
   let resolve = (_: R) => {};
@@ -118,7 +119,10 @@ export class PriorityBasedMessageQueue {
     return this.queues[priority];
   }
 
-  constructor(private defaultPriority: CoValuePriority) {}
+  constructor(
+    private defaultPriority: CoValuePriority,
+    private peerRole: Peer["role"],
+  ) {}
 
   public push(msg: SyncMessage) {
     const { promise, resolve, reject } = promiseWithResolvers<void>();
@@ -130,6 +134,7 @@ export class PriorityBasedMessageQueue {
 
     this.enqueuedCounter.add(1, {
       priority,
+      role: this.peerRole,
     });
 
     return promise;
@@ -144,6 +149,7 @@ export class PriorityBasedMessageQueue {
 
     this.dequeuedCounter.add(1, {
       priority,
+      role: this.peerRole,
     });
 
     return this.queues[priority]?.shift();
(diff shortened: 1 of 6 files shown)


packages/cojson/src/PriorityBasedMessageQueue.ts
```diff
@@ -82,35 +85,79 @@ export class LinkedList<T> {
...
+    this.pushCounter = metrics
+      .getMeter("cojosn")
+      .createCounter(`${prefix}.pushed`, {
+        description: "Number of messages pushed to the queue",
+        valueType: ValueType.INT,
+        unit: "1",
+      });
+
+    this.pullCounter.add(0, this.attrs);
+    this.pushCounter.add(0, this.attrs);
```
Elfo404, line 122: this makes sure that those metrics are generated (and emitted) as soon as the queue is created. this is to avoid edge cases where one series reset is delayed, which would cause spikes or dips when queried - and it also more correctly represents the actual state of the queue after a restart.

gdorsi: Great explaination, thanks!

Could you add it into the code as a comment?

Elfo404: yep. good point. (is also covered by a test btw)

packages/cojson/src/PriorityBasedMessageQueue.ts
```diff
@@ -82,35 +85,84 @@ export class LinkedList<T> {
...
+     * Optional attributes to be added to the generated metrics.
+     * By default the metrics will have the priority as an attribute.
+     */
+    attrs?: Record<string, string | number>,
+  ) {
+    this.queues = [
+      meteredList({ priority: CO_VALUE_PRIORITY.HIGH, ...attrs }),
+      meteredList({ priority: CO_VALUE_PRIORITY.MEDIUM, ...attrs }),
+      meteredList({ priority: CO_VALUE_PRIORITY.LOW, ...attrs }),
+    ];
```
gdorsi, line 159: suggestion: how about using directly an object here?

```ts
this.queues = {
  [CO_VALUE_PRIORITY.HIGH]: meteredList({ priority: CO_VALUE_PRIORITY.HIGH, ...attrs }),
  [CO_VALUE_PRIORITY.MEDIUM]: meteredList({ priority: CO_VALUE_PRIORITY.MEDIUM, ...attrs }),
  [CO_VALUE_PRIORITY.LOW]: meteredList({ priority: CO_VALUE_PRIORITY.LOW, ...attrs }),
}
```

gdorsi (approved):
LGTM

packages/cojson/src/PriorityBasedMessageQueue.ts
Elfo404: something a la https://github.com/garden-co/jazz/pull/1834 would work?

Elfo404: @gdorsi if we then need to (rightfully so) cache the result of object.entries we are basically just at square 1 with an extra step, at that point i think i prefer this implementation. wdyt?

gdorsi: > @gdorsi if we then need to (rightfully so) cache the result of object.entries we are basically just at square 1 with an extra step, at that point i think i prefer this implementation. wdyt?

Agree

PR #1785, opened 2025-04-02
changed the constructor parameters of the BrowserPasskeyAUth class iin jazz-browser's PasskeyAuth.ts to enable extension of the class

joshua1:
So that we can have this and override the default methods like signUp, users can extend signup adding more proprties to profile on signup
even sending verification code and checking for email verification status on login
  ```
export class AppBrowserPasskeyAuth extends BrowserPasskeyAuth { 
  constructor(
    protected crypto: CryptoProvider,
    protected authenticate: AuthenticateAccountFunction,
    protected authSecretStorage: AuthSecretStorage,
    public appName: string,
    public appHostname: string = window.location.hostname,
  ) {
    super(crypto, authenticate, authSecretStorage, appName, appHostname);
  }
}
```

### DIFF
--- a/packages/jazz-browser/src/auth/PasskeyAuth.ts
+++ b/packages/jazz-browser/src/auth/PasskeyAuth.ts
@@ -19,9 +19,9 @@ import {
  */
 export class BrowserPasskeyAuth {
   constructor(
-    private crypto: CryptoProvider,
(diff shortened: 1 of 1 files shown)


gdorsi (approved):
LGTM, thanks!

PR #1786, opened 2025-04-02
fix(coValueState): support CoValue re-load after it is marked as unavailable and improve timeout

gdorsi:


packages/cojson/src/tests/sync.test.ts
```diff
@@ -956,27 +956,6 @@ test.skip("When a peer's incoming/readable stream closes, we remove the peer", a
     */
 });
 
-test("If we start loading a coValue before connecting to a peer that has it, it will load it once we connect", async () => {
```
gdorsi, line 959: This doesn't make much sense and probably related to some old workaround.

If there are no peers it should be correct to mark to CoValue as unavailable.

Hopefully the changes done on sync should cover the case where the workaround is aimed.

packages/cojson/src/coValueState.ts
```diff
@@ -176,11 +176,12 @@ export class CoValueState {
   async loadFromPeers(peers: PeerState[]) {
     const state = this.state;
 
-    if (state.type !== "unknown" && state.type !== "unavailable") {
+    if (state.type === "loading" || state.type === "available") {
```
gdorsi, line 179: Changed to this check because it is easier to reason about

packages/cojson/src/coValueState.ts
```diff
@@ -176,11 +176,12 @@ export class CoValueState {
...
       return;
     }
 
     if (peers.length === 0) {
+      this.moveToState(new CoValueUnavailableState());
```
gdorsi, line 184: This fixes some cases where the load is stuck in waiting

packages/cojson/src/coValueState.ts
```diff
@@ -192,7 +193,11 @@ export class CoValueState {
 
       // If we are in the loading state we move to a new loading state
       // to reset all the loading promises
-      if (this.state.type === "loading" || this.state.type === "unknown") {
+      if (
+        this.state.type === "loading" ||
+        this.state.type === "unknown" ||
+        this.state.type === "unavailable"
```
gdorsi, line 199: This change fixes the case where unavailable values weren't reloaded

packages/cojson/src/coValueState.ts
```diff
@@ -319,9 +337,10 @@ async function loadCoValueFromPeers(
             type: "not-found-in-peer",
             peerId: peer.id,
           });
+          resolve();
         }
-      }, CO_VALUE_LOADING_CONFIG.TIMEOUT);
-      await coValueEntry.state.waitForPeer(peer.id);
+      }, timeoutDuration);
+      await Promise.race([promise, coValueEntry.state.waitForPeer(peer.id)]);
```
gdorsi, line 343: Using Promise.race to be 100% sure that the timeout also moves the load to the next peer

aeplay (approved):

PR #1788, opened 2025-04-03
inspector: various fixes and css refactoring

trishalim:
- fix: console error due to nested button <button> elements
- move css out of grid-view.tsx and value-renderer.tsx
- add placeholder to Account Secret field
- add Card and Grid components
- fix: text inside grid items get cut off

image


gdorsi (approved):
LGTM!

PR #1793, opened 2025-04-03
This adds a helper to the jazz-svelte library to enable use of InboxSender as available in the jazz-react

joshua1:
This adds a helper to the jazz-svelte library to enable use of InboxSender as available in the jazz-react


packages/jazz-svelte/src/lib/tests/useInboxSender.svelte.test.ts
```diff
@@ -0,0 +1,137 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, screen, fireEvent, cleanup } from '@testing-library/svelte';
+import { InboxSender } from 'jazz-tools';
+import { getJazzContext, type RegisteredAccount } from '../jazz.svelte';
+import type { ID } from 'jazz-tools';
+import { default as TestComponent } from './components/useInboxSender.svelte';
+
+// Mock dependencies
+vi.mock('./jazz.svelte', () => ({
```
gdorsi:
```suggestion
vi.mock('../jazz.svelte', () => ({
```

packages/jazz-svelte/src/lib/tests/useInboxSender.svelte.test.ts
```diff
@@ -0,0 +1,137 @@
...
+  const mockInboxOwnerID = 'recipient456' as ID<RegisteredAccount>;
+  const mockSendMessage = vi.fn();
+  const mockInboxSender = {
+    owner: { id: mockInboxOwnerID },
+    sendMessage: mockSendMessage
+  };
+
+  beforeEach(() => {
+    // Mock the context
+    vi.mocked(getJazzContext).mockReturnValue({
```
gdorsi, line 30: I generally prefer to avoid mocks when possible.

First, because without mocks we are pushed on making Jazz more testable and also because, without mocks, the test is less aware of the implementation details and works using the public API.

To write this test without mocks you can refer to:
- [useInboxSender.test.ts](https://github.com/garden-co/jazz/blob/main/packages/jazz-react-core/src/tests/useInboxSender.test.ts) as example for testing the Inbox APIs
- [useAcceptInvite.svelte.test.ts](https://github.com/garden-co/jazz/blob/main/packages/jazz-svelte/src/lib/tests/useAcceptInvite.svelte.test.ts) as example for testing Svelte components without mocking the Jazz APIs

What do you think? 
How do you feel about giving it a try?

gdorsi (commented):
THanks for the contribution!

Code-wise LGTM.

Left one suggestion on the tests.

PR #1805, opened 2025-04-07
fix(inspector): show unknown mime type instead of application/octet-stream

### LINKED ISSUE #1738: Incorrect mime type shown in inspector
Image
This should be `unknown`

trishalim:
closes #1738

I also added the embedded inspector to the filestream example and displayed the covalue ID. This example was useful for this task.


gdorsi (approved):
Thanks!

PR #1834, opened 2025-04-08
chore: make queues an object

Elfo404:


packages/cojson/src/PriorityBasedMessageQueue.ts
```diff
@@ -175,8 +178,8 @@ export class PriorityBasedMessageQueue {
   }
 
   public pull() {
-    const priority = this.queues.findIndex((queue) => queue.length > 0);
-
-    return this.queues[priority]?.shift();
+    return Object.values(this.queues)
```
gdorsi: Can we cache this into a variable?

e.g. `this.queueList = Object.values(this.queues)` in the constructor

This function is heavily called, so less memory allocations we do the better.

gdorsi (approved):
LGTM

Thanks!

PR #1836, opened 2025-04-08
1830 feat: creating opengraph-image.tsx on homepage

### LINKED ISSUE #1830: Programatically Created Top Level Home Page Open Graph & Twitter Image Base
Get the first instance of a programatically created image for Open Graph & Twitter working for the homepage

https://nextjs.org/docs/app/api-reference/file-conventions/metadata/opengraph-image#generate-images-using-code-js-ts-tsx

Sammii-HK:
Implementing the first programatically created twitter image for the home page

process.cwd happens at the _project directory_ which is why the font is in the homepage/public folder

See [docs](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/opengraph-image#generate-images-using-code-js-ts-tsx):

>   // Font loading, process.cwd() is Next.js project directory
  const interSemiBold = await readFile(
    join(process.cwd(), 'assets/Inter-SemiBold.ttf')
  )
  
[Preferred font](https://vercel.com/docs/og-image-generation#limitations_)
> To maximize the font parsing speed, ttf or otf are preferred over woff

[Recommended Size](https://vercel.com/docs/og-image-generation#technical-details)
> Recommended OG image size: 1200x630 pixels

To preview the open graph image head to: https://jazz-homepage-git-1830-progmatically-created-homepa-ef4948-gcmp.vercel.app/opengraph-image?3a03271e6f670fc6

I had to add width and height properties to the logo svg as tailwind css classes eg 'w-7' have no effect due to the image being created early at build

Refactoring marketing copy into constant for single source of truth

NB: Anselm has approved the removal of the '.' at the end of the headline


homepage/design-system/src/app/components/organisms/OpenGraphImage.tsx
```diff
@@ -0,0 +1,49 @@
...
+  width: 1200,
+  height: 630,
+};
+
+export const imageContentType = "image/png";
+
+export default async function OpenGraphImage({ title }: { title: string }) {
+  const manropeSemiBold = await readFile(
+    join(process.cwd(), "public/fonts/Manrope-SemiBold.ttf"),
+  );
```
gdorsi, line 16: **suggestion(non-blocking):** shall we cache this file so we don't read from the fs on each request?

gdorsi (approved):
LGTM, thanks!

What happens when `public/fonts/Manrope-SemiBold.ttf` is moved or removed?
Does the build fail?

Asking because someone might delete this file breaking unknowingly OpenGraph

Sammii-HK:
> LGTM, thanks!
> 
> 
> 
> What happens when `public/fonts/Manrope-SemiBold.ttf` is moved or removed?
> 
> Does the build fail?
> 
> 
> 
> Asking because someone might delete this file breaking unknowingly OpenGraph 

Yeah if the file is moved it breaks the build on pre-render(from my memory), the font file has to be in the project directory of the open graph image, otherwise I would keep it in the design system with the other fonts 
This also means that anytime we reuse the open graph component in another project, eg garden computing homepage, we need to keep the font placement consistent, eg public/fonts (for now)

We can place the font anywhere in the project directory, I did have it in a top level 'assets' folder but after a discussion with Anselm to keep things neater we decided it was best placed in public.

Not sure if you have any ideas on how to prevent this possible regression from happening? Please refer to my PR description for the relevant doc links

Sammii-HK:
> LGTM, thanks!
> 
> What happens when `public/fonts/Manrope-SemiBold.ttf` is moved or removed? Does the build fail?
> 
> Asking because someone might delete this file breaking unknowingly OpenGraph

Here is a build failing on `pre render` because the font path does not exist

https://vercel.com/gcmp/jazz-homepage/5ronAwxWRcTg9morjhJYyatxZXo2?filter=errors

homepage/design-system/src/app/components/organisms/OpenGraphImage.tsx
Sammii-HK: If you could hint at what that may look like I would appreciate it, I am currently just following the [docs](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/opengraph-image#generate-images-using-code-js-ts-tsx) verbatim

gdorsi: Ah, all right then.

The Image is probably cached somewhere (maybe even generated at build-time) so we don't need this kind of optimization

gdorsi:
> Here is a build failing on pre render because the font path does not exist
> https://vercel.com/gcmp/jazz-homepage/5ronAwxWRcTg9morjhJYyatxZXo2?filter=errors

Awesome, I think this is enough to block regressions.

homepage/design-system/src/app/components/organisms/OpenGraphImage.tsx
Sammii-HK: To quote the docs

> By default, generated images are [statically optimized](https://nextjs.org/docs/app/building-your-application/rendering/server-components#static-rendering-default) (generated at build time and cached) unless they use [Dynamic APIs](https://nextjs.org/docs/app/building-your-application/rendering/server-components#server-rendering-strategies#dynamic-apis) or uncached data.

trishalim (approved):

PR #1842, opened 2025-04-09
fix(inspector): CoMaps showing as table instead of grid

### LINKED ISSUE #1843: Inspector shows a table with undefined rows for a CoMap
Schema:
```ts
export class AccountRoot extends CoMap {
  dreams = co.ref(Dreams)
  dreams_archive = co.ref(DreamsArchive)
  preferences = co.ref(Preferences)
  settings = co.ref(Settings)
  user_dream_tags = co.ref(UserDreamTags)
}

export class BaseAccount extends Account {
  profile = co.ref(AccountProfile)
  root = co.ref(AccountRoot)
}
```

The inspector showing a table, when CoMaps should be grid view. There are also fields shown that are not from AccountRoot. They are from the settings field.

Image

trishalim:
This is how we're checking for Record-like CoMaps:

```
          const children = Object.values(snapshot).slice(0, 10);

          if (
            children.every(
              (c) => typeof c === "string" && c.startsWith("co_"),
            ) &&
            children.length > 3
          ) {
            extendedType = "record";
          }
```

It takes the first 10 fields, and if all these fields point to a reference (not a primitive), and if there are more than 3 fields, then it's considered a Record-like CoMap.

This is not correct because a Record-like CoMap can also have primitive values. And it also falsely identifies normal CoMaps as Records if all the fields are references, like the following:

```
export class AccountRoot extends CoMap {
  dreams = co.ref(Dreams)
  dreams_archive = co.ref(DreamsArchive)
  preferences = co.ref(Preferences)
  settings = co.ref(Settings)
  user_dream_tags = co.ref(UserDreamTags)
}
```

I've not found a way to detect Record-like CoMaps, so the fix is to remove this checking, and we'll show Records like a regular CoMap, which looks fine imo.

fixes #1843

### DIFF
--- a/packages/jazz-inspector/src/viewer/use-resolve-covalue.ts
+++ b/packages/jazz-inspector/src/viewer/use-resolve-covalue.ts
@@ -93,17 +93,8 @@ export async function resolveCoValue(
       extendedType = "account";
     } else if (isGroup(snapshot)) {
       extendedType = "group";
-    } else {
-      // This check is a bit of a hack
-      // There might be a better way to do this
-      const children = Object.values(snapshot).slice(0, 10);
-      if (
-        children.every((c) => typeof c === "string" && c.startsWith("co_")) &&
-        children.length > 3
-      ) {
-        extendedType = "record";
-      }
     }
+    // TODO: check for Record-like CoMaps
   }
 
   return {
@@ -139,17 +130,8 @@ function subscribeToCoValue(
           extendedType = "account";
         } else if (isGroup(snapshot)) {
           extendedType = "group";
-        } else {
-          const children = Object.values(snapshot).slice(0, 10);
-          if (
-            children.every(
-              (c) => typeof c === "string" && c.startsWith("co_"),
-            ) &&
-            children.length > 3
-          ) {
-            extendedType = "record";
-          }
         }
+        // TODO: check for Record-like CoMaps
       } else if (type === "costream") {
         const coStream = detectCoStreamType(value as RawCoStream);
 
--- a/packages/jazz-inspector/src/viewer/use-resolve-covalue.ts
+++ b/packages/jazz-inspector/src/viewer/use-resolve-covalue.ts
@@ -94,7 +94,6 @@ export async function resolveCoValue(
     } else if (isGroup(snapshot)) {
       extendedType = "group";
     }
-    // TODO: check for Record-like CoMaps
   }
 
   return {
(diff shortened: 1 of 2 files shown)


packages/jazz-inspector/src/viewer/use-resolve-covalue.ts
```diff
@@ -93,17 +93,8 @@ export async function resolveCoValue(
       extendedType = "account";
     } else if (isGroup(snapshot)) {
       extendedType = "group";
-    } else {
-      // This check is a bit of a hack
-      // There might be a better way to do this
-      const children = Object.values(snapshot).slice(0, 10);
-      if (
-        children.every((c) => typeof c === "string" && c.startsWith("co_")) &&
```
trishalim, line 101: If the type of the value of the record-like comap is a primitive and not a reference, then this becomes false.

packages/jazz-inspector/src/viewer/use-resolve-covalue.ts
```diff
@@ -93,17 +93,8 @@ export async function resolveCoValue(
...
-        children.length > 3
```
trishalim, line 102: This doesn't make sense because you can have record-like comaps with less than 3 fields

aeplay (approved):

packages/jazz-inspector/src/viewer/use-resolve-covalue.ts
```diff
@@ -93,17 +93,8 @@ export async function resolveCoValue(
...
-      ) {
-        extendedType = "record";
-      }
     }
+    // TODO: check for Record-like CoMaps
```
gdorsi: Currently there isn't a way to identify a Record-like CoMap and as you have mentioned in the PR description the current visualization looks fine.

So I would suggest to remove this TODO and if in the future we find issues for maps with many fields, do fixes for both normal maps and Record-like CoMap.

gdorsi (approved):
LGTM, thanks!

packages/jazz-inspector/src/viewer/use-resolve-covalue.ts
gdorsi:
```suggestion
```

packages/jazz-inspector/src/viewer/use-resolve-covalue.ts
```diff
@@ -139,17 +130,8 @@ function subscribeToCoValue(
...
-          if (
-            children.every(
-              (c) => typeof c === "string" && c.startsWith("co_"),
-            ) &&
-            children.length > 3
-          ) {
-            extendedType = "record";
-          }
         }
+        // TODO: check for Record-like CoMaps
```
gdorsi:
```suggestion
```

Issue #1802, opened 2025-04-04
Jazz React `useIsAuthenticated()` does not keep in sync with Clerk


fi5u:
After logging in with the Jazz Clerk integration, when the Clerk auth session has expired (after 7 days on the free tier), Jazz's `useIsAuthenticated()` does not respect the new Clerk logged out state. This can be observed when using the `<UserButton />` component from @clerk/clerk-react – 7 days after logging in, the `<UserButton />` becomes disabled despite `useIsAuthenticated()` returns `true`.

# To reproduce

- Set up a Jazz React app using the Clerk example
- Place the `<UserButton />` in a post-auth part of the app
- Sign on the app up with a user
- Wait 7 days (providing session duration is set to 7 days in the Clerk dashboard)
- Notice that the `<UserButton />` component is no longer accessible

# Expectation

I would expect the Jazz Session to monitor Clerk's logged-in status and revert to the logged-out state when the Clerk auth session expires.

gdorsi:
Thanks for filing this bug, going to look into it!

gdorsi:
@fi5u fix released with Jazz 0.13.4

PR #1848, opened 2025-04-09
Fix Clerk Logout Issues

### LINKED ISSUE #1658: Auth clerk: calling Clerk.signOut from doesn't logout the user from Jazz
Reported here: https://discord.com/channels/1139617727565271160/1313981497010552852/1349696811652350035

> I think I needed to call both the `logOut` function from `useAccount` hook and also the `signOut` function from the `useAuth` hook. Now it works.

The user should only call `logOut`, and not `signOut` also. Trying to reproduce.

Maybe somehow related to @boorad reporting the need to tap the logout button twice to actually logout ?

pax-k:
Addresses the underlying logout issue with Clerk.

1. `packages/jazz-tools`: Added a delay before anonymous account creation to prevent race conditions and improved logout flow to properly handle context transitions

2. `examples/chat-rn-expo-clerk`: Added an auth state listener for immediate navigation redirects, providing a smoother authentication experience

Fixes #1658 #1802

### DIFF
--- a/packages/jazz-auth-clerk/src/index.ts
+++ b/packages/jazz-auth-clerk/src/index.ts
@@ -38,18 +38,13 @@ export class JazzClerkAuth {
   }
 
   onClerkUserChange = async (clerkClient: Pick<MinimalClerkClient, "user">) => {
-    const jazzIsAuthenticated = this.authSecretStorage.isAuthenticated;
-
     if (!clerkClient.user) {
-      if (jazzIsAuthenticated) {
-        await this.authSecretStorage.clear();
-      }
       return;
     }
 
-    if (jazzIsAuthenticated) {
-      return;
-    }
+    const isAuthenticated = this.authSecretStorage.isAuthenticated;
+
+    if (isAuthenticated) return;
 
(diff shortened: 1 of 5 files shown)


packages/jazz-auth-clerk/src/index.ts
```diff
@@ -38,13 +38,18 @@ export class JazzClerkAuth {
   }
 
   onClerkUserChange = async (clerkClient: Pick<MinimalClerkClient, "user">) => {
+    const jazzIsAuthenticated = this.authSecretStorage.isAuthenticated;
+
     if (!clerkClient.user) {
+      if (jazzIsAuthenticated) {
```
gdorsi: We need to ensure that if the user is already authenticated id doesn't get logged out during the load.

Reason: clerkClient.user might be undefined because we are still loading clerk.

Also this change affects both Web and Native, and it needs to be tested on both.

gdorsi (commented):
I like the fix!

We need to add some tests to ensure we aren't breaking anything.

For now having some unit tests would be ok, I can work on adding some integration tests in a separate branch.

PR #1850, opened 2025-04-09
fix(inspector): handle displaying a CoList of primitives

### LINKED ISSUE #1844: Inspecting a co.literal results to an error
Image

trishalim:
Before: inspecting a CoList of primitives would result into an error because we were treating the value of each item as a covalue id to subscribe to.

After:
This is a colist of co.strings
image

fixes #1844

### DIFF
--- a/.changeset/tender-goats-chew.md
+++ b/.changeset/tender-goats-chew.md
@@ -1,0 +1,5 @@
+---
+"jazz-inspector": patch
+---
+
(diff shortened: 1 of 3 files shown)


bensleveritt (approved):
Looks good. Thanks~!

gdorsi (approved):
Awesome, thanks!

PR #1866, opened 2025-04-10
adjusting footer title styling on dark mode

Sammii-HK:
Titles for the footer were unclear on dark mode, improving styling to reflect the style on light mode to differentiate headers from links

Screenshot 2025-04-10 at 10 22 43

### DIFF
--- a/homepage/design-system/src/app/components/organisms/Footer.tsx
+++ b/homepage/design-system/src/app/components/organisms/Footer.tsx
@@ -57,7 +57,9 @@ export function Footer({
               key={index}
               className="flex flex-col gap-2 text-sm col-span-6 md:col-span-2"
             >
-              <h2 className="font-medium">{section.title}</h2>
+              <h2 className="font-medium dark:text-stone-700 cursor-default">
+                {section.title}
+              </h2>
               {section.links.map((link, linkIndex) => (
                 <FooterLink
                   key={linkIndex}


homepage/design-system/src/app/components/organisms/Footer.tsx
```diff
@@ -57,7 +57,9 @@ export function Footer({
               key={index}
               className="flex flex-col gap-2 text-sm col-span-6 md:col-span-2"
             >
-              <h2 className="font-medium">{section.title}</h2>
+              <h2 className="font-medium dark:text-stone-700 cursor-default">
```
Elfo404, line 60: out of curiosity, why `cursor-default` instead of the default `text` that would be applied here?

Sammii-HK: For me it is an a11y and UX issue:

This is an edit summary grabbed from google which I feel explains my point well:

------


For heading text in a web page or application, whether to use a text cursor (caret) or the default cursor (arrow) depends on the context and user interaction expected. If the heading is meant to be editable or selectable, a text cursor is appropriate, otherwise, a default cursor is generally preferred. 
Here's a more detailed breakdown:

## Text Cursor (Caret):
Purpose: Indicates the current position for inserting or editing text. 
Usage: Use when the heading is meant to be editable, such as in a text editor or a form field. 
Example: A heading in a document editor where the user can type or edit the text. 

## Default Cursor (Arrow):
Purpose: Indicates the general cursor position and is the default cursor style in many applications. 
Usage: Use when the heading is not meant to be edited or interacted with directly, such as a static title or a section heading in a webpage. 
Example: The title of a blog post or the heading of a section on a website. 

### **When to consider a text cursor:**

Editability:
If the heading is intended to be editable by the user, use a text cursor to clearly indicate where the user can start typing.

Copying:
If the user might need to copy and paste the heading text, a text cursor can help with selection.

Form Fields:
If the heading is part of a form field where the user can enter text, use a text cursor to indicate the input area. 


### **When to consider a default cursor:**

Static Content:
If the heading is simply a title or section heading that doesn't require user interaction, use the default cursor. 

Read-Only:
If the heading is for display purposes only and not meant to be edited, use the default cursor. 

Navigation:
_If the heading is part of a navigation menu or a link, use the default cursor, which can be changed to a pointer cursor on hover to indicate interactivity. 
In summary: Use a text cursor for headings that are intended to be editable or interacted with directly, and use the default cursor for headings that are simply for display or navigation._ 



I feel that site wide there are many instances where a `caret` is being used when they are static, un-interactive headers/titles/text and should be the default cursor.

NB: you can still select the text with a default cursor

PS: I know many sites don't adhere to this but for me it makes the page look messy and jarring

trishalim: I'm wary of overriding browser defaults as it creates unexpected behavior

I've noticed Linear overrides their button cursor to be default instead of pointer in order to have a native feel. I also checked Notion landing page, and they have cursor default for their text. 

This is probably more of a preference thing so I'm fine either way

Sammii-HK: It's mainly on these headers as it makes the links less prominent with the cursor change, as they are all so close together.

I worked at ASOS who are hot on a11y and web standards and if you go to their footer they do this for the footer link section headings,

Notion also do the cursor default for the footer link section headings

Sammii-HK: I also don't agree with Linear overriding the button to be cursor default, this is bad UX and a11y, and it should be a pointer

trishalim (approved):

homepage/design-system/src/app/components/organisms/Footer.tsx
trishalim: Makes sense ✅

Elfo404: yeah, to be honest i also don't like the default cursor on buttons. agree with everything above though, i didn't see this was one of the small footer titles, it totally makes sense for this to not be a caret 👍

PR #1877, opened 2025-04-10
Relax the WebSocket types on jazz-nodejs

### LINKED ISSUE #1869: Relax the WebSocket types on jazz-nodejs
When doing:

```ts
    import { WebSocket } from "ws";

    const { worker } = await startWorker({ WebSocket, synServer });
```

It can happen that the types from `ws` are not compatible with our WebSocket type.

```
Type 'typeof WebSocket' is not assignable to type '{ new (url: string | URL, protocols?: string | string[]): WebSocket; prototype: WebSocket; readonly CONNECTING: 0; readonly OPEN: 1; readonly CLOSING: 2; readonly CLOSED: 3; }'.
  Types of property 'prototype' are incompatible.
    Property 'dispatchEvent' is missing in type 'import("/Users/me/code/my-project/node_modules/@types/ws/index").default' but required in type 'WebSocket'.ts(2322)
```

We should relax the type definition so users won't get these useless messages.

jmsv:
new `AnyWebSocketConstructor` type based on `AnyWebSocket` used in place of `typeof WebSocket`

fixes #1869

### DIFF
--- a/packages/jazz-nodejs/src/webSocketWithReconnection.ts
+++ b/packages/jazz-nodejs/src/webSocketWithReconnection.ts
@@ -1,10 +1,13 @@
 import { Peer } from "cojson";
-import { createWebSocketPeer } from "cojson-transport-ws";
+import {
+  AnyWebSocketConstructor,
+  createWebSocketPeer,
+} from "cojson-transport-ws";
 
 export function webSocketWithReconnection(
   peer: string,
   addPeer: (peer: Peer) => void,
-  ws?: typeof WebSocket,
+  ws?: AnyWebSocketConstructor,
 ) {
   let done = false;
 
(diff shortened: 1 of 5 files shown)


packages/jazz-nodejs/src/webSocketWithReconnection.ts
```diff
@@ -25,7 +28,7 @@ export function webSocketWithReconnection(
     timer = setTimeout(() => {
       const wsPeer: Peer = createWebSocketPeer({
         id: "upstream",
-        websocket: new WebSocketConstructor(peer) as any,
+        websocket: new WebSocketConstructor(peer),
```
gdorsi, line 31: ❤️

PR #1879, opened 2025-04-10
alert component styling

trishalim:
Test here: https://jazz-homepage-git-improvement-alert-styling-gcmp.vercel.app/docs/react/upgrade/0-11-0
You need to scroll down a bit

This is a component we use mostly in the dashboard. We made some improvements there, so I've copied it over and improved on it a bit more. 

image


bensleveritt (approved):
It's definitely something we want in the docs

trishalim:
I am abandoning this because the component we need for the docs need to be way less prominent
https://github.com/garden-co/jazz/issues/2314

PR #1880, opened 2025-04-10
Remove catchall-route for docs

aeplay:
This PR achieves several things:

- moves a bunch of things that are "pure content" into a top-level content folder so it can be easily edited without understanding the full app structure
- replaces the catchall `[[...slug]]` route segment for docs with a simpler `[topic]/[subtopic]` structure. This introduces some redundancies in the page.tsx's (factored out as much as possible) but unlocks two things:
  - we'll be able to have precise opengraph-images for doc pages with full metadata & mdx available (future PR)
  - we can get rid of the context hack to supply the table of contents. The layout is now pulled into the page.tsx's at each level, where the mdx & TOC is available


trishalim (approved):

bensleveritt (approved):
Works well enough for me

PR #1882, opened 2025-04-10
fix: handle null properties in the CoMap deep loading

gdorsi:


packages/jazz-tools/src/tests/deepLoading.test.ts
```diff
@@ -610,6 +610,34 @@ describe("Deep loading with unauthorized account", async () => {
...
+    class Lv3 extends CoMap {
+      string = co.string;
+    }
+
+    const map = Lv1.create(
+      { lv2: Lv2.create({ lv3: Lv3.create({ string: "hello" }, bob) }, bob) },
+      bob,
+    );
+
+    map.lv2!.lv3 = null;
```
gdorsi, line 632: Found that when the property is written via setter the value becomes `null`

Instead when an optional value is set as null in the constructor is casted to `undefined`

aeplay (approved):

PR #1883, opened 2025-04-10
Add opengraph images for docs

aeplay:
Also moves where H1s get filtered out of tables-of-content so we can get the actual title for both metadata and images


mculotta120 (dismissed):

aeplay:
Closing, needs to be precisely rebased

PR #1886, opened 2025-04-10
fix: handle Clerk expiration and fix logout

### LINKED ISSUE #1658: Auth clerk: calling Clerk.signOut from doesn't logout the user from Jazz

### LINKED ISSUE #1802: Jazz React `useIsAuthenticated()` does not keep in sync with Clerk

gdorsi:
fixes #1658
fixes #1802 

This PR turns upside down the Clerk auth management, making `clerk` our source of truth for the auth state.

This improves:
- The logout flow is much more simple. We clear the credentials when the user is not authenticated anymore in clerk.
- SignOut made from Clerk now works
- When the Clerk session expires the user is logged out from Jazz


examples/chat-rn-expo-clerk/app/_layout.tsx
```diff
@@ -33,6 +44,17 @@ export default function RootLayout() {
...
+  useEffect(() => {
+    if (fontsLoaded) {
+    } else {
+      SplashScreen.preventAutoHideAsync();
+    }
+  }, [fontsLoaded]);
+
+  if (!fontsLoaded) {
+    return null;
+  }
```
gdorsi, line 56: Picked these changes from #1848 as before the user wasn't redirected to the auth screen on logout.

packages/jazz-auth-clerk/src/index.ts
```diff
@@ -37,13 +38,47 @@ export class JazzClerkAuth {
...
+    return clerkClient.addListener((event) => {
+      const user = (event as Pick<MinimalClerkClient, "user">).user ?? null;
+
+      if (!isClerkAuthStateEqual(previousUser, user) || this.isFirstCall) {
+        this.onClerkUserChange({ user });
+        previousUser = user;
+        this.isFirstCall = false;
+      }
+    });
+  }
```
gdorsi, line 70: Moved the logic from the framework adapters here to reduce duplicates and make it more testable.

packages/jazz-auth-clerk/src/index.ts
```diff
@@ -37,13 +38,47 @@ export class JazzClerkAuth {
...
+
+  registerListener(clerkClient: MinimalClerkClient) {
+    let previousUser: MinimalClerkClient["user"] | null =
+      clerkClient.user ?? null;
+
+    // Need to use addListener because the clerk user object is not updated when the user logs in
+    return clerkClient.addListener((event) => {
+      const user = (event as Pick<MinimalClerkClient, "user">).user ?? null;
+
+      if (!isClerkAuthStateEqual(previousUser, user) || this.isFirstCall) {
```
gdorsi, line 64: This is part of the bugfix, we now call `onClerkUserChange` only on the first trigger or when the auth state really changes.

PR #1892, opened 2025-04-11
preserve side nav scroll state

trishalim:


homepage/homepage/lib/navigation-items.ts
```diff
@@ -0,0 +1,32 @@
+import { NavItemProps } from "gcmp-design-system/src/app/components/organisms/Nav";
+
+export const navigationItems: NavItemProps[] = [
```
aeplay: Should we move this into the new top-level `content` folder?

aeplay (approved):
Very nice, thanks!

PR #1901, opened 2025-04-11
fix: improve peer reconciliation

### LINKED ISSUE #1871: Improve the reconciliation between peers when calling `addPeer`
Currently we are facing an issue in production when the connection between edge servers drops and they try to reconnect.

This happens because the reconciliantion system implemented inside `addPeer` is quite naive:
- https://github.com/garden-co/jazz/blob/main/packages/cojson/src/sync.ts#L321

We could improve it by:
- Reviewing carefully the messages sent for each entry state, and do different things when re-connection to a pre-existing peer
- Excecute the reconcilation with low priority, to not bring down the core server

gdorsi:
fixes #1871


packages/cojson/src/sync.ts
```diff
@@ -285,27 +291,85 @@ export class SyncManager {
...
+  }
+
+  async addPeer(peer: Peer) {
     const prevPeer = this.peers[peer.id];
     const peerState = new PeerState(peer, prevPeer?.knownStates);
     this.peers[peer.id] = peerState;
 
     if (prevPeer && !prevPeer.closed) {
       prevPeer.gracefulShutdown();
+      await prevPeer.processMessages?.catch((e) => {});
```
gdorsi: We need to wait for the previous peer to process all the incoming messages before starting to process the new  peer messages.

This is probably why, on frequent reconnections, the sync server starts to have troubles.

packages/cojson/src/sync.ts
```diff
@@ -285,27 +291,85 @@ export class SyncManager {
...
         type: "COMBINE_WITH",
-        id,
+        id: coValue.id,
         value: coValue.knownState(),
       });
     }
   }
 
-  addPeer(peer: Peer) {
+  async startPeerReconciliation(peer: PeerState) {
```
gdorsi, line 240: Before we were sending the content without waiting for the `known` state from the server.

Done some changes so we just send a `load` request for each available coValue to the server.

Added some logic to make the server request the content if they notice that they miss some content from the client.

The downside is that we now wait for the server response before sending the content, but we should send less messages around (in theory)

packages/cojson/src/sync.ts
gdorsi: Quick fix, doesn't handle more than one peers in line to be added.

Will improve in the final version of the PR

packages/cojson/src/tests/sync.test.ts
```diff
@@ -654,32 +662,228 @@ test("When we connect a new server peer, we try to sync all existing coValues to
...
+      msg: {
+        action: "known",
+        asDependencyOf: undefined,
+        ...group.core.knownState(),
+      },
+    },
+    {
+      from: "client",
+      msg: map.core.newContentSince(undefined)?.[0],
+    },
```
gdorsi: We send the same content twice 🤔

packages/cojson/src/tests/sync.test.ts
```diff
@@ -654,32 +662,228 @@ test("When we connect a new server peer, we try to sync all existing coValues to
...
+        ...map.core.knownState(),
+      },
+    },
+    {
+      from: "client",
+      msg: {
+        action: "content",
+        ...map.core.newContentSince(knownStateBefore)?.[0],
+      },
+    },
```
gdorsi: We send the same content twice 🤔

packages/jazz-tools/src/coValues/account.ts
```diff
@@ -288,11 +288,15 @@ export class Account extends CoValueBase implements CoValue {
...
-    return this.create<A>({
+    const account = await this.create<A>({
       creationProps: options.creationProps,
       crypto: as._raw.core.node.crypto,
       peersToLoadFrom: [connectedPeers[0]],
     });
+
+    await account.waitForAllCoValuesSync();
+
+    return account;
```
gdorsi, line 299: Needed this change because the content reconciliation now requires more than one roundtrip.

packages/cojson/src/sync.ts
```diff
@@ -189,65 +190,6 @@ export class SyncManager {
...
-      this.trySendToPeer(peer, {
-        action: "load",
-        ...coValue.knownState(),
-      }).catch((e: unknown) => {
-        logger.error("Error sending load", { err: e });
-      });
-    }
-  }
-
-  async tellUntoldKnownStateIncludingDependencies(
```
gdorsi, line 228: These two methods have been optimized away. See below.

packages/cojson/src/PeerState.ts
```diff
@@ -150,4 +153,35 @@ export class PeerState {
     this.peer.outgoing.close();
     this.closed = true;
   }
+
+  async processIncomingMessages(callback: (msg: SyncMessage) => Promise<void>) {
```
gdorsi, line 157: Moved this logic inside PeerState to handle the tracking of incomingMessagesProcessingPromise internally

packages/cojson/src/sync.ts
```diff
@@ -189,56 +190,35 @@ export class SyncManager {
...
       coValue
         .getDependedOnCoValues()
-        .map((dependentCoID) =>
-          this.tellUntoldKnownStateIncludingDependencies(
-            dependentCoID,
-            peer,
-            asDependencyOf || id,
-          ),
-        ),
+        .map((id) => this.sendNewContentIncludingDependencies(id, peer)),
```
gdorsi, line 208: Refactored to send the `known` message inside `sendNewContentIncludingDependencies`

This way we can optimize away the known message when we have new content to sync.

packages/cojson/src/sync.ts
```diff
@@ -189,56 +190,35 @@ export class SyncManager {
...
+      peer.optimisticKnownStates.get(id),
     );
 
-    if (!peer.toldKnownState.has(id)) {
+    if (newContentPieces) {
+      for (const piece of newContentPieces) {
+        this.trySendToPeer(peer, piece).catch((e: unknown) => {
+          logger.error("Error sending content piece", { err: e });
+        });
+      }
```
gdorsi, line 220: Simplified this part, before:
```ts
     const optimisticKnownStateBefore =
        peer.optimisticKnownStates.get(id) || emptyKnownState(id);

      const sendPieces = async () => {
        let lastYield = performance.now();
        for (const [_i, piece] of newContentPieces.entries()) {
          this.trySendToPeer(peer, piece).catch((e: unknown) => {
            logger.error("Error sending content piece", { err: e });
          });

          if (performance.now() - lastYield > 10) {
            await new Promise<void>((resolve) => {
              setTimeout(resolve, 0);
            });
            lastYield = performance.now();
          }
        }
      };

      sendPieces().catch((e) => {
        logger.error("Error sending new content piece, retrying", { err: e });
        peer.optimisticKnownStates.dispatch({
          type: "SET",
          id,
          value: optimisticKnownStateBefore ?? emptyKnownState(id),
        });
        return this.sendNewContentIncludingDependencies(id, peer);
      });
```

Removed the scheduling code, now we have scheduling inside the outgoing messages.
Also the error management was quite unreachable.

packages/cojson/src/sync.ts
```diff
@@ -248,66 +228,101 @@ export class SyncManager {
...
+      }).catch((e: unknown) => {
+        logger.error("Error sending load", { err: e });
       });
     }
   }
 
-  addPeer(peer: Peer) {
+  nextPeer: Map<PeerID, Peer> = new Map();
+
+  async addPeer(peer: Peer) {
```
gdorsi, line 311: Major fix here.

Before we weren't waiting for the previous peer to completely close and we lacked support multiple subsequent reconnections.

packages/cojson/src/sync.ts
```diff
@@ -403,30 +402,23 @@ export class SyncManager {
...
 
       if (eligiblePeers.length === 0) {
-        // If the load request contains a header or any session data
-        // and we don't have any eligible peers to load the coValue from
-        // we try to load it from the sender because it is the only place
-        // where we can get informations about the coValue
-        if (msg.header || Object.keys(msg.sessions).length > 0) {
-          entry.loadFromPeers([peer]).catch((e) => {
-            logger.error("Error loading coValue in handleLoad", { err: e });
-          });
```
gdorsi, line 422: Removed this part because sending a `known` message gives us what we need for this situation (content syncing) and there is no need for requesting the CoValue load.

packages/cojson/src/sync.ts
```diff
@@ -461,27 +448,23 @@ export class SyncManager {
...
         id: msg.id,
         header: false,
         sessions: {},
       });
     }
-
-    if (entry.state.type === "available") {
-      await this.tellUntoldKnownStateIncludingDependencies(msg.id, peer);
-      await this.sendNewContentIncludingDependencies(msg.id, peer);
-    }
```
gdorsi, line 493: Moved up in the else/if

packages/cojson/src/sync.ts
```diff
@@ -493,31 +476,6 @@ export class SyncManager {
       value: knownStateIn(msg),
     });
 
-    if (entry.state.type === "unknown" || entry.state.type === "unavailable") {
-      if (msg.asDependencyOf) {
```
gdorsi, line 506: This was very likely a workaround.

I think it's not needed anymore

packages/cojson/src/sync.ts
```diff
@@ -729,10 +686,8 @@ export class SyncManager {
       if (peer.erroredCoValues.has(coValue.id)) continue;
 
       if (peer.optimisticKnownStates.has(coValue.id)) {
-        await this.tellUntoldKnownStateIncludingDependencies(coValue.id, peer);
         await this.sendNewContentIncludingDependencies(coValue.id, peer);
       } else if (peer.isServerOrStoragePeer()) {
-        await this.subscribeToIncludingDependencies(coValue.id, peer);
         await this.sendNewContentIncludingDependencies(coValue.id, peer);
```
gdorsi, line 717: Before https://github.com/garden-co/jazz/pull/1741/files we needed to send a `load` message to the server to trigger the load of the CoValue from storage before sending content.

Now it's not required anymore.

Also the subscription is handled inside sendNewContentIncludingDependencies

packages/cojson/src/tests/sync.load.test.ts
```diff
@@ -0,0 +1,336 @@
...
+      ),
+    ).toMatchInlineSnapshot(`
+      [
+        "client -> LOAD Map sessions: empty",
+        "server -> CONTENT Group header: true new: After: 0 New: 3",
+        "client -> KNOWN Group sessions: header/3",
+        "server -> CONTENT Map header: true new: After: 0 New: 1",
+        "client -> KNOWN Group sessions: header/3",
+        "client -> KNOWN Map sessions: header/1",
+        "client -> KNOWN Map sessions: header/1",
```
gdorsi, line 44: Still having a duplicated KNOWN message here, in the end as ACK. 🤔

packages/cojson/src/tests/sync.test.ts
```diff
@@ -33,454 +40,6 @@ beforeEach(async () => {
   jazzCloud = setupSyncServer();
 });
 
-test("Node replies with initial tx and header to empty subscribe", async () => {
```
gdorsi, line 36: Replaced these tests with the ones inside `sync.load` and `sync.peerReconciliation`

packages/jazz-tools/src/tests/groupsAndAccounts.test.ts
```diff
@@ -155,7 +155,7 @@ describe("Group inheritance", () => {
     const mapAsReaderAfterUpdate = await TestMap.load(mapInGrandChild.id, {
       loadAs: reader,
     });
-    expect(mapAsReaderAfterUpdate?.title).toBe("In Grand Child");
+    expect(mapAsReaderAfterUpdate).toBe(null);
```
gdorsi, line 158: This is actually what should happen when the user lose access to the map.

Probably because of the optimizations it happens with less event loop iterations.

tests/e2e/tests/RetryUnavailable.test.ts
```diff
@@ -21,8 +21,6 @@ test.describe("Retry unavailable states", () => {
       timeout: 20_000,
     });
 
-    await setTimeout(1000);
```
gdorsi, line 24: Had to remove this timeout, probably because since now the peer reconciliation doesn't send the content directly it takes more time.

packages/cojson-storage-sqlite/src/tests/sqlite.test.ts
```diff
@@ -0,0 +1,321 @@
...
+  expect(map2.get("hello")).toBe("world");
+
+  expect(
+    toSimplifiedMessages(
+      {
+        Map: map.core,
+        Group: group.core,
+      },
+      node2Sync.messages,
+    ),
```
gdorsi, line 115: Created a test utility to be able to use snapshot tests against the sync messages.

I've basically summarized all the important bits and removed all the randomized ids and sessions

emmyoh (commented):
Your comments (both in code and PR) were super useful. No feedback on how to better document this 👍

packages/cojson-storage-sqlite/src/tests/sqlite.test.ts
aeplay: love it

packages/cojson-storage-sqlite/src/tests/sqlite.test.ts
```diff
@@ -0,0 +1,321 @@
...
+  ).toMatchInlineSnapshot(`
+    [
+      "client -> LOAD Map sessions: empty",
+      "storage -> KNOWN Group sessions: header/3",
+      "storage -> CONTENT Group header: true new: After: 0 New: 3",
+      "client -> KNOWN Group sessions: header/3",
+      "storage -> KNOWN Map sessions: header/1",
+      "storage -> CONTENT Map header: true new: After: 0 New: 1",
```
gdorsi, line 123: We can probably optimize away those `storage -> KNOWN` messages because the content is enough to update the knownState

packages/cojson/src/sync.ts
aeplay: bravo!

packages/jazz-tools/src/coValues/account.ts
aeplay: thanks for catching it

packages/cojson/src/sync.ts
```diff
@@ -257,66 +237,101 @@ export class SyncManager {
...
-    const coValue = this.local.expectCoValueLoaded(id);
+  async startPeerReconciliation(peer: PeerState) {
+    const coValuesOrderedByDependency: CoValueCore[] = [];
 
-    await Promise.all(
-      coValue
-        .getDependedOnCoValues()
-        .map((id) => this.sendNewContentIncludingDependencies(id, peer)),
-    );
+    const requested = new Set<string>();
```
aeplay:
```suggestion
    const gathered = new Set<string>();
```

(`requested` sounds like you're already doing some syncing, when you're just gathering & ordering them)

packages/cojson/src/sync.ts
```diff
@@ -257,66 +237,101 @@ export class SyncManager {
...
 
-    const newContentPieces = coValue.newContentSince(
-      peer.optimisticKnownStates.get(id),
-    );
+    const buildOrderedCoValueList = (coValue: CoValueCore) => {
+      if (requested.has(coValue.id)) {
```
aeplay:
```suggestion
      if (gathered.has(coValue.id)) {
```

packages/cojson/src/sync.ts
```diff
@@ -257,66 +237,101 @@ export class SyncManager {
...
-      const optimisticKnownStateBefore =
-        peer.optimisticKnownStates.get(id) || emptyKnownState(id);
-
-      const sendPieces = async () => {
-        let lastYield = performance.now();
-        for (const [_i, piece] of newContentPieces.entries()) {
-          this.trySendToPeer(peer, piece).catch((e: unknown) => {
-            logger.error("Error sending content piece", { err: e });
-          });
+      requested.add(coValue.id);
```
aeplay:
```suggestion
      gathered.add(coValue.id);
```

aeplay (approved):
Just have one variable rename suggestion, otherwise all LGTM! Thank you very much

emmyoh (approved):

PR #1903, opened 2025-04-14
add user presence avatars to multicursor example

### LINKED ISSUE #1851: Add user presence avatars to multicursor example
Image

trishalim:
fixes #1851


examples/multi-cursors/src/components/Container.tsx
```diff
@@ -1,29 +1,87 @@
...
-            x: move.position.x,
-            y: move.position.y,
-          },
-        });
-      }}
-      remoteCursors={Object.values(cursors?.perSession ?? {})}
-      name={getName(me?.profile?.name, me?.sessionID)}
-    />
+    <>
+      <div className="absolute top-4 right-4 bg-white p-2 rounded-lg shadow">
```
jmsv, line 54: would it be possible to render avatars inside the SVG canvas rather than overlaying HTML?

trishalim: James and I decided to keep the HTML because 
- it's fixed to the top right, and that would require some calculations 
- users looking at this example might not be working with an SVG canvas

jmsv (approved):

examples/multi-cursors/src/components/Container.tsx
```diff
@@ -1,29 +1,90 @@
...
+    </span>
+  );
+}
+
 /** A higher order component that wraps the canvas. */
 function Container({ cursorFeedID }: { cursorFeedID: ID<CursorFeed> }) {
   const { me } = useAccount();
   const cursors = useCoState(CursorFeed, cursorFeedID, { resolve: true });
 
+  const remoteCursors = useMemo(
```
bensleveritt: @jmsv How do we feel about useMemo here?

bensleveritt (approved):

examples/multi-cursors/src/components/Container.tsx
trishalim: Had the same question so I will just remove it

PR #1904, opened 2025-04-14
 Establishing colour design tokens

Sammii-HK:
This PR establishes a few key items for a semantically named design system

It establishes colour variables which may be changed for light and dark mode, meaning you can say `text-hightlight` which means on light mode the colour is darker and more prominent, and on dark-mode it is set to be a lighter colour, this means developers do not need to concern themselves with the intricacies of colour systems but only need to know the token `highlight` which handles everything under the hood in tailwind

The main patterns which were established throughout the design are now distilled into 3 main design tokens:
1. `default`
1. `highlight` - text is 'more prominent'
1. `muted` - text is 'less prominent'
As expressed before, these tokens have the light and dark mode 'baked in', and remove the need for specifying a light and dark colour
Going forward these may be added to, but to create a cohesive design system my stance will always be to question why something 'unique' should be created and if it could be brought into the outstanding system, before adding extending to the theme.

The theme also establishes semantic naming such as:
- `primary`
- `secondary`
- `info`
- `highlight`
- `success`
- `warning`
- `tip`

this is the beginning on a design system which encapsulates all of the companies colours into a cohesive system, these are mainly unused as of yet but going forward alerts etc will use tokens such as `success` and `info` etc.
The tokens that currently hold precedence are `primary` and `secondary`, though mainly `primary`, as `secondary` is computed from this variable
`primary` for the current design system is established as the 'jazz blue' but is also tied into being the `primary` colour of the design system, so going forward whenever wanting to achieve the `*-blue` variable in css this should now be `*-primary` to ensure everything is encapsulated from the design variables and may be amended in a single place, created a single source of truth for all css and allow themes to be created/extended upon easily

Tailwind also allows you to configure colours for specific items such as `text` and `background`, so within the design system the colour for `text-highlight` is not the same as `bg-highlight` meaning they can be used in conjunction with one another, this also applies to `muted`, [see the configuration reference](https://v2.tailwindcss.com/docs/theme#configuration-reference)


bensleveritt:
Any chance we could trouble you for a demo/image of the tokens?

Sammii-HK:
> Any chance we could trouble you for a demo/image of the tokens?

Happy to create a doc around the tokens and usage, this PR was mainly introducing tokens were there were already patterns emerging from the design and condensing these into the tokens, eg, colour pairings for visual text hierarchy using colours, with light and dark mode supplied in a single token

If you are more interested in digging into the tokens in-situ, the inspector is going to be the best way to see how colours and tokens are reused across the site, all tokens are visible on the home page, with text hierarchy, `bg-highlight` on the svgs and quote highlights and everything plugged into `primary`

The fun comes mainly with (mostly) everything being plugged in to the `primary` colour, allowing themes to be spun off from this one variable with ease

bensleveritt:
I’m mostly interested in something visual to get a sense of the colours. We can get to usage later

examples/form/src/DraftIndicator.tsx
```diff
@@ -7,7 +7,7 @@ export function DraftIndicator() {
 
   if (me?.root.draft?.hasChanges) {
     return (
-      <div className="absolute -top-1 -right-1 bg-blue-500 border-2 border-white w-3 h-3 rounded-full dark:border-stone-925">
+      <div className="absolute -top-1 -right-1 bg-blue-500 border-2 border-highlight w-3 h-3 rounded-full dark:border-stone-925">
```
trishalim: I think you didn't mean to change this one. It's not using the design system.

examples/form/tailwind.config.ts
```diff
@@ -52,18 +52,6 @@ const config: Config = {
     formsPlugin,
     plugin(({ addBase }) =>
       addBase({
-        ":root": {
-          "--gcmp-border-color": stonePalette[200],
-          "--gcmp-invert-border-color": stonePalette[900],
-        },
```
trishalim, line 58: The borders don't look good in dark mode 

image

examples/organization/src/index.css
```diff
@@ -1,3 +1,5 @@
+@import "gcmp-design-system/src/app/globals.css";
```
trishalim: The example apps need to work on their own because we can clone them into their own repo / use them as templates like this:
`create-jazz-app --example organization`

So it's best to keep them independent from our design system.

homepage/design-system/src/app/components/atoms/logos/JazzLogo.tsx
```diff
@@ -22,7 +23,7 @@ export function JazzLogo({
...
-        fill="#3313F7"
+        fill={COLORS.BLUE}
```
trishalim: My personal preference is a class `fill-primary`  but up to you

trishalim:
There are more places where the jazz blue color needs to be changed such as favicon, jazz cloud latency map on the landing page, landing page diagram, example thumbnails on the examples page. 

I'd also like to see the full color palette for the blue. `--color-primary-dark` is a bit too indigo 🤔

Changing the brand color deserves its own separate PR I think.

homepage/design-system/src/app/components/atoms/Button.tsx
```diff
@@ -56,10 +56,10 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
 
     const variantClasses = {
       primary:
-        "bg-blue border-blue text-white font-medium bg-blue hover:bg-blue-800 hover:border-blue-800",
+        "bg-primary border border-primary text-white font-medium hover:bg-highlight hover:border-primary hover:text-primary dark:hover:bg-highlight dark:hover:text-primary",
       secondary:
-        "text-stone-900 border font-medium hover:border-stone-300 hover:dark:border-stone-700 dark:text-white",
-      tertiary: "text-blue underline underline-offset-4",
+        "text-stone-900 border font-medium hover:border-primary hover:text-primary hover:bg-highlight hover:dark:border-primary dark:text-white dark:hover:text-primary",
```
trishalim, line 61: Was this intentional?

primary
image

secondary
image

mculotta120:
Unrelated to this PR, I noticed in dark mode some of the colors were a bit difficult to read and ran an accessibility audit. I created this issue as a result, I'm not sure if this is a good opportunity to fix now while doing color work? Or leave as a follow up as it was already an issue. https://github.com/garden-co/jazz/issues/1911

trishalim:
@mculotta120 The code highlighting uses different css vars and it's already an issue so let's just address separately. Thanks for making the issue!

trishalim:
> I’m mostly interested in something visual to get a sense of the colours. We can get to usage later

What do you all think about deploying the homepage/design-system project, and having this in there?
image
 This way we can see all the colors from a vercel preview

examples/organization/src/index.css
Sammii-HK: Should we then think about properly packaging up our design system as an npm package so that it may be imported and used everywhere allowing all apps to be created with a similar aesthetic except when people specifically choose otherwise?

This just allows the apps to use the base colours we have set for `color`, `text` & `border`, but I can easily remove it,
the only thing which seemed to use the design system was setting a specific border colour as `--gcmp-border-color` which I have removed, as it seemed overkill for what it was actually achieving

homepage/design-system/src/app/components/atoms/Button.tsx
Sammii-HK: On hover state yes, otherwise it looks like this:

Screenshot 2025-04-15 at 08 21 30

examples/form/src/DraftIndicator.tsx
Sammii-HK: I will look into this, thanks for flagging

examples/organization/src/index.css
trishalim: > properly packaging up our design system as an npm package

Yes, but more for the dashboard than the example apps. 

I agree the border color was overkill. A class name is fine imo. I think that's the simplest way, and we try to keep the example apps minimal.

Sammii-HK:
> There are more places where the jazz blue color needs to be changed such as favicon, jazz cloud latency map on the landing page, landing page diagram, example thumbnails on the examples page.
> 
> I'd also like to see the full color palette for the blue. `--color-primary-dark` is a bit too indigo 🤔
> 
> Changing the brand color deserves its own separate PR I think.

In the interest of moving quickly and addressing the accessibility while creating an encompassing design system, this is the best way to achieve that

I am happy to improve on the colour system, it also needs to be added on to have `dark` and `light` values for all colours, `--color-primary-dark` may look out of place on it's own but in use for a link hover it is fit for purpose, I am happy to adjust the colour but we should not just look at a harmonious palette but also create semantic colours for specific use cases, happy to tweak going forward, this is just the prelude


Screenshot 2025-04-15 at 08 45 54
Screenshot 2025-04-15 at 08 45 57



This also allows us to spot anything which isn't encompassed within the design system and any outlying `blue` will be amended to `primary`, I will be focusing on switching out any blue I see for the next few days, and will likely be it's own follow up PR,

I will also publish the design system as a package on another PR and introduce it into the example apps 

PS. Will fix example apps with a class but will still work on publishing the design package

examples/form/tailwind.config.ts
Sammii-HK: They probably don't look good in light either, I am going to work on packaging up the design system to resolve this

Sammii-HK: This is now resolved

trishalim (approved):
Lgtm. Thanks so much! 🚀

PR #1920, opened 2025-04-15
Clarify peer known state

aeplay:
All this really does is
- replace the "dispatch pattern" with direct method calls (more redundancy in `PeerState`, but clearer flow in `sync.ts` where it really matters IMO)
- more strictly represent & enforce the `assumeInfallible` assumption for storage peers


aeplay:
(waiting for @gdorsi to merge main into his branch before taking this out of draft, to get a reasonable diff - but otherwise ready for review)

gdorsi (approved):

PR #1921, opened 2025-04-15
creating Icons for jazz specific CoX terms, which act like lucide icons

Sammii-HK:
Custom Lucide Icons for CoFeed & CoRecord

Screenshot 2025-04-15 at 15 54 54

NB: A caveat is that as these icons adhere to the strict 24x24 size with 1px padding rules set out by tailwind, they appear shorter than the other icons which are not as wide, my solution is to create a subset of jazz specific svgs which are all hand spaced accordingly as a set, and then exported so then nothing needs to be done after and everything will be cohesive


bensleveritt:
CoRecord is cute, like that
CoFeed might be better explained as an array of things [...], it's more list than object

trishalim:
Love it! especially the record one 🥹

Right now I'm not able to detect Record types though. It's hard to differentiate them from a CoMap so now they are just showing as CoMap. Maybe in the future.

trishalim:
Can we please also have an icon for ImageDefinition and FileStream 🙏

Sammii-HK:
> Can we please also have an icon for ImageDefinition and FileStream 🙏

I will encapsulate these in a follow up PR

aeplay:
> CoRecord is cute, like that CoFeed might be better explained as an array of things [...], it's more list than object

Interesting, I suggested object because to me a CoFeed is primarily a map from account -> value, with the possibility to get the whole history of values

bensleveritt:
> > CoRecord is cute, like that CoFeed might be better explained as an array of things [...], it's more list than object
> 
> 
> 
> Interesting, I suggested object because to me a CoFeed is primarily a map from account -> value, with the possibility to get the whole history of values

Ah, I guess this goes to show there's different mental models at play. For me it's a feed of events, primarily for access to the latest item. Accounts and session cuts are a secondary detail in my head

trishalim:
[...] makes more sense to me

image

aeplay (approved):
Moving the discussion about what CoFeeds are intuitively to Discord (it's an important one!) and approving this PR in the meantime since I'm happy with the way the icons are exposed

PR #1930, opened 2025-04-16
Use Turbo to determine Vercel deployments

bensleveritt:


bensleveritt:
Just going to add my notes here to come back to.

Projects in Vercel don't match what Turbo thinks are the names of the projects, so `process.env.APP_NAME` is nearly always wrong. To compensate for this adding project name to the env var `JAZZ_PROJECT_NAME` to the Vercel project can be picked up by the script.

Any attempt to use Turbo's `=filter=...[SHA]` seems to fail as the checkout is shallow and we can't use Git history to determine if the project has changed.

Further investigation needs to look at if there other ways to tell if a project has been changed.

bensleveritt:
https://github.com/garden-co/jazz/pull/2440 Is a better and simpler solution

Issue #1856, opened 2025-04-09
TwoSlash hover is clipped by container


bensleveritt:
![Image](https://github.com/user-attachments/assets/944ebece-a956-4a3d-a3ae-912b509f9cab)

PR #1944, opened 2025-04-17
update styles to handle mobile devices

mculotta120:
## fixes
[1856 ](https://github.com/garden-co/jazz/issues/1856)- Previous overflow fix wasn't quite right for mobile devices 

### Before
overflow-before

### After
overflow-after


trishalim (changes requested):
it's not working for me

i think we need overflow-x-visible so that the smallest breakpoint is targeted

PR #1946, opened 2025-04-18
feat: allow to assign the writeOnly role to everyone

gdorsi:
Adding support for writeOnly roles to everyone.

This is achieved by letting the accounts create their own writeKey and reveal it to the other group members


packages/cojson/src/coValues/group.ts
```diff
@@ -258,29 +258,38 @@ export class RawGroup<
...
 
-      this.set(
-        `${currentReadKey.id}_for_${EVERYONE}`,
-        currentReadKey.secret,
-        "trusting",
-      );
+      if (role === "writeOnly") {
+        if (previousRole === "reader" || previousRole === "writer") {
+          this.rotateReadKey();
+        }
```
gdorsi, line 283: Changing from reader/writer to writeOnly means that we remove the read rights, hence we need to rotate keys.

mculotta120 (approved):

aeplay (approved):
bellissimo!

PR #1952, opened 2025-04-19
Make design-system publishable

aeplay:


Sammii-HK:
Can't quite work out if we are using a locally transpiled version of the package in the apps or not,

If you import the app completely as an npm package will the "src/components" be in the path? Ideally when we build the package we want the atoms etc top level after the 'design-system' namespace

Sammii-HK (approved):

aeplay:
Agreed @Sammii-HK - a second step would be to separate this into the published package (only containing config and components) and then a "kitchen-sink" app that shows off everything in the published package and how to consume it

PR #1972, opened 2025-04-22
fix: do not crash when loading a CoValue core without the group in the sync server

gdorsi:


packages/cojson/src/coValueCore.ts
```diff
@@ -121,14 +121,35 @@ export class CoValueCore {
     this.header = header;
     this._sessionLogs = internalInitSessions;
     this.node = node;
+  }
+
+  groupInvalidationSubscription?: () => void;
+
+  subscribeToGroupInvalidation() {
```
gdorsi, line 128: Deferring the group subscription to the first `getCurrentContent` call.

This should both:
- remove the dependency requirement when loading a new CoValue in handleNewContent
- totally skip the group content loading on the sync server giving us a small perf boost

packages/cojson/src/coValueCore.ts
```diff
@@ -121,14 +121,35 @@ export class CoValueCore {
...
+            this.notifyUpdate("immediate");
+          },
+          false,
+        );
+      } else {
+        logger.error("CoValueCore: Owner group not available", {
+          id: this.id,
+          groupId,
         });
+      }
```
gdorsi, line 152: For extra safety, I've switched an error log instead of throwing an error.

The group here is required to invalidate the content when the permission change, so the group availability shouldn't be a strong requirement.

aeplay (approved):

Issue #1973, opened 2025-04-23
Add request invite orgs example


mculotta120:
Reference: https://github.com/garden-co/jazz/issues/1936

Add the workflow for request invite to the Organizations example

trishalim:
This conversation might be relevant https://discord.com/channels/1139617727565271160/1364231448219095061/1372894075627507773

Issue #1953, opened 2025-04-20
Not enough permissions to load / subscribe to CoValue


bdparrish:
### Description
A user that has been invited to a CoValue that does not have access to all CoValue children fails the invite - `Not enough permissions to load / subscribe to CoValue`

### Reproduce
1. Create root CoValue
2. Root CoValue has a field that is a CoList - `channels` in this example
3. The CoList group extends the root CoValue group
4. Add a CoValue to the CoList that is only accessible to one of the current members of the CoList
5. Invite a new user
6. Observe that the invite fails because the new user cannot load / subscribe to the CoValue in the CoList that was added  in step 4

### Expectation
I would expect the invited user has access to all available CoValues, CoList, etc that is the root CoValue group or extends the root CoValue group.

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/88a6dbe2-0ff0-4295-ac74-214f0d2aa63d" />

bdparrish:
schema.tsx

``` javascript
import { Account, CoList, CoMap, Group, co } from "jazz-tools";

export class Item extends CoMap {
    name = co.string;
}

export class ListOfItems extends CoList.Of(co.ref(Item)) { }

export class Organization extends CoMap {
    name = co.string;
}

export class List extends CoMap {
    name = co.string;
    items = co.ref(ListOfItems)
    organization = co.ref(Organization)
}

export class ListOfLists extends CoList.Of(co.ref(List)) { }

export class BaseAccountRoot extends CoMap {
    lists = co.ref(ListOfLists);
    currentList = co.optional.ref(List)
    currentItem = co.optional.ref(Item)
}

export class BaseAccount extends Account {
    root = co.ref(BaseAccountRoot);

    async migrate(this: BaseAccount, creationProps?: { name: string }) {
        const meGroup = Group.create({ owner: this })

        if (this.root === undefined) {
            this.root = BaseAccountRoot.create(
                {
                    lists: ListOfLists.create([], meGroup)
                },
                meGroup,
            );
        }
    }
}

```

addList
```javascript
const addItem = (name: string, organization: Organization) => {
    const orgGroup = organization._owner as Group
    const listGroup = Group.create({ owner: me })
    orgGroup.extend(listGroup)

    const list = List.create({
      name,
      organization
    }, { owner: listGroup })

    const item = Item.create(
      {
        name: 'todo',
      },
      { owner: listGroup }
    )

[...]

PR #1974, opened 2025-04-23
#1973 - Add request invite to organization example

mculotta120:
## what this does:

Adds ability to request invites to organizations in the Organizations example

#1973


mculotta120:
I believe I'm encountering the same bug described here: https://github.com/garden-co/jazz/issues/1953

PR #1987, opened 2025-04-23
fix: skip non-schema related keys when calling CoMap.toJSON

gdorsi:


packages/jazz-tools/src/coValues/coMap.ts
```diff
@@ -308,36 +308,48 @@ export class CoMap extends CoValueBase implements CoValue {
    * @category Content
    */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  toJSON(_key?: string, seenAbove?: ID<CoValue>[]): any[] {
-    const jsonedFields = this._raw.keys().map((key) => {
+  toJSON(_key?: string, seenAbove?: ID<CoValue>[]): any {
```
mculotta120: Optional: I was a bit confused by this parameter name. Would something like processedValues be more descriptive?

mculotta120 (commented):
I just had an optional comment as the parameter name was somewhat confusing

mculotta120 (approved):

PR #2008, opened 2025-04-26
[Expo Adapter] Allow to create new connections

tobiaslins:
image

https://github.com/expo/expo/issues/28176#issuecomment-2062361077


gdorsi:
This change has been integrated in the new expo storage driver.

Thanks for the suggestion!

Issue #1806, opened 2025-04-07
Add $skipInvalid to handle lists/records with inaccessible items more gracefully


gdorsi:
## Problem

When resolving a list where some items are inaccessible to the current user (similar to a Google Docs shared folder with varying permissions), the current system fails the entire list resolution if any single item is inaccessible. There is no built-in way to access only the permitted items.

## Current workaround

Users currently need to implement a manual workaround that catches errors and handles individual items:

```ts
const list = useCoState(ListOfTasks, id, { resolve: true }) // Not using $each
function preloadList() {
  list.ensureLoaded({ resolve: { $each: { childTasks: { $each: true } } } }).catch(err => {
     logger.warn(err); 
  })
}
// Optional, just useful if you need to do some preload
useEffect(() => {
  preloadList()
}, [id])
return list?.map(item => {
  if (!item) return null;
  return <Task key={item.id} item={item} />
})
```

## Proposed solution

Add an option to resolve lists with null values in place of inaccessible elements:

```ts
const list = useCoState(ListOfTasks, id, { resolve: { $each: true, $skipInvalid: false } }) 
```

This approach preserves the original list structure by keeping null placeholders rather than shifting indices, which would be more complex to implement and could cause reference management issues.

Issue #1750, opened 2025-03-28
Failed to load coValue from peer / loading a huge coList


aheissenberger:
**Error:**
```
Failed to load coValue from peer 
Object { coValueId: "co_z4RZerBy5kg6QnwfEi6N5po1nMC", peerId: "indexedDB", peerRole: "storage" }
```

I used the example from guided tour and created 4.000 issues under one project with a [worker process under node](https://github.com/aheissenberger/jazz-data-sharing-app-demo/blob/b6db7154329e002b09adc0d8adbd450b996ba022/src/worker/cli.ts)

My [github example repository](https://github.com/aheissenberger/jazz-data-sharing-app-demo/tree/master)

When I scroll through the issues of the project to the end of the list. The app hangs every xx lines, loads some data, even when all data was synced. On the first scrolls I see many `Failed to load coValue from peer` errors in the browser console and sometimes it is not possible to get all items.

My second problem is related to the [worker process under node](https://github.com/aheissenberger/jazz-data-sharing-app-demo/blob/b6db7154329e002b09adc0d8adbd450b996ba022/src/worker/cli.ts) which fails to delete the issues:
`pnpm jazz-worker --project-id=co_XXXXXXXXX --delete`
("jazz-worker" is defined in `packages.json`)

From my logging it looks like the worker is not able to get the coList `project.issues` synced. With small number of items up to 100 it works without any problem.

The process to create X issues is:
`pnpm jazz-worker --project-id=co_XXXXXXXXX --count=4000`

gdorsi:
Thanks for filing the issue and set up a repro!

4rknova:
Hi @aheissenberger,

Thank you for reporting this issue.
I was able to repro and it seems that this is due to sync timeouts.
I will investigate further.

aheissenberger:
@4rknova This is not a fix of the underlying problems in jazz. Using paging with max 50 items per page is a work-a-round by avoiding handling more rows where 5.000 rows is a very very low number and no problem for react. I tried the same with [gunjs](https://gun.eco) and had no issues with blocked react render queue and no errors at all.

gdorsi:
@aheissenberger 

Well, can't disagree with you. 

We just wanted to give you a quick solution, but if you really want to load this much data all at once, you will have to wait for a proper fix to land.

Honestly we still have alot to optimize, so we will surely get there in no time!

Thanks again for the repro, will use it as a first performance goal 😄

gdorsi:
[...]

PR #2015, opened 2025-04-28
A new subscription engine for jazz-tools

### LINKED ISSUE #2079: Rewrite the jazz-tools subscription system for perf and resolve features

gdorsi:
This PR introduces a completely rewritten subscription engine for jazz-tools.

It aims to solve some longstanding issues we have with the current engine:
- the current subscription engine adds alot of overhead when loading large lists of refs (see https://github.com/garden-co/jazz/issues/1750)
- it's hard to build new features on top of it (e.g. https://github.com/garden-co/jazz/issues/1806)
- it emits too many updates

The changes should be backward compatible, and so not require any upgrade guide.

fixes #2079

An informal test on large lists (around 20k items) shows a 15x performance improvement


packages/jazz-tools/src/subscribe/CoValueCoreSubscription.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (this.unsubscribed) return;
+
+        if (value !== "unavailable") {
+          this.subscribe(value.getCurrentContent());
+        } else {
+          this.listener("unavailable");
+          this.subscribeToState();
+        }
+      });
+    }
```
gdorsi, line 29: This will probably simplified by the new CoValueCore subscription system

packages/jazz-tools/src/subscribe/SubscriptionScope.ts
```diff
@@ -0,0 +1,523 @@
...
+    if (this.value.type === "unloaded") return false;
+
+    // If the value is in error, we send the update regardless of the children statuses
+    if (this.value.type !== "loaded") return true;
+
+    for (const value of this.childValues.values()) {
+      // We don't wait for autoloaded values to be loaded, in order to stream updates
+      // on autoloaded lists or records
+      if (value.type === "unloaded" && !this.autoloaded.has(value.id)) {
+        return false;
```
gdorsi, line 197: The check on `autoloaded` here is an important UX detail for the React/frameworks integration.

It makes the autoloaded deps "stream" results as opposed as waiting for all the autoloaded deps before triggering a single update.

We could make the behavoir configurable to support the cases where it's better to reduce the number of updates (e.g. when subscribing in a worker)

aeplay (approved):
LGTM, beautiful

packages/jazz-tools/src/subscribe/SubscriptionScope.ts
aeplay: I think this is basically equivalent to the current autoload behaviour - and you're exactly correct about the tradeoff: stuff appears quicker but there is a danger of overloading the framework with updates to render.

Let's see what makes sense once we remove autoload. Probably this interacts with pagination as well where you typically want to wait for a whole page of items, but not more.

MentalGear:
Indeed, by default skipping non-permitted items of a list and returning a valid list even if of only one item would be much appreciated.

( related discussion https://discord.com/channels/1139617727565271160/1139617728580305018/1369652275727044630 )

PR #2050, opened 2025-04-30
fix(colist): correct successor order when handling conflicts

gdorsi:


packages/cojson/src/tests/coList.test.ts
```diff
@@ -128,11 +126,11 @@ test("appendItems at index", () => {
...
   expect(content.toJSON()).toEqual([
     "first",
-    "hello",
-    "world",
     "second",
     "third",
     "fourth",
+    "hello",
+    "world",
   ]);
```
gdorsi, line 134: This test is most likely what induced the reverse successors thing.

So it looks like that by fixing the conflicts management we can break the `append at index` operation.

PR #2061, opened 2025-04-30
fix(sync): remove redundant content messages when loading a coValue from multiple peers

### LINKED ISSUE #2059: Discovery: rendundant content messages sent to server peers during the initial load
As highlighted in [this discussion](https://github.com/garden-co/jazz/pull/2057/files#r2068715420) we probably have some redundancy on the initial load phase when loading from storage and server peers.

We should:
- Try to reproduce the issue in a real application
- If it's reproducible provide a fix

The issue seems to be related to `syncCoValue` that syncs content to peers with an unkwnown known state.

When we have never sent any message about a specific coValue to a peer we set the known state as empty, leading us to send content to that peer assuming that it doesn't know anything about it.

While it doesn't compromise the correctness of the sync it probably heavily affect it's performance.

gdorsi:
fixes #2059 

Removes all the unnecessary content messages exchanged during the application bootstrap.

Difference quite noticeable on the network logs:

| Before | After |
|--|--|
| Screenshot 2025-04-30 at 23 14 41 | Screenshot 2025-04-30 at 23 15 03 | 

Visible on reload when the data is on storage

Production invite: https://todo-demo.jazz.tools/#/invite/project/co_zUCiwXkkeWibV3FVrtruH7sAwFK/inviteSecret_zEDusFyjri7wXU9YFgdr63kTXoGqNDhDgHjgxXMkrS1DH

Preview invite: https://jazz-todo-git-fix-redundant-messages-gcmp.vercel.app/#/invite/project/co_zUCiwXkkeWibV3FVrtruH7sAwFK/inviteSecret_zEDusFyjri7wXU9YFgdr63kTXoGqNDhDgHjgxXMkrS1DH


packages/cojson/src/PeerState.ts
```diff
@@ -226,6 +226,7 @@ export class PeerState {
       peerId: this.id,
       peerRole: this.role,
     });
+    this.peer.crashOnClose = false;
```
gdorsi, line 239: This fixes some errors I've started to get when closing nodes with a storage peer.

Don't know why it didn't happen before

aeplay (approved):

PR #2069, opened 2025-05-01
createJazzPlugin support custom ProseMirror schema

jmsv:
infer custom schema from view/doc

enables stuff like unordered/ordered lists as per [ProseMirror Basic Example](https://prosemirror.net/examples/basic/)

Screen Recording 2025-05-01 at 10 36 59


bensleveritt (approved):
Yeah, looks great. Thanks!

PR #2074, opened 2025-05-02
Combine CoValueCore and CoValueState Part 1 & 2

### LINKED ISSUE #2077: Combine CoValueState and CoValueCore to centralise both verified and syncing state

aeplay:
Fixes #2077 

Tried to combine CoValueCore and CoValueState, but realised that it would be nice to split verifying logic out of CoValueCore first.

Then, as a first step towards combining CoValueCore and CoValueState, I represented unavailable CoValues with "has core, but no verified state" instead of "has no core".

At that point some tests about migrations were failing, so I decided to tackle the cursed node bootstrap first, continuing in #2108 for clarity


aeplay:
Note: as of [e40e01f](https://github.com/garden-co/jazz/pull/2074/commits/e40e01fce0c1f7d2f783da50b739323da91dfafe) two tests related to migrations on loaded accounts still fail (in jazz-tools and jazz-nodejs)

Going to attempt a clean up of node bootstrap to understand better what's going on.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -95,42 +65,102 @@ export type DecryptedTransaction = {
...
-  _sessionLogs: Map<SessionID, SessionLog>;
-  _cachedContent?: RawCoValue;
-  listeners: Set<(content?: RawCoValue) => void> = new Set();
-  _decryptionCache: {
+  readonly node: LocalNode;
+  private readonly crypto: CryptoProvider;
+  private _verified: VerifiedState | null;
+  get verified() {
+    return this._verified;
+  }
```
gdorsi, line 97: **request:** could you document via comment what `verified` state means?

More specifically, what we can do without a verified state?

packages/cojson/src/coValueState.ts
```diff
@@ -219,17 +220,22 @@ export class CoValueState {
...
-  internalMarkMagicallyAvailable(coValue: CoValueCore) {
+  internalMarkMagicallyAvailable(
+    verified: VerifiedState,
+    { forceOverwrite = false }: { forceOverwrite?: boolean } = {},
+  ) {
     const previousState = this.highLevelState;
-    this.core = coValue;
+    this.core.internalShamefullyCloneVerifiedStateFrom(verified, {
+      forceOverwrite,
+    });
```
gdorsi: question: why are we cloning the verified state when marking the coValue as available?

packages/cojson/src/coValueCore/verifiedState.ts
```diff
@@ -0,0 +1,382 @@
...
+        [...this.sessions.entries()].map(([sessionID, sessionLog]) => [
+          sessionID,
+          {
+            lastSignature: sessionLog.lastSignature,
+            lastHash: sessionLog.lastHash,
+            streamingHash: sessionLog.streamingHash.clone(),
+            signatureAfter: { ...sessionLog.signatureAfter },
+            transactions: [...sessionLog.transactions],
+          } satisfies SessionLog,
+        ]),
```
gdorsi: **nit:** we can make this a little faster this way

```ts
const clonedSessions = new Map();

for (let [sessionID, sessionLog] of this.sessions) {
  clonedSessions.set(sessionID,  {
    lastSignature: sessionLog.lastSignature,
    lastHash: sessionLog.lastHash,
    streamingHash: sessionLog.streamingHash.clone(),
    signatureAfter: { ...sessionLog.signatureAfter },
    transactions: sessionLog.transactions.slice(),
  } satisfies SessionLog);
}

 return new VerifiedState(
    this.id,
    this.crypto,
    this.header,
    clonedSessions
 )
```

Since this will probably called alot it's probably worth it to focus a little on reducing the memory footprint of this clone operation.

gdorsi (approved):

packages/cojson/src/sync.ts
```diff
@@ -486,9 +489,8 @@ export class SyncManager {
 
       peer.updateHeader(msg.id, true);
 
-      coValue = new CoValueCore(msg.header, this.local);
-
-      entry.markAvailable(coValue, peer.id);
+      entry.markAvailable(msg.header, peer.id);
```
gdorsi: **question:** would it be possible to call this `markAvailable` after processing the sessions?

While working to the new subscription I've found a bug where we emit an update with an empty coValue before emitting the content because we call `markAvailable` here before loading the content.

With the old version of the code the fix has been easy, now it would require more thinking.

aeplay: I think the ideal way to resolve this problem is by pulling precise peer state into CoValueCore and letting subscribers skip all updates until local state matches peer known state (either storage, or server or max(storage, server)).

I can try to do that in a follow up PR, estimate 0.5 days

packages/cojson/src/coValueCore/verifiedState.ts
aeplay: Thank you!

packages/cojson/src/sync.ts
gdorsi: LGTM

Opened an issue to track: https://github.com/garden-co/jazz/issues/2117

In the meantime will skip the failing tests.

aeplay (commented):
> question: why are we cloning the verified state when marking the coValue as available? 

Abundance of caution - we don't trust the caller to provide a verified state that's not used elsewhere

packages/cojson/src/coValueState.ts
aeplay: Abundance of caution - we don't trust the caller to provide a verified state that's not used elsewhere

PR #2080, opened 2025-05-02
fix(docs): body wider than page container

trishalim:
Before: the TOC is outside the container. It's not aligned with the top nav.
image

After
image

This fixes the docs body width, but cuts off the twoslash tooltips
image

It used to be able to overflow outside of the docs content
image

This doesn't happen too often. I suggest we keep each line in the code samples shorter to minimize this. I've not found a solution that caters to both problems.

### DIFF
--- a/homepage/homepage/lib/docMdxContent.tsx
+++ b/homepage/homepage/lib/docMdxContent.tsx
@@ -42,7 +42,7 @@ export async function getDocMetadata(framework: string, slug?: string[]) {
 
 function DocProse({ children }: { children: React.ReactNode }) {
   return (
-    <Prose className="overflow-x-hidden lg:overflow-x-visible lg:flex-1 pb-8 pt-[calc(61px+2rem)] md:pt-8 md:max-w-3xl mx-auto">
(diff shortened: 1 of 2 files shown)


homepage/homepage/lib/docMdxContent.tsx
```diff
@@ -42,7 +42,7 @@ export async function getDocMetadata(framework: string, slug?: string[]) {
 
 function DocProse({ children }: { children: React.ReactNode }) {
   return (
-    <Prose className="overflow-x-hidden lg:overflow-x-visible lg:flex-1 pb-8 pt-[calc(61px+2rem)] md:pt-8 md:max-w-3xl mx-auto">
+    <Prose className="overflow-hidden pb-8 pt-[calc(61px+2rem)] md:pt-8 md:max-w-3xl mx-auto">
```
jmsv, line 45: where does `calc(61px+2rem)` come from?

jmsv (approved):
lgtm!

homepage/homepage/lib/docMdxContent.tsx
trishalim: 61px is the height of the top nav
2rem is pt-8
So I added both since the top nav is position absolute on mobile

PR #2108, opened 2025-05-03
Simplify LocalNode bootstrap and more

aeplay:
This significantly simplifies LocalNode bootstrap, by making the notion of "the current account" much less stateful, letting us quickly switch the controlling agent of a LocalNode without having to make weird ephemeral copies of the LocalNode and copying state back and forth.

After this, it was easier to continue with [Combine CoValueCore and CoValueState Part 2](https://github.com/garden-co/jazz/pull/2109)


packages/cojson/src/localNode.ts
```diff
@@ -445,17 +456,15 @@ export class LocalNode {
...
-      group.core
-        .testWithDifferentAccount(
-          new ControlledAgent(inviteAgentSecret, this.crypto),
-          this.crypto.newRandomSessionID(inviteAgentID),
-        )
-        .getCurrentContent(),
+    const { oldAgentSecret, oldSessionID } = this.switchControl(
+      inviteAgentSecret,
+      this.crypto.newRandomSessionID(inviteAgentID),
     );
```
gdorsi, line 472: Calling switchControl in a fully loaded node is definitely too unpredictable and disruptive for the higher level of abstraction (mostly jazz-tools)

I think the previous version of spawning a LocalNode is a little easier to manage than a temporary ownership transfer.

packages/cojson/src/localNode.ts
```diff
@@ -92,64 +146,47 @@ export class LocalNode {
...
-    const account = setupNode.createAccount(initialAgentSecret);
+    const createdAccount = node.createAccount(initialAgentSecret);
 
-    const nodeWithAccount = account.core.node.testWithDifferentAccount(
-      account,
-      crypto.newRandomSessionID(account.id),
+    node.switchControl(
+      initialAgentSecret,
+      crypto.newRandomSessionID(createdAccount.id),
     );
```
gdorsi, line 168: Considering that `createAccount` calls `switchControl` twice internally, wouldn't make more sense to simply do something like this:

```ts
const node = new LocalNode(
      initialAgentSecret,
      crypto.newRandomSessionID(crypto.getAgentID(throwawayAgent)),
      crypto,
    );
    
  const account = expectGroup(
      this.createCoValue(
        accountHeaderForInitialAgentSecret(agentSecret, this.crypto),
      ).getCurrentContent(),
    );
    
   account.set(accountAgentID, "admin", "trusting");

    const readKey = this.crypto.newRandomKeySecret();

    const sealed = this.crypto.seal({
      message: readKey.secret,
      from: this.crypto.getAgentSealerSecret(agentSecret),
      to: this.crypto.getAgentSealerID(accountAgentID),
      nOnceMaterial: {
        in: account.id,
        tx: account.core.nextTransactionID(),
      },
    });

    account.set(`${readKey.id}_for_${accountAgentID}`, sealed, "trusting");

    account.set("readKey", readKey.id, "trusting");
 
 const controlledAccount = new ControlledAccount(account as RawAccount, agentSecret);
```

We also probably don't need `createAccount` as `withNewlyCreatedAccount` should be the go-to function to create new accounts (because it handles migrations)

aeplay:
Switched back to a model where we use ephemeral nodes and clone verified state - at least that is a lot cleaner and more explicit now.

We need `createAccount` because we need to be able to programatically create valid Account CoValues without running migrations or setting up a full separate context with peers, for example for `npx jazz-run account create`

gdorsi:
> Switched back to a model where we use ephemeral nodes and clone verified state - at least that is a lot cleaner and more explicit now.
> 
> We need `createAccount` because we need to be able to programatically create valid Account CoValues without running migrations or setting up a full separate context with peers, for example for `npx jazz-run account create`

Won't it be enough to call `withNewlyCreatedAccount` without any params for those use cases?

packages/cojson/src/localNode.ts
```diff
@@ -93,63 +126,49 @@ export class LocalNode {
...
-      crypto.newRandomSessionID(account.id),
+    const node = new LocalNode(
+      initialAgentSecret,
+      crypto.newRandomSessionID(createdAccount.id),
+      crypto,
     );
 
-    const accountOnNodeWithAccount =
-      nodeWithAccount.account as RawControlledAccount<Meta>;
+    node.cloneVerifiedStateFrom(setupNode);
```
gdorsi, line 142: What about simplifying this code this way:

```ts
    const node = new LocalNode(
      initialAgentSecret,
      crypto.newRandomSessionID(createdAccount.id),
      crypto,
    );
    
    node.createAccount(initialAgentSecret);
```

Can this work?

gdorsi, line 170: What about simplifying this code by getting rid of `setupNode`:

```ts
    const node = new LocalNode(
      initialAgentSecret,
      crypto.newRandomSessionID(createdAccount.id),
      crypto,
    );
    
    node.createAccount(initialAgentSecret);
```

Can this work?

packages/cojson/src/localNode.ts
```diff
@@ -93,63 +126,49 @@ export class LocalNode {
...
-      for (const coValueEntry of nodeWithAccount.coValuesStore.getValues()) {
+      for (const coValueEntry of node.coValuesStore.getValues()) {
         if (coValueEntry.isAvailable()) {
-          void nodeWithAccount.syncManager.requestCoValueSync(
-            coValueEntry.core,
-          );
+          void node.syncManager.requestCoValueSync(coValueEntry.core);
         }
       }
     }
```
gdorsi, line 202: Probably this is not needed anymore, will try to cover it with a test and see if deleting this code introduces some regressions.

packages/cojson/src/localNode.ts
aeplay: Tried and it breaks stuff, but I'll combine createAccount and withNewlyCreatedAccount

aeplay:
> > Switched back to a model where we use ephemeral nodes and clone verified state - at least that is a lot cleaner and more explicit now.
> > We need `createAccount` because we need to be able to programatically create valid Account CoValues without running migrations or setting up a full separate context with peers, for example for `npx jazz-run account create`
> 
> Won't it be enough to call `withNewlyCreatedAccount` without any params for those use cases?

I tried to combine these things but got stuck (broke something about syncing?): https://github.com/garden-co/jazz/pull/2122

Merging this for now and maybe we can do that as a follow up - but would like to move back to Zod

PR #2122, opened 2025-05-06
Attempt to simplify account/node creation even further

aeplay:


gdorsi:
Replaced by https://github.com/garden-co/jazz/pull/2139

PR #2123, opened 2025-05-06
Fix the InvalidSignature issues on RNQuickCrypto

### LINKED ISSUE #2124: RNQuickCrypto: InvalidSignature errors when loading signatures from PureJSCrypto and WASMCrypto
Repro branch here: https://github.com/garden-co/jazz/pull/2123

gdorsi:
Repro steps:
- run `cd examples/chat-rn-expo && pnpm install && pnpm ios` to start the example app
- ~~if an account is already stored the loading should fail with an "InvalidSignature" error~~
- load the chat co_zTD6yGYgLJf3gM72d7So4cgySb1
- the loading should fail with an "InvalidSignature" error

Fix: https://github.com/margelo/react-native-quick-crypto/pull/690

After fix timings:
```
PureJSCrypto  (2x) 2621ms, 2606ms
RNQuickCrypto (2x)  534ms,  535ms
```

Also, there are some DX improvements for `ios` builds, for when folks start using XCode 16.3 with Nitro Modules

fixes #2124


boorad:
@gdorsi I rebased, added a commit that makes DX a bit better for certain XCode versions, but didn't make any headway.  

After recent merges, I don't get the InvalidSignature error (maybe because you commented out my crap code in RNQuickCrypto).  But I do get 
```
 ERROR  Error withLoadedAccount {"err": [Error: Account has no profile]}
```
which seems to be from recent work on `cojson/src/localNode.ts` by you and @aeplay.  Any way you can take a look at why we don't have the "default profile" behavior like we used to?

gdorsi:
@boorad Opened a PR to not throw when the profile is missing: https://github.com/garden-co/jazz/pull/2155

The profile is missing probably because it wasn't synced before destroying the node (it can happen if the node is immediately closed).

We can fix that, but for now I think we can just not throw if the profile is missing as it is not really needed anywhere by Jazz.

gdorsi:
Merged the fix here. Could you try again @boorad and see if everything works?

packages/jazz-react-native-core/src/crypto/RNQuickCrypto.ts
boorad, line 1: @gdorsi when calling `verify()` with `signerID`, is that the "public" key?

The reason I ask is that I think the bug might be in my ed25519 implementation using the "private" key.  See [this](https://github.com/paulmillr/noble-curves/blob/main/src/abstract/edwards.ts#L93) from the original `@noble/curves` repo.  In a RNQC test, when I use `priv` it works, but when I use `pub` it doesn't.

aeplay: verify should use the public key, yes

PR #2139, opened 2025-05-07
feat: simplify LocalNode account creation flow

gdorsi:
Changes:
- Simplify LocalNode account creation flow by removing the tempNodes steps and forced sync
- Removed createAccount method and created an equivalent test utility
- Removed resolveAccountAgentAsync as unused


packages/cojson/src/localNode.ts
```diff
@@ -133,6 +133,59 @@ export class LocalNode {
...
+
+    const node = new LocalNode(
+      initialAgentSecret,
+      crypto.newRandomSessionID(accountID as RawAccountID),
+      crypto,
+    );
+
+    for (const peer of peersToLoadFrom) {
+      node.syncManager.addPeer(peer);
+    }
```
gdorsi, line 160: Moved the addPeer before the coValue creation in order to sync the coValues with just a content message.

If we add the peer after the flow would be "LOAD -> KNOWN empty -> CONTENT" because the sync would fall into the peer reconciliation flow.

aeplay (approved):

packages/cojson/src/tests/sync.load.test.ts
```diff
@@ -77,12 +77,12 @@ describe("loading coValues from server", () => {
...
         "client -> server | LOAD Group sessions: header/3",
-        "client -> server | LOAD Map sessions: header/1",
         "server -> client | KNOWN Group sessions: empty",
-        "client -> server | CONTENT Group header: true new: After: 0 New: 3",
+        "client -> server | LOAD Map sessions: header/1",
         "server -> client | KNOWN Map sessions: empty",
-        "client -> server | CONTENT Map header: true new: After: 0 New: 1",
+        "client -> server | CONTENT Group header: true new: After: 0 New: 3",
         "server -> client | KNOWN Group sessions: header/3",
+        "client -> server | CONTENT Map header: true new: After: 0 New: 1",
```
gdorsi, line 85: Order of messages shifted due to changes on the test utils.

gdorsi:
Had one test failing because now `withNewlyCreatedAccount` resolves faster (no more 500ms of wait) and it's more likely that when it's done the account isn't uploaded.

Shall we increase the retries when a value is unavailable?

Issue #1332, opened 2025-02-10
Type validation fails when assigning a value to a deeply resolved property


gdorsi:
Example:

<img src="https://uploads.linear.app/acccaaae-2ae4-487e-a83c-7148ce16c659/523a52d2-bedc-4930-8ac4-855a26e71525/ee61163b-f846-434c-8ecb-287a19b930e1?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2FjY2NhYWFlLTJhZTQtNDg3ZS1hODNjLTcxNDhjZTE2YzY1OS81MjNhNTJkMi1iZWRjLTQ5MzAtOGFjNC04NTVhMjZlNzE1MjUvZWU2MTE2M2ItZjg0Ni00MzRjLThlY2ItMjg3YTE5YjkzMGUxIiwiaWF0IjoxNzM5MTg0MzMyLCJleHAiOjMzMzA5NzQ0MzMyfQ.x-tQWzlsLTVcoAoMI7ivLQoYR0K-ACzxkjUbEwf9D8k " alt="Screenshot 2025-02-10 at 11.42.02.png" width="577" data-linear-height="588" />

PR #2150, opened 2025-05-07
Zod support (using a schema translation layer)

aeplay:
Notable changes that seem unrelated to the Zod translation layer but were necessary to make it work:

- ID<T> is now just `string`
  - this required moving the per-account entries in a CoFeed into a sub-object (like the sessions already are)
- **Breaking:** We completely got rid of the `co<T>` field marker type (this unlocks correct narrowing on unions, and made possible some previously impossible assignments)
- **Breaking:** had to upgrade React native to support zod's export map
- **Breaking:** _refs and _edits are `| undefined` in a lot of places to make certain assignments possible
- getting rid of all kinds of type-level schema registration, opting for explicit passing of AccountSchemas instead
  - **Breaking:** useAccount() now needs to be passed the AccountSchema if it's a custom one
- ensured that `CoList` and `CoFeed` have correct variance (i.e. `CoList<T>` is assignable to `CoList<T | null>`)

TODO:

- [x] check stuff is actually nullable
- [ ] see if we can keep _refs not-null
- [ ] implement CoMap.set()
- [x] make members, by a property again, but just make it Account
- [ ] in case we init a CoValue and pass some shallowly loaded CoValues, the return type of .create is too optimistic

Random notes about stuff I've encountered
- Why are all the different chat apps so different? Why does chat-rn-expo-clerk not use useCoState for chat??
- Some fully loaded vs partially loaded types are more precise now, so #1332 is more frequent/painful now

Important to mention:

- explain that most Zod constraints are not enforced yet


aeplay:
Current state:

- [x] `co.map`, `co.record`, `co.list`, `co.feed`*, `co.filestram` & `co.image` work
- [ ] figuring out next: `co.account`
- [ ] change the instance code of CoFeeds to fix currently broken type inference on per-account values
- [ ] update all the examples
- [ ] update the docs

PR #2155, opened 2025-05-08
fix(auth): removed throw error when the profile is unavailable after a login

gdorsi:
Related to: https://github.com/garden-co/jazz/pull/2123#issuecomment-2860586004

Kinda realized that we don't really need that the profile must be available on login.
I think that would be better to let users authenticate and let them face the issue when loading the profile.

### DIFF
--- a/packages/cojson/src/tests/sync.auth.test.ts
+++ b/packages/cojson/src/tests/sync.auth.test.ts
@@ -1,16 +1,11 @@
 import { assert, beforeEach, describe, expect, test } from "vitest";
 
-import { expectMap } from "../coValue";
-import { expectAccount } from "../coValues/account";
 import { WasmCrypto } from "../crypto/WasmCrypto";
 import { LocalNode } from "../localNode";
-import { toSimplifiedMessages } from "./messagesTestUtils";
 import {
   SyncMessagesLog,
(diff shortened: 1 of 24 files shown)


packages/cojson/src/tests/sync.auth.test.ts
```diff
@@ -169,19 +164,82 @@ describe("LocalNode auth sync", () => {
...
+      [
+        "creation-node -> server | CONTENT Account header: true new: After: 0 New: 4",
+        "auth-node -> server | LOAD Account sessions: empty",
+        "server -> creation-node | KNOWN Account sessions: header/4",
+        "server -> auth-node | CONTENT Account header: true new: After: 0 New: 4",
+        "auth-node -> server | KNOWN Account sessions: header/4",
+        "auth-node -> server | LOAD Profile sessions: empty",
+        "server -> auth-node | KNOWN Profile sessions: empty",
+        "auth-node -> server | LOAD Profile sessions: empty",
+        "server -> auth-node | KNOWN Profile sessions: empty",
```
gdorsi, line 242: As you can see in this case the Profile is never uploaded to the sync server.

Working on a fix

aeplay (approved):

PR #2159, opened 2025-05-08
fix: wait for storage sync before resolving the new account creation

### LINKED ISSUE #2161: ReactNative:  Error withLoadedAccount {"err": [Error: Account has no profile]}
I get 
```
 ERROR  Error withLoadedAccount {"err": [Error: Account has no profile]}
```
which seems to be from recent work on `cojson/src/localNode.ts` by you and @aeplay.  Any way you can take a look at why we don't have the "default profile" behavior like we used to?

_Originally posted by @boorad in https://github.com/garden-co/jazz/issues/2123#issuecomment-2860586004_

gdorsi:
Changes:
- Make waitForSync work on storage peers, adding content ack and switching to optimistic/known states
- Skip closed peers and not subscribed client peers when calling waitForSync
- Wait for storage sync before resolving the new account creation

This also fixes a critical bug on waitForSync by fixing `combineOptimisticWith` which should update only the optimistic state.

fixes #2161

### DIFF
--- a/packages/cojson/src/PeerState.ts
+++ b/packages/cojson/src/PeerState.ts
@@ -23,15 +23,7 @@ export class PeerState {
     });
 
     this._knownStates = knownStates?.clone() ?? new PeerKnownStates();
-
-    // We assume that exchanges with storage peers are always successful
-    // hence we don't need to differentiate between knownStates and optimisticKnownStates
-    if (peer.role === "storage") {
-      this._optimisticKnownStates = "assumeInfallible";
-    } else {
-      this._optimisticKnownStates =
-        knownStates?.clone() ?? new PeerKnownStates();
-    }
+    this._optimisticKnownStates = knownStates?.clone() ?? new PeerKnownStates();
   }
 
   /**
@@ -52,13 +44,9 @@ export class PeerState {
    * The main difference with knownState is that this is updated when the content is sent to the peer without
    * waiting for any acknowledgement from the peer.
    */
-  readonly _optimisticKnownStates: PeerKnownStates | "assumeInfallible";
+  readonly _optimisticKnownStates: PeerKnownStates;
 
   get optimisticKnownStates(): ReadonlyPeerKnownStates {
-    if (this._optimisticKnownStates === "assumeInfallible") {
-      return this.knownStates;
-    }
-
     return this._optimisticKnownStates;
   }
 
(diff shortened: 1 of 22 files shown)


packages/cojson/src/PeerState.ts
```diff
@@ -76,53 +64,33 @@ export class PeerState {
...
+    this._optimisticKnownStates.combineWith(id, value);
   }
 
   combineOptimisticWith(id: RawCoID, value: CoValueKnownState) {
-    if (this._optimisticKnownStates === "assumeInfallible") {
-      this._knownStates.combineWith(id, value);
-    } else {
-      this._optimisticKnownStates.combineWith(id, value);
-    }
+    this._optimisticKnownStates.combineWith(id, value);
```
gdorsi, line 76: This should update only the optimistic state.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -656,15 +652,15 @@ export class CoValueCore {
...
-    );
+    if (a.madeAt !== b.madeAt) {
+      return a.madeAt - b.madeAt;
+    }
+
+    if (a.txID.sessionID === b.txID.sessionID) {
+      return a.txID.txIndex - b.txID.txIndex;
+    }
+
+    return 0;
```
gdorsi, line 663: Fixes an edge case where two updates on two different sessions are happening at the same time.

After the latest optimizations was generating some false negatives on the tests.

Basically this way if the timestamp is the same but the sessions are not the insert order is preserved.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -1006,6 +1002,10 @@ export class CoValueCore {
       const waitingForPeer = new Promise<void>((resolve) => {
         const markNotFound = () => {
           if (this.peers.get(peer.id)?.type === "pending") {
+            logger.warn("Timeout waiting for peer to load coValue", {
+              id: this.id,
+              peerID: peer.id,
+            });
```
gdorsi, line 1008: Restored the warn on timeout.

packages/cojson/src/streamUtils.ts
```diff
@@ -43,32 +37,11 @@ export function connectedPeers(
...
-          );
-          yield msg;
-        }
-      })(),
-      channel,
-    ];
-  } else {
-    return [channel.wrap(), channel];
-  }
+  return [channel.wrap(), channel];
```
gdorsi, line 46: Removed the debug code

packages/cojson/src/sync.ts
```diff
@@ -622,28 +630,24 @@ export class SyncManager {
...
+      if (this.requestedSyncs.has(coValue.id)) {
+        this.syncCoValue(coValue);
+      }
+    });
 
-      this.requestedSyncs.set(coValue.id, promise);
-      return promise;
-    }
+    this.requestedSyncs.add(coValue.id);
   }
```
gdorsi, line 646: Simplified to not use promises (weren't used)

Done to debug, and kept the refactoring because it simplifies the flow.

packages/cojson/src/sync.ts
```diff
@@ -393,10 +399,12 @@ export class SyncManager {
...
-        // Should move the state to loading
-        this.local.loadCoValueCore(msg.id, peer.id).catch((e) => {
-          logger.error("Error loading coValue in handleLoad", { err: e });
-        });
+        // Syncronously updates the state loading is possible
+        coValue
+          .loadFromPeers(this.getServerAndStoragePeers(peer.id))
+          .catch((e) => {
+            logger.error("Error loading coValue in handleLoad", { err: e });
+          });
```
gdorsi, line 407: Removed one layer of abstraction, helped on debugging and kept like this because it makes the flow easier to navigate

aeplay (approved):

PR #2163, opened 2025-05-08
feat: package jazz-inspector as custom element

gdorsi:
Adding a new package `jazz-inspector-element` to make the inspector mountable as custom element.

Screenshot 2025-05-08 at 17 50 27

## Usage

By default jazz-inspector-element relies on `Account.getMe()` to link itself to the Jazz context, so it's enough to mount the html element in a page with a loaded Jazz app:

```ts
import "jazz-inspector-element"

document.body.appendChild(document.createElement("jazz-inspector"))
```

OR

```ts
import "jazz-inspector-element" // somewhere in your app modules
```
and
```html
<jazz-inspector />
```

### DIFF
--- a/packages/jazz-inspector-element/src/main.ts
+++ b/packages/jazz-inspector-element/src/main.ts
@@ -1,0 +1,3 @@
+if (typeof window !== "undefined" && process.env.NODE_ENV === "development") {
+  import("./element.js");
+}
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -15,6 +15,7 @@
       "jazz-browser-media-images",
       "jazz-expo",
       "jazz-inspector",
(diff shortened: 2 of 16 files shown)


packages/jazz-inspector-element/src/main.ts
```diff
@@ -0,0 +1,3 @@
+if (typeof window !== "undefined" && process.env.NODE_ENV === "development") {
+  import("./element.js");
+}
```
gdorsi, line 3: Doing this to register the element and load react only in the browser and in dev mode.

The check on window is required to work on svelte-kit without asking the user to do checks.

bensleveritt (approved):
Looks amazing, great job!

trishalim (approved):

PR #2174, opened 2025-05-09
feat(svelte):  new CoState and AccountCoState APIs

gdorsi:
Introduces CoState and AccountCoState classes as new primitives to subscribe to Jazz state in Svelte using https://runed.dev/docs as inspiration.

It definitely looks more ergonomic, and it's easier to guess where `$derived` is required

~~Didn't manage to make reactive ids work, will move them in a separate PR to get some feedback on where I'm doing things wrong.~~

### DIFF
--- a/examples/file-share-svelte/src/lib/components/FileItem.svelte
+++ b/examples/file-share-svelte/src/lib/components/FileItem.svelte
@@ -61,6 +61,8 @@
     </div>
     <div class="flex-grow">
       {#if isAdmin}
+        <!-- Jazz values are reactive, but they are not recognized as reactive by Svelte -->
+        <!-- svelte-ignore binding_property_non_reactive -->
         <input  class="font-medium text-gray-900 w-full py-1" type="text" bind:value={file.name} />
       {:else}
         <h3 class="font-medium text-gray-900">{file.name}</h3>
--- a/examples/file-share-svelte/src/lib/components/FileItem.svelte
+++ b/examples/file-share-svelte/src/lib/components/FileItem.svelte
@@ -3,9 +3,8 @@
   import { SharedFile } from '$lib/schema';
   import { FileStream } from 'jazz-tools';
   import { File, FileDown, Trash2, Link2 } from 'lucide-svelte';
-  import { useAccount } from 'jazz-svelte';
   import { toast } from 'svelte-sonner';
-  import { formatFileSize } from '$lib/utils';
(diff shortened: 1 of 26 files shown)


examples/file-share-svelte/src/lib/components/FileItem.svelte
```diff
@@ -66,14 +55,16 @@
...
-  <div class="flex items-center space-x-4">
+  <div class="flex items-center space-x-4 flex-grow">
     <div class="flex h-10 w-10 items-center justify-center rounded-lg bg-blue-100 text-blue-600">
       <File class="h-6 w-6" />
     </div>
-    <div>
-      <a href="/file/{file.id}" class="hover:text-blue-600 hover:underline">
+    <div class="flex-grow">
+      {#if isAdmin}
+        <input  class="font-medium text-gray-900 w-full py-1" type="text" bind:value={file.name} />
```
gdorsi: Svelte complains here:

```
[svelte] binding_property_non_reactive`bind:value={file.name}` 
(src/lib/components/FileItem.svelte:64:73) is binding to a non-reactive property
https://svelte.dev/e/binding_property_non_reactive
```

but the reactivity works, because it's powered by Jazz instead of Svelte (it's not fine-grained thou, probably)

examples/file-share-svelte/src/routes/+page.svelte
```diff
@@ -1,32 +1,38 @@
...
+  const me = new AccountCoState({
+    resolve: {
+      profile: true,
+      root: {
+        sharedFiles: {
+          $each: true
+        },
+      }
+    }
+  });
```
gdorsi, line 17: Using deep load in the correct way is more intuitive, because deconstructing looks wrong when used on a class instance.

packages/jazz-svelte/src/lib/tests/components/CoState/UpdateNestedValue.svelte
```diff
@@ -0,0 +1,36 @@
...
+
+  let props: Props = $props();
+
+  const person = $derived(
+    new CoState(Person, props.id, {
+      resolve: {
+        dog: true
+      }
+    })
+  );
```
gdorsi: Had to wrap this with `$derived` to make it react to id changes.

Tried with `() => props.id` but didn't work.

Will make a separate PR for reactive ids

aeplay (approved):

PR #2190, opened 2025-05-12
homepage: improve CTA, move problem section down

trishalim:
image


homepage/homepage/components/home/EarlyAdopterSection.tsx
```diff
@@ -8,7 +8,7 @@ export function EarlyAdopterSection() {
     <Card className="p-4 md:py-16">
       <div className="lg:max-w-3xl md:text-center mx-auto space-y-6">
         <p className="uppercase text-primary tracking-widest text-sm font-medium ">
-          Become an early adopter
```
trishalim, line 11: This makes jazz sound like it's not prod ready. We still do say it's "early" in the paragraph below.

aeplay (approved):

PR #2201, opened 2025-05-13
fix(inspector): colist shows too many items after navigating from a longer colist

### LINKED ISSUE #2193: Inspector shows stale data
When navigating around the Inspector, I can be viewing a long list (10 of 90), and then click to a coValue that has a shorter list (only 2), and see the two proper rows, but also the stale data in rows 3-10, or undefined.

https://github.com/user-attachments/assets/5999cd38-8192-4ab1-b06c-6a0ffdaa01d2

trishalim:
closes #2193

### DIFF
--- a/packages/jazz-inspector/src/viewer/use-resolve-covalue.ts
+++ b/packages/jazz-inspector/src/viewer/use-resolve-covalue.ts
@@ -195,7 +195,7 @@ export function useResolvedCoValues(
       const unsubscribe = subscribeToCoValue(coValueId, node, (newResult) => {
         if (isMounted) {
           setResults((prevResults) => {
-            const newResults = [...prevResults];
+            const newResults = prevResults.slice(0, coValueIds.length);
             newResults[index] = newResult;
             return newResults;
           });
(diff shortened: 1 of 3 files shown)


boorad (approved):
LGTM 👍

packages/jazz-inspector/src/viewer/use-resolve-covalue.ts
```diff
@@ -195,7 +195,7 @@ export function useResolvedCoValues(
       const unsubscribe = subscribeToCoValue(coValueId, node, (newResult) => {
         if (isMounted) {
           setResults((prevResults) => {
-            const newResults = [...prevResults];
+            const newResults = prevResults.slice(0, coValueIds.length);
```
boorad, line 198: Is there any way you can add a test that shows this failing w/o the fix, and passing with it?

trishalim: Good idea. There are no tests at all, but I will prioritize adding them https://github.com/garden-co/jazz/issues/2219

PR #2227, opened 2025-05-14
fix(flake): stick with `pnpm@9.x`

sectore:
to be in sync with `pnpm` version [defined in `package.json`](https://github.com/garden-co/jazz/blob/12e983785824c21dba29c89555aafeb16d91d13a/package.json#L6) (currently `9.15.0`). In other case
`pnpm@10` is used and `pnpm i` or other commands won't work properly

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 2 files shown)


gdorsi (approved):
Thanks!

PR #2247, opened 2025-05-15
feat: add $onError to catch errors on resolve

### LINKED ISSUE #1806: Add $skipInvalid to handle lists/records with inaccessible items more gracefully

gdorsi:
fixes #1806 

We have a big pain point now.

When deep loading values, if anything in the tree has an error (due to access rights or load issues) the whole loading returns an error. (usually results in `null` returned)

We do this to keep Jazz type-safe. When the types say that something is not-nullable we keep that promise.

Now we are trying to come up with an API that helps to deal with this problem, especially when working with shared values which access can be revoked.

Our first idea has been to add a `$skipInvalid` option to resolve, but after adding the docs we've realized that language-wise it might be misleading:

```ts
const user = await User.load(map.id, {
  resolve: {
    friends: { $each: { dog: true }, $skipInvalid: true },
  },
});
```

So we are coming up with a new idea:
```ts
const user = await User.load(map.id, {
  resolve: {
    friends: { $each: { dog: true, $onError: null } },
  },
});
```

These two would do the same thing, which is replace any list item inside of `friends` that have errors with `null`

This means that even if we don't have access to a `dog` the dog owner is going to be `null` for us.

We are going to use `$onError: null` because in the future we want to support also `$onError: "skip"` to skip elements in the lists/records instead of setting them to `null` and `$onError: (err) => value` to let the user handle error values and manually recover from errors.

[Link to the docs](https://jazz-homepage-git-feat-skipinvalid-gcmp.vercel.app/docs/react/using-covalues/subscription-and-loading?_vercel_share=sqSpt7FmSGk3aOlx8HkoMXYteo2hn90H)

### DIFF
--- a/examples/organization/src/Layout.tsx
+++ b/examples/organization/src/Layout.tsx
@@ -3,7 +3,7 @@ import { UserIcon } from "lucide-react";
 
 export function Layout({ children }: { children: React.ReactNode }) {
   const { me, logOut } = useAccount({
-    resolve: { root: { draftOrganization: true } },
+    resolve: { profile: true },
   });
 
   return (
@@ -16,7 +16,20 @@ export function Layout({ children }: { children: React.ReactNode }) {
             <span className="bg-stone-500 pt-1 size-6 flex items-center justify-center rounded-full">
               <UserIcon size={20} className="stroke-white" />
             </span>
-            {me?.profile?.name}
+            <label htmlFor="profile-name" className="sr-only">
+              Profile name
+            </label>
+            <input
+              id="profile-name"
+              type="text"
+              value={me?.profile.name ?? ""}
+              className="rounded-md shadow-sm dark:bg-transparent text-sm py-1.5 px-3"
+              onChange={(e) => {
+                if (me) {
+                  me.profile.name = e.target.value;
+                }
+              }}
+            />
           </span>
 
           <button
--- a/examples/organization/src/schema.ts
+++ b/examples/organization/src/schema.ts
@@ -1,4 +1,5 @@
-import { Account, CoList, CoMap, Group, co } from "jazz-tools";
+import { Account, CoList, CoMap, Group, Profile, co } from "jazz-tools";
+import { getRandomUsername } from "./util";
 
 export class Project extends CoMap {
   name = co.string;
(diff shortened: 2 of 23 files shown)


examples/organization/src/Layout.tsx
```diff
@@ -16,7 +16,20 @@ export function Layout({ children }: { children: React.ReactNode }) {
...
+              id="profile-name"
+              type="text"
+              value={me?.profile.name ?? ""}
+              className="rounded-md shadow-sm dark:bg-transparent text-sm py-1.5 px-3"
+              onChange={(e) => {
+                if (me) {
+                  me.profile.name = e.target.value;
+                }
+              }}
+            />
```
gdorsi, line 32: Changed to setup the profile name to simplify testing

examples/organization/src/schema.ts
```diff
@@ -38,43 +39,49 @@ export class JazzAccountRoot extends CoMap {
 export class JazzAccount extends Account {
   root = co.ref(JazzAccountRoot);
 
-  async migrate() {
-    if (!this._refs.root) {
-      const draftOrganizationOwnership = {
-        owner: Group.create({ owner: this }),
-      };
-      const draftOrganization = DraftOrganization.create(
+  async migrate(this: JazzAccount) {
```
gdorsi, line 42: Updated the migrate flow as per latest best practices

packages/jazz-tools/src/subscribe/SubscriptionScope.ts
```diff
@@ -220,7 +224,7 @@ export class SubscriptionScope<D extends CoValue> {
     const error = this.errorFromChildren;
     const value = this.value;
 
-    if (error) {
+    if (error && !this.skipInvalidEnabled) {
```
gdorsi: It's been dead easy to add the feature in in the runtime 🚀

homepage/homepage/content/docs/using-covalues/subscription-and-loading.mdx
```diff
@@ -465,6 +465,190 @@ project?.owner // => always null
...
+const friends = await Friends.load(source.id, {
+  resolve: {
+    $each: { $onError: null }
+  },
+  loadAs: me,
+});
+
+// Thanks to $onError catching the errors, the list is loaded
+// because we have access to friends
+expect(friends).not.toBeNull();
```
bensleveritt: If you put the output in comments it will make it easier for the reader to connect the assertions to the result (rather than having to infer it from the surrounding comments and code)

bensleveritt (approved):
This is a great improvement to the loading API, nice work~!

homepage/homepage/content/docs/using-covalues/subscription-and-loading.mdx
gdorsi: Done: https://github.com/garden-co/jazz/pull/2247/commits/16d5553ccd621a9f3876e1be12f56c006e19cc6f

PR #2264, opened 2025-05-16
feat(storage): implement content streaming

gdorsi:
This PR makes the file loading from storage non-blocking.

I have some profiling, but a video gives a better idea of the perf improvement:

Before

https://github.com/user-attachments/assets/0ca2bf88-4ce9-459e-822e-dee6c6431185

After

https://github.com/user-attachments/assets/0bc7e879-dd8d-4090-92a9-1cfd474634ff

### DIFF
--- a/packages/cojson-storage/src/managerAsync.ts
+++ b/packages/cojson-storage/src/managerAsync.ts
@@ -9,7 +9,12 @@ import {
   logger,
 } from "cojson";
 import { collectNewTxs, getDependedOnCoValues } from "./syncUtils.js";
-import type { DBClientInterfaceAsync, StoredSessionRow } from "./types.js";
+import type {
+  DBClientInterfaceAsync,
+  SignatureAfterRow,
+  StoredCoValueRow,
+  StoredSessionRow,
+} from "./types.js";
 import NewContentMessage = CojsonInternalTypes.NewContentMessage;
 import KnownStateMessage = CojsonInternalTypes.KnownStateMessage;
 import RawCoID = CojsonInternalTypes.RawCoID;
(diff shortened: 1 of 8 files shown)


packages/cojson-storage/src/managerAsync.ts
```diff
@@ -42,113 +47,98 @@ export class StorageManagerAsync {
...
 
+    let contentStreaming = false;
     for (const sessionRow of allCoValueSessions) {
-      newCoValueKnownState.sessions[sessionRow.sessionID] = sessionRow.lastIdx;
+      const signatures = await this.dbClient.getSignatures(sessionRow.rowID, 0);
+
+      if (signatures.length > 0) {
+        contentStreaming = true;
+        signaturesBySession.set(sessionRow.sessionID, signatures);
+      }
```
gdorsi, line 89: Detecting content streaming by checking if there are any signatures stored.

packages/cojson-storage/src/managerAsync.ts
```diff
@@ -42,113 +47,98 @@ export class StorageManagerAsync {
...
     }
 
-    const newContentMessages: CojsonInternalTypes.NewContentMessage[] = [
-      {
-        action: "content",
+    /**
+     * If we are going to send the content in streaming, we send before a known state message
+     * to let the peer know how many transactions we are going to send.
+     */
+    if (contentStreaming) {
```
gdorsi, line 96: We really need to send the known state only if we are going to do streaming, otherwise it's useless because the content we will send already contains our known state.

packages/cojson-storage/src/managerAsync.ts
```diff
@@ -157,54 +147,56 @@ export class StorageManagerAsync {
...
+          contentMessage,
           sessionRow,
           firstNewTxIdx: idx,
           signature: signature.signature,
         });
 
         idx = signature.idx + 1;
+
+        if (signatures.length > 1) {
+          await this.sendContentMessage(coValueRow, contentMessage);
```
gdorsi, line 159: Now going to send the content messages as they are loaded.

This way we don't block the main thread use the fetching time between two content messages to process and validate it on the LocalNode.

packages/cojson-storage/src/managerAsync.ts
```diff
@@ -157,54 +147,56 @@ export class StorageManagerAsync {
...
+  ) {
+    if (Object.keys(contentMessage.new).length === 0) {
+      return;
+    }
 
     const dependedOnCoValuesList = getDependedOnCoValues({
       coValueRow,
-      newContentMessages,
+      newContentMessages: [contentMessage],
     });
```
gdorsi, line 185: Going to check the dependencies before sending each content message.

This way we ensure that the order is respected while having the possiblity of doing streaming.

packages/cojson-storage/src/managerSync.ts
```diff
@@ -47,79 +52,81 @@ export class StorageManagerSync {
...
-      }
-    }
-  }
-
-  private collectCoValueData(
-    peerKnownState: CojsonInternalTypes.CoValueKnownState,
-    messageMap: OutputMessageMap = {},
-    asDependencyOf?: CojsonInternalTypes.RawCoID,
+  async sendNewContent(
+    coValueKnownState: CojsonInternalTypes.CoValueKnownState,
```
gdorsi, line 56: Same thing as above using sync code.

packages/cojson-storage/src/managerSync.ts
```diff
@@ -153,56 +144,69 @@ export class StorageManagerSync {
...
+            action: "content",
+            id: coValueRow.id,
+            header: coValueRow.header,
+            new: {},
+            priority: cojsonInternals.getPriorityFromHeader(coValueRow.header),
+          } satisfies CojsonInternalTypes.NewContentMessage;
+
+          // Introduce a delay to not block the main thread
+          // for the entire content processing
+          await new Promise((resolve) => setTimeout(resolve));
```
gdorsi, line 167: Added a syntetic delay to unblock the main thread when processing large files.

Since the `handleSyncMessage` in the sync storages isn't awaited this also means that we are going to load the rest of the chunks in background.

aeplay (approved):

PR #2269, opened 2025-05-19
fix: make the profile access on Group members trigger updates correctly

gdorsi:


examples/organization/src/components/OrganizationMembers.tsx
```diff
@@ -22,18 +22,13 @@ export function OrganizationMembers({
...
-  accountId,
+  account,
   role,
   group,
-}: { accountId: ID<Account>; role: string; group: Group }) {
-  const account = useCoState(Account, accountId, {
-    resolve: {
-      profile: true,
-    },
-  });
```
gdorsi, line 33: Changed to test the autoload

PR #2273, opened 2025-05-19
#1514 - Connecting CoValue docs

### LINKED ISSUE #1514: Add docs for when different CoValue types reference the same data
[https://discord.com/channels/1139617727565271160/1139621689882321009/1343962276918460497](https://discord.com/channels/1139617727565271160/1139621689882321009/1343962276918460497)

Question

> Let's take an app that reviews restaurants as an example. With a relational DB I have a restaurants table, reviews table and users table. To show reviews for a specific restaurant I specify that in the where clause. Same when showing the reviews from a specific user. Now with coValues my first instinct was to store the user's reviews in the user's account. This would allow to show them on the user's profile page easily. But when I want to show reviews from all user's on the restaurant page it becomes cumbersome (potentially). Or I can do it the other way around and showing the data on the user's profile page becomes a bit more complicated. Maybe I can store the data in both places like with a noSQL database? (and sorry if this is just rambling, trying to wrap my head around architecture best practices)

Answer

> You can see Jazz data as a graph. If you want a list of all the restaurant reviews, when a user creates a review it can be linked to the restaurant using a list:
>
> ```ts
> class ReviewsFeed extends CoFeed.Of(co.ref(Review)) {}
> class Restaurant extends CoMap {
>   reviews = co.ref(ReviewsFeed)
> }
> ```
>
> I'm suggesting a CoFeed because you probably want users only be able to add their reviews to the Restaurant.
>
> With CoFeed reviews can only be added and not removed.
>
> CoList instead can be pictured as a normal Array, where values can be added and removed.
>
> For the permissions you can work out something like this:
>
> ```ts
> function createReview(restaurant: Restaurant, description: string, stars: number) {
>   const reviewGroup = Group.create()
>   const review = Review.create({ restaurant, description, stars }, reviewGroup);
>   const restaurantGroup = c._owner.castAs(Group); // Assuming that you have created the restaurant assigning a group
>   reviewGroup.extend(restaurantGroup, "reader"); // Every member from the restaurant can only read the review
>   restaurant.reviews.push(review);
>   const currentAccount = await Account.getMe().ensureLoaded({ root: { reviews: [] }})
>   currentAccount.root.reviews.push(review) // If you want to have the list of all reviews from a given account
>   return review;
> }
> ```

mculotta120:
Closes [#1514](https://github.com/garden-co/jazz/issues/1514)

Adds documentation and example for references and shared CoValues


bensleveritt (approved):
Thank you!

PR #2282, opened 2025-05-19
Add missing discriminator

Wizzel1:
I noticed that the discriminator in `z.discriminatedUnion` is missing which caused errors in my project


bensleveritt:
Superficially looks good to me, but I'll let @aeplay check it too. Thanks for your contribution!

Wizzel1:
> Superficially looks good to me, but I'll let @aeplay check it too. Thanks for your contribution!

No problem, sorry for messing up the formatting. Do you guys have a script for that? I tried running the biome script but that didn't properly format the markdown.

Wizzel1:
@bensleveritt seems like this has been resolved with the latest version. I can't use `z.discriminatedUnion` without explicitly passing the discriminator field name.

gdorsi (approved):
Thanks for the fix!

homepage/homepage/content/docs/schemas/covalues.mdx
```diff
@@ -207,16 +212,17 @@ import { co, z } from "jazz-tools";
...
+}); 
 
 const SliderWidget = co.map({
   type: z.literal("slider"),
   min: z.number(),
   max: z.number(),
 });
 
-const WidgetUnion = z.discriminatedUnion([ButtonWidget, SliderWidget]);
+const WidgetUnion = z.discriminatedUnion("type", [ButtonWidget, SliderWidget]);
```
gdorsi, line 223: We actually missed this one, thanks!

PR #2284, opened 2025-05-19
#1604 - Add docs for soft delete

### LINKED ISSUE #1604: Document the Soft Deletion pattern

mculotta120:
Closes [1604](https://github.com/garden-co/jazz/issues/1604)
Adds docs for soft delete pattern


trishalim:
I think this belongs to CoList (because you have the helper method there) and CoMap (because you have the `deleted` property on the CoMap) pages.

Can you please also make it use Zod, and use the example schema ("gardening project") from those pages so that it flows well into these new sections. 🙏

homepage/homepage/content/docs/using-covalues/colists.mdx
```diff
@@ -111,6 +111,40 @@ tasks[0].status = "complete";        // Update properties of references
...
+Then, create methods in CoList classes, which explicitly filter out items where the deleted property is true, ensuring these items are not surfaced in standard queries.
+
+<CodeGroup>
+```ts twoslash
+import { co, z, Loaded, Group } from "jazz-tools";
+
+export const Task = co.map({
+  name: z.string(),
+  deleted: z.boolean(), 
+}).withHelpers((Self) => ({
```
trishalim: This is simple enough that we don't need a helper method

homepage/homepage/content/docs/using-covalues/colists.mdx
```diff
@@ -111,6 +111,40 @@ tasks[0].status = "complete";        // Update properties of references
...
+    }
+  })
+);
+
+const ListOfTasks = co.list(Task)
+  .withHelpers((Self) => ({
+    getAll(list: Loaded<typeof Self>) {
+      return Array.from(list.values()).filter(
+        (task): task is Loaded<typeof Task> =>
+          !task?.deleted && task?._owner.castAs(Group).myRole() !== undefined,
```
trishalim: I am guessing the checking is for when I don't have permissions? Makes sense, but for the sake of keeping this example super simple, can we remove the role check?

homepage/homepage/content/docs/using-covalues/colists.mdx
```diff
@@ -111,6 +111,40 @@ tasks[0].status = "complete";        // Update properties of references
 ```
 </CodeGroup>
 
+### Soft Deletion
+When an object needs to be "deleted," instead of removing it from the database, you can set a deleted flag is set to true.
+
+Then, create methods in CoList classes, which explicitly filter out items where the deleted property is true, ensuring these items are not surfaced in standard queries.
```
trishalim: This is good! I think we can make it even shorter

"You can do a soft deletion by using a `deleted` flag, then creating a helper method that explicitly filters out items where the deleted property is true."

homepage/homepage/content/docs/using-covalues/comaps.mdx
```diff
@@ -266,6 +266,22 @@ project.startDate = "2025-03-15"; // ✗ Type error: expected Date
...
+
+Implementing a soft deletion pattern by using a `deleted` flag allows you to maintain data for potential recovery and auditing. 
+
+Simply add an optional `deleted` boolean to the CoValue you want to delete in the schema.
+
+<CodeGroup>
+```tsx
+export class Project extends co.map({
+    name = z.string;
+    deleted = z.optional.boolean;
```
trishalim: The code can speak for itself here. Instead of having "Simply add an optional...", you can add ` // [!code ++]` to this line to highlight it.

`deleted = z.optional.boolean; // [!code ++]`

trishalim:
LGTM aside from the weird `Array.from(list.values()).filter()` in the helper method. Maybe this can be improved after Anselm does the changes for the helpers, so let's wait for that?

homepage/homepage/components/docs/DocsNav.tsx
bensleveritt, line 1: Is this formatting the automatic one from Biome?

homepage/homepage/content/docs/using-covalues/colists.mdx
```diff
@@ -111,6 +111,33 @@ tasks[0].status = "complete";        // Update properties of references
...
+        (task): task is Loaded<typeof Task> =>
+          !task?.deleted
+      );
+    }
+  })
+);
+```
+</CodeGroup>
+
+There are several benefits to utilizing this soft delete pattern, including the ability to recover deleted data, it helps maintain data integrity for relationships, and potentially support audit trails.
```
bensleveritt: Could be rewritten as bullet-points, just to make it clearer

homepage/homepage/content/docs/using-covalues/comaps.mdx
```diff
@@ -266,6 +266,20 @@ project.startDate = "2025-03-15"; // ✗ Type error: expected Date
...
+
+<CodeGroup>
+```tsx
+export class Project extends co.map({
+    name = z.string;
+    deleted = z.optional.boolean; // [!code ++]
+})
+```
+</CodeGroup>
+When an object needs to be "deleted," instead of removing it from the database, the deleted flag is set to true.
```
bensleveritt: It's probably best we avoid the term 'database'. I think:

"When an object needs to be "deleted", instead of removing it from the system, the `deleted` flag is set to `true`. This gives us a property to omit it in the future."

bensleveritt (changes requested):
Looks good, just a few notes

homepage/homepage/components/docs/DocsNav.tsx
mculotta120: @bensleveritt - Sorry, I think that updated when I ran the linting checker. I can remove these changes

homepage/homepage/content/docs/using-covalues/colists.mdx
```diff
@@ -154,6 +154,45 @@ tasks[0].status = "complete";        // Update properties of references
...
+  name: z.string(),
+  deleted: z.boolean(),
+});
+
+export const ListOfTasks = co.list(Task);
+
+export function getCurrentTasks(
+  list: Loaded<typeof ListOfTasks>
+): Loaded<typeof Task>[] {
+  return Array.from(list.values()).filter(
```
trishalim: Does `list.filter` work instead of `Array.from(list.values()).filter`?

homepage/homepage/content/docs/using-covalues/colists.mdx
```diff
@@ -154,6 +154,45 @@ tasks[0].status = "complete";        // Update properties of references
...
+
+export const Task = co.map({
+  name: z.string(),
+  deleted: z.boolean(),
+});
+
+export const ListOfTasks = co.list(Task);
+
+export function getCurrentTasks(
+  list: Loaded<typeof ListOfTasks>
```
trishalim: `Loaded` works but we are now recommending `co.loaded` as best practice

homepage/homepage/content/docs/using-covalues/colists.mdx
bensleveritt: Yep, fixed

homepage/homepage/content/docs/using-covalues/colists.mdx
bensleveritt: Fixed

trishalim (approved):

PR #2286, opened 2025-05-20
fix: zod & RN experiments

boorad:
### Workaround
To resolve runtime errors ("Cannot call a class as a function", "Cannot read property 'S'/'default' of undefined") when using Zod v4 with React Native Hermes, we implemented the following:

1. Babel Configuration (`babel.config.js`):
   - Configured `babel-preset-expo` to use non-loose mode for ES6 class transformations (`'@babel/plugin-transform-classes': { loose: false }`). Class properties remain in loose mode (default).
1. Custom Babel Plugin (`babel-plugin-transform-zod.js`):
   - This plugin targets Zod's `core.$constructor` function.
   - It removes the `Object.defineProperty(ClassName, "name", ...)` call for the internally generated class (`_`).
   - It also removes the `static [Symbol.hasInstance]` method from this same internal class.
1. Metro Configuration (`metro.config.js`):
   - Ensured Zod is processed by Babel by adding it to `transformIgnorePatterns` (e.g., `"node_modules/(?!(...|zod)/)"`).

### For Zod:
We've encountered runtime issues when using Zod v4 in a React Native (Expo) environment with the Hermes JavaScript engine. Specifically:

1. The dynamic class generation in `core.$constructor`, particularly the `super()` call when a Zod-generated class (like `$ZodType`) is the `Parent`, combined with the custom `static [Symbol.hasInstance]` method, leads to "TypeError: Cannot call a class as a function" on Hermes.
1. The `Object.defineProperty(ClassName, "name", ...)` call within `$constructor` for the dynamically generated classes results in "TypeError: Cannot read property 'S' of undefined" or "TypeError: Cannot read property 'default' of undefined" on Hermes.

We've worked around this by creating a Babel plugin that, at build time, removes both the `static [Symbol.hasInstance]` method and the `Object.defineProperty(ClassName, "name", ...)` call from the classes generated by `$constructor`. We also ensure ES6 class transformations are not in 'loose' mode via Babel.

Could you investigate if Zod's dynamic class generation, particularly the `Symbol.hasInstance` implementation and class naming strategy, could be made more directly compatible with Hermes? Perhaps offering an option for standard `instanceof` behavior or an alternative to dynamic naming could alleviate these issues for React Native Hermes users.

### DIFF
--- a/examples/chat-rn-expo/babel-plugin-transform-zod.js
+++ b/examples/chat-rn-expo/babel-plugin-transform-zod.js
@@ -1,0 +1,81 @@
+// Custom Babel plugin to transform Zod's dynamic class generation
+// This makes it more compatible with Hermes
+
+module.exports = function (babel) {
+  const { types: t } = babel;
+
+  return {
+    name: "transform-zod-multi-fixes",
+    visitor: {
+      FunctionDeclaration(path) {
+        // Part 1: Target Zod's $constructor function
+        if (path.node.id?.name === "$constructor") {
+          path.traverse({
+            CallExpression(callPath) {
+              const { node } = callPath;
+              const callee = node.callee;
+              if (
+                t.isMemberExpression(callee) &&
+                t.isIdentifier(callee.object, { name: "Object" }) &&
+                t.isIdentifier(callee.property, { name: "defineProperty" }) &&
+                node.arguments.length >= 3 &&
+                t.isIdentifier(node.arguments[0]) &&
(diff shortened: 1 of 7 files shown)


boorad:
We can probably close this in favor of https://github.com/colinhacks/zod/pull/4466

I'll leave open for now, but if someone stumbles across this in a while, feel free to close.

boorad:
These workarounds are no longer required due to https://github.com/colinhacks/zod/pull/4482

PR #2298, opened 2025-05-20
feat: add support for SSR rendering

gdorsi:
Adds a barebone support for SSR

Notable changes:
- Removed `useAccountOrGuest` and changed the `useAccount` return type. More [here](https://github.com/garden-co/jazz/pull/2298/files#r2098539633)
- Removes the experimental prefix from the `enableSSR` prop to the `jazz-react` provider

TODO:
- Fix all the typescript errors
- Start writing the upgrade guide


packages/jazz-react-core/src/hooks.ts
```diff
@@ -187,17 +165,14 @@ function useAccount<
     resolve?: ResolveQueryStrict<A, R>;
   },
 ): {
-  me: Loaded<A, true> | Loaded<A, R> | undefined | null;
+  me: Loaded<A, R> | undefined | null;
+  agent: AnonymousJazzAgent | InstanceOfSchema<A>;
   logOut: () => void;
 } {
```
gdorsi: Replaced `useAccountOrGuest` by always returning the agent in `useAccount`

To be compatible with the ssr mode also changed the `me` return type to always be nullable, because regardless of the guestMode in SSR we don't have accounts.

I think this way the guest mode would be way less painful to manage, because the `guest` value can only be used in the `loadAs` options, and having an `agent` value that works in both modes is better.

Also having to deal with all the errors in guest mode makes developing apps that support both modes very very hard.

Maybe we can just add an `isGuest` prop, to help devs to detect which mode is active.

Even better, we could rename the mode into `readOnly` and return `isReadOnly`.

gdorsi:
Implemented as part of https://github.com/garden-co/jazz/pull/2552

PR #2299, opened 2025-05-20
feat: deprecate extend methods

### LINKED ISSUE #2189: Groups as members: make Group.addMember support groups as inputs
The idea is to replace the `Group.extend`/`Group.revokeExtend` APIs with `Group.addMember(group` and `Group.removeMember(group` as it "group as members" would be easier to explain/understand than the Group inheritance concept.

To keep in consideration:
- The ability of extending a group is currently more limited than the ability of adding an account (see https://github.com/garden-co/jazz/issues/1331)
- It might clash with https://github.com/garden-co/jazz/issues/1515

emmyoh:
Closes https://github.com/garden-co/jazz/issues/2189.

Changes the public API in the `jazz-tools` package, deprecating `Group.extend` and `Group.revokeExtend` (replaced by `Group.addMember` and `Group.removeMember` respectively); the internal API in `cojson` is unchanged for now.

Documentation preview: https://jazz-homepage-git-emil-group-inheritance-evolution-gcmp.vercel.app/docs/react/groups/inheritance

---

todo:
- ~~add unextend tests in `packages/jazz-tools/src/tests/groupsAndAccounts.test.ts`~~
  - ~~also add more `addMember` tests~~
    - ~~`member` is not `Group` but `role` is `undefined` (type error)~~
    - ~~`member` is not `Group` but `role` is `inherit` (type error)~~
    - ~~`member` is `Group` but `role` is `writeOnly` (type error and should throw)~~
- ~~test affected example apps by hand~~
  - ~~music player example app~~
    - ~~adding tracks to playlists~~
    - ~~removing tracks from playlists~~
- ~~rewrite group inheritance documentation in terms of adding & removing groups as members~~
- ~~changeset~~


packages/jazz-tools/src/coValues/group.ts
```diff
@@ -147,12 +147,40 @@ export class Group extends CoValueBase implements CoValue {
...
+  addMember(
+    member: Group,
+    role?: "reader" | "writer" | "admin" | "inherit",
+  ): void;
+  addMember(
+    member: Group | Everyone | Account,
+    role?: AccountRole | "inherit",
+  ) {
+    if (member !== "everyone" && member._type === "Group") {
+      this._raw.extend(member._raw, role === "writeOnly" ? "writer" : role);
```
gdorsi: In the case of `writeOnly` we should throw an error, not fallback to `writer`.

homepage/homepage/content/docs/groups/inheritance.mdx
```diff
@@ -1,153 +1,190 @@
...
+Groups can be added to other groups using the `addMember` method.
 
-When a group extends another group, members of the parent group will become automatically part of the child group.
+When a group is added as a member of another group, members of the parent group will become part of the child group.
 
-## Basic Usage
+## Basic usage
 
-Here's how to extend a group:
+Here's how to add a child group to a parent group:
```
gdorsi: Child/parent doesn't sound aligned with the interpretation of groups as members.

How about wording it in something like:

> Here's how to add a group as member of another group

homepage/homepage/content/docs/groups/inheritance.mdx
```diff
@@ -1,153 +1,190 @@
...
-// This way track becomes visible to the members of playlist
-trackGroup.extend(playlistGroup);
+// Tracks are now visible to the members of playlist
+trackGroup.addMember(playlistGroup);
 ```
 </CodeGroup>
 
-When you extend a group:
+When you add groups as members:
 - Members of the parent group get access to the child group
```
gdorsi: How about:

```suggestion
- Members of the added group become members of the container group
```

homepage/homepage/content/docs/groups/inheritance.mdx
```diff
@@ -1,153 +1,190 @@
...
-- Their roles are inherited (with some exceptions, see [below](#role-inheritance-rules))
+- Their roles are inherited (with some exceptions, see [below](#the-rules-of-role-inheritance))
 - Removing a member from the parent group also removes their access to child groups
```
gdorsi:
```suggestion
 - Revoking access from the member group also removes their access to the container groups
```

homepage/homepage/content/docs/groups/inheritance.mdx
```diff
@@ -1,153 +1,190 @@
...
 </CodeGroup>
 
-The "override role" works in both directions:
+Members of the grandparent group will get access to all descendant groups based on their roles.
+
+## Roles
 
+### The rules of role inheritance
+
+If the account is already a member of the child group, it will get the more permissive role:
```
gdorsi:
```suggestion
If the account is already a member of the container group, it will get the more permissive role:
```

packages/jazz-tools/src/coValues/group.ts
```diff
@@ -227,6 +261,12 @@ export class Group extends CoValueBase implements CoValue {
     return this;
   }
 
+  /** @category Identity & Permissions
+   * Revokes membership from members a parent group.
+   * @deprecated Use `removeMember` instead.
+   * @param parent The group that will lose access to this group.
+   * @returns This group.
+   */
   async revokeExtend(parent: Group) {
```
gdorsi, line 282: How about removing these methods instead of deprecating them?

@aeplay

packages/jazz-tools/src/tests/groupsAndAccounts.test.ts
```diff
@@ -250,6 +250,10 @@ describe("Group inheritance", () => {
 
     // @ts-expect-error - Even though readerInvite is a valid role for an account, we don't allow it to not create confusion when using the intellisense
     group.addMember(account, "readerInvite");
+    // @ts-expect-error
+    group.addMember(account, "inherit");
+    // @ts-expect-error
+    group.addMember(account, undefined);
```
gdorsi, line 268: Could you add a brief description of why the type checking should fail?

It would to understand more quickly what these lines are testing.

gdorsi (commented):
Code wise looks good.

Docs also quite good.

Unsure about using the parent/child words in the docs to describe the roles between groups.

homepage/homepage/content/docs/groups/inheritance.mdx
```diff
@@ -1,153 +1,190 @@
...
+const parentGroup = Group.create();
+parentGroup.addMember(bob, "reader");
+// ---cut---
+// Remove member from parent
+await parentGroup.removeMember(bob);
 
-// Works only if I'm a member of companyGroup
-teamGroup.extend(companyGroup); 
+// Bob loses access to both parent and child groups
 ```
```
gdorsi, line 171: This might be a good opportunity to also explain that removing a member from the container group doesn't revoke access if they also have access to a member group.

packages/jazz-tools/src/coValues/group.ts
emmyoh: IMO deprecating these methods allows people at least one version upgrade before switching over (conveniently reduces work if they have a lot they've already had to change). They should see a warning in their IDE.

homepage/homepage/content/docs/groups/inheritance.mdx
```diff
@@ -156,10 +194,12 @@ await childGroup.revokeExtend(parentGroup);
...
-```typescript
+```ts twoslash
+import { Group } from "jazz-tools";
+// ---cut---
 const childGroup = Group.create();
 const parentGroup = Group.create();
-childGroup.extend(parentGroup);
+childGroup.addMember(parentGroup);
 
 console.log(childGroup.getParentGroups()); // [parentGroup]
```
gdorsi: For consistency, how about renaming this to `childGroup.getMemberGroups()`

packages/jazz-tools/src/coValues/group.ts
bensleveritt: I'm in favour of deprecating, but I think we need an issue tracking it

.changeset/empty-pears-carry.md
```diff
@@ -0,0 +1,6 @@
+---
+"jazz-example-music-player": minor
+"jazz-tools": minor
```
gdorsi: Since we are only deprecating, this one can be a patch.

PR #2316, opened 2025-05-21
fix-file-share-svelte

bensleveritt:
Follow up to 0.14, fixes the remaining type issues in the example app

Also turns on type-checking on build, to make things easier to spot next time


gdorsi (approved):
Thanks

PR #2318, opened 2025-05-21
#1206 - Document find unique

mculotta120:
Adds docs for `findUnique` method


gdorsi:
This API is going to be deprecated in favor of `upsertUnique` and `loadUnique`.

Closing the PR

PR #2339, opened 2025-05-22
Create Svelte version of chat example

joeinnes:
Takes the chat example available in other frameworks and ports it to Svelte.


4rknova:
Manual test from 2 clients
image

examples/form/vite.config.js
4rknova, line 1: Looks like this shouldn't be part of this changeset

examples/chat-svelte/src/app.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <link rel="icon" href="%sveltekit.assets%/jazz-icon.png" />
```
4rknova: MIME type is missing

examples/chat-svelte/src/lib/utils.ts
4rknova, line 1: When I tested with 2 users, the second one was assigned the username "Anonymous User"
which suggests that this might not be working as intended. I did not debug further so not sure where the issue is.

(see: https://github.com/garden-co/jazz/pull/2339#issuecomment-2955882507)

4rknova (changes requested):

examples/form/vite.config.js
joeinnes: Removed!

examples/chat-svelte/src/app.html
joeinnes: Added!

examples/chat-svelte/src/lib/utils.ts
joeinnes: We discussed on this. I added a `trim()` to the URL param parsing, and a test to ensure that random user names are assigned.

examples/chat-svelte/src/routes/+layout.svelte
```diff
@@ -7,7 +7,7 @@
   import { getRandomUsername } from '$lib/utils';
   let { children } = $props();
   const { url } = $derived(page);
-  const defaultProfileName = $derived(url.searchParams.get('user') ?? getRandomUsername());
+  const defaultProfileName = $derived(url.searchParams.get('user')?.trim() ?? getRandomUsername());
```
trishalim: I think `const defaultProfileName = getRandomUsername());` is enough

The whole logic about the name from the URL is specifically for the chat demos in the homepage where we want to show users as Alice and Bob. The react chat app has a bunch of logic related to those demos, which is quite problematic imo.

joeinnes: OK!

trishalim:
I had to create a separate branch to fix the lock file issues https://github.com/garden-co/jazz/pull/2528

PR #2351, opened 2025-05-23
fix: fix chat-rn-expo example app

gdorsi:
To fix the issues on the chat-rn-expo example app I've started from scratch outside the monorepo.
This helped me to find an issue we had on jazz-expo with npm and find why out setup was failing (we had an override for React 18 which is not compatible with the latest version of Expo)

Changes:
- Made the chat-rn-expo more minimal, removing tailwind and navigation related deps
- Upgraded the monorepo to react 19
- Fixed the peerDependencies for jazz-expo (npm had troubles with react)


examples/password-manager/src/3_vault.tsx
```diff
@@ -36,17 +50,13 @@ const VaultPage: React.FC = () => {
...
   }, []);
 
-  const items = me.root?.folders?.flatMap(
+  const items = me?.root.folders.flatMap(
     (folder) =>
       folder?.items?.filter(
         (item): item is Exclude<typeof item, null> => !!item,
       ) || [],
   );
-  const folders = useCoState(co.list(Folder), me.root?._refs.folders?.id, {
```
gdorsi, line 45: Found that passing the schema this way to `useCoState` creates infinte loops in React 19.

PR #2373, opened 2025-05-27
Set default package manager based on command runner

lukahartwig:
This PR sets the default package manager to the one used to run `create-jazz-app`.

So if you run `pnpm dlx create-jazz-app` it would select `pnpm` by default which allows you to just press enter.


bensleveritt (approved):
Thanks for your contribution @lukahartwig! 

Tested manually. I'll do a follow up PR with some automatic tests

Issue #2376, opened 2025-05-29
group.makePublic() as an alias


bensleveritt:
`group.addMember("everyone", "reader");` could be more clear about intention.

Proposal: `group.makePublic()`

PR #2377, opened 2025-05-29
Add makePublic to group

bensleveritt:
Completes: #2376

- [x] Add to an example app
- [x] Add to docs
- [x] Add changeset

### DIFF
--- a/.changeset/curvy-suits-watch.md
+++ b/.changeset/curvy-suits-watch.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Adds `makePublic(role)` alias for `group.addMember("everyone", role)`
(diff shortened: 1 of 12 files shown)


trishalim:
does this need a changeset?

gdorsi (commented):
Thanks @bensleveritt!

Code-wise looks good.

To ensure that an API is valuable and gives a good DX I usually try:
- To use it in some examples or in a private project
- Add some minimal docs

Sometimes this process makes me realize some pitfalls or improvments that I can do on the new APIs.

This one is quite simple, but I would like for us to try this process on every new API just to ensure a better quality for our output.

What do you think?

aeplay (approved):

trishalim (approved):

gdorsi (approved):

PR #2386, opened 2025-05-31
[website] Use `/next` import for analytics

tobiaslins:
Use Next.js import for Vercel Analytics to get pageviews per route as well


bensleveritt:
@trishalim Looks correct to me, but perhaps you'd be able to tell better?

trishalim (approved):
Thanks!

PR #2391, opened 2025-06-02
fix: support for recursive props on co.profile and for co.image inside z.discriminatedUnion

### LINKED ISSUE #2385: Bug: discriminatedUnion fails when a comap with an Image prop is part of the possible schemas
Repro:
```ts
 test.only("create a Record with a discriminated union", () => {
      const PersonTypeA = co.map({
        type: z.literal("a"),
        name: z.string(),
      });

      const PersonTypeB = co.map({
        type: z.literal("b"),
        profilePicture: z.optional(co.image()),
      });

      const PersonRecord = co.record(
        z.string(),
        z.discriminatedUnion("type", [PersonTypeA, PersonTypeB]),
      );

      const person = PersonTypeB.create({
        type: "b",
        profilePicture: undefined,
      });

      const record = PersonRecord.create({
        a: person,
      });

      if (record.a?.type === "a") {
        expect(record.a.name).toEqual("John");
      } else if (record.a?.type === "b") {
        expect(record.a.profilePicture).toEqual(undefined);
      }
    });
```

gdorsi:
fixes #2385

### DIFF
--- a/packages/jazz-tools/src/implementation/zodSchema/schemaTypes/AccountSchema.ts
+++ b/packages/jazz-tools/src/implementation/zodSchema/schemaTypes/AccountSchema.ts
@@ -74,12 +74,8 @@ export type DefaultProfileShape = {
 
 export type CoProfileSchema<
   Shape extends z.core.$ZodLooseShape = DefaultProfileShape,
-> = Omit<CoMapSchema<Shape & DefaultProfileShape>, "create"> & {
-  create: (
-    init: Simplify<CoMapInitZod<Shape & DefaultProfileShape>>,
-    options: { owner: Exclude<Group, Account> } | Exclude<Group, Account>,
-  ) => CoMapInstance<Shape & Simplify<DefaultProfileShape>>;
-};
+  Config extends z.core.$ZodObjectConfig = z.core.$ZodObjectConfig,
+> = CoMapSchema<Shape & DefaultProfileShape, Config, Group>;
 
 // less precise verion to avoid circularity issues and allow matching against
 export type AnyAccountSchema<
--- a/packages/jazz-tools/src/implementation/zodSchema/zodCo.ts
+++ b/packages/jazz-tools/src/implementation/zodSchema/zodCo.ts
@@ -1,4 +1,5 @@
 import { CoValueUniqueness } from "cojson";
+import { ZodObject } from "zod/v4";
 import {
   Account,
   AccountCreationProps,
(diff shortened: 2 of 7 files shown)


packages/jazz-tools/src/implementation/zodSchema/schemaTypes/AccountSchema.ts
```diff
@@ -74,12 +74,8 @@ export type DefaultProfileShape = {
...
 export type CoProfileSchema<
   Shape extends z.core.$ZodLooseShape = DefaultProfileShape,
-> = Omit<CoMapSchema<Shape & DefaultProfileShape>, "create"> & {
-  create: (
-    init: Simplify<CoMapInitZod<Shape & DefaultProfileShape>>,
-    options: { owner: Exclude<Group, Account> } | Exclude<Group, Account>,
-  ) => CoMapInstance<Shape & Simplify<DefaultProfileShape>>;
-};
+  Config extends z.core.$ZodObjectConfig = z.core.$ZodObjectConfig,
+> = CoMapSchema<Shape & DefaultProfileShape, Config, Group>;
```
gdorsi, line 78: Omit breaks the recursive props, so worked around it.

packages/jazz-tools/src/implementation/zodSchema/zodCo.ts
```diff
@@ -26,68 +27,49 @@ import {
...
+      return coSchema.subscribe(...args);
+    },
+    findUnique: (...args: any[]) => {
+      return coSchema.findUnique(...args);
+    },
+    catchall: (index: z.core.$ZodType) => {
+      return enrichCoMapSchema(baseCatchall(index));
+    },
+    withHelpers: (helpers: (Self: z.core.$ZodType) => object) => {
+      return Object.assign(schema, helpers(schema));
```
gdorsi, line 56: The object spread breaks both the recursive props and misses some not-enumerable props required for `z.discriminatedUnion` to work.

Refactored `catchall` to not use the spread while keeping the schema immutability.
Moved to `Object.assign` for withHelpers.

packages/jazz-tools/src/implementation/zodSchema/zodCo.ts
```diff
@@ -26,68 +27,49 @@ import {
...
+    },
+  }) as unknown as CoMapSchema<Shape>;
+
+  // Needs to be derived from the enriched schema
+  const coSchema = zodSchemaToCoSchema(enrichedSchema) as any;
```
gdorsi, line 61: This should also optimize a little the create, because we don't create 1 class x create anymore.

packages/jazz-tools/src/implementation/zodSchema/zodCo.ts
```diff
@@ -249,19 +231,13 @@ export const coProfileDefiner = <
...
   });
-  return {
-    ...base,
-    // enforce that the owner is a group
-    create: ((init: any, options: { owner: Group } | Group) => {
-      return base.create(init, options);
-    }) as CoProfileSchema<Shape>["create"],
-  } as CoProfileSchema<Shape>;
+
+  return coMapDefiner(ehnancedShape) as CoProfileSchema<Shape>;
```
gdorsi, line 240: Same thing here, the spread breaks alot of things so moved away from it.

aeplay (approved):
Very good!

PR #2397, opened 2025-06-02
fix: make checks on the discriminator field on z.discriminatedUnion more specific and less strict

### LINKED ISSUE #2372: Creating a coMap value with a z.optional(z.date()) field throws error when passing undefined as date.
I have this schema:

```ts
export const NotificationPayload = co.map({
  ...
  triggerDate: z.optional(z.date()),
});
```

when I try to create a new object without a date from it, I am getting an error: 

```ts
export async function sendNotificationPayload(
  triggerDate?: Date,
) {
    const payload = NotificationPayload.create(
      {
        //...
        triggerDate: triggerDate,
      },
      { owner: Group.create({ owner: sender }) }
    );
}
```

`Cannot read property 'toISOString' of undefined`

I am at jazz-expo 0.14.17

gdorsi:
fixes https://github.com/afanslau/repro-discriminated-union/pull/1/files

fixes #2372

### DIFF
--- a/packages/jazz-tools/src/implementation/zodSchema/unionUtils.ts
+++ b/packages/jazz-tools/src/implementation/zodSchema/unionUtils.ts
@@ -11,19 +11,22 @@ export function schemaUnionDiscriminatorFor(
         "z.union() of collaborative types is not supported, use z.discriminatedUnion() instead",
       );
     }
-    if (![...schema._zod.disc.keys()].every((key) => typeof key === "string")) {
+
+    const discriminator = schema._zod.def.discriminator;
+    const field = schema._zod.disc.get(discriminator);
+
+    if (!field) {
       throw new Error(
-        "z.discriminatedUnion() of collaborative types with non-string discriminator key is not supported",
+        "z.discriminatedUnion() of collaborative types with non-existent discriminator key is not supported",
       );
     }
-    if (
-      ![...schema._zod.disc.values()].every((v) =>
-        [...v.values].every((value) => typeof value === "string"),
-      )
-    ) {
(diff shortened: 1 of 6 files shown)


packages/jazz-tools/src/implementation/zodSchema/unionUtils.ts
```diff
@@ -11,19 +11,22 @@ export function schemaUnionDiscriminatorFor(
...
-      throw new Error(
-        "z.discriminatedUnion() of collaborative types with non-string discriminator value is not supported",
-      );
+
+    for (const value of field.values) {
+      if (typeof value !== "string" && typeof value !== "number") {
+        throw new Error(
+          "z.discriminatedUnion() of collaborative types with non-string or non-number discriminator value is not supported",
+        );
+      }
```
gdorsi, line 29: Strict checks only on the main discriminator

packages/jazz-tools/src/implementation/zodSchema/unionUtils.ts
```diff
@@ -46,49 +49,52 @@ export function schemaUnionDiscriminatorFor(
...
-          return (
-            (discriminatorDef._zod.def as z.core.$ZodLiteralDef).values[0] ===
-            discriminatorValue
-          );
-        });
-        if (remainingOptions.length === 1) {
-          return zodSchemaToCoSchema(remainingOptions[0]!);
+        }
+
+        if (match) {
```
gdorsi, line 92: The first field that matches all the discriminators passes.

Any non-defined implicit discriminator is skipped in the check

Issue #1135, opened 2025-01-06
RFC: Per CoValue migration hooks


gdorsi:
Having the migration hook only on Account makes some migrations very complex to write.<br><br>To make things easier we should support the migration hook in every CoValue:

```ts
export class Playlist extends CoMap {
  version = co.json<2>();
  title = co.string;
  tracks = co.ref(ListOfTracks);

  async migration() {
    const version = this.version ?? 2;

    if (version < 2) {
      // Perform the migration
    } 
  }
}
```

We must guarantee that this hook is always executed before the CoValue is loaded in the app.<br><br>On non-account values the hook is not executed on creation.

aeplay:
The trickiest question here is: what if you encounter a CoValue that a migration should be run on that you don't have write access to?

Note how this situation can already happen with the account-only migrations, it's just more hidden

bensleveritt:
[Queried on discord](https://discord.com/channels/1139617727565271160/1139621689882321009/1379125008084762745)

gdorsi:
> The trickiest question here is: what if you encounter a CoValue that a migration should be run on that you don't have write access to?
> 
> Note how this situation can already happen with the account-only migrations, it's just more hidden

A possible fix for this problem would be to make it possible to do "local-only" migrations.

The users with read-only rights should be able to modify the coValue locally, and drop the changes as soon as they get an external update.

I think this can be considered a problem on his own and live in a separate issue.

PR #2405, opened 2025-06-02
feat: add support for coMap migrations

### LINKED ISSUE #2374: Bug: castAs doesn't support Zod schemas
When doing so the API throws

gdorsi:
Part of #1135 
Fixes #2374 

Design decisions:
- ~Running the migrations also on `.create` / this way they are applied consistently and can also be used as creation "triggers"~
- Limited the migrations to be sync. Async migrations might cause some performance issues if not managed correctly.
- Add support only for CoMap because on the others types it feels less useful and hard to track when the migration should run.

Docs: https://jazz-homepage-git-feature-co-map-migrations-gcmp.vercel.app/docs/react/using-covalues/comaps?_vercel_share=xNNVoJ49NPRFYf3zki8XnG44ySHhIhVD

### DIFF
--- a/packages/jazz-tools/src/implementation/zodSchema/runtimeConverters/zodSchemaToCoSchema.ts
+++ b/packages/jazz-tools/src/implementation/zodSchema/runtimeConverters/zodSchemaToCoSchema.ts
@@ -61,19 +61,6 @@ export function tryZodSchemaToCoSchema<S extends z.core.$ZodType>(
         }
       };
 
-      if ("migration" in schema) {
-        const migration = schema.migration;
-        if (typeof migration !== "function") {
-          throw new Error("migration must be a function");
-        }
-        (coSchema.prototype as Account).migrate = async function (
-          this,
-          creationProps,
-        ) {
-          await migration(this, creationProps);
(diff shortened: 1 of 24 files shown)


packages/jazz-tools/src/implementation/zodSchema/runtimeConverters/zodSchemaToCoSchema.ts
```diff
@@ -61,19 +61,6 @@ export function tryZodSchemaToCoSchema<S extends z.core.$ZodType>(
...
-        if (typeof migration !== "function") {
-          throw new Error("migration must be a function");
-        }
-        (coSchema.prototype as Account).migrate = async function (
-          this,
-          creationProps,
-        ) {
-          await migration(this, creationProps);
-        };
-      }
```
gdorsi, line 75: Changed the approach to patch the schema prototype directly in the definer functions.

aeplay (approved):

emmyoh (approved):

bensleveritt (approved):
Yep, all good

PR #2408, opened 2025-06-02
Remove api references from build

### LINKED ISSUE #2407: Remove API Reference from build
We're not using it, and we should remove it from the process.

bensleveritt:
Closes #2407


trishalim (approved):
Great!

PR #2409, opened 2025-06-02
fix: expo logo color on dark mode

trishalim:


bensleveritt (approved):
Yup, looks right now!

PR #2411, opened 2025-06-03
fix: fix id reactivity in CoState for svelte > 5.33

gdorsi:
Changes:
- fix id reactivity in CoState for svelte > 5.33 by using a $version state rune
- changed the vitest configuration to follow the svelte suggestion (the vitest plugin doesn't work with the latest svelte version)
- changed the test setup to work around the lack of support in svelte-kit for vitest workspaces


packages/jazz-svelte/src/lib/jazz.class.svelte.ts
```diff
@@ -172,12 +156,21 @@ import type {
...
+    #versionValue = 0
+    #triggerUpdate() {
+      this.#versionValue++;
+      this.#version = this.#versionValue;
+    }
   
     get current() {
-      this.#subscribe();
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      this.#version;
```
bensleveritt: Do we need this to trigger the reactivity?

bensleveritt:
Where do they mention the vitest change? Are they dropping support for Svelte, or is it just a temporary thing (do they even talk about it?

gdorsi:
> Where do they mention the vitest change? Are they dropping support for Svelte, or is it just a temporary thing (do they even talk about it?

We were using an hacky setup to get it working with the monorepo.
It breaks everytime we upgrade svelte, so I've decided to go with the "mainstream" and just opt-out from vitest workspaces for `jazz-svelte`

There is already an issue somewhere in the svelte repo tracking the bug.

packages/jazz-svelte/src/lib/jazz.class.svelte.ts
gdorsi: Yes

gdorsi: Added a comment

gdorsi:
Updated to the latest svelte version and the changes are breaking the nested updates tests...

Will have to isolate the bugs and file them to the svelte team.

Going to revert the changes on jazz.class.svelte.ts and keep the changes on vite (so upgrading svelte will be easier)

bensleveritt (approved):
Looks good, thanks!

PR #2426, opened 2025-06-03
feat: `_createdAt` and `_lastUpdatedAt`

### LINKED ISSUE #2410: Feature: add _createdAt and _lastUpdatedAt getters on CoMap
Since we already process the transactions it should be doable to add those getters to CoMap at zero cost.

emmyoh:
Closes #2410.

### DIFF
--- a/packages/cojson/src/coValues/coMap.ts
+++ b/packages/cojson/src/coValues/coMap.ts
@@ -50,8 +50,6 @@ export class RawCoMapView<
   /** @internal */
   latestTxMadeAt: number;
   /** @internal */
-  earliestTxMadeAt: number;
-  /** @internal */
   ops: {
     [Key in keyof Shape & string]?: MapOp<Key, Shape[Key]>[];
   };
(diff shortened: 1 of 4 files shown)


packages/cojson/src/coValues/coMap.ts
```diff
@@ -75,6 +77,7 @@ export class RawCoMapView<
     this.id = core.id as CoID<this>;
     this.core = core;
     this.latestTxMadeAt = 0;
+    this.earliestTxMadeAt = Number.MAX_SAFE_INTEGER;
```
gdorsi: We can read the creation date from the header (core.verified.header.createdAt) 

This way we can get the value even if the CoMap is empty, and fallback to the max integer only of the comap is empty and the createdAt in the header is not set (should never happen, because the CoValues that have no createdAt in the header have always some content)

packages/jazz-tools/src/tests/coMap.test.ts
```diff
@@ -1569,4 +1569,28 @@ describe("CoMap migration", () => {
     expect(loaded?.friend?.name).toEqual("Charlie");
     expect(loaded?.friend?.version).toEqual(2);
   });
+  describe("Time", () => {
+    test("created time and last updated time", async () => {
```
gdorsi, line 1582: We should also cover the case of empty coMaps

packages/cojson/src/coValues/coMap.ts
emmyoh: Thanks so much! So, you're saying we can assume `createdAt` will always be in the header, but if it isn't we can return `Number.MAX_SAFE_INTEGER`?

PR #2439, opened 2025-06-04
fix: clarify `Group.addMember` error message

### LINKED ISSUE #2375: Bug: improve error messages for addMember failures
Now it simply says "Failed to set role""

It should explain the reason why it failed.

emmyoh:
Closes #2375.

### DIFF
--- a/packages/cojson/src/coValues/group.ts
+++ b/packages/cojson/src/coValues/group.ts
@@ -292,8 +292,11 @@ export class RawGroup<
       this.set(account, role, "trusting");
 
       if (this.get(account) !== role) {
+        const myRole = this.myRole();
         throw new Error(
-          `Failed to set role due to insufficient permissions (role of current account is ${this.myRole()})`,
+          myRole === "admin"
+            ? "Administrators cannot demote other administrators in a group"
+            : `Failed to set role due to insufficient permissions (role of current account is ${myRole})`,
         );
       }
 
@@ -351,8 +354,11 @@ export class RawGroup<
       this.set(memberKey, role, "trusting");
 
       if (this.get(memberKey) !== role) {
+        const myRole = this.myRole();
         throw new Error(
-          `Failed to set role due to insufficient permissions (role of current account is ${this.myRole()})`,
+          myRole === "admin"
+            ? "Administrators cannot demote other administrators in a group"
+            : `Failed to set role due to insufficient permissions (role of current account is ${myRole})`,
         );
       }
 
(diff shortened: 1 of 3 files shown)


packages/cojson/src/coValues/group.ts
```diff
@@ -292,7 +292,7 @@ export class RawGroup<
       this.set(account, role, "trusting");
 
       if (this.get(account) !== role) {
-        throw new Error("Failed to set role");
+        throw new Error("Failed to set role due to insufficient permissions");
```
gdorsi: **suggestion:** how about specifying the current role of the account?

It would help in getting a better understanding of the error

emmyoh: Great idea. On it

emmyoh: > **suggestion:** how about specifying the current role of the account?
> 
> It would help in getting a better understanding of the error

Like [this](https://github.com/garden-co/jazz/pull/2439/commits/780961d46fad570ee57a2465eed424cfc9eb6d10), yeah?

packages/cojson/src/tests/permissions.test.ts
```diff
@@ -81,7 +81,7 @@ test("Admins can't demote other admins in a group (high level)", () => {
   );
 
   expect(() => groupAsOtherAdmin.addMemberInternal(admin.id, "writer")).toThrow(
-    "Failed to set role",
+    "Failed to set role due to insufficient permissions (role of current account is admin)",
```
gdorsi: This error message is misleading.

I think we should have a specific message for this case, especially because without reading the docs is hard to understand why an admin has insufficient permissions.

emmyoh: Good idea. Will do this this morning

packages/cojson/src/coValues/group.ts
```diff
@@ -292,7 +292,12 @@ export class RawGroup<
       this.set(account, role, "trusting");
 
       if (this.get(account) !== role) {
-        throw new Error("Failed to set role");
+        const myRole = this.myRole();
+        throw new Error(
+          myRole === "admin"
+            ? "Administrators cannot demote other administrators in a group"
```
gdorsi: This is valid only when both roles are admin.

If I'm admin, and the other member is not an admin (which I don't recall a case for this to happen) we should at least trigger a more generic error.

There might be a bug in the permissions and we don't want to give the wrong error message.

emmyoh: Should now be fixed. Thanks

gdorsi (approved):
LGTM, thanks!

PR #2448, opened 2025-06-05
fix(react-native): close db connections and keep the same instance of storage over JazzProvider renders

gdorsi:
This PR brings two fixes:
- Initialize storage inside a `useMemo` / the changes on storage trigger a Jazz ctx reset generating stiability issues when the JazzProvider re-renders (especially on StrictMode/dev where the render happens twice)
- Close the dbs when the ctx is closed. Keeping the connection open was probably what was causing random errors on `expo-sqlite` on dev.

### DIFF
--- a/packages/jazz-expo/src/provider.tsx
+++ b/packages/jazz-expo/src/provider.tsx
@@ -5,7 +5,7 @@ import {
   AnyAccountSchema,
   CoValueFromRaw,
 } from "jazz-tools";
-import React from "react";
+import React, { useMemo } from "react";
 import { ExpoSecureStoreAdapter } from "./storage/expo-secure-store-adapter.js";
 import { ExpoSQLiteAdapter } from "./storage/expo-sqlite-adapter.js";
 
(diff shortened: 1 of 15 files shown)


packages/jazz-expo/src/provider.tsx
```diff
@@ -14,14 +14,13 @@ export function JazzProvider<
     | (AccountClass<Account> & CoValueFromRaw<Account>)
     | AnyAccountSchema,
 >(props: JazzProviderProps<S>) {
-  // Destructure kvStore and pass everything else via rest
-  const { kvStore, storage, ...rest } = props;
+  const storage = useMemo(() => {
+    return props.storage ?? new ExpoSQLiteAdapter();
+  }, [props.storage]);
```
gdorsi, line 19: Thinking about changing the storage prop to be `() => SQLiteDatabaseDriverAsync`

This way we wouldn't reuse the same storage instance on different ctx creations.

packages/jazz-expo/src/provider.tsx
```diff
@@ -14,14 +14,13 @@ export function JazzProvider<
...
-      {...rest}
-      storage={storage ?? new ExpoSQLiteAdapter()}
-      kvStore={kvStore ?? new ExpoSecureStoreAdapter()}
-    />
-  );
+  const kvStore = useMemo(() => {
+    return props.kvStore ?? new ExpoSecureStoreAdapter();
+  }, [props.kvStore]);
+
+  return <JazzProviderCore {...props} storage={storage} kvStore={kvStore} />;
```
boorad, line 25: oh shit... I see the problem here.  Good catch.

boorad (approved):
👍

PR #2452, opened 2025-06-05
feat(jazz-expo): exported API to clear local data

### LINKED ISSUE #2454: Directly export `clearLocalData` from `jazz-expo`
Currently, you can do something like the following:

```ts
import { ExpoSQLiteAdapter, JazzProvider } from "jazz-expo";
import React, { StrictMode } from "react";

export const storage = new ExpoSQLiteAdapter();

export default function App() {
    return (
    <StrictMode>
      <JazzProvider
        storage={storage}
        sync={{
          peer: `wss://cloud.jazz.tools/?key=example@garden.co`,
        }}
      >
      </JazzProvider>
    </StrictMode>
  );
}
```

Using the exported `storage` variable, elsewhere in the app you can call `storage.clearLocalData()` to clear the SQLite database used by Jazz.

However, we also want the ability to do something like the following:

```ts
import { clearLocalData } from "jazz-expo";
import React from "react";

export default function MyDeleteEverythingButton() {
  return (
    <Button onPress={clearLocalData}>
      Reset local data
    </Button>
  );
}
```
(code example by @gdorsi)

This would use the SQLite database of the `JazzProvider` in the app; only one `JazzProvider` is supported per app, so we can assume there are 0 databases or just 1.

emmyoh:
# Draft (see https://github.com/garden-co/jazz/pull/2452#issuecomment-2947997836)

Closes #2454.

- Adds `clearLocalData` function to exports from `jazz-expo`; it clears the database used by an app's `JazzProvider`
- Adds a 'Reset' button to the `chat-rn-expo` example app

Steps to test locally:
1. In `examples/chat-rn-expo`, run `pnpm start` and start a simulator using Expo Go
2. In the simulator, create a new chat, and write a message
3. Save the chat ID
4. Turn off your machine's network connection
5. Logout, then access the chat by its ID again
  - The chat should load from your local storage successfully
6. Click the 'Reset' button
7. Attempt to access the chat by its ID again
  - The chat should fail to load since the local storage has been cleared


emmyoh:
Marking as a draft PR, as testing shows this does not consistently behave correctly. On occasion (with no clear way to consistently reproduce), deleting the DB will fail as it's open, but attempting to close will fail as it's already closed (which seems to me to be logically inconsistent). The database that cannot be deleted or closed is indeed the same database (at least on storage), as the file paths in the two errors are the same. Attempting to individually perform the operations in the React Native dev-tools shows the same errors. This makes me think multiple database handles exist in memory, with no clear way to close & delete using a specific one.

---

I will see if https://github.com/garden-co/jazz/pull/2448 getting merged fixes this.

---

Update: This is fixed.

PR #2461, opened 2025-06-06
updates richtext-prosemirror example app layout

4rknova:
Before
image
After
image

### DIFF
--- a/examples/richtext-prosemirror/src/Editor.tsx
+++ b/examples/richtext-prosemirror/src/Editor.tsx
@@ -49,7 +49,7 @@ export function Editor() {
 
   return (
     <div className="flex flex-col">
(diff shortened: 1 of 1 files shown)


examples/richtext-prosemirror/src/Editor.tsx
```diff
@@ -49,28 +49,26 @@ export function Editor() {
 
   return (
     <div className="flex flex-col">
-      <div className="flex-1 flex flex-col md:flex-row gap-4 p-8">
+      <div className="flex-1 flex flex-col md:flex-col gap-4 p-8">
```
trishalim: you can just remove `md:flex-col` because there is a `flex-col` right before

trishalim (approved):

PR #2465, opened 2025-06-06
Worker group tools in jazz-run

aeplay:
Adds two commands:

`npx jazz-run group create -o OWNER_ID -s OWNER_SECRET`

And

`npx jazz-run group create-with-admin --name "MyApp"`

- creates an admin account to manage members
- creates group with admin as owner
- adds a first member account to use as actual worker account ("writer" rights)

### DIFF
--- a/packages/jazz-run/src/greateWorkerGroup.ts
+++ b/packages/jazz-run/src/greateWorkerGroup.ts
@@ -1,0 +1,42 @@
+import { AgentSecret } from "cojson";
+import { createWebSocketPeer } from "cojson-transport-ws";
+import { WasmCrypto } from "cojson/crypto/WasmCrypto";
+import {
+  Account,
+  Group,
+  createJazzContextFromExistingCredentials,
+  randomSessionProvider,
+} from "jazz-tools";
+import { WebSocket } from "ws";
+
+export const createWorkerGroup = async ({
+  owner,
+  ownerSecret,
+  peer: peerAddr,
+}: { owner: string; ownerSecret: string; peer: string }) => {
+  const crypto = await WasmCrypto.create();
+
+  const peer = createWebSocketPeer({
+    id: "upstream",
+    websocket: new WebSocket(peerAddr),
+    role: "server",
+  });
+
+  const context = await createJazzContextFromExistingCredentials({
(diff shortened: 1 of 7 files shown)


packages/jazz-run/src/index.ts
```diff
@@ -39,6 +41,78 @@ const accountCommand = Command.make("account").pipe(
...
+  Options.withAlias("s"),
+  Options.withDescription("The secret of the owner"),
+  Options.withDefault(process.env.JAZZ_WORKER_SECRET),
+);
+
+const createGroupCommand = Command.make(
+  "create",
+  {
+    owner: ownerOption,
+    ownerSecret: ownerSecretOption,
```
gdorsi, line 60: Shall we use an env variable instead?

Passing a secret via CLI doesn't seem safe considering that it will be stored in the shell history.

packages/jazz-run/src/index.ts
```diff
@@ -39,6 +41,78 @@ const accountCommand = Command.make("account").pipe(
...
+    peer: peerOption,
+  },
+  ({ owner, ownerSecret, peer }) => {
+    return Effect.gen(function* () {
+      if (!owner) {
+        throw new Error("Owner is required");
+      }
+      if (!ownerSecret) {
+        throw new Error("Owner secret is required");
+      }
```
gdorsi, line 70: How about we make this optional and make it possible to pass the amount of workers?

packages/jazz-run/src/index.ts
aeplay: Note that it has env vars as a fallback, but yeah we should completely remove it as a CLI argument

packages/jazz-run/src/index.ts
Elfo404: > How about we make this optional

as in let the newly created account create the group, and then demote itself once it adds `owner` as an admin?

packages/jazz-run/src/greateWorkerGroup.ts
frolic, line 1: there's a typo in the filename btw!

frolic:
Why closed ooc?

PR #2467, opened 2025-06-07
Add w-full to the pagefind container div

### LINKED ISSUE #2466: Bug: x-overflow issue on docs pages since introducing search (especially on small screens)
Overflow seems to be caused here:

https://github.com/garden-co/jazz/blob/b2e2b91dd74f2a0d36d7148e64482343fef81540/homepage/homepage/components/docs/DocsLayout.tsx#L69

Div has no width so is overflowing. Setting `className="w-full"` should fix this.

joeinnes:
Fixes #2466


trishalim (approved):
Thanks for catching that!

PR #2473, opened 2025-06-09
useForm hook

trishalim:
Preview: https://form-demo-git-feat-form-with-save-button-gcmp.vercel.app/

Pain points
- I have to specify the resolve query in the type but also in useCoState
- No obvious way to clone different types of CoValues
  - How about `BubbleTeaOrder.clone()` which does a deep clone? 
- No obvious way to copy over a CoValue's value to another CoValue (rich text and list)
  - How about `order.applyDiff(order2)`? 
- In `covalue.applyDiff(covalue2)`, it's not obvious where the value is being applied to

Todo
- Try with ImageDefinition, FileStream, CoFeed

Ideas
- We can extend this pattern to have edit drafts

### DIFF
--- a/examples/form/index.html
+++ b/examples/form/index.html
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Jazz | Form example</title>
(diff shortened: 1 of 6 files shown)


trishalim:
This was good for research but not a good enough pattern without making an API change. We will continue to find a better solution.

PR #2492, opened 2025-06-10
Add error boundary for inspector

shortdiv:
[nit] I was working with getting familiarized with the jazz inspector today and noticed that the accounts drop down would be stuck in a loading state if an account was invalid and errors were logged in the console. It's easily identifiable when you open the console, but I figured it's an easy fix that could be helpful for more visibility for a customer when things don't work as expected.

Not sure if I missed an error component in the design system but open to edits and feedback in that regard!

Before:
Screenshot 2025-06-10 at 5 53 36 PM
After:
Screenshot 2025-06-10 at 5 54 20 PM

### DIFF
--- a/examples/inspector/src/viewer/new-app.tsx
+++ b/examples/inspector/src/viewer/new-app.tsx
@@ -94,24 +94,24 @@ export default function CoJsonViewerApp() {
           },
         });
       } catch (err: any) {
-        const invalidAccountIndex = accounts.findIndex(
-          (acc) => acc.id === currentAccount.id,
-        );
-        accounts.splice(invalidAccountIndex, 1);
+        if (err.toString().includes("invalid id")) {
+          setAccounts(accounts.filter((acc) => acc.id !== currentAccount.id));
+          //remove from localStorage
+          localStorage.removeItem("lastSelectedAccountId");
+          localStorage.setItem(
+            "inspectorAccounts",
+            JSON.parse(localStorage.inspectorAccounts).filter(
+              (acc: Account) => acc.id != currentAccount.id,
+            ),
+          );
+          setCurrentAccount(null);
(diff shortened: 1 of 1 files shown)


trishalim (commented):
Nice! In the UI, is it possible for you to get the message to show only the message part "Error: trying to load covalue with invalid id"?

aeplay (approved):

examples/inspector/src/viewer/new-app.tsx
```diff
@@ -80,23 +81,47 @@ export default function CoJsonViewerApp() {
...
+            console.log("Not running any migration in inspector");
+          },
+        });
+      } catch (err: any) {
+        const invalidAccountIndex = accounts.findIndex(
+          (acc) => acc.id === currentAccount.id,
+        );
+        accounts.splice(invalidAccountIndex, 1);
+
+        setAccounts(accounts);
```
gdorsi: The inspector doesn't have local persistence so an account might fail to load because of connectivity issues.

We can't be sure that's a broken account, and since this effects always run on the inital load we might end up removing from the list some valid accounts.

We should find a way to delete only the "never loaded accounts"

gdorsi: **nit:** `splice` mutates the array so setAccounts does have no effect here. A `useState` friendly way of handling this kind of updates would be:

```ts
setAccounts(accounts.filter(acc => acc.id !== currentAccount.id))
```

`useState` detects changes using referential identity, `filter`creates a new array, so setAccounts can detect the change by looking at the different reference and enqueue a render.

gdorsi (commented):
Thanks for the contribution!

gdorsi (approved):
Awesome, thanks!

PR #2493, opened 2025-06-11
feat: add `upsertUnique` and `loadUnique` APIs to CoMap

### LINKED ISSUE #2256: Bug: when using findUnique the load is slow because of the unavailable values retry
When using `findUnique` the retry mechanism for unavailable values is slowing down the initial load. 

Currently, is needed to implement the following pattern:

```typescript
let eventId = Event.findUnique(identifier, workspace.id);
let activeEvent = await Event.load(eventId as unknown as ID<Event>);
if (!activeEvent) {
  activeEvent = Event.create(
    {
      title: sourceData.title,
      identifier: sourceData.identifier,
      external_id: sourceData._id,
    },
    workspace
  );
} else {
  activeEvent.applyDiff({
    title: sourceData.title,
    identifier: sourceData.identifier,
    external_id: sourceData._id,
  });
}
```

## Proposed Solutions

Two potential improvements could address this issue:

1. **Add a `skipRetry` option to `load`**:
   This would allow users to bypass the retry mechanism when they expect data might not exist yet.

2. **Implement an `upsertUnique` method**:
   This would handle the create/applyDiff logic internally and always return a CoValue, simplifying the common pattern shown above.

This would simplify the user's workflow to:

```typescript
const activeEvent = await Event.upsertUnique(
  sourceData.identifier,
  workspace.id,
  {
    title: sourceData.title,
    identifier: sourceData.identifier,
    external_id: sourceData._id,
  },
  workspace
);
```

emmyoh:
Closes #2256.

New docs: https://jazz-homepage-git-emil-findunique-slow-gcmp.vercel.app/docs/react/using-covalues/comaps#uniqueness


packages/jazz-tools/src/coValues/coMap.ts
```diff
@@ -564,6 +563,80 @@ export class CoMap extends CoValueBase implements CoValue {
...
+    creationOptions?:
+      | {
+          owner: Account | Group;
+        }
+      | Account
+      | Group,
+    loadOptions?: {
+      resolve?: RefsToResolveStrict<M, R>;
+      loadAs?: Account | AnonymousJazzAgent;
+      skipRetry?: boolean;
```
gdorsi: These options look a bit redundant and can be simplified this way:

```ts
    this: CoValueClass<M>,
    options: {
      value: Simplify<CoMapInit<M>>,
      unique: CoValueUniqueness["uniqueness"],
      owner: Account | Group, // required so we don't need ownerId
      resolve?: RefsToResolveStrict<M, R>;
    }
```

Explaination:
- we don't want people to set `skipRetry: false` as per issue description it causes undesired delays
- loadAs can be inferred from `owner._loadedAs`
- we don't need `ownerId` if the owner is passed

Suggesting an options wrapper for everything because with 4 arguments better to make them named.

packages/jazz-tools/src/coValues/coMap.ts
```diff
@@ -564,6 +563,80 @@ export class CoMap extends CoValueBase implements CoValue {
...
+      const { owner, uniqueness } = parseCoValueCreateOptions(createOptions);
+      const raw = instance.rawFromInit(init, owner, uniqueness);
+      Object.defineProperties(instance, {
+        id: {
+          value: raw.id,
+          enumerable: false,
+        },
+        _raw: { value: raw, enumerable: false },
+      });
+      map = instance;
```
gdorsi: Would it be possible to de-dupe this logic?

We could make a standalone `createCoMap` function to use here and in the create.

packages/jazz-tools/src/coValues/coMap.ts
```diff
@@ -564,6 +563,80 @@ export class CoMap extends CoValueBase implements CoValue {
...
+    } else {
+      map.applyDiff(init as Partial<CoMapInit<M>>);
+    }
+
+    return map as M;
```
gdorsi: We need to call `loadCoValueWithoutMe` again to ensure that the applied update respects the `resolve` param.

packages/jazz-tools/src/subscribe/CoValueCoreSubscription.ts
```diff
@@ -10,22 +10,25 @@ export class CoValueCoreSubscription {
...
-        } else {
-          this.listener("unavailable");
-          this.subscribeToState();
-        }
-      });
+          if (value.isAvailable()) {
+            this.subscribe(value.getCurrentContent());
+          } else {
+            this.listener("unavailable");
+            if (!skipRetry) this.subscribeToState();
```
gdorsi: Could you explain the logic here?

packages/jazz-tools/src/tests/coMap.test.ts
```diff
@@ -469,6 +474,66 @@ describe("CoMap resolution", async () => {
...
+      name: z.string(),
+      age: z.number(),
+      dog: Dog,
+    });
+
+    // Disconnect the current account
+    const currentAccount = Account.getMe();
+    currentAccount._raw.core.node.syncManager.getPeers().forEach((peer) => {
+      peer.gracefulShutdown();
+    });
```
gdorsi: Considering that the load is called on the same node where the value is created we don't need to disconnect anything

packages/jazz-tools/src/tests/coMap.test.ts
```diff
@@ -469,6 +474,66 @@ describe("CoMap resolution", async () => {
...
```
gdorsi: Same here

packages/jazz-tools/src/tests/coMap.test.ts
```diff
@@ -543,6 +608,135 @@ describe("CoMap resolution", async () => {
...
+    expect(resolved).toBe(false);
+
+    await new Promise((resolve) =>
+      setTimeout(
+        resolve,
+        cojsonInternals.CO_VALUE_LOADING_CONFIG.RETRY_DELAY - 100,
+      ),
+    );
+
+    expect(resolved).toBe(true);
```
gdorsi: This kind of time-based tests are very likely to be flaky on CI when the machines are slow.

A possible way to test this is to set the `cojsonInternals.CO_VALUE_LOADING_CONFIG.RETRY_DELAY` to a very high value and check the the load resolves to `unavailable` (should be null here)

packages/jazz-tools/src/tests/coMap.test.ts
```diff
@@ -543,6 +608,135 @@ describe("CoMap resolution", async () => {
...
+    promise.then(() => {
+      resolved = true;
+    });
+
+    await new Promise((resolve) =>
+      setTimeout(
+        resolve,
+        cojsonInternals.CO_VALUE_LOADING_CONFIG.RETRY_DELAY - 100,
+      ),
+    );
```
gdorsi: For this one instead, we can:
- still set `cojsonInternals.CO_VALUE_LOADING_CONFIG.RETRY_DELAY - 100` to a very high value
- reconnect the `currentAccount` after 100ms 
- expect that the promise always resolve with the value

This is also timing-related, but should always succed. In slow devices might succeed because the load didn't start (very unlikely) but it's and acceptable compromise if we add more low-level tests

packages/jazz-tools/src/tests/coMap.test.ts
```diff
@@ -1201,6 +1395,85 @@ describe("Creating and finding unique CoMaps", async () => {
...
+      });
+    }
+    expect(activeEvent).toEqual({
+      title: sourceData.title,
+      identifier: sourceData.identifier,
+      external_id: sourceData._id,
+    });
+  });
+
+  test("upserting", async () => {
```
gdorsi: We need additional tests for:
- upserting on an already available value
- upserting with a resolve on a new value
- upserting with a resolve on a new value where the passed refereces don't match the given resolve
- upserting with a resolve on an existing value
- upserting with a resolve on an existing value where the passed refereces don't match the given resolve

packages/jazz-tools/src/tests/load.test.ts
```diff
@@ -69,6 +69,96 @@ test("retry an unavailable a value", async () => {
...
+  });
+
+  expect(resolved).toBe(false);
+
+  await new Promise((resolve) =>
+    setTimeout(
+      resolve,
+      cojsonInternals.CO_VALUE_LOADING_CONFIG.RETRY_DELAY - 100,
+    ),
+  );
```
gdorsi: Same issues flagged above

packages/jazz-tools/src/tests/load.test.ts
```diff
@@ -69,6 +69,96 @@ test("retry an unavailable a value", async () => {
...
+  promise.then(() => {
+    resolved = true;
+  });
+
+  await new Promise((resolve) =>
+    setTimeout(
+      resolve,
+      cojsonInternals.CO_VALUE_LOADING_CONFIG.RETRY_DELAY - 100,
+    ),
+  );
```
gdorsi: Same issues flagged above

packages/cojson/src/localNode.ts
```diff
@@ -419,11 +424,12 @@ export class LocalNode {
   subscribe<T extends RawCoValue>(
     id: CoID<T>,
     callback: (update: T | "unavailable") => void,
+    skipRetry?: boolean,
   ): () => void {
     let stopped = false;
     let unsubscribe!: () => void;
 
-    this.load(id)
+    this.load(id, skipRetry)
```
gdorsi, line 432: This change should be covered with a test in `cojson` to have more visibility on it's effect.

Check the other retry-related tests on the `sync.load.test.ts` file

homepage/homepage/content/docs/using-covalues/subscription-and-loading.mdx
```diff
@@ -239,6 +239,69 @@ Task.subscribe(taskId, {}, (task: co.loaded<typeof Task>) => {
 ```
 </CodeGroup>
 
+### Retrying unavailable items
```
gdorsi: I don't think we really need a section on the retry, should be a detail that users should not care about.

homepage/homepage/content/docs/using-covalues/comaps.mdx
```diff
@@ -445,3 +445,107 @@ console.log(isProjectActive(project)); // false
 console.log(formatProjectDuration(project, "short")); // "3 days"
 ```
 </CodeGroup>
+
+#### Upserting
+
+Consider a scenario where one needs to update a CoMap if it exists, or create it if it doesn't:
```
gdorsi: The docs should focus more on the real value of the API, which is not upserting, but make it possible to refer a coValue with human-firendly strings.

- Example from learnAnything on how to do slugs [link](https://github.com/learn-anything/attempt/blob/e6ab7a8aa44cfc77c50b1e3c7e5bbd17dd7b8905/web/app/routes/_layout/_pages/(topic)/%24.tsx#L22)
- Example from our cursor example on how to make a "global" covalue [link](https://github.com/garden-co/jazz/blob/main/examples/multi-cursors/src/utils/loadCursorContainer.ts#L49)

The second one should be ported to upsertUnique

packages/jazz-tools/src/coValues/coMap.ts
```diff
@@ -564,6 +587,87 @@ export class CoMap extends CoValueBase implements CoValue {
...
+        unique: options.unique,
+      });
+    } else {
+      map.applyDiff(options.value as Partial<CoMapInit<M>>);
+      map =
+        (await loadCoValueWithoutMe(this, mapId, {
+          ...options,
+          loadAs: options.owner._loadedAs,
+          skipRetry: true,
+        })) ?? map;
```
gdorsi: This revalidation should be done in both cases, and there should be no fallback to `map`

The intent is to stay true to the user request, and ensure that we return a resolved value or nothing.
Doing otherwise means that we may return a runtime value that doesn't match the Typescript type.

packages/jazz-tools/src/tests/coMap.test.ts
```diff
@@ -1197,8 +1379,303 @@ describe("Creating and finding unique CoMaps", async () => {
...
+      owner: workspace,
+    });
+
+    expect(updatedProject).toEqual(initialProject);
+    expect(updatedProject.name).toEqual("My updated project");
+    expect(myOrg.projects.length).toBe(1);
+    expect(myOrg.projects.at(0)?.name).toEqual("My updated project");
+  });
+
+  test("upserting a non-existent value with resolve not matching value", async () => {
```
gdorsi: This test doesn't match what I've requested (probably wasn't clear from the note)

It should be a test with two different accounts, one creates the projects list (readable by everyone), waitsForSync,  and the other does the upsert without having the list downloaded locally.

We should also add a test where the the user that does the upsert doesn't have read access to the list (I expect to return null)

packages/jazz-tools/src/subscribe/CoValueCoreSubscription.ts
gdorsi: > In the scenario where the CoValue is unavailable and we're not interested in retrying,
>  we are not interested in watching for changes since we're taking this to mean
> the CoValue doesn't exist.

This actually depends on where the API is used. For useCoState for example we actually want to stay subscribed.

Also we actually don't need this logic for APIs like load, because we unsubscribe as soon as we return the value.
It may be good to do `this.subscribeToState();` before triggering the listener, this way if the listerner triggers an unsubscribe there is no risk of having a pending listener and a conequent memory leak.

homepage/homepage/content/docs/using-covalues/comaps.mdx
```diff
@@ -461,3 +461,113 @@ console.log(isProjectActive(project)); // false
 console.log(formatProjectDuration(project, "short")); // "3 days"
 ```
 </CodeGroup>
+
+#### Uniqueness
+
+Identifying CoMaps by their CoValue IDs can be inconvenient if using an externally-generated identifier is preferrable, such as when:
```
aeplay:
```suggestion
CoMaps are typically created with a CoValue ID that acts as an opaque UUID, by which you can then load them. However, there are situations where it is preferable to load CoMaps using a custom identifier:
```

The sentence is a bit convoluted, here's a simplification suggestion

homepage/homepage/content/docs/using-covalues/comaps.mdx
```diff
@@ -461,3 +461,113 @@ console.log(isProjectActive(project)); // false
...
+- The CoMaps have user-generated identifiers, such as a slug
+- The CoMaps have identifiers referring to equivalent data in an external system
+- The CoMaps have human-readable & application-specific identifiers
+  - If an application has CoValues used by every user, referring to it by a unique name (eg, `"my-global-comap"`) can be more convenient than using a CoValue ID
```
aeplay:
```suggestion
  - If an application has CoValues used by every user, referring to it by a unique *well-known* name (eg, `"my-global-comap"`) can be more convenient than using a CoValue ID
```

"well-known" is a typical way to refer to this kind of abstraction

homepage/homepage/content/docs/using-covalues/comaps.mdx
```diff
@@ -461,3 +461,113 @@ console.log(isProjectActive(project)); // false
...
+});
+// ---cut---
+// This will not work as `learning-jazz` is not a CoValue ID
+const myTask = await Task.load("learning-jazz");
+```
+</CodeGroup>
+
+To make it possible to use human-readable identifiers Jazz lets you to define a `unique` property on CoMaps.
+
+Then the CoValue id will be generated based on the `unique` property and the owner of the CoMap.
```
aeplay:
```suggestion
Then the CoValue ID is deterministically derived from the `unique` property and the owner of the CoMap.
```

homepage/homepage/content/docs/using-covalues/comaps.mdx
```diff
@@ -461,3 +461,113 @@ console.log(isProjectActive(project)); // false
...
+  {
+    value: {
+      text: "Let's learn some Jazz!",
+    },
+    unique: "learning-jazz",
+    owner: project._owner,
+  }
+);
+```
+</CodeGroup>
```
aeplay, line 573:
```suggestion
</CodeGroup>

**Caveats:**

- The `unique` parameter acts as an *immutable* identifier - i.e. the same `unique` parameter in the same `Group` will always refer to the same CoValue.

  - To make dynamic renaming possible, you can create an indirection where a stable CoMap identified by a specific value of `unique` is simply a pointer to another CoMap with a normal, dynamic CoValue ID. This pointer can then be updated as desired by users with the corresponding permissions.
  
- This way of introducing identifiers allows for very fast lookup of individual CoMaps by identifier, but it doesn't let you enumerate all the CoMaps identified this way within a `Group`. If you also need enumeration, consider using a global `co.record()` that maps from identifier to a CoMap, which you then do lookups in (this requires at least a shallow load of the entire `co.record()`, but this should be fast for up to 10s of 1000s of entries)
```

This is kind of a logical implication, but worth spelling it out

aeplay (approved):

Issue #2494, opened 2025-06-11
applyDiff out of bounds for strings with emoji


bensleveritt:

bensleveritt:
Completed in https://github.com/garden-co/jazz/pull/2495

PR #2495, opened 2025-06-11
Fix: `applyDiff` out of bounds for strings with emoji

bensleveritt:
`coText.applyDiff` uses string index for comparison, which with emoji is incorrect. This PR adds splitting into grapheme string array first, then calculating the difference.


aeplay (approved):

bensleveritt:
Closes #2494

PR #2496, opened 2025-06-11
integrates inspector to all example apps

### LINKED ISSUE #2462: Add jazz-inspector to all example apps

4rknova:


trishalim (changes requested):
The tests are all failing due to this error
```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with <ROOT>/examples/passkey-svelte/package.json
```

This might be fixed by doing a `pnpm i` on root, and then committing the `pnpm-lock.yaml` changes that result from that. Lmk if that doesn't work

trishalim (approved):

PR #2501, opened 2025-06-11
fix for clerk example app .env

4rknova:


4rknova:
Updated environment variables in Vercel deployment

trishalim (approved):

trishalim:
Thanks Nikos!
Next, let's figure out how to provide the playwright test an API key

PR #2508, opened 2025-06-11
organization example app: Invitation creation moved to button click event

### LINKED ISSUE #2503: Create invitation onClick, rather than pre-emptively in Org. example
https://github.com/garden-co/jazz/blob/c496f49bb06572cbe54a92b86f97464ea42dbb54/examples/organization/src/components/InviteLink.tsx#L14

We currently create an invitation whenever the Organization id is updated (when the button has been rendered), but this is inefficient. Instead we should only create the invitation is there isn't already one when the button is clicked.

4rknova:


examples/organization/src/components/InviteLink.tsx
```diff
@@ -27,11 +21,17 @@ export function InviteLink({
...
-    if (inviteLink) {
-      navigator.clipboard.writeText(inviteLink).then(() => {
-        setCopyCount((count) => count + 1);
-      });
+    // Create invite link only if it doesn't exist yet
+    const linkToUse = inviteLink || createInviteLink(organization, "writer");
+
+    // Store the link if it was just created
+    if (!inviteLink) {
+      setInviteLink(linkToUse);
```
trishalim: You don't need to store the invite link in a state.

bensleveritt (approved):
This makes sense to me, presuming we don't need to keep invitations between clicks

PR #2511, opened 2025-06-12
Add docs for accept invite

Wizzel1:
I was searching for a way to handle invites programatically until I found the `acceptInvite` method. I thought it wouldn't hurt to include this in  the docs.

### DIFF
--- a/homepage/homepage/content/docs/groups/sharing.mdx
+++ b/homepage/homepage/content/docs/groups/sharing.mdx
@@ -99,8 +99,7 @@ const Organization = co.map({
 });
 const account = {} as unknown as Account;
 const organizationId = "";
-const inviteSecret ="";
-
+const inviteSecret = "inviteSecret_z";
 // ---cut---
 await account.acceptInvite(
   organizationId,
--- a/homepage/homepage/content/docs/groups/sharing.mdx
+++ b/homepage/homepage/content/docs/groups/sharing.mdx
@@ -86,7 +86,8 @@ useAcceptInvite({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


homepage/homepage/content/docs/groups/sharing.mdx
```diff
@@ -85,8 +85,25 @@ useAcceptInvite({
...
+
+const Organization = co.map({
+  name: z.string(),
+});
+
+await account.acceptInvite(
+  organizationId,
+  inviteSecret,
+  Organization
+);
```
gdorsi: Our examples are typechecked, so all the variables need to be declared.
Using `// ---cut---` to hide the hacks to make it work.

```suggestion
```ts twoslash
import { co, z, Account } from "jazz-tools";

const Organization = co.map({
  name: z.string(),
});
const account = {} as unknown as Account;
const organizationId = "";
const inviteSecret ="";

// ---cut---
await account.acceptInvite(
  organizationId,
  inviteSecret,
  Organization
);
```

gdorsi (commented):
Thanks for the contribution!

Suggested some fixes to the code example.

homepage/homepage/content/docs/groups/sharing.mdx
```diff
@@ -85,8 +85,29 @@ useAcceptInvite({
   },
 });
 ```
```
gdorsi, line 87:
```suggestion
```
</CodeGroup>
```

File: homepage/homepage/content/docs/groups/sharing.mdx [hunk 3]
```diff
@@ -85,8 +85,29 @@ useAcceptInvite({
...
-</CodeGroup>
```
gdorsi, line 89:
```suggestion
</CodeGroup>

```

PR #2513, opened 2025-06-12
chore: ✏️ update README and .gitignore files

### LINKED ISSUE #2371: Example apps should have .env in .gitignore
Feedback here: https://discord.com/channels/1139617727565271160/1139617728580305018/1376550079816536205

### Today
Example apps have very varied .gitignores, some have .env included, others do not. Although most of the example apps don’t have any secrets, developers might use the example apps as a starting point for an application which *does* have secrets, and then inadvertently end up committing their secrets to git.

### Plan
Go through all .gitignore files and add .env

joeinnes:
* Added instructions for renaming `.env.example` to `.env` in multiple README files.
* Updated `.gitignore` files to include `.env` and `.env.*` while excluding `.env.example` and `.env.test`.
* Removed `.env` file from `examples/multi-cursors`.
* Improved consistency across example projects.
Fixes #2371


examples/multi-cursors/README.md
```diff
@@ -1,63 +1,85 @@
 # Jazz Multi-Cursors Example
 
 Track user presence on a canvas with multiple cursors and out of bounds indicators.
-Live version: [https://multi-cursors.demo.jazz.tools/](https://multi-cursors.demo.jazz.tools/)
+Live version: [https://multi-cursors-demo.jazz.tools/](https://multi-cursors-demo.jazz.tools/)
```
trishalim: The old url was the correct one :)

examples/multi-cursors/.env
```diff
@@ -1,3 +0,0 @@
-VITE_CURSOR_FEED_ID=multi-cursors-250425-1708
-VITE_GROUP_ID=co_zXE8C8sd9QxEbxnt3neRvFRPFUc
-VITE_OLD_CURSOR_AGE_SECONDS=36000
```
trishalim: How about we just rename this to .env.example
This way, you only need to rename that to be able to run this (instead of finding the values in the console)

joeinnes: Good idea, will do that! I had not considered that we’re using a shared API key so every instance of this example is essentially part of the same app.

examples/multi-cursors/README.md
joeinnes: I wrote this before af388c1, will fix, thanks!

trishalim (approved):
Thanks so much!

PR #2514, opened 2025-06-12
feat: improve missing deps recovery & parallelize storage reads

### LINKED ISSUE #1917: Do not require depended-on CoValues to be synced/loaded before dependent ones
Right now a lot of code *requires* that depended-on CoValues ('dependencies') such as:

- owning `Group`
- `Account` members of a `Group`
- extended `Groups`

...are synced/loaded before the CoValue that depends on them.

While we may keep proactive syncing of dependencies as a *performance optimisation* to save roundtrips, the system should be able to deal with missing dependencies, represent the resulting incomplete state cleanly, and should request dependencies to be synced/loaded ad-hoc when needed.

gdorsi:
Started by working on improving performance by parallelizing read operations on storage, while working I've found some gaps in the current deps recovery system.

Now we handle correctly missing parent groups (was breaking the app before) and a coValue is marked as available only after all their required deps are loaded.

fixes #1917

The storage reads parallelization is bringing around 1s gain when loading 1k tasks

Before:
Screenshot 2025-06-12 at 17 10 53

After:
Screenshot 2025-06-12 at 17 08 55

### DIFF
--- a/packages/cojson-storage-indexeddb/src/CoJsonIDBTransaction.ts
+++ b/packages/cojson-storage-indexeddb/src/CoJsonIDBTransaction.ts
@@ -39,7 +39,7 @@ export class CoJsonIDBTransaction {
   startedAt = performance.now();
   isReusable() {
     const delta = performance.now() - this.startedAt;
-    return !this.done && delta <= 20;
+    return !this.done && !this.failed && delta <= 100;
   }
 
   getObjectStore(name: StoreName) {
--- a/packages/cojson-storage-indexeddb/src/tests/storage.indexeddb.test.ts
+++ b/packages/cojson-storage-indexeddb/src/tests/storage.indexeddb.test.ts
@@ -103,7 +103,9 @@ test("should sync and load data from storage", async () => {
     [
       "client -> LOAD Map sessions: empty",
       "storage -> CONTENT Group header: true new: After: 0 New: 3",
+      "client -> KNOWN Group sessions: header/3",
       "storage -> CONTENT Map header: true new: After: 0 New: 1",
+      "client -> KNOWN Map sessions: header/1",
     ]
   `);
 });
@@ -174,7 +176,9 @@ test("should send an empty content message if there is no content", async () => 
     [
       "client -> LOAD Map sessions: empty",
       "storage -> CONTENT Group header: true new: After: 0 New: 3",
+      "client -> KNOWN Group sessions: header/3",
       "storage -> CONTENT Map header: true new: ",
+      "client -> KNOWN Map sessions: header/0",
     ]
   `);
 });
@@ -255,8 +259,11 @@ test("should load dependencies correctly (group inheritance)", async () => {
     [
       "client -> LOAD Map sessions: empty",
       "storage -> CONTENT ParentGroup header: true new: After: 0 New: 4",
+      "client -> KNOWN ParentGroup sessions: header/4",
       "storage -> CONTENT Group header: true new: After: 0 New: 5",
+      "client -> KNOWN Group sessions: header/5",
       "storage -> CONTENT Map header: true new: After: 0 New: 1",
+      "client -> KNOWN Map sessions: header/1",
     ]
   `);
 });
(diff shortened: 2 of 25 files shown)


packages/cojson-storage-indexeddb/src/CoJsonIDBTransaction.ts
```diff
@@ -39,7 +39,7 @@ export class CoJsonIDBTransaction {
   startedAt = performance.now();
   isReusable() {
     const delta = performance.now() - this.startedAt;
-    return !this.done && delta <= 20;
+    return !this.done && !this.failed && delta <= 100;
```
gdorsi, line 42: increasing the time delta to batch more reads in the same transaction

packages/cojson-storage-indexeddb/src/tests/storage.indexeddb.test.ts
```diff
@@ -255,8 +259,11 @@ test("should load dependencies correctly (group inheritance)", async () => {
     [
       "client -> LOAD Map sessions: empty",
       "storage -> CONTENT ParentGroup header: true new: After: 0 New: 4",
+      "client -> KNOWN ParentGroup sessions: header/4",
       "storage -> CONTENT Group header: true new: After: 0 New: 5",
+      "client -> KNOWN Group sessions: header/5",
       "storage -> CONTENT Map header: true new: After: 0 New: 1",
+      "client -> KNOWN Map sessions: header/1",
```
gdorsi, line 266: Showing up because of the parallelization, before these were delayed more and not recorded

packages/cojson-storage/src/syncUtils.ts
```diff
@@ -47,60 +40,16 @@ export function getDependedOnCoValues(
...
-  } catch (e) {
-    return [];
-  }
+  const id = contentMessage.id;
+  const sessionIDs = Object.keys(contentMessage.new) as SessionID[];
+  const transactions = Object.values(contentMessage.new).map(
+    (entry) => entry.newTransactions,
+  );
+
+  return cojsonInternals.getDependedOnCoValuesFromRawData(
```
gdorsi, line 49: Since this util is now compatible with content messages we can reuse the same logic from cojson to process deps

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -229,13 +233,34 @@ export class CoValueCore {
...
+        if (this.isAvailable()) {
+          this.notifyUpdate("immediate");
+        }
+      });
+
+      this.missingDependencies.add(dependency);
+    }
+  }
+
+  provideHeader(header: CoValueHeader, fromPeerId: PeerID) {
```
gdorsi, line 256: Renamed "markAvailable" into "provideHeader"

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -919,51 +944,20 @@ export class CoValueCore {
...
-    }
-
-    if (header.ruleset.type === "group") {
-      if (isAccountID(header.ruleset.initialAdmin)) {
-        deps.add(header.ruleset.initialAdmin);
-      }
-
-      if (this.verified) {
-        for (const id of getGroupDependentKeyList(
-          expectGroup(this.getCurrentContent()).keys(),
```
gdorsi, line 953: This was consuming the group content before, leading us to throw when the parent groups weren't available.

Refactored into a dependency check that reads only from the raw transactions.

packages/cojson/src/sync.ts
```diff
@@ -565,13 +544,18 @@ export class SyncManager {
       if (isAccountID(accountId)) {
         const account = this.local.getCoValue(accountId);
 
+        // We can't verify the transaction without the account, so we skip it and ask for a correction
         if (!account.isAvailable()) {
-          account.loadFromPeers([peer]);
```
gdorsi, line 569: We are already loading the missing dependencies above

packages/cojson/src/sync.ts
```diff
@@ -565,13 +544,18 @@ export class SyncManager {
...
-          invalidStateAssumed = true;
+          if (!coValue.correctionsRequested.has(sessionID)) {
+            coValue.correctionsRequested.add(sessionID);
```
gdorsi: This is to avoid inifinite loops of corrections requests

packages/cojson/src/sync.ts
```diff
@@ -565,13 +544,18 @@ export class SyncManager {
...
+            invalidStateAssumed = true;
```
gdorsi: We should time the correction request to when the account is loaded.

Going to do this in a separate PR.

PR #2519, opened 2025-06-12
Prevent submitting form if account id isn't accompanied by a secret

shortdiv:
Noticed that the console was blowing up an error when I tried to load an invalid secret and/or I submitted the form with only an account id and not secret. This adds an extra check so we prevent this from happening

Screenshot 2025-06-12 at 11 15 04 AM

### DIFF
--- a/examples/inspector/src/viewer/new-app.tsx
+++ b/examples/inspector/src/viewer/new-app.tsx
@@ -54,6 +54,12 @@ export default function CoJsonViewerApp() {
   const [coValueId, setCoValueId] = useState<CoID<RawCoValue> | "">("");
   const { path, addPages, goToIndex, goBack, setPage } = usePagePath();
 
+  const KNOWN_ERRORS = [
+    "trying to load covalue with invalid id",
+    "account has no profile",
+    "invalid sealer secret format",
+  ];
+
   useEffect(() => {
     localStorage.setItem("inspectorAccounts", JSON.stringify(accounts));
   }, [accounts]);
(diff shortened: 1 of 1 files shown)


examples/inspector/src/viewer/new-app.tsx
```diff
@@ -369,7 +382,11 @@ function AddAccountForm({
         onChange={(e) => setSecret(e.target.value)}
         placeholder="sealerSecret_ziz7NA12340abcdef123789..."
       />
-      <Button className="mt-3" type="submit">
+      <Button
+        className="mt-3"
+        type="submit"
+        disabled={secret.length == 0 || id.length == 0}
```
trishalim: Instead of disabling the button, can you please add `required` to both `Input`s above. This way, the form won't submit if either is empty.

examples/inspector/src/viewer/new-app.tsx
```diff
@@ -54,6 +54,12 @@ export default function CoJsonViewerApp() {
   const [coValueId, setCoValueId] = useState<CoID<RawCoValue> | "">("");
   const { path, addPages, goToIndex, goBack, setPage } = usePagePath();
 
+  const KNOWN_ERRORS = [
+    "trying to load covalue with invalid id",
+    "account has no profile",
+    "invalid sealer secret format",
```
trishalim, line 60: Am I supposed to see this message when I put an account ID and an invalid secret? I'm not seeing it 🤔

shortdiv: yeah :/ hmm lemme check again!

shortdiv: This is how I'm testing:
[Video test of incorrect sealer secret](https://github.com/user-attachments/assets/7973a6be-33e6-427d-8280-49fb5fa353c8)

examples/inspector/src/viewer/new-app.tsx
shortdiv: Sure thing!

PR #2520, opened 2025-06-13
chore: upgrade examples to tailwind v4

trishalim:
#2521

This does **not** include
- react native examples
- prosekit example
- the examples that use shadcn are handled here https://github.com/garden-co/jazz/pull/2530


gdorsi (approved):
Thanks!

PR #2524, opened 2025-06-13
Add "use client" directive for Next.js

tobiaslins:
I haven't verified if tsc leaves this at the top of the file - but we need something like this


gdorsi:
Implemented as part of https://github.com/garden-co/jazz/pull/2552/commits/fc342306e2eb468b063ff9256118dbb74c40f01e using a bundle banner.

Thanks for the suggestion!

PR #2543, opened 2025-06-16
fix: remove RN Buffer polyfill for ProgressiveImg

boorad:
There is a [new error](https://github.com/craftzdog/react-native-buffer/issues/14) with a RN `Buffer` polyfill, not sure if it was a change in Hermes or RN or what...  However, Jazz `<ProgressiveImg>` was affected.

The workaround removed one use of `Buffer`, and so I chose to remove the others, reducing polyfills and making setup at least a little easier.

fixes GCO-570

### DIFF
--- a/packages/jazz-react-native-core/src/media.tsx
+++ b/packages/jazz-react-native-core/src/media.tsx
@@ -41,7 +41,11 @@ export function useProgressiveImg({
           });
 
           // Create data URL
-          const base64 = Buffer.from(combinedArray).toString("base64");
+          const base64 = btoa(
+            Array.from(combinedArray, (byte) => String.fromCharCode(byte)).join(
+              "",
+            ),
+          );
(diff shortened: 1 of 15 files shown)


packages/jazz-react-native-core/src/media.tsx
```diff
@@ -41,7 +41,11 @@ export function useProgressiveImg({
           });
 
           // Create data URL
-          const base64 = Buffer.from(combinedArray).toString("base64");
+          const base64 = btoa(
+            Array.from(combinedArray)
+              .map((byte) => String.fromCharCode(byte))
```
gdorsi: **nit:** we can use the Array.from callback to optimize this and generate only one array

```suggestion
            Array.from(combinedArray, (byte) => String.fromCharCode(byte))
```

gdorsi (approved):
LGTM!

Could you add a changeset?

PR #2544, opened 2025-06-17
Adds caching to the highlighter

bensleveritt:
Which does seem to make things a lot faster on subsequent runs. Just need to check that code blocks update as expected.


trishalim:
Tested it. 

Problem: I'm now getting the connection error on `pnpm dev` while it tries (and fails) to load the font. Previously, it was only a `pnpm build` issue. :( 

Here it is without the right fonts
image

On the bright side, it definitely loads faster now.

PR #2549, opened 2025-06-17
feat: add FileStream.loadAsBase64 API

gdorsi:
Since base64 is commonly used to handle FileStream values in React Native I've decided to make a first-class API to load a FileStream directly in base64 encoding

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/coFeed.ts
+++ b/packages/jazz-tools/src/tools/coValues/coFeed.ts
@@ -774,6 +774,44 @@ export class FileStream extends CoValueBase implements CoValue {
     });
   }
 
+  static async loadAsBase64(
+    id: ID<FileStream>,
+    options?: {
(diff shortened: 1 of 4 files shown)


packages/jazz-tools/src/tools/coValues/coFeed.ts
```diff
@@ -774,6 +774,44 @@ export class FileStream extends CoValueBase implements CoValue {
...
+      for (const byte of chunk) {
+        output.push(String.fromCharCode(byte));
+      }
+    }
+
+    const base64 = btoa(output.join(""));
+
+    if (options?.dataURL) {
+      return `data:${chunks.mimeType};base64,${base64}`;
+    }
```
gdorsi, line 810: Wondering if dataURL should be enabled by default 🤔

aeplay (approved):

PR #2568, opened 2025-06-23
feat(inspector): improve CoPlainText view

### LINKED ISSUE #2202: CoText display
`instructions` is a CoPlainText that says "hey"

Image

Image

- [ ] show a "CoPlainText" or "CoRichText" type + an icon
- [ ] in a list/grid, show text content with a link to the covalue, instead of just the covalue ID
- [ ] improve page display

trishalim:
image
closes #2202

### DIFF
--- a/packages/jazz-tools/src/inspector/viewer/co-plain-text-view.tsx
+++ b/packages/jazz-tools/src/inspector/viewer/co-plain-text-view.tsx
@@ -1,0 +1,17 @@
+import { JsonObject } from "cojson";
+import { RawDataCard } from "./raw-data-card.js";
+
+export function CoPlainTextView({
+  data,
+}: {
+  data: JsonObject;
+}) {
+  if (!data) return;
+
+  return (
+    <>
+      <p>{Object.values(data).join("")}</p>
+      <RawDataCard data={data} />
+    </>
+  );
+}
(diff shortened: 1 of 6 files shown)


packages/jazz-tools/src/inspector/viewer/co-plain-text-view.tsx
```diff
@@ -0,0 +1,17 @@
...
+  data,
+}: {
+  data: JsonObject;
+}) {
+  if (!data) return;
+
+  return (
+    <>
+      <p>{Object.values(data).join("")}</p>
+      <RawDataCard data={data} />
```
gdorsi, line 14: Do we need to show the "row data" for this data type?

gdorsi (commented):
Could you add a small e2e test for this feature?

trishalim:
> Could you add a small e2e test for this feature?

Yes!

packages/jazz-tools/src/inspector/viewer/co-plain-text-view.tsx
trishalim: Good question 🤔

aeplay: I don't think we do, but it doesn't really matter. We'll re-address this when we think about the CoValue visualisations most likely.

aeplay (approved):

PR #2575, opened 2025-06-25
feat: storage as load/store API

gdorsi:
Changes from using storage as any other peer into having an imperative API to sync with storage with the follwing interface:

```ts
/**
 * The StorageAPI is the interface that the StorageSync and StorageAsync classes implement.
 *
 * It uses callbacks instead of promises to have no overhead when using the StorageSync and less overhead when using the StorageAsync.
 */
export interface StorageAPI {
  load(
    id: string,
    // This callback is fired when data is found, might be called multiple times if the content requires streaming (e.g when loading files)
    callback: (data: NewContentMessage) => void,
    done?: (found: boolean) => void,
  ): void;
  store(
    data: NewContentMessage[] | undefined,
    handleCorrection: (correction: CoValueKnownState) => void,
  ): void;
  getKnownState(id: string): CoValueKnownState;
  waitForSync(id: string, knownState: CoValueKnownState): Promise<void>;
  close(): void;
}
```

and is used in our sync code this way:
```ts
 node.storage.load(
      this.id,
      (data) => {
        // Loads the content in memory and triggers subscriptions with server peers
        node.syncManager.handleNewContent(data);
      },
      (found) => {
        if (!found) {
          this.markNotFoundInPeer("storage");
        }

        // Handle follow-up code here, e.g. loadFromPeers if the value is not in storage
      },
    );
```

```ts
  storage.store(data, (correction) => { // May be fired if our assumptions on the data stored is wrong
    if (!coValue.verified) return;

    const newContentPieces = coValue.verified.newContentSince(correction);

    if (!newContentPieces) return;

    storage.store(newContentPieces, (response) => {
      logger.error(
        "Correction requested by storage after sending a correction content",
        {
          response,
        },
      );
    });
  });
```

This change is in preparation for future improvements, including a backpressure system for sync (we can now wait for the storage response) and more storage-related features (e.g. doing reconciliation checks without loading data in memory or loading data in bulk)

It also brings a 10% improvement on performance when loading many values from storage.


packages/cojson-storage-sqlite/package.json
```diff
@@ -7,8 +7,7 @@
   "license": "MIT",
   "dependencies": {
     "better-sqlite3": "^11.7.0",
-    "cojson": "workspace:0.15.3",
-    "cojson-storage": "workspace:*"
+    "cojson": "workspace:*"
```
gdorsi, line 10: Moved `cojson-storage` inside `cojson` to facilitate integration testing for the sync.

We have now storage tests based on libsql for different sync scenarios, all with messages logs.

packages/cojson-storage-sqlite/src/index.ts
```diff
@@ -1 +1,39 @@
...
+  closeDb() {
+    this.db.close();
+  }
+}
+
+export function getBetterSqliteStorage(filename: string) {
+  const db = new BetterSqliteDriver(filename);
+
+  return getSqliteStorage(db);
+}
```
gdorsi, line 39: Now creating a sqlite adapter is this easy and the API will probably stay like this.

packages/cojson/package.json
```diff
@@ -28,6 +28,7 @@
   "version": "0.15.3",
   "devDependencies": {
     "@opentelemetry/sdk-metrics": "^2.0.0",
+    "libsql": "^0.5.13",
```
gdorsi, line 31: For the storage tests

packages/cojson/src/storage/knownState.ts
```diff
@@ -0,0 +1,70 @@
+import { getIsUploaded } from "../SyncStateManager.js";
+import { RawCoID } from "../ids.js";
+import { CoValueKnownState, emptyKnownState } from "../sync.js";
+
+export class StorageKnownState {
```
gdorsi, line 10: Used to track how much data we have in Storage and to wait for sync with storage.

packages/cojson/src/tests/sync.storage.test.ts
```diff
@@ -219,51 +206,115 @@ describe("client syncs with a server with storage", () => {
...
+    client.addStorage({
+      ourName: "client",
+      storage,
+    });
+
+    const mapOnClient2 = await loadCoValueOrFail(client.node, largeMap.id);
+
+    await mapOnClient2.core.waitForSync();
+
+    // TODO: The client should wait for the full load from the storage peer before subscribing to the core server
```
gdorsi: Going to fix this with a follow-up PR

https://github.com/garden-co/jazz/issues/2584

aeplay (approved):

PR #2577, opened 2025-06-25
Fix GCO-600: image original size

Takeno:
### What this Does
At the moment, the originalSize saved in ImageDefinition is [calculated](https://github.com/garden-co/jazz/blob/main/packages/jazz-tools/src/browser-media-images/index.ts#L26-L39) on the original image. But in case the maxSize argument is used, the original image [will not be uploaded](https://github.com/garden-co/jazz/blob/main/packages/jazz-tools/src/browser-media-images/index.ts#L123), causing the originalSize property useless.

### Why Are We Doing This?
This PR adjusts the the originalSize using the highest resolution available.

### Scope / Boundaries
Includes:
- [x] Bug fixing 05eab4e2a99097cb91a50bab00f3d5ef4e0a690b
- [x] Tests inside browser 05eab4e2a99097cb91a50bab00f3d5ef4e0a690b
- [x] Small refactoring 5d2ff7b992e0b47a72bbd0394061d64bb8fdf51c

### Testing Instructions
Creating an Image 1000x1000 at source with `maxSize: 512` must create an Image resource with `originalSize: [512, 512]`.

### Known Issues / Open Questions (if any)
- [ ] Double check the headless browser testing pattern

### Related Links
- [Linear GCO-600](https://linear.app/garden-co/issue/GCO-600/bug-images-originalsize-is-not-coherent-with-resized-image-if-maxsize)


packages/jazz-tools/src/browser-media-images/index.ts
```diff
@@ -18,113 +18,108 @@ export async function createImage(
...
+  imageBlobOrFile: Blob | File,
+): Promise<{ width: number; height: number }> {
+  const { width, height } = await new Promise<{
+    width: number;
+    height: number;
+  }>((resolve, reject) => {
+    const img = new Image();
+    img.onload = () => resolve({ width: img.width, height: img.height });
+    img.onerror = () => reject(new Error("Failed to load image"));
+    img.src = URL.createObjectURL(imageBlobOrFile);
```
gdorsi, line 84: Once the image is loaded it's good to call [URL.revokeObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL_static) to free up the reference.

gdorsi (approved):
Looks good!

One minor issue to fix and the PR is good to merge.

Could you add a changeset?

packages/jazz-tools/src/browser-media-images/index.ts
Takeno: Good catch, thanks!

PR #2587, opened 2025-06-28
feat: add RN BetterAuth

boorad:
### What this Does
Changes that verify the BetterAuth package for React also works with RN and Expo apps.

### Why Are We Doing This?
As a result of testing Jazz & BetterAuth with an external app, these changes:
- import hooks from `jazz-tools/react-core`, their original source, and better for sharing across platforms
- separated types for `newAuthClient`, making type derivation easier for `tsc` and users of the function
- sync'd React and RN code a bit more
- removed yalc, which wasn't working properly.
- cleaned up unused imports, added `type` to others

### Scope / Boundaries
This should very minimally affect React, React Native, and Expo apps.

### Testing Instructions
We need to talk about how to test this without having a proliferation of example apps with every permutation of framework, auth, crypto, etc..  My only testing for this PR is in a closed-source app.

### Known Issues / Open Questions (if any)
N/A

### Related Links
fixes GCO-620


packages/jazz-tools/src/react/provider.tsx
```diff
@@ -74,6 +74,7 @@ export function JazzReactProvider<
 
         if (contextManager.propsChanged(props)) {
           contextManager.createContext(props).catch((error) => {
+            console.log(error.stack);
```
boorad, line 77: This is a weird line that I added, as part of keeping the react and react-native code bases in provider the same.  I can go either way.  Keep both or remove both.  The other one is right above, in react-native-core/provider.tsx.  That edit has the `satisfies` clause, to also keep both code bases in sync.

packages/jazz-auth-betterauth/src/react.ts
```diff
@@ -0,0 +1,133 @@
...
+ */
+export function createUseBetterAuthHook<T extends ClientOptions>(hooks: {
+  useJazzContext: () => {
+    authenticate: AuthenticateAccountFunction;
+    [key: string]: any;
+  };
+  useAuthSecretStorage: () => AuthSecretStorage;
+  useIsAuthenticated: () => boolean;
+}) {
+  const { useJazzContext, useAuthSecretStorage, useIsAuthenticated } = hooks;
```
gdorsi: These hooks are actually the same between react and react-native and you can import them directly from `jazz-tools/react-core` instead having to provide a factory method.

For this reason I think that the whole implementation of better-auth can be shared between react and react native by just importing the jazz hooks from `jazz-tools/react-core`

boorad: Yeah, I knew they were similar. But you are right, they are the same.

However, when I revert most of the code (in the latest commit), the RN app complains with an error in the screenshot below.  This is what I saw initially when I tried to use the React BetterAuth directly.  Also, changing the import of the hooks to `jazz-tools/react-core` from `jazz-tools/react` makes no difference.

I have not yet found the difference between React provider / context and React Native provider / context, or why the context value in RN is `undefined`.  If you don't know anything off the top of your head, I'll continue to look tomorrow.

spicy screenshot 2025-07-01 22 44 29

gdorsi (approved):
LGTM 🚀 

Could you add a changeset?

packages/jazz-auth-betterauth/src/react.ts
gdorsi: Dunno.

Checked the jazz-tools dist, and both `useJazzContext` and `JazzContext` are correctly imported from `jazz-tools/react-core`

The only other possible issue is with some double version of jazz-tools, it may happen if you are testing better-auth outside our monorepo.

We should consider moving the `better-auth` packages into `jazz-tools`

Haven't done that yet because I still don't consider them stable enough to be part of `jazz-tools` (there are some bits I need to double-check, like "do we really need that AuthProvider?" but haven't got the time yet)

PR #2590, opened 2025-06-30
 fix: server subscription when streaming from storage

### LINKED ISSUE #2584: Bug: streaming from storage triggers content download from the server peers

gdorsi:
### What this Does

To optimize the loading of large CoValue sessions, we split the content in many small blocks and we call it streaming.

Currently we have a bug when streaming content from storage, that leads to having content be pushed also from the sync server even thou it's available on storage.

In [this test for](https://github.com/garden-co/jazz/blob/feat/storage-api/packages/cojson/src/tests/sync.storage.test.ts#L304) example there should be no content sent from the server.

This PR fixes this bug adding `streamingTarget` prop to NewContentMessage, to make it possible to subscribe to the server peer by telling the final known state after the streaming and not the current one.

Fixes #2584 

### Testing Instructions
- Open the filestream demo https://jazz-filestream-git-fix-storage-streaming-gcmp.vercel.app/
- Open the network inspector
- Upload a file larger than 100Kb
- Wait for the full sync with the server
- Reload the page
- Check the WebSocket messages, there should no content messages going to the sync server

The difference is noticeable when compared the messages sent on https://file-upload.demo.jazz.tools/ using the same flow.


packages/cojson/src/coValue.ts
```diff
@@ -18,7 +15,7 @@ export type CoID<T extends RawCoValue> = RawCoID & {
 export interface RawCoValue {
   /** The `CoValue`'s (precisely typed) `CoID` */
   id: CoID<this>;
-  core: AvailableCoValueCore;
+  core: CoValueCoreVerified;
```
gdorsi, line 18: Renamed because it isn't related to Available state, which requires also the CoValue deps, but only a CoValue with a VerifiedState

packages/cojson/src/storage/knownState.ts
```diff
@@ -48,16 +49,30 @@ export class StorageKnownState {
...
+    if (isInSync(coValue.knownState(), this.getKnownState(id))) {
       return Promise.resolve();
     }
 
     const requests = this.waitForSyncRequests.get(id) || new Set();
     this.waitForSyncRequests.set(id, requests);
 
     return new Promise<void>((resolve) => {
-      requests.add({ knownState, resolve });
+      const knownState = coValue.knownState();
```
gdorsi: Before the `waitForSync` was getting the initial `knownState` of the in-memory coValue and so not working when waiting from the content to be pulled from the storage

packages/cojson/src/sync.ts
```diff
@@ -482,28 +485,38 @@ export class SyncManager {
...
+        msg.streamingTarget,
+      );
+
+      if (msg.streamingTarget) {
+        peer?.combineWith(msg.id, {
+          id: msg.id,
+          header: true,
+          sessions: msg.streamingTarget,
+        });
+      }
```
gdorsi, line 516: Here we track how much content the peer has, and since we know that the other peer is going to stream content to us, we track the streamingTarget here.

tests/browser-integration/src/browserSync.test.ts
```diff
@@ -182,9 +179,6 @@ describe("Browser sync", () => {
       AccountSchema: CustomAccount,
     });
 
-    // TODO: Wait for sync doesn't work on the IndexedDB storage peer as it just waits for the content to be pushed
-    await new Promise((resolve) => setTimeout(resolve, 500));
-
```
gdorsi, line 187: Improved the API and stability around storage, so we can now remove these workarounds from the browser tests.

packages/cojson/src/coValueCore/verifiedState.ts
```diff
@@ -336,6 +351,43 @@ export class VerifiedState {
...
+      const entries = Object.entries(this.streamingKnownState);
+      let outdated = true;
+
+      for (const [sessionID, txs] of entries) {
+        if ((knownState.sessions[sessionID as SessionID] ?? 0) < txs) {
+          newSessions[sessionID as SessionID] = txs;
+          outdated = false;
+        } else {
+          newSessions[sessionID as SessionID] = txs;
+        }
```
nrainhart: Nit:
```suggestion
        newSessions[sessionID as SessionID] = txs;
        if ((knownState.sessions[sessionID as SessionID] ?? 0) < txs) {
          outdated = false;
        }
```

packages/cojson/src/storage/knownState.ts
nrainhart: Nit: the result of `coValue.knownState()` could be extracted into a constant to avoid calling the method twice (here and in L53)

nrainhart (approved):
I've got very little context on jazz's architecture and abstractions, but I didn't notice any flagrant errors in this PR. Left only a few minor comments

PR #2596, opened 2025-06-30
fix #1914

### LINKED ISSUE #1914: Make project name optional if an example is provided
When running `npx create-jazz-app --example passkey-svelte`, you are prompted for a project name, and you're not able to skip it.

Make this optional, and default to the example name, or the example name affixed with `-app` like `passkey-svelte-app`

satendra03:
**What this Does**
Improves the project name handling in `create-jazz-app`. When using `--example` or `--starter`, it now defaults to `<template-name>-app` (e.g., `chat` → `chat-app`).

**Why Are We Doing This?**
This solves the issue where the CLI would prompt for a project name even when a valid example or starter is provided. It aligns with the expected behavior for a smoother developer experience.

**Testing Instructions**
Run: `npx create-jazz-app --example chat`

Skip project name
✅ Should create folder named chat-app and continue setup

Try with invalid example like passkey-svelte
❌ Should fail early due to missing example directory

**Related Links**
[GitHub Issue](https://github.com/garden-co/jazz/issues/1914)

### DIFF
--- a/packages/create-jazz-app/src/index.ts
+++ b/packages/create-jazz-app/src/index.ts
@@ -493,10 +493,24 @@ async function promptUser(
   }
 
   if (!partialOptions.projectName) {
+    // Determine a default project name if possible
+    let defaultProjectName = undefined;
+    if (partialOptions.example) {
+      // Use the example name, affixed with -app if not already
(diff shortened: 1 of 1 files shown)


gdorsi (approved):
Thanks for the contribution!

PR #2600, opened 2025-07-01
Make all React Native deps in `jazz-tools` optional peer dependencies

### LINKED ISSUE #2576: Remove react-native-nitro-modules from the jazz-tools deps

nrainhart:
### What this Does

Follow-up to https://github.com/garden-co/jazz/pull/2595. Closes https://github.com/garden-co/jazz/issues/2576

Converts all React Native dependencies in `jazz-tools` into optional peer dependencies, since these dependencies are already required to be installed in the host project.

This PR also updates the project setup docs for Expo to include all required dependencies.

### Why Are We Doing This?

Some users have reported issues with React Native dependencies in non-RN projects. This also reduces the download size when installing `jazz-tools` noticeably, given that some of this RN dependencies are quite heavy (e.g. `op-sqlite` weights 100Mb).

### Testing Instructions

- [x] Confirmed the React Native and Expo Project setup docs include all required peer dependencies
- [x] Confirmed `chat-rn` and `chat-rn-expo` include the required peer dependencies
- [x] Manually tested that both the `chat-rn` and `chat-rn-expo` examples continue to work.

### Related Links
- GitHub issue: https://github.com/garden-co/jazz/issues/2576

### DIFF
--- a/packages/jazz-tools/package.json
+++ b/packages/jazz-tools/package.json
@@ -161,6 +161,7 @@
     "prosemirror-schema-basic": "^1.2.2",
     "prosemirror-state": "^1.4.3",
     "prosemirror-transform": "^1.9.0",
+    "react-native-fast-encoder": "^0.2.0",
     "zod": "3.25.28"
   },
   "scripts": {
@@ -198,7 +199,6 @@
     "react": "*",
     "react-dom": "*",
     "react-native": "*",
-    "react-native-fast-encoder": "^0.2.0",
     "react-native-mmkv": "^3.2.0",
     "react-native-nitro-modules": "0.25.2",
     "react-native-quick-crypto": "1.0.0-beta.16",
@@ -230,9 +230,6 @@
       "optional": true
     },
     "react-native": {
-      "optional": true
-    },
-    "react-native-fast-encoder": {
       "optional": true
     },
     "react-native-mmkv": {
--- a/packages/jazz-tools/package.json
+++ b/packages/jazz-tools/package.json
@@ -161,7 +161,6 @@
     "prosemirror-schema-basic": "^1.2.2",
     "prosemirror-state": "^1.4.3",
     "prosemirror-transform": "^1.9.0",
-    "react-native-fast-encoder": "^0.2.0",
     "zod": "3.25.28"
   },
   "scripts": {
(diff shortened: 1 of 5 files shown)


packages/jazz-tools/package.json
```diff
@@ -164,7 +162,6 @@
     "prosemirror-state": "^1.4.3",
     "prosemirror-transform": "^1.9.0",
     "react-native-fast-encoder": "^0.2.0",
```
nrainhart: There are no usages of this dependency in our codebase, so I'm not sure why/when it is required. @gdorsi @Takeno do you happen to know?

Takeno: It has been introduced here: https://github.com/garden-co/jazz/commit/0d5ee3ed075dab7a37bcc22c0bbd01f151cb320c

/cc @boorad

nrainhart: Looks like this is only necessary for [Quick Crypto](https://jazz.tools/docs/react-native-expo/project-setup/providers#quick-crypto) support? If so I'll make this an optional peer dependency as well and update that doc to require installing this dependency as well.

@boorad can you confirm?

boorad: yep, I'd like to try this optional peers setup to not drag too much into a build until you begin to use it...

gdorsi (approved):
Great, thanks!

Could you add a changeset?

PR #2601, opened 2025-07-02
Update pinned react version to 19.1.0

nrainhart:
### What this Does
Updates our pinned react version (and the react version of all examples & the homepage) to 19.1.0 and updates `chat-rn`'s `Podfile.lock`.

### Why Are We Doing This?
I tried running our chat-rn example and ran into this error:
```
Incompatible React versions: The "react" and "react-native-renderer" packages must have the exact same version. Instead got:
- react:
19.0.0
- react-native-renderer: 19.1.0
```

### Testing Instructions
- In main: run `cd examples/chat-rn && pnpm ios` and notice the react version mismatch error
- In this branch: run `cd examples/chat-rn && pnpm ios` again and notice the example starts


nrainhart:
Turns out chat-rn-expo requires `react@19.0.0`, because expo only supports react-native up to 0.79.x (support for 0.80.x is only available in [canary releases](https://expo.dev/changelog/react-native-80#support-is-available-on-the-expo-canary-channel)), and that version of RN only supports react up to 19.0.0.

So, this leaves us in an awkward position:
- chat-rn requires react@19.1.0 because it's already on react-native@0.80.0
- chat-rn-expo requires react@19.0.0

And we enforce a single version of react across the monorepo because we use [pnpm.overrides](https://github.com/garden-co/jazz/pull/2601/files#diff-de3ff16f7e05b9062b4df4a745813aacb1b739332af5ecb12c60421b6dd7e1c1L186).

What was the motivation for using `pnpm.overrides`? I wonder if we should switch to using a [pnpm catalog](https://pnpm.io/catalogs) for most cases (as we do in a few places already), and opt out of using catalogs for the few apps that require a particular version of react.

boorad (approved):
LGTM

PR #2602, opened 2025-07-02
feat: refactor Peer communication and schedule incoming messages on sync

gdorsi:
# Description

We have noticed that under load our sync server tends to accumulate incoming messages and stops responding until a big chunk of them is processed.

This leads to low throughput and to make the service look like not operative.

In this PR I'm refactoring the internal message scheduling in order to migrate all the async code in the hot-paths into something that we fully control:
- Changed the incoming messages scheduling to use a single queue for all peers, that schedules processing of messages doing a round-robin over the peers
- Refactored the peer communication to be callback-based, switching to a sync style of communication. This makes possible to complete the entire processing of an incoming message syncronously on the sync server, making it easy to do collaborative scheduling on the incoming messages
- Implemented collaborative scheduling on the incoming messages. When processing exceeds 50ms we pause the processing using `setTimeout`, giving to the event loop the space to handle the rest of the tasks

While testing these changes I've also observerd some preformance improvements in both the sync server processing and the browser loading time.

The change on the peer communication has also affected the testing API of jazz-tools, now everything happen in sync so the multi-user tests should become easier to write.

I've created a stress-test application that I've used to check the performance and the behavoir of the sync server under load.

It's inside `tests/stress-test`:
- `pnpm build:packages # ensure that the packages dist are aligned with the code` 
- `cd tests/stress-test`
- `pnpm dev`

It spawns both a vite dev build and a local sync server.
`chrome://inspect/#devices` can be used to connect the Chrome dev tools to the sync server and do profiling.

## Tests

## Checklist


packages/cojson-transport-ws/src/BatchedOutgoingMessages.ts
```diff
@@ -1,48 +1,152 @@
...
+  waitForWebSocketBufferedAmount,
+  waitForWebSocketOpen,
+} from "./utils.js";
+
+const { CO_VALUE_PRIORITY } = cojsonInternals;
 
 export const MAX_OUTGOING_MESSAGES_CHUNK_BYTES = 25_000;
 
-export class BatchedOutgoingMessages {
+export class BatchedOutgoingMessages
```
gdorsi, line 22: Moved the PriorityBasedMessageQueue inside this class, so now outgoing messages prioritization is done here.

Now we can fully control the outgoing messages scheduling from this class, and I've changed to scheduling from debouncing to a simple delay of 10ms before start the messages processing.

This makes the batching and prioritization more effective as subsequent operations are more likely to be synced and reduces its overhead because we don't do anymore setTimeout/clearTimeout on each outgoing.push.

packages/cojson-transport-ws/src/createWebSocketPeer.ts
```diff
@@ -166,18 +88,19 @@ export function createWebSocketPeer({
     expectPings,
     pingTimeout,
     () => {
-      incoming
-        .push("PingTimeout")
-        .catch((e) =>
-          logger.error("Error while pushing ping timeout", { err: e }),
-        );
+      incoming.push("Disconnected");
```
gdorsi, line 91: Removed the PingTimeout event, so now we only have "Disconnected" which is more generic and works for every situation.

This change simplified the Peer code in general, as we need to check only for a single exception.

packages/cojson-transport-ws/src/createWebSocketPeer.ts
```diff
@@ -206,50 +129,42 @@ export function createWebSocketPeer({
...
-              websocket.close();
-            },
-            { once: true },
-          );
-        } else if (websocket.readyState === 1) {
-          websocket.close();
-        }
-      },
-    },
+    outgoing,
```
gdorsi, line 166: BatchedOutgoingMessages implements the OutgoingPeerChannel interface and can be used directly as outgoing channel.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -236,12 +236,41 @@ export class CoValueCore {
   }
 
   missingDependencies = new Set<RawCoID>();
+
+  // Checks if the current CoValueCore is already a missing dependency of the given CoValueCore
+  checkCircularDependencies(dependency: CoValueCore) {
```
gdorsi, line 225: The changes I've done in the tests have highlighted that the previous fix on the circular deps didn't catch all the possible cases of missing circular deps.

So I've implemented a more strict algorithm to mark the missing dependencies, by checking before if the current CoValue is already marked as missing dependency of the dependency CoValue.

packages/cojson/src/storage/StoreQueue.ts
```diff
@@ -0,0 +1,56 @@
+import { LinkedList } from "../PriorityBasedMessageQueue.js";
+import { logger } from "../logger.js";
+import { CoValueKnownState, NewContentMessage } from "../sync.js";
+
+type StoreQueueEntry = {
+  data: NewContentMessage[];
+  correctionCallback: (data: CoValueKnownState) => void;
+};
+
+export class StoreQueue {
```
gdorsi, line 10: Just extracted this logic from storageAsync an added some tests on it

packages/cojson/src/streamUtils.ts
```diff
@@ -1,47 +1,89 @@
...
+  const channel = new ConnectedPeerChannel();
 
-  return [channel.wrap(), channel];
+  return [channel, channel];
+}
+
+export class ConnectedPeerChannel
+  implements IncomingPeerChannel, OutgoingPeerChannel
+{
+  buffer: (SyncMessage | DisconnectedError)[] = [];
```
gdorsi, line 50: Changed the Peer channels to use a callback-based communication instead of relying on async iterators.

Used a buffer to collect all the messages pushed before the connection with the syncManager.

packages/cojson-transport-ws/src/BatchedOutgoingMessages.ts
```diff
@@ -1,48 +1,156 @@
...
+  }
+
+  private async processQueue() {
+    const { websocket } = this;
+
+    this.processing = true;
+
+    // Delay the initiation of the queue processing to accumulate messages
+    // before sending them, in order to do prioritization and batching
+    await new Promise<void>((resolve) => setTimeout(resolve, 10));
```
aeplay: I'm dubious of this: doesn't it always add 10ms delay?

Do we get "enough" batching if we wait for 1ms? Can we skip batching if the queue is "deep enough"?

packages/cojson/src/IncomingMessagesQueue.ts
```diff
@@ -0,0 +1,138 @@
...
+        callback(msg, peer);
+      } catch (err) {
+        logger.error("Error processing message", { err });
+      }
+
+      const currentTimer = performance.now();
+
+      // We check if we have blocked the main thread for too long
+      // and if so, we schedule a timer task to yield to the event loop
+      if (currentTimer - lastTimer > 50) {
```
aeplay: Doesn't this mean that under load, the response to the first message will only be sent out after at least 50ms? Can we try yielding more often and seeing how that affects performance? Maybe make it an env var so we can experiment in prod?

packages/cojson-transport-ws/src/BatchedOutgoingMessages.ts
gdorsi: > I'm dubious of this: doesn't it always add 10ms delay?

Only for the first message of the batch.
Choosen 10ms because I thought it as a negligible delay for a batch.

> Do we get "enough" batching if we wait for 1ms?

We get a smaller window, but still effective when handling stress situations.

>  Can we skip batching if the queue is "deep enough"?

The queue start growing after this line. Without this await the queue can't be longer than 1 item.

EDIT: It can actually grow when the outgoing buffer is full or when the WebSocket is connecting

packages/cojson/src/IncomingMessagesQueue.ts
gdorsi: I can setup a config that can be modified from the outside.

>  the response to the first message will only be sent out after at least 50ms?

We could remove the await from the outgoing messages and the response would be scheduled immediately when the buffer is free.

> Can we try yielding more often and seeing how that affects performance? 

In my stress tests 50ms looked a good compromise, but can do some more tests to see if things are changing when lowering the budget.

aeplay (approved):

Issue #2404, opened 2025-06-02
Zod: define a better boundary between zod and Jazz schemas


gdorsi:
- We can apply to Co schema Only co. modifiers (e.g. co.optional)
- CoValue schemas should not work with Zod modifiers (should flag an error if possible)

We don't need CoValue schemas to be valid Zod schemas, it would be nice in the future to have some `toZodSchema` APIs to play well with form libraries and the Zod ecosystem.

PR #2607, opened 2025-07-04
Add TanStack Form example

nrainhart:
### What this Does
Add a [TanStack Form](https://tanstack.com/form/latest) example to our examples library. This example is based off our existing [form](https://github.com/garden-co/jazz/tree/main/examples/form) example, but shows how to integrate with form libraries (which are usually not very friendly with non-JSON objects, such as CoValues) and leverage TanStack Form's support for schema validation libraries to validate the form using Jazz's Zod-based schemas.

https://github.com/user-attachments/assets/f883d696-7c0d-4296-872b-f88ea7a92f31

### Why Are We Doing This?
This started off as me researching our form-building experience with Jazz, but after discussing it with @gdorsi we thought this could make sense as an example.

### Open Questions
I initially wanted to make this an Svelte project (in fact, I've got an early prototype [here](https://github.com/garden-co/jazz/tree/feat/form-with-save-button-tanstack-form/examples/tanstack-form)) but ended up spending more time playing with Zod validations (instead of working through some styling quirks product of the Svelte migration) and thus kept the example in React. If you think it'd be more valuable to make this an Svelte example, let me know and I can spend a bit more time working on it!

### Testing Instructions
Added an e2e example to confirm the basic create & edit functionality works as expected.

### Related Links
- Linear pitch: https://linear.app/garden-co/project/research-our-form-building-experience-2209b5f5212b/overview


packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
```diff
@@ -146,6 +146,8 @@ export type CoMapSchema<
     ): CoMapSchema<Shape, Config, Owner>;
 
     getCoSchema: () => typeof CoMap;
+
+    getZodSchema: () => z.ZodObject<Shape, Config>;
```
nrainhart, line 150: I think it'd make sense to expose a CoValueSchema's underlying Zod schema, specially considering we want to define a better boundary between Zod and Jazz schemas ([link](https://linear.app/garden-co/project/zod-define-a-better-boundary-between-zod-and-jazz-schemas-22b19d1b8000/overview)).

This lets users integrate with other tools that use Zod schemas (in this case, TanStack Form). Otherwise, we're forced to re-define the schema for form validations.

I didn't spend too much time on this because I'm interested in folks' feedback: do you think this is something it makes sense to add now? If so, I can first open a PR modifying [all of our schema types](https://github.com/garden-co/jazz/blob/feat/tanstack-form-example/packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes) and then rebase this PR on top

examples/tanstack-form/src/OrderForm.tsx
```diff
@@ -0,0 +1,210 @@
...
+    error: "Please select your preferred base tea.",
+  }),
+  deliveryDate: z.date("Plese select a delivery date."),
+  // TanStack Form doesn't support CoList fields, so we need to convert them to arrays
+  addOns: z
+    .array(z.enum(BubbleTeaAddOnTypes))
+    .min(1, "Please select at least one add-on."),
+  // TanStack Form doesn't support CoPlainText fields, so we need to convert them to strings
+  instructions: z.string().optional(),
+});
```
gdorsi, line 24: This one is interesting!
Let's keep this use-case in mind when working on the next itereations of our schemas

Also, we could use extend to do the operation in a more Zod-like way:

```ts

// TanStack Form can leverage Jazz's Zod schema to validate the form
const baseSchema = BubbleTeaOrder.getZodSchema();
const orderFormSchema = baseSchema.extend({
  baseTea: baseSchema.shape.baseTea.refine((value) => value, {
    error: "Please select your preferred base tea.",
  }),
  deliveryDate: baseSchema.shape.deliveryDate.refine((value) => value, {
    error: "Plese select a delivery date.",
  }),
  // TanStack Form doesn't support CoList fields, so we need to convert them to arrays
  addOns: z
    .array(z.enum(BubbleTeaAddOnTypes))
    .min(1, "Please select at least one add-on."),
  // TanStack Form doesn't support CoPlainText fields, so we need to convert them to strings
  instructions: z.string().optional(),
});
```

On the comments, maybe `getZodSchema` could do the CoSchema -> Zod conversion internally for all the defined fields.

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
gdorsi: I think we should add this API as part of https://github.com/garden-co/jazz/issues/2404 

After the splitting we are going to need this API more, and the implementation will also probably differ after those changes.

gdorsi (approved):

.github/workflows/playwright.yml
```diff
@@ -30,6 +30,7 @@ jobs:
           "examples/inspector",
           "examples/music-player",
           "examples/organization",
+          "examples/tanstack-form",
```
gdorsi, line 33: Added the test to the CI, but it looks like this doesn't pass

examples/tanstack-form/src/OrderForm.tsx
nrainhart: > maybe getZodSchema could do the CoSchema -> Zod conversion internally for all the defined fields

Yeah! I was thinking the same

nrainhart: > Also, we could use extend to do the operation in a more Zod-like way

I actually used `extend` first, but it looks like Zod now recommends using destructuring (see [their docs](https://zod.dev/api?id=extend))

gdorsi: Ok, convinced me at "It's more tsc-efficient"

Thanks for pointing it out!

.github/workflows/playwright.yml
nrainhart: Fixed!

nrainhart:
@gdorsi without `CoValueSchema.getZodSchema` this example is quite verbose and not something I'd like to showcase. I'd prefer to wait until we add that API (including the CoSchema -> Zod conversion for all the defined fields) and then adjust this example and merge it. If you think we should merge this sooner, lmk

PR #2609, opened 2025-07-05
User age calc fix

jeffgca:
### What this Does

In the svelte example, the function getUserAge calculates the user's age by only comparing years, meaning the answer will be wrong for most of the year for people born late in the year. This new implementation should be more accurate.

### Scope / Boundaries
Includes:
- [x] Core feature functionality

### Testing Instructions

Create a svelte app, then enter your birthday as Dec 31, 1995. The app will say you are 30 years old when in fact you would sitll be 29.


gdorsi (approved):
Thanks!

PR #2612, opened 2025-07-07
perf(ci): batch the e2e tests execution in 2 workflow runs

gdorsi:
Currently we execute each e2e test in a separate workflow, but now we have 13 workflows and it's starting to become too much.

To optimize the execution, and reuse the installation/build across the different e2e tests, I've changed the strategy to run more than one tests in a single run and use the [same sharding approach as vitest](https://vitest.dev/guide/cli.html#shard) to split the execution into multiple workflows


.github/workflows/playwright-homepage.yml
```diff
@@ -1,52 +0,0 @@
-name: Playwright Tests - Homepage
```
gdorsi, line 1: This already runs on Vercel

PR #2617, opened 2025-07-07
Implement current session logic in Rust and use as WASM

aeplay:
# Description

Moves the session management logic (sign & encrypt transactions) in the CryptoProvider, to reduce the communication overhead between WASM/Native and Javascript.

This PR is part of a bigger migration, where we move more logic into Rust for performance & memory management reasons.

TODO:
- [x] Prune the unused Crypto APIs (realized that leads to too many edits, to do in a follow-up PR)
- [x] Move jazz-crypto-rs here, to have a single WASM instance
- [x] Add more dedicated unit tests on the CryptoProvider implementations
- [x] Ensure that RNQuickCrypto continues to work

Load any of the examples and check that they work!

## Tests

## Checklist


gdorsi (approved):
Let's gooo!

PR #2621, opened 2025-07-08
fix: fixes InvalidSignature errors that could happen during streaming

gdorsi:
# Description

Found that when stale content is reiceved on storage we still update the lastSignature, leading to overriding the valid lastSignature with one that is referred to an old transaction.

Reproduced the InvalidSignature bug using `tests/stress-test` app, restarting the sync server while having a page with 10k tasks open.

In practice, we may still send non-required content when both the client and the sync-server are loading a coValue in streaming.

This can cause one of the two nodes to send some content blocks, which leads to updating the lastSignature with one of the chunk signatures.

In production this bug can be reproduced in a similar way.

## Tests

## Checklist

### DIFF
--- a/packages/cojson/src/storage/storageAsync.ts
+++ b/packages/cojson/src/storage/storageAsync.ts
@@ -298,6 +298,10 @@ export class StorageApiAsync implements StorageAPI {
 
     const actuallyNewTransactions = newTransactions.slice(actuallyNewOffset);
 
+    if (actuallyNewTransactions.length === 0) {
+      return sessionRow?.lastIdx || 0;
+    }
+
     let newBytesSinceLastSignature =
       (sessionRow?.bytesSinceLastSignature || 0) +
       actuallyNewTransactions.reduce(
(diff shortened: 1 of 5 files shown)


packages/cojson/src/storage/storageAsync.ts
```diff
@@ -298,6 +298,10 @@ export class StorageApiAsync implements StorageAPI {
 
     const actuallyNewTransactions = newTransactions.slice(actuallyNewOffset);
 
+    if (actuallyNewTransactions.length === 0) {
+      return sessionRow?.lastIdx || 0;
+    }
```
gdorsi, line 303: This is the actual fix

packages/cojson/src/storage/syncUtils.ts
```diff
@@ -23,17 +23,17 @@ export function collectNewTxs({
...
     contentMessage.new[sessionRow.sessionID] = sessionEntry;
+  } else {
+    sessionEntry.lastSignature = signature;
   }
 
   for (const tx of newTxsInSession) {
     sessionEntry.newTransactions.push(tx.tx);
   }
-
-  sessionEntry.lastSignature = signature;
```
gdorsi, line 36: This is a small refactoring to make the code cleaner

aeplay (approved):

PR #2623, opened 2025-07-09
Feat/design system shadcn integration

Sammii-HK:
# Description
clear up buttons docs
add height consistency across components, also setting height for buttons with lower and upper case buttons

## Tests

## Checklist


Elfo404 (approved):
🚀

Issue #2303, opened 2025-05-21
Support incognito mode for `createImage`


bensleveritt:
`jazz-browser-media-images` uses `pica` under the hood to resize the images we store for ImageDefinitions. When in an incognito `pica` throws due to canvas fingerprinting restrictions.

We need to think of a solution that works in incognito mode (as the developer has no idea where their apps will be used).

Also window-less environments (server-side, tests).

bensleveritt:
https://github.com/corbindavenport/photostack/issues/138 has been suggested as a 'fix', but I don't like the idea of patching `pica` for this

tobiaslins:
@bensleveritt what if we just move to a self-rolled image resizer :)
I've just vibe-coded one in a minute - also works in incognito
https://kzmq8p4ikbcp8l5tg66r.lite.vusercontent.net/

joeinnes:
Comparison:

| Solution | Bundle Size Δ | Performance (Speed) | Incognito Mode | Non-browser | Concerns |
| --- | --- | --- | --- | --- | --- |
| Do nothing | 0kB (baseline: 35.1kB) | 1.0&times; | ❌ | ❌ | This won't fix the issue, or add support for non-browser environments |
| Patch pica | 0kB | 1.0&times;[^1] | ✅ | ❌ | Seems like a maintenance nightmare, and the patch itself looks to be pretty fragile in case of browser changes in anti-fingerprinting behaviour |
| image-js | +355.5 kB | 0.93&times; | ✅ | ❌ | Couldn't get this working in non-browser, plus it's a heavy dependency |
| @tobiaslins's vibe-coded 'modern-resize' | -31.3 kB | 1.59&times; |  ✅ | ❌ | Very lightweight, but relies on canvas, and not sure whether it would have weird edge-case quirks which have already been addressed in more battle-tested libraries |
| @silvia-odwyer/photon | -6.5 kB | TBD |  ✅ | ❌ |  I ended up giving up on this. I couldn't get it to load properly, but at any rate, although the JS would be smaller than pica, there's a 2MB (😱) WASM you need to run this anyway |
| pica-gpu | -28.4 kB | 1.86&times; |  ✅ | ❌ | Upscaling was... not good. Also due to differences between 2D and WebGL Canvas contexts, you have to fiddle around with flipping images. Requires a (relatively) modern browser. Fast and lightweight. Not well battle tested, and can't handle non-browser environments, but so far I like this suggestion best. |

[...]

PR #2624, opened 2025-07-09
New Image management API

Takeno:
# Description

This PR introduces a refactoring of Image management on creation and consumption.

### 1. New `jazz-tools/media` package - BC

All imperative functions have been moved into a single package, providing a unified entry point for all supported platforms.
The `createImage()` function has a new signature environment-agnostic, and the environment-specific implementations are exposed in the package.json with keys `react-native`, `node`, and `default` (browser).
New utility functions are now exposed from the same package, providing users with a way to consume images in a vanilla or imperative way.

**Breaking Changes:** the `createImage` has been moved from `jazz-tools/browser-media-image` and `jazz-tools/react-native-images` to `jazz-tools/media`.

### 2. Refactoring `createImage` - BC
The new options enable users to opt into built-in features: placeholder generation, progressive loading preparation, and max-size resizes.
The internal keys for resizing (`${width}x${height}`) have been designed to maintain backward compatibility and are ready for future integrations. 
The new function `createImageFactory` is now exposed, allowing users to make their custom implementation using image manipulation libraries.

**Breaking Changes:** A few keys of the options have been changed. 

### 3. New React, React-Native, Svelte components - BC

A new `<Image>` component has been introduced for the three supported frameworks. This allows users an easy way to use the saved images and enjoy the opt-in features out of the box. The three components are exported from their respective packages.
```ts
import {Image} from 'jazz-tools/react';
import {Image} from 'jazz-tools/react-native';
import {Image} from 'jazz-tools/svelte';
```

**Breaking Changes:** The old `ProgressiveImg` component and hook have been dropped.

### 4. `pica` replaced for browser resizing
The [`pica`](https://github.com/nodeca/pica) library was introduced for better resizing outcomes. The library doesn't work in Incognito mode (#2303, a known bug that the author won't fix), and several more bugs are pending (the last commit was 4 years ago).

Other pure JavaScript implementations would introduce a considerable increment in bundle size,  and `wasm` solutions create more friction on setup in final projects.

The chosen solution is to utilize trivial resizing via canvas as a built-in implementation, and to use `createImageFactory` to enable users to leverage custom image libraries tailored to their specific use cases.

**Breaking Changes:** No BC.

### 5. New progressive loading strategy

A new algorithm has been introduced and tested to reduce the in-memory loading of file streams for progressive loading. Now, the first already loaded image is returned (if present), and the best-fit image load starts.

**Breaking Changes:** No BC

### 6. React Native performance - Out of scope

We're watching https://github.com/mrousavy/react-native-nitro-image library as a future replacement for resizing and Image visualization. The library is still very young, but promising for very fast performance. As more tests must be conducted, they will come in a new dedicated PR.
Additionally, it [requires](https://github.com/mrousavy/react-native-nitro-image?tab=readme-ov-file#installation) the new architecture. Even if it is stable and enabled by default, we might wait a couple of new releases before replace it.

With the new API, the future replacement of the internal library for resizing and image visualization should not introduce breaking changes.

## Open topics
- [ ] Confirm the default options for `createImage`: placeholder, progressive
- [ ] Double-check the progressive loading strategy in `highestResAvailable`

# Manual testing instructions

Use `createImage` from the `jazz-tools/media` package
Use `<Image>` components from framework-related packages

# Tests

Tests on core parts have been added and/or updated.
No tests are in place for React Native and Svelte implementations.

# Additional notes

- All changes have been squashed into separate commits to facilitate review. Check the commits' names and the file diff from them. From now on, requested changes will be pushed in separate new commits to help facilitate the conversation. 
- Some console logs for debugs and re-renders checks will be removed after the reviews

# Checklist

- [ ] Remove the console.log


packages/jazz-tools/src/media/create-image.ts
```diff
@@ -0,0 +1,141 @@
...
+async function createImage(
+  imageBlobOrFile: SourceType,
+  options: CreateImageOptions,
+  impl: CreateImageImpl,
+): Promise<ImageDefinition> {
+  // Get the original size of the image
+  const { width: originalWidth, height: originalHeight } =
+    await impl.getImageSize(imageBlobOrFile);
+
+  const def: any = {
```
aeplay: `any` is a bit lazy

packages/jazz-tools/src/media/index.browser.ts
```diff
@@ -0,0 +1,144 @@
...
+
+// using createImageBitmap is ~10x slower than Image object
+// Image object: 640 milliseconds
+// createImageBitmap: 8128 milliseconds
+function getImageFromBlob(blob: Blob): Promise<HTMLImageElement> {
+  return new Promise((resolve, reject) => {
+    const img = new Image();
+    img.onload = () => {
+      resolve(img);
+      URL.revokeObjectURL(img.src);
```
aeplay, line 42: interesting - does revoking it this early not cause problems down the line? If not, an elegant solution to avoid leaking object URLs

packages/jazz-tools/src/media/index.browser.ts
```diff
@@ -0,0 +1,144 @@
...
+  ctx.drawImage(image, 0, 0, width, height);
+
+  return new Promise<Blob>((resolve, reject) => {
+    canvas.toBlob((blob) => {
+      if (!blob) {
+        reject(new Error("Failed to convert canvas to blob"));
+        return;
+      }
+      resolve(blob);
+    }, "image/png");
```
aeplay: we should probably use the same mime type as the input blob/file otherwise complex images might end up bigger than the original!

particularly making sure that for jpgs we keep jpg. For jpgs maybe even set a quality < 1?

seems supported everywhere: https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob#browser_compatibility

In the future we might resize to webp where possible...

packages/jazz-tools/src/media/index.native.ts
```diff
@@ -0,0 +1,138 @@
...
+    throw new Error(
+      "ImageResizer is not installed, please run `npm install @bam.tech/react-native-image-resizer`",
+    );
+  }
+
+  const { uri } = await ImageResizer.createResizedImage(
+    filePath,
+    8,
+    8,
+    "JPEG",
```
aeplay: Same here, let's keep the same mime type if possible

packages/jazz-tools/src/media/index.native.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  const blob = await fetch(filePath).then((res) => res.blob());
+  const arrayBuffer = await toArrayBuffer(blob);
+
+  return FileStream.createFromArrayBuffer(arrayBuffer, blob.type, undefined, {
+    owner,
+  });
+}
+
+// TODO: look for more efficient way to do this
```
aeplay: maybe add a note that https://developer.mozilla.org/en-US/docs/Web/API/Blob/arrayBuffer isn't available in React Native yet

homepage/homepage/content/docs/using-covalues/imagedef/react.mdx
```diff
@@ -1,26 +1,25 @@
 import { CodeGroup } from "@/components/forMdx";
 
 export const metadata = {
-  description: "ImageDefinition is a CoValue for managing images, storage of multiple resolutions, and progressive loading."
+  description: "ImageDefinition is a CoValue for store images with bult-in UX features."
```
aeplay:
```suggestion
  description: "ImageDefinition is a CoValue for storing images with built-in UX features."
```

homepage/homepage/content/docs/using-covalues/imagedef/react.mdx
```diff
@@ -1,26 +1,25 @@
...
 };
 
 # ImageDefinition
 
-`ImageDefinition` is a specialized CoValue designed specifically for managing images in Jazz applications. It extends beyond basic file storage by supporting multiple resolutions of the same image and progressive loading patterns.
+`ImageDefinition` is a specialized CoValue designed specifically for managing images in Jazz applications. It extends beyond basic file storage by supporting blurry placeholder, a built-in max size resizing, and progressive loading patterns.
```
aeplay:
```suggestion
`ImageDefinition` is a specialized CoValue designed specifically for managing images in Jazz applications. It extends beyond basic file storage by supporting a blurry placeholder, built-in resizing, and progressive loading patterns.
```

homepage/homepage/content/docs/using-covalues/imagedef/react.mdx
```diff
@@ -78,194 +82,229 @@ The `createImage()` function:
...
+
+console.log({
+  width: image.width,
+  height: image.height,
+  image: image.image,
+});
+```
+</CodeGroup>
+
+If the image was generated with progressive loading, and you want to access the best resolution, `loadImageBySize` function is here the utility function you need. It will load the image of the best resolution that fits the wanted width and height.
```
aeplay:
```suggestion
If the image was generated with progressive loading, and you want to access the best-fit resolution, use `loadImageBySize`. It will load the image of the best resolution that fits the wanted width and height.
```

homepage/homepage/content/docs/using-covalues/imagedef/react.mdx
```diff
@@ -78,194 +82,229 @@ The `createImage()` function:
...
+console.log({
+  width: image.width,
+  height: image.height,
+  image: image.image,
+});
+```
+</CodeGroup>
+
+
+But you know, the CoValue are alive and they will change over time. If want to listen to the changes and retrieve the best resolution _loaded_ that fits the wanted width and height, you can use the `subscribe` and the `highestResAvailable` function.
```
aeplay:
```suggestion
If want to dynamically listen to the _loaded_ resolution that best fits the wanted width and height, you can use the `subscribe` and the `highestResAvailable` function.
```

homepage/homepage/content/docs/using-covalues/imagedef/react.mdx
```diff
@@ -78,194 +82,229 @@ The `createImage()` function:
...
+    if(blob) {
+      const url = URL.createObjectURL(blob);
+      // ...
+    }
+  }
 });
+```
+</CodeGroup>
+
+
```
aeplay, line 318: Maybe add a section here about how to create "resized *copies*" of an image?

something like `createImage(await existing.toBlob(), { maxSize: 500 })` (pseudo-code)

aeplay (changes requested):
Thank you for implementing this big feature, the extensive testing and docs!

I just have a couple small asks, see comments below.

For the docs, I only commented on React, please apply to other envs as necessary.

packages/jazz-tools/src/media/index.browser.ts
Takeno: Yes, it doesn't cause any problem. Once the image is loaded, the browser has already decoded it, and the URL is no longer used. For our use case, it is perfect. In case you revoke the URL on the `` tag, this will prevent users from interacting with the image, such as Save As or Open in a new tab.

Additional [Evidence](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob#getting_a_file_representing_the_canvas): 
> Note that we don't immediately revoke the object URL after the image has loaded, because doing so would make the image unusable for user interactions (such as right-clicking to save the image or opening it in a new tab).

packages/jazz-tools/src/media/index.browser.ts
Takeno: You're right, I forgot to keep the mime type. Updated in 8aad598a9

homepage/homepage/content/docs/using-covalues/imagedef/react.mdx
Takeno: Added in 1fbc09a57

packages/jazz-tools/src/media/create-image.ts
Takeno: Creating a partial object without any constraints was a little bit tricky due to the catchall on the type. That's why I used `any` and later use `.parse`:
```ts
  const imageCoValue = ImageDefinition.create(
    ImageDefinition.parse(def),
    options?.owner,
  );
```

Any combination I used such as 
```ts
type PartialImageDefinition = {
  originalSize: [number, number];
  progressive: boolean;
  placeholderDataURL: string | undefined;
} & {
  [key: string]: FileStream | undefined;
}
```
TS expects `originalSize` to be `FileStream | undefined`.

packages/jazz-tools/src/media/index.native.ts
Takeno: Updated in 3158b74

packages/jazz-tools/src/media/create-image.ts
```diff
@@ -0,0 +1,160 @@
...
+} from "jazz-tools";
+
+export type SourceType = Blob | File | string;
+
+export type CreateImageOptions = {
+  owner?: Group | Account;
+  placeholder?: "blur" | false; // default "blur"
+  maxSize?: number; // | [number, number];
+  progressive?: boolean;
+};
```
gdorsi, line 36: **suggestion:** it would be useful to have inlne docs for these options, so users and AIs can understand which ones enable by reading them

packages/jazz-tools/src/react/tests/media/image.test.tsx
```diff
@@ -0,0 +1,519 @@
...
+      );
+
+      const { container } = render(<Image imageId={im.id} alt="test" />, {
+        account,
+      });
+
+      const img = container.querySelector("img");
+      expect(img).toBeDefined();
+      expect(img!.getAttribute("width")).toBe(null);
+      expect(img!.getAttribute("height")).toBe(null);
```
gdorsi, line 52: Why should these be `null` 

The originalSize is available, so these attributes should be defined

packages/jazz-tools/src/react-native/media/image.tsx
```diff
@@ -0,0 +1,107 @@
...
+      };
+    }, [image?.originalSize, width, height]);
+
+  useEffect(() => {
+    if (!image) return;
+
+    let lastBestImage: FileStream | string | undefined =
+      image.placeholderDataURL;
+
+    const unsub = image.subscribe({}, (update) => {
```
gdorsi: I think it would be better to not have the image as dependency and fully subscribe to the image internally.

```suggestion
    const unsub = ImageDefinition.subscribe(imageId, {}, (update) => {
```

This way the dependencies are more stable and the component re-renderings more predictable.

packages/jazz-tools/src/react/media/image.tsx
```diff
@@ -0,0 +1,112 @@
...
+  useEffect(
+    () => () => {
+      revokeObjectURL(src);
+    },
+    [src],
+  );
+
+  if (!image) {
+    return null;
+  }
```
gdorsi: We should return an image when the dev provides both dimensions to reduce the risk of layout shifts.

Would also help when doing SSR on the React web side.

packages/jazz-tools/src/svelte/media/image.svelte
```diff
@@ -0,0 +1,78 @@
...
+});
+
+$effect(() => {
+  const image = imageState.current;
+  if (!image) return;
+
+  src = image.placeholderDataURL;
+  let lastBestImage: FileStream | null = null;
+
+  const unsub = image.subscribe({}, (update) => {
```
gdorsi: Same as in React.

I think it would be better to subscribe from the id for a more stable reactivity on the effect

packages/jazz-tools/src/react/tests/media/image.test.tsx
Takeno: It might break the layout. If the user didn't ask for the original sizes explicitly, then we preserve the browser's behaviour without any attribute. That's why I introduced the literal "original" for width/height.

PR #2627, opened 2025-07-09
feat: add `co.optional` and `co.discriminatedUnion`

nrainhart:
# Description

Add `co.` equivalents of `z.optional` and `z.discriminatedUnion` to be used with CoValue schemas. The motivation for having these functions is to prevent nesting CoValue schemas inside Zod functions:
```typescript
const Dog = co.map({
  children: z.optional(Person),
})
```
We want to disallow this pattern because it doesn't work in many cases (e.g. it doesn't make sense to nest a coMap schema inside of a zod object schema).

For now the new functions are just thin wrappers around the Zod functions (with improved type safety), but eventually they'll be handled differently internally.

## Open questions

I left a question in [this doc](https://www.notion.so/gcmp/Proposal-Quick-changes-for-a-type-safe-Schema-API-22b3ad00c3958092a193ed8bc75f6861?source=copy_link#22b3ad00c395806eadf0c9faf4e75563) regarding having both `z.` and `co.` counterparts of `optional` and `discriminatedUnion`. It'd be great if folks can leave your opinions in that topic when reviewing this PR.

## Tests

## Checklist

- [ ] I've updated the jsDoc comments to the public APIs I've modified, or added them when missing (waiting until we confirm we want `co.optional` and `co.discriminated` instead of just two functions that work for zod and covalue schemas)


packages/cursor-docs/.cursor/docs/llms-full.md
```diff
@@ -2136,7 +2136,7 @@ See the corresponding sections for [creating](/docs/using-covalues/filestreams#c
 
 ### Unions of CoMaps (declaration)
 
-You can declare unions of CoMaps that have discriminating fields, using `z.discriminatedUnion()`.
+You can declare unions of CoMaps that have discriminating fields, using `co.discriminatedUnion()`.
```
nrainhart, line 2139: Not sure if this doc is auto-generated

packages/jazz-tools/package.json
```diff
@@ -164,6 +164,7 @@
     "zod": "3.25.28"
   },
   "scripts": {
+    "check": "tsc --noEmit",
```
nrainhart, line 166: I found it useful to have this command to quickly validate type errors. If folks don't like it, I can remove it

packages/jazz-tools/src/tools/implementation/zodSchema/zodCo.ts
```diff
@@ -390,3 +390,21 @@ export const coRichTextDefiner = (): RichTextSchema => {
...
+export const coDiscriminatedUnionDefiner = <
+  T extends readonly [
+    AnyCoSchema & z.core.$ZodTypeDiscriminable,
+    ...(AnyCoSchema & z.core.$ZodTypeDiscriminable)[],
+  ],
+>(
+  discriminator: string,
+  schemas: T,
+): z.ZodDiscriminatedUnion<T> => {
+  return z.discriminatedUnion(discriminator, schemas as any);
```
gdorsi: **suggestion:** how about we add the load/subscribe methods to this schema?

Could be implemented this way: https://discord.com/channels/1139617727565271160/1139621689882321009/1390736173478969376

We can avoid passing the resolve

gdorsi (approved):
LGTM

Added a minor suggestion

packages/jazz-tools/src/tools/implementation/zodSchema/zodCo.ts
nrainhart: I'm in the middle of a refactor right now, and I think adding the load/subscribe methods to the discriminated union schema will be cleaner once that's in.

If I see the refactor's taking too long I can backport the load/subscribe methods to this PR

nrainhart: Here's the refactor I was talking about: https://github.com/garden-co/jazz/pull/2636. Working on adding load/subscribe on top of that now

nrainhart: Done! https://github.com/garden-co/jazz/pull/2637

I'll merge https://github.com/garden-co/jazz/pull/2636 and https://github.com/garden-co/jazz/pull/2637 into this branch and then into main

PR #2629, opened 2025-07-09
feat: prevent using Zod functions with CoValue schemas

nrainhart:
# Description

Re-defines Zod functions that receive schemas as arguments to not allow passing CoValue schemas.

Some things worth mentioning:
- `z.object` (and variations like `z.strictObject` and `z.looseObject`) continue to be very lenient with their inputs, as tightening the types prevent having recursive schemas
- `z.intersection` and `z.record` can now be re-exported, given they work only with Zod schemas

@aeplay I know you're not sold on going with this approach, but it was so easy to do that I couldn't resist. We don't need to merge it if we think there's a better alternative.

If we decide to merge this, we need to keep in mind this is a breaking change.

## Tests

## Checklist

- [x] I've updated the part of the docs that are affected the PR changes (already done as part of https://github.com/garden-co/jazz/pull/2627)


packages/jazz-tools/src/tools/implementation/zodSchema/zodReExport.ts
```diff
@@ -41,3 +51,57 @@ export {
...
+}
+
+export function discriminatedUnion<
+  T extends readonly [
+    NonCoZodType & core.$ZodTypeDiscriminable,
+    ...(NonCoZodType & core.$ZodTypeDiscriminable)[],
+  ],
+>(discriminator: string, schemas: T): ZodDiscriminatedUnion<T> {
+  return zodDiscriminatedUnion(discriminator, schemas as any);
+}
```
gdorsi: Better to keep optional and discriminatedUnion out for now.

Not support interop with Co schemas would be a breaking change because it works currently.

packages/jazz-tools/src/tools/implementation/zodSchema/zodReExport.ts
```diff
@@ -41,3 +51,57 @@ export {
...
+export function record<
+  Key extends core.$ZodRecordKey,
+  Value extends NonCoZodType,
+>(
+  keyType: Key,
+  valueType: Value,
+  params?: string | core.$ZodRecordParams,
+): ZodRecord<Key, Value> {
+  return zodRecord(keyType, valueType, params);
+}
```
gdorsi: minor: Could you check if these wrappers are breaking the Zod recursive types?

```ts
const object = z.object({ 
  get child() { return z.optional(record) }
})
const record = z.record(z.string, object)
```

nrainhart: From what I can see, Zod's inference with recursive types does not work for records (even before adding these wrappers). This example works if you manually add the type (both with and without the wrapper):
```typescript
const object = z.object({
  get child(): z.ZodOptional<typeof record> {
    return z.optional(record);
  },
});
const record = z.record(z.string(), object);
```

Buuuut even if you use the type annotation, type inference will break when you nest the recursive zod schema into a covalue, because we don't yet support records and intersections:
```typescript
const CoObject = co.map({
  field: object,
});

CoObject.create({
  root: {
	// Type error: Type '{ field: { children: undefined; }; }' is not assignable to type 'undefined'
    children: {
      field: {
        children: undefined,
      },
    },
  },
});
```

I'll keep `z.record` and `z.intersection` commented out for now, and try to add support for them later in the week.

PR #2637, opened 2025-07-11
feat: add load & subscribe to Discriminated union schemas

nrainhart:
# Description

You can now use load & subscribe in schemas generated with `co.discriminatedUnion`.

Added these methods to the `SchemaUnion` coValue class as well.

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/schemaUnion.ts
+++ b/packages/jazz-tools/src/tools/coValues/schemaUnion.ts
@@ -11,9 +11,9 @@ import {
   Resolved,
   SubscribeListenerOptions,
   SubscribeRestArgs,
-  subscribeToCoValueWithoutMe,
   loadCoValueWithoutMe,
   parseSubscribeRestArgs,
+  subscribeToCoValueWithoutMe,
 } from "../internal.js";
 
 /**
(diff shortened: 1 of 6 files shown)


packages/jazz-tools/src/tools/coValues/schemaUnion.ts
```diff
@@ -99,4 +110,58 @@ export abstract class SchemaUnion extends CoValueBase implements CoValue {
...
+  /**
+   * Load a `SchemaUnion` with a given ID, as a given account.
+   *
+   * @category Subscription & Loading
+   */
+  static load<M extends SchemaUnion, const R extends RefsToResolve<M> = true>(
+    this: CoValueClass<M>,
+    id: ID<M>,
+    options?: {
+      resolve?: RefsToResolveStrict<M, R>;
```
gdorsi: Resolve type doesn't work with SchemaUnion

How about not including it at all?

gdorsi (approved):
Minor comment

packages/jazz-tools/src/tools/coValues/schemaUnion.ts
nrainhart: Sure, I can leave it our for now and we can add it back once https://github.com/garden-co/jazz/issues/2639 is fixed

nrainhart: [`2475a46` (#2637)](https://github.com/garden-co/jazz/pull/2637/commits/2475a465781f44c6c65500ce071694c03384237e) ✅

PR #2641, opened 2025-07-11
Expose current Account id in unauthorized error message

Takeno:
# Description
Previously, when an Account requested a CoValue they couldn't access, the error message displayed was: `The current user is not authorized to access this value from co_valueId on path xyz`.

To help with debugging, especially with server workers, the Account id is shown in the error message like:
`The current user (co_accountId) is not authorized to access this value from co_valueId on path xyz`

## Tests

## Checklist

### DIFF
--- a/.changeset/green-jobs-smash.md
+++ b/.changeset/green-jobs-smash.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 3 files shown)


gdorsi (approved):
LGTM

Could you add a changeset?

PR #2646, opened 2025-07-15
Add community-maintained VueJS bindings packages and examples

nicolasembleton:
# Description

Following the discussion on discord#vue and the desire to have a community-supported set of packages for various bindings that do not have official support. This is the case for this release, which adds back VueJS bindings, updated to the API signature up to jazz-tools@0.15.4, but tested and working up to main@HEAD

Things that are included in this PR:

1. VueJS bindings updated to 0.15.4, matching React APIs 1-to-1, to make it easier to maintain moving forward while trailing React's, and make it easier for newcomers to look at  React code / tutorials and use Vue bindings because the APIs are matching
2. Restores an updated, community- prefixed version of Todo and Chat examples
3. Adds a community- prefixed clerk auth example
4. Adds a Vue port of the Inspector to make it easy to use the inspector in a Vue context
5. Adds a few tests to avoid regressions from VueJS reactivity / proxy behavior which work quite differently from React's

Only community-* packages have been touched in this PR, and the branch is up-to-date with main. 
The API signatures sync up to 0.15.4, there may be more work if there were changes to React bindings afterwards. Therefore I'm marking the bindings to be version 0.15.4. I'll make another PR soon thereafter to catch-up with recent releases.

Simply run the normal tests, they will be included in the suite.

## Tests

## Checklist


packages/community-jazz-vue/src/inspector/JazzInspector.vue
```diff
@@ -0,0 +1,26 @@
+<template>
+  <VueJazzInspector :position="position" />
+</template>
```
gdorsi: We have created the [custom element version](https://jazz.tools/docs/vanilla/inspector#embedding-the-inspector-widget-into-your-app) of the Inspector to be used by all the non-React frameworks.

The goal was to keep maintaining only one inspector while still supporting all the frameworks.

How about moving to a Vue wrapper component around [the custom element](https://jazz.tools/docs/vanilla/inspector#embedding-the-inspector-widget-into-your-app) provided by jazz-tools?

packages/community-jazz-vue/src/tests/useAcceptInvite.test.ts
```diff
@@ -0,0 +1,55 @@
...
+import { describe, expect, it } from "vitest";
+import { createInviteLink, useAcceptInvite } from "../index.js";
+import { createJazzTestAccount, linkAccounts } from "../testing.js";
+import { waitFor, withJazzTestSetup } from "./testUtils.js";
+
+describe("useAcceptInvite", () => {
+  it("should accept the invite", async () => {
+    class TestMap extends CoMap {
+      value = coField.string;
+    }
```
gdorsi: This should be ported to Zod based schemas

```suggestion
    const TestMap = co.map({
      value: z.string()
    })
```

packages/community-jazz-vue/src/tests/useAccount.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+class AccountSchema extends Account {
+  root = coField.ref(AccountRoot);
+
+  migrate() {
+    if (!this._refs.root) {
+      this.root = AccountRoot.create({ value: "123" }, { owner: this });
+    }
+  }
+}
```
gdorsi: This also should be imported to Zod schemas

packages/community-jazz-vue/src/tests/useAccount.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+declare module "../provider" {
+  interface Register {
+    Account: AccountSchema;
+  }
+}
```
gdorsi: This should not be required anymore

packages/community-jazz-vue/src/tests/useCoState.test.ts
```diff
@@ -0,0 +1,179 @@
...
+  await setupJazzTestSync();
+});
+
+cojsonInternals.setCoValueLoadingRetryDelay(300);
+
+describe("useCoState", () => {
+  it("should return the correct value", async () => {
+    class TestMap extends CoMap {
+      content = coField.string;
+    }
```
gdorsi: This also should be imported to Zod schemas

packages/community-jazz-vue/src/auth/useDemoAuth.ts
```diff
@@ -0,0 +1,43 @@
+import { DemoAuth } from "jazz-tools";
+import { computed, markRaw, ref, watch } from "vue";
+import { useAuthSecretStorage, useJazzContext } from "../composables.js";
+import { useIsAuthenticated } from "./useIsAuthenticated.js";
+
+export function useDemoAuth() {
```
gdorsi: We can remove this auth provider.

It's legacy and doesn't make much sense with the current auth system

packages/community-jazz-vue/src/tests/useCoState.test.ts
nicolasembleton: Good catch.

packages/community-jazz-vue/src/inspector/JazzInspector.vue
nicolasembleton: Excellent. Let me take a look at it. Did not realize this existed, but it definitely make sense. Will make the Vue wrapper a lot lighter and require much less maintenance.

examples/community-clerk-vue/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc -b && vite build",
+    "preview": "vite preview",
+    "format-and-lint": "biome check .",
+    "format-and-lint:fix": "biome check . --write",
+    "test:e2e": "playwright test",
+    "test:e2e:ui": "playwright test --ui"
```
mculotta120, line 12: I had some problems running the playwright tests, but they look like they may have been brought over from the original clerk example. Please feel free to remove those, or you can update so they run in this version.

mculotta120 (changes requested):
Just one comment about the playwright tests in the clerk example

packages/community-jazz-vue/src/inspector/JazzInspector.vue
nicolasembleton: I spent some time trying to make a Vue Wrapper work, but the 2 reactive systems were conflicting each other and I couldn't get it to work without recursive reactive calls, so I ended up just removing the Vue inspector in favor for the VanillaJS one, as other bindings do today.

packages/community-jazz-vue/src/tests/useAcceptInvite.test.ts
nicolasembleton: This has been updated across all tests.

packages/community-jazz-vue/src/auth/useDemoAuth.ts
nicolasembleton: This is done. Removed the tests as well.

examples/community-clerk-vue/package.json
nicolasembleton: Right, so it turned out the tests were indeed just copied and not adapted from it. I spent some time getting the Clerk to behave during the tests, which failed in some cases (the logout button just did not want to work during the expiration test), so I used the Clerk Playwright helpers in these instances. The tests now work.

mculotta120 (approved):

nicolasembleton:
I have merged upstream main and resolved the type issues that prevented the build from happening. 
There were a few tiny bugs / issues that were left and were addressed as well.

PR #2647, opened 2025-07-15
fix: catch errors from CoValue loading and treat it as "unavailable"

Takeno:
# Description
The `CoValue.load` method throws an error if the ID is invalid. The rejected promise was not acknowledged by `CoValueCoreSubscription`, resulting in the load promise never being settled, and an uncaught rejection was logged.

Now subscription listens to node loading rejection, and it treats it as "unavailable". Any high-level covalue loadings now return null if the ID looked for is invalid.

```ts
await Account.load("test") // should return null + console.error

const account = useCoState(Account, "test"); // should return null + console.error
```

## Tests

## Checklist

### DIFF
--- a/.changeset/tender-pears-march.md
+++ b/.changeset/tender-pears-march.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+fix unhandled rejection on CoValue.load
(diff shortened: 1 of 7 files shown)


gdorsi (approved):

gdorsi:
Good job, thanks!

PR #2650, opened 2025-07-16
refactor: CoValue schemas are no longer Zod schemas

### LINKED ISSUE #2404: Zod: define a better boundary between zod and Jazz schemas

nrainhart:
# Description

Closes https://github.com/garden-co/jazz/issues/2404

This PR introduces a clearer boundary between CoValue and Zod schemas: Zod schemas can only be composed with other Zod schemas, and CoValues can (mostly) be composed with both CoValues and Zod schemas.

This PR does not aim to improve performance for inferred loaded types.

These are the most notable changes:
- Upgraded Zodv3 to the latest stable version
- Dropped support for using `z.optional` and `z.discriminatedUnion` with collaborative schemas
  - This greatly simplified CoValue schema creation, as we don't need to recursively process a schema's fields
- CoValue schemas cannot be nested inside Zod schemas. This is enforced at the type-level, except for `z.object` and `z.strictObject`, where we have runtime checks on schema creation 
- CoMap schemas are no longer an instance of `z.$ZodTypeDiscriminable` (and by extension, of `z.$ZodType`), so you cannot call Zod methods like `.partial()` or `.extend()` in them
	- the only exception is `optional()`, which I added to all CoValue schemas
- Fixed bug with deeply nested discriminated unions (https://github.com/garden-co/jazz/pull/2650/commits/86e360e378b86154ff04ee11bebf938e4b1af319)
- Fixed bug with cyclic discriminated union references (https://github.com/garden-co/jazz/pull/2650/commits/e76fe343da6bc65895ec8b0a9fab974e7df67cc5#diff-f6513eaa9fb206d6a0cab3dccf43cdd0587146237735481e52f90171805fa108R173)
- People will need to replace Zod types in cyclic references with their CoValue schema counterparts (e.g. `z.ZodOptional` to `co.Optional`) (https://github.com/garden-co/jazz/pull/2650/commits/bfaed991b441afb1bdbaa419dd749c7fe01bc9ff)
  - I exported all schema types inside the `co.` namespace (e.g. `co.Optional`, `co.List`, etc.)
  - Also removed the non-namespaced CoValue schema exports (https://github.com/garden-co/jazz/pull/2650/commits/fd8afb96a795279e8d38d49ba8957054e9645dd7) 

Also did some general cleanup:
- removed `withHelpers`
- removed the deprecated `createCoValueObservable` function

## Tests

## Checklist


packages/jazz-tools/src/tools/coValues/account.ts
```diff
@@ -268,9 +268,13 @@ export class Account extends CoValueBase implements CoValue {
       inviteSecret,
     );
 
-    return loadCoValue(anySchemaToCoSchema(coValueClass), valueID, {
-      loadAs: this,
-    }) as Resolved<InstanceOrPrimitiveOfSchema<S>, true> | null;
+    return loadCoValue(
+      coValueClassFromCoValueClassOrSchema(coValueClass),
```
gdorsi, line 272: **suggestion:** how about we move this conversion inside `subscribeToCoValue`?

This would reduce the need of calling `coValueClassFromCoValueClassOrSchema` around by a great deal

packages/jazz-tools/src/tools/coValues/account.ts
```diff
@@ -254,7 +254,7 @@ export class Account extends CoValueBase implements CoValue {
     return value._owner.getRoleOf(this.id) === "admin";
   }
 
-  async acceptInvite<S extends CoValueOrZodSchema>(
+  async acceptInvite<S extends CoValueClassOrSchema>(
     valueID: string,
     inviteSecret: InviteSecret,
     coValueClass: S,
```
gdorsi, line 260: **suggestion:** what if we start accepting only "Schema" as input for our public APIs?

This would be another breaking change, but would simplify the types by alot.

packages/jazz-tools/src/tools/coValues/interfaces.ts
```diff
@@ -310,7 +310,7 @@ export function subscribeToCoValue<
  * Going to be removed in the next minor version.
  */
 export function createCoValueObservable<
```
gdorsi: We can remove this one.

packages/jazz-tools/src/tools/implementation/zodSchema/runtimeConverters/coValueSchemaTransformation.ts
```diff
@@ -0,0 +1,155 @@
...
+): schema is CoValueSchemaFromCoreSchema<AnyCoreCoValueSchema> {
+  return isAnyCoValueSchema(schema) && "getCoValueClass" in schema;
+}
+
+/**
+ * Convert a {@link CoreCoValueSchema} into a CoValue schema.
+ *
+ * @returns The CoValue schema matching the provided CoreCoValueSchema
+ */
+export function hydrateCoreCoValueSchema<S extends AnyCoreCoValueSchema>(
```
gdorsi, line 58: Could you document the difference between a CoreCoValueSchema and a CoValueSchemaFromCoreSchema and why we have the distinction?

I think it would be really helpful for other people to gain context more easily on what is happening here.

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/AccountSchema.ts
```diff
@@ -90,7 +120,7 @@ export function enrichAccountSchema<Shape extends BaseAccountShape>(
       // @ts-expect-error
       return coValueClass.subscribe(...args);
     },
-    withHelpers: (helpers: (Self: z.core.$ZodType) => object) => {
+    withHelpers: (helpers: (Self: AnyZodOrCoValueSchema) => object) => {
```
gdorsi: How about we remove withHelpers @aeplay?

packages/jazz-tools/src/tools/tests/coMap.test.ts
```diff
@@ -210,7 +210,7 @@ describe("CoMap", async () => {
       const Person = co.map({
         name: z.string(),
         age: z.number(),
-        get friend(): z.ZodOptional<typeof Person> {
+        get friend(): co.Optional<typeof Person> {
```
gdorsi, line 213: I like this!

How about we make it possible for every CoSchema type?

Like `co.List<typeof Person>`?

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
```diff
@@ -2,160 +2,181 @@ import { CoValueUniqueness } from "cojson";
...
+
+  getCoValueClass: () => typeof CoMap;
+
+  optional(): CoOptionalSchema<CoMapSchema<Shape, CatchAll, Owner>>;
+}
+
+export function createCoreCoMapSchema<
+  Shape extends z.core.$ZodLooseShape,
+  CatchAll extends AnyZodOrCoValueSchema | unknown = unknown,
+>(shape: Shape, index?: CatchAll): CoreCoMapSchema<Shape, CatchAll> {
```
gdorsi: Noticed that reaching the shape via `map.def.shape` is not valid anymore.

I think it was an handy way to reach inner schemas, and would even go far suggesting that we should support `map.shape` to make it more handy.

Any reason to remove it or it has just been part of the type definition simplification?

gdorsi (commented):
Great job!

Since we are introducing some breaking changes, we should also provide an upgrade guide.

In that is also important to mention that we are upgrading our Zod dependency, so if people have zod installed they will need to upgrade to v4 in order to not have it bundled twice.

gdorsi:
Getting this error on `packages/jazz-tools/src/tools/tests/coMap.test-d.ts`

Screenshot 2025-07-18 at 11 51 42

For some reason the test-d tests are not being checked by Vitest now

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
nrainhart: `map.def` went away when CoValue schemas stopped being Zod schemas, but I like the idea of adding `map.shape`. I'll include it in this PR

packages/jazz-tools/src/tools/tests/coMap.test.ts
nrainhart: Should I remove [these type exports](https://github.com/garden-co/jazz/blob/main/packages/jazz-tools/src/tools/exports.ts#L107) then? To avoid having two ways of referencing the same type

gdorsi: Yes, that would work.

Just let's keep it in mind as part of the upgrade guide.

nrainhart:
> Getting this error on packages/jazz-tools/src/tools/tests/coMap.test-d.ts

Fixed the error by converting most CoValue schemas into classes (instead of types/interfaces). Tbh I don't remember where I got the idea that Typescript performs more aggressive caching on class methods' types (maybe you mentioned this to me @gdorsi?), but the evidence appears to show that it works. Plus, I think the code is now more readable as well (because we avoid having the interface separate from the schema object's creation).

I kept only `CoMapSchema` as an interface, because we rely on structural manipulations extensively to derive the Account and Image schemas.

homepage/homepage/content/docs/upgrade/0-16-0.mdx
```diff
@@ -0,0 +1,91 @@
...
+  }
+});
+```
+</CodeGroup>
+
+### Unsupported Zod methods have been removed from CoMap schemas
+
+CoMap schemas no longer incorrectly inherit Zod methods like `.extend()` and `.partial()`. These methods previously appeared to work but could behave unpredictably. They’ve now been disabled.
+
+In their place, .optional() has been added, and more Zod-like methods will be introduced in future releases.
```
gdorsi: We are actually keeping .optional, because it was already available on co.map

homepage/homepage/content/docs/upgrade/0-16-0.mdx
```diff
@@ -0,0 +1,91 @@
+import { CodeGroup } from '@/components/forMdx'
+
+# Jazz 0.16.0 - Cleaner separation between Zod and CoValue schemas
+
+This release introduces a cleaner separation between Zod and CoValue schemas and simplifies how you access internal schemas.
+While most applications won’t require extensive refactors, some breaking changes will require action.
```
gdorsi: I think we should explain our motives for the change.

It makes it easier to spot errors in the schema definition, avoiding possible footguns happening when mixing Zod and co schemas in the wrong way.

And make an example of a bad mix

homepage/homepage/content/docs/upgrade/0-16-0.mdx
nrainhart: Reworded!

homepage/homepage/content/docs/upgrade/0-16-0.mdx
nrainhart: Added a "Motivation" section

gdorsi (approved):
Stellar job!

PR #2656, opened 2025-07-18
feat: Prevent resolving discriminated union fields

nrainhart:
# Description

https://discord.com/channels/1139617727565271160/1139621689882321009/1395835683268071616

Loading fields from discriminated unions is not currently supported, but our `resolve` APIs allowed doing so. This PR is a quick fix to prevent users from getting into an invalid state.

## Tests

## Checklist

- [x] I've updated the part of the docs that are affected the PR changes (no need)

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/deepLoading.ts
+++ b/packages/jazz-tools/src/tools/coValues/deepLoading.ts
@@ -2,10 +2,22 @@ import { SessionID } from "cojson";
 import { ItemsSym } from "../internal.js";
 import { type Account } from "./account.js";
 import { CoFeedEntry } from "./coFeed.js";
-import { type CoKeys, type CoMap } from "./coMap.js";
+import { type CoKeys } from "./coMap.js";
 import { type CoValue, type ID } from "./interfaces.js";
 
 type NotNull<T> = Exclude<T, null>;
+
+/**
+ * Used to check if T is a union type.
+ *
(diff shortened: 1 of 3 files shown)


packages/jazz-tools/src/tools/coValues/deepLoading.ts
```diff
@@ -2,11 +2,23 @@ import { SessionID } from "cojson";
...
+ * Used to check if T is a union type.
+ *
+ * If T is a union type, the left hand side of the extends becomes a union of function types.
+ * The right hand side is always a single function type.
+ */
+type IsUnion<T, U = T> = (T extends any ? (x: T) => void : never) extends (
+  x: U,
+) => void
+  ? false
+  : true;
```
gdorsi, line 20: Interesting hack!

So if I understand this correctly, both the extends are there in order to force the Typescript compiler to explode every possible type of the union, and the second generic parameter is there to make the compiler infer only one of the possibilities of then union types.

Am I correct?

gdorsi (approved):

packages/jazz-tools/src/tools/coValues/deepLoading.ts
nrainhart: Close. The left-hand side of the top-most extends explodes all possibilities of the union, and returns the union of functions receiving each possible type as argument. The right hand side returns a function that receives the union type as argument. U = T is there to “freeze” the original union type T before the conditional starts distributing it

PR #2660, opened 2025-07-22
feat: introduce persistent peers

### LINKED ISSUE #2450: Bug: wait for the WebSocket connection to be restored before resolving load and waitForSync
Currently load fails immediately if there is no connection with the server and the value is not on storage.

This is especially painful when working with server workers, where the connection might be temporary down and the load immediately returns undefined

The same issue is shared with `waitForSync` which skips the closed peers, making it seems that a value is synced immediately when the connection is down.

gdorsi:
# Description

Introduce the persistent peers. Used to mark the WebSocket connections to the server as persistent, and wait for reconnection before failing load.

This improves how Jazz behaves on unstable connections:

* when calling `waitForSync` while the WebSocket is down, it now waits for reconnection instead of resolving immediately
* loading a coValue while offline now doesn't return immediately as unavailable but waits for reconnection

This is a core improvement required for HTTP requests. [https://github.com/garden-co/jazz/pull/2626](https://github.com/garden-co/jazz/pull/2626)
Workers are going to be deployed in serverless environments where the WebSocket connection might be frozen.

This means that when the request is fired, the worker might be disconnected.
With these changes, they will wait for reconnection for 30s + 3s + 30s (one timeout + one retry) instead of failing any load immediately.

fixes #2450

To test the change from a UX standpoint, I've opened a Jazz app where I have data and nuked the IndexedDB.

After that, by navigating the app and using the DevTools to toggle the offline state, it's possible to observe what happens on a flaky connection.

**Before:**
The "unavailable" state was shown after 3s, but the content was still correctly loaded when going online (thanks to the peer reconciliation process).

**After:**
It takes more time to show the "unavailable" state.

While the difference isn't that big on client apps, this new behavior makes it much easier to work with Workers in serverless environments.

## Tests

* [x] Tests have been added and/or updated
* [ ] Tests have not been updated, because: 
* [ ] I need help with writing tests

## Checklist

* [x] I've updated the part of the docs that are affected by the PR changes
* [x] I've generated a changeset, if a version bump is required
* [x] I've updated the jsDoc comments for the public APIs I've modified, or added them when missing

### DIFF
--- a/packages/cojson/src/sync.ts
+++ b/packages/cojson/src/sync.ts
@@ -88,7 +88,7 @@ export interface Peer {
   outgoing: OutgoingPeerChannel;
   role: "server" | "client";
   priority?: number;
-  deletePeerStateOnClose?: boolean;
+  persistent?: boolean;
 }
 
 export function combinedKnownStates(
@@ -157,8 +157,7 @@ export class SyncManager {
 
   getServerPeers(excludePeerId?: PeerID): PeerState[] {
     return this.getPeers().filter(
-      (peer) =>
-        peer.role === "server" && peer.id !== excludePeerId && !peer.closed,
+      (peer) => peer.role === "server" && peer.id !== excludePeerId,
     );
   }
 
@@ -353,7 +352,7 @@ export class SyncManager {
       unsubscribeFromKnownStatesUpdates();
       this.peersCounter.add(-1, { role: peer.role });
 
-      if (peer.deletePeerStateOnClose && this.peers[peer.id] === peerState) {
+      if (!peer.persistent && this.peers[peer.id] === peerState) {
         delete this.peers[peer.id];
       }
     });
@@ -794,8 +793,8 @@ export class SyncManager {
 
     const peerState = this.peers[peerId];
 
-    // The peer has been closed, so it isn't possible to sync
-    if (!peerState || peerState.closed) {
+    // The peer has been closed and is not persistent, so it isn't possible to sync
+    if (!peerState) {
       return;
     }
 
@@ -830,7 +829,7 @@ export class SyncManager {
     return this.local.storage?.waitForSync(id, this.local.getCoValue(id));
   }
 
-  waitForSync(id: RawCoID, timeout = 30_000) {
+  waitForSync(id: RawCoID, timeout = 60_000) {
     const peers = this.getPeers();
 
     return Promise.all(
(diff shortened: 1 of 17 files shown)


packages/cojson/src/sync.ts
```diff
@@ -830,7 +829,7 @@ export class SyncManager {
     return this.local.storage?.waitForSync(id, this.local.getCoValue(id));
   }
 
-  waitForSync(id: RawCoID, timeout = 30_000) {
+  waitForSync(id: RawCoID, timeout = 60_000) {
```
gdorsi, line 832: Increasing the timeout to make it more reasonable.

In the past we needed this because the whole system was less predictable, but with the recent improvements we can be more confident on not getting "stuck".


Thinking about removing it completely, or making it even longer just as a safeguard to stuck errors.

tests/browser-integration/src/commands.ts
```diff
@@ -36,9 +36,11 @@ const disconnectAllClientsCommand: BrowserCommand<[url: string]> = async (
   syncServer.disconnectAllClients();
 };
 
-const setOfflineCommand: BrowserCommand<
-  [url: string, active: boolean]
-> = async (ctx, url, active) => {
+const setOnlineCommand: BrowserCommand<[url: string, active: boolean]> = async (
```
nrainhart, line 39: 😅

packages/cojson/src/tests/testUtils.ts
```diff
@@ -646,12 +650,30 @@ export type SyncTestMessage = {
...
   });
 
+  // If the persistent option is not provided, we default to true for the server and false for the client
+  // Trying to mimic the real world behavior of the sync server
+  if (opts.persistent === undefined) {
+    if (opts.peer1.role === "server") {
+      peer1.persistent = true;
+    } else {
+      peer1.persistent = false;
+    }
```
nrainhart:
```suggestion
    peer1.persistent = opts.peer1.role === "server";
```

packages/cojson/src/tests/testUtils.ts
```diff
@@ -646,12 +650,30 @@ export type SyncTestMessage = {
...
+
+    if (opts.peer2.role === "server") {
+      peer2.persistent = true;
+    } else {
+      peer2.persistent = false;
+    }
```
nrainhart:
```suggestion
    peer2.persistent = opts.peer2.role === "server";
```

nrainhart (approved):
Changes look good! Tests in particular are really easy to follow ✨

PR #2665, opened 2025-07-24
fix: property update when assigning an optional reference on CoMap

gdorsi:
# Description

While working on the Rock/Paper/Scissors I found a bug on the reference updates.

When a resolved optional undefined reference is updated, accessing the value was returining `null` instead of the updated value.

This PR fixes the bug.

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
+++ b/packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
@@ -341,16 +341,14 @@ export class SubscriptionScope<D extends CoValue> {
       this.resolve = {};
     }
 
-    if (this.resolve.$each || key in this.resolve) {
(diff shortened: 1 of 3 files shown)


packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
```diff
@@ -341,16 +341,14 @@ export class SubscriptionScope<D extends CoValue> {
       this.resolve = {};
     }
 
-    if (this.resolve.$each || key in this.resolve) {
-      return;
-    }
```
gdorsi, line 346: This method is accessed when the child load haven't been tried yet.

By returning early here we were preventing the property access to trigger the load of the new child id.

nrainhart (approved):

PR #2673, opened 2025-07-25
fix: ensure that transactions are synced in the correct order

gdorsi:
# Description

Update the local sync logic to preserve insert update on CoValues transactions.

This way if the tab is reloaded in the middle of sync, the synced state won't link to non-persisted CoValues.

It also prepares the code to do some research on a WAL implementation on clients

Changes in detail:
- Changed the local transactions sync flow to collect the changes instead of diffing on sync
- When syncing local transactions, content is pushed directly to storage & peers. The reconciliation process handles any edge cases in content that misses dependencies
- Changed the correction flow on storage, that works by pulling the correction immediately instead of adding it to the storage queue
- Refactored storage API to accept a content msg instead of a list of content msg

Tested by reloading the stress-test app when generating 5k tasks.

The list points to the correct amount of tasks, and there are no unavailable references:
Screenshot 2025-07-25 at 15 42 03

## Tests

## Checklist


packages/cojson-storage-sqlite/src/tests/storage.sqlite.test.ts
```diff
@@ -374,7 +377,9 @@ test("should recover from data loss", async () => {
     [
       "client -> CONTENT Group header: true new: After: 0 New: 3",
       "client -> CONTENT Map header: true new: After: 0 New: 1",
-      "client -> CONTENT Map header: false new: After: 1 New: 3",
+      "client -> CONTENT Map header: false new: After: 3 New: 1",
+      "storage -> KNOWN CORRECTION Map sessions: header/4",
+      "storage -> CONTENT Map header: false new: After: 1 New: 3",
```
gdorsi: This is expected.

The test checks what happens when the storage loses some transactions

Before: newContentSince was sending all the content since the last knownState of storage
Now: new local transactions are pushed without checking the knownState and corrected in case of errors

packages/cojson-storage-indexeddb/src/tests/storage.indexeddb.test.ts
```diff
@@ -179,8 +179,11 @@ test("should load dependencies correctly (group inheritance)", async () => {
...
   ).toMatchInlineSnapshot(`
     [
-      "client -> CONTENT Group header: true new: After: 0 New: 5",
-      "client -> CONTENT ParentGroup header: true new: After: 0 New: 4",
+      "client -> CONTENT Group header: true new: After: 0 New: 3",
+      "client -> CONTENT ParentGroup header: true new: After: 0 New: 3",
+      "client -> CONTENT Group header: false new: After: 3 New: 1",
+      "client -> CONTENT ParentGroup header: false new: After: 3 New: 1",
+      "client -> CONTENT Group header: false new: After: 4 New: 1",
       "client -> CONTENT Map header: true new: After: 0 New: 1",
```
gdorsi, line 185: This reflects the main change on the local transactions sync.

Before syncing the parent extension on Group we first sync the ParentGroup.

packages/cojson/src/coValueContentMessage.ts
```diff
@@ -0,0 +1,62 @@
...
+  Transaction,
+  VerifiedState,
+} from "./coValueCore/verifiedState.js";
+import { MAX_RECOMMENDED_TX_SIZE } from "./config.js";
+import { Signature } from "./crypto/crypto.js";
+import { RawCoID, SessionID } from "./ids.js";
+import { getPriorityFromHeader } from "./priority.js";
+import { NewContentMessage } from "./sync.js";
+
+export function createContentMessage(
```
gdorsi, line 12: Started to centralize all the functions to manage the content messages, to reduce the code duplication and make the logic around that easier to understand.

packages/cojson-storage-sqlite/src/tests/storage.sqlite.test.ts
```diff
@@ -374,6 +375,8 @@ test("should recover from data loss", async () => {
     [
       "client -> CONTENT Group header: true new: After: 0 New: 3",
       "client -> CONTENT Map header: true new: After: 0 New: 1",
+      "client -> CONTENT Map header: false new: After: 3 New: 1",
+      "storage -> KNOWN CORRECTION Map sessions: header/4",
```
gdorsi, line 379: This happens because we know push new changes directly instead of diffing.

This test covers data loss, so it's ok to have an extra roundtrip here.

packages/cojson/src/storage/storageAsync.ts
```diff
@@ -290,38 +320,31 @@ export class StorageApiAsync implements StorageAPI {
     storedCoValueRowID: number,
   ) {
     const newTransactions = msg.new[sessionID]?.newTransactions || [];
+    const lastIdx = sessionRow?.lastIdx || 0;
```
gdorsi, line 323: This is just a light refactoring to make the code more readable and gain more understanding of it.

packages/cojson/src/storage/storageAsync.ts
```diff
@@ -330,7 +353,7 @@ export class StorageApiAsync implements StorageAPI {
       sessionID,
       lastIdx: newLastIdx,
       lastSignature: msg.new[sessionID].lastSignature,
-      bytesSinceLastSignature: newBytesSinceLastSignature,
+      bytesSinceLastSignature,
     };
 
     const sessionRowID: number = await this.dbClient.addSessionUpdate({
```
gdorsi, line 359: End of the light refactoring

packages/cojson/src/storage/storageSync.ts
```diff
@@ -272,35 +297,29 @@ export class StorageApiSync implements StorageAPI {
     storedCoValueRowID: number,
   ) {
     const newTransactions = msg.new[sessionID]?.newTransactions || [];
+    const lastIdx = sessionRow?.lastIdx || 0;
```
gdorsi, line 300: Start of a light refactoring

packages/cojson/src/storage/storageSync.ts
```diff
@@ -312,7 +331,7 @@ export class StorageApiSync implements StorageAPI {
       sessionID,
       lastIdx: newLastIdx,
       lastSignature: msg.new[sessionID].lastSignature,
-      bytesSinceLastSignature: newBytesSinceLastSignature,
+      bytesSinceLastSignature,
```
gdorsi, line 334: End of the light refactoring

packages/cojson/src/tests/SyncStateManager.test.ts
```diff
@@ -38,14 +38,6 @@ describe("SyncStateManager", () => {
     const updateSpy: GlobalSyncStateListenerCallback = vi.fn();
     const unsubscribe = subscriptionManager.subscribeToUpdates(updateSpy);
 
-    await client.node.syncManager.syncCoValue(map.core);
-
-    expect(updateSpy).toHaveBeenCalledWith(
-      peerState.id,
-      emptyKnownState(map.core.id),
-      { uploaded: false },
-    );
```
gdorsi, line 47: I've removed the `syncCoValue` function, and it's not possible anymore to replicate this intermediate state.

It's not really needed in the end functionality so just removed the check.

packages/cojson/src/localNode.ts
```diff
@@ -739,8 +739,8 @@ export class LocalNode {
   }
 
   gracefulShutdown() {
-    this.storage?.close();
     this.syncManager.gracefulShutdown();
+    return this.storage?.close();
```
nrainhart, line 748: It could be useful to add an explicit return type to `gracefulShutdown()` and explain what it can be used for

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -380,15 +380,15 @@ export class CoValueCore {
   }
 
   knownStateWithStreaming(): CoValueKnownState {
-    if (this.isAvailable()) {
+    if (this.verified) {
```
nrainhart, line 383: We no longer care about checking missing dependencies?

packages/cojson/src/tests/StorageApiAsync.test.ts
```diff
@@ -0,0 +1,829 @@
...
+      const finalKnownState = storage.getKnownState(group.id);
+      expect(finalKnownState).toEqual(group.core.verified.knownState());
+
+      node.setStorage(storage);
+      const groupOnNode = await loadCoValueOrFail(node, group.id);
+
+      expect(groupOnNode.get("everyone")).toEqual("writer");
+    });
+
+    test("should handle log an error when the correction callback returns undefined", async () => {
```
nrainhart: Nit
```suggestion
    test("should log an error when the correction callback returns undefined", async () => {
```

nrainhart: There are a few other like this

nrainhart (approved):
Looks good from my end. Left one question, but it's mostly for my education.

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: Nice spot!

This method is used in the exchange between peers, to share their current state.

When sending content from peerA to peerB, with the isAvailable check in case of missing dependencies peerB would reply with an empty known state, making peerA think that peerB never reiceved the content.

This changes the behavoir of `waitForSync` which now resolves when the value content is acknowledged, even if there are still some dependencies to sync.

Normally this shouldn't happen, because the peerA always sends the dependencies before sending the content, unless they think that peerB already has the dependencies but they have lost them because of *reasons*

packages/cojson/src/localNode.ts
gdorsi: Thanks for the suggestion, done!

packages/cojson/src/tests/StorageApiAsync.test.ts
gdorsi: Nice spot, used LLMs to fix all the test titles on StorageAsync and StorageSync

PR #2675, opened 2025-07-25
Feat/quint add full button suite

Sammii-HK:
# Description

got the entire button suite from the design system ported over to quint, and the icon, ported over and improved docs and refactored tailwind usage and css while doing so,  note the nav has not had any thought to it, nor the dropdown to view all intent/variants but this will happen as I introduce these components into quint 

## Tests

## Checklist


packages/quint-ui/src/components/button.tsx
```diff
@@ -2,130 +2,365 @@ import { mergeProps } from "@base-ui-components/react/merge-props";
...
+
+export function Button({ render = <button />, ...props }: ButtonProps) {
+  const element = useRender({
+    render,
+    props: mergeProps<"button">(
+      {
+        className: button({
+          size: props.size || "md",
+          variant: props.variant || "default",
+          intent: props.intent || "default",
```
Elfo404: these shouldn't be needed as you are defining them as `defaultVariants` down below

```suggestion
          size: props.size,
          variant: props.variant,
          intent: props.intent,
```

Elfo404 (approved):

PR #2676, opened 2025-07-28
fix: jazz-run package.json exports

Gabrola:
# Description
Fixes the order of the export conditions and replaces `import` condition with `default`

## Tests

## Checklist

### DIFF
--- a/.changeset/smooth-pears-grin.md
+++ b/.changeset/smooth-pears-grin.md
@@ -1,0 +1,5 @@
+---
+"jazz-run": patch
+---
+
+Fix jazz-run package.json exports
--- a/packages/jazz-run/package.json
+++ b/packages/jazz-run/package.json
@@ -6,12 +6,12 @@
   "version": "0.16.0",
   "exports": {
     "./startSyncServer": {
-      "import": "./dist/startSyncServer.js",
-      "types": "./dist/startSyncServer.d.ts"
+      "types": "./dist/startSyncServer.d.ts",
+      "default": "./dist/startSyncServer.js"
     },
     "./createWorkerAccount": {
-      "import": "./dist/createWorkerAccount.js",
-      "types": "./dist/createWorkerAccount.d.ts"
+      "types": "./dist/createWorkerAccount.d.ts",
+      "default": "./dist/createWorkerAccount.js"
     }
   },
   "scripts": {


gdorsi (approved):
Thanks!

gdorsi:
@Gabrola could you share what prompted the fix?

Any issues with a specific bundler/CDN?

Gabrola:
@gdorsi Node.js couldn't find the exports in a CJS project because the export was under `import` and not `default` or `require`. Node >= 22 can `require` ESM as you might already know

Gabrola:
@gdorsi if you're curious here's the error
```
node:internal/modules/cjs/loader:664
      throw e;
      ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './createWorkerAccount' is not defined by "exports" in node_modules/jazz-run/package.json
    at exportsNotFound (node:internal/modules/esm/resolve:314:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:604:13)
    at resolveExports (node:internal/modules/cjs/loader:657:36)
    at Function._findPath (node:internal/modules/cjs/loader:749:31)
    at Function._resolveFilename (node:internal/modules/cjs/loader:1387:27)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1057:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1062:22)
    at Function._load (node:internal/modules/cjs/loader:1211:37)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:235:24) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}

Node.js v22.15.1
```

gdorsi:
Sorry for the late reply @Gabrola 

Going to merge this as it doesn't break anything, but we currently don't compile to CJS in jazz-run so it's probably not going to work on your Node.js version.

Adding a cjs compilation should be fairly easy, if you want to give it a try feel free to open another PR.

Thanks for the contribution ❤️

Gabrola:
@gdorsi no worries! It actually works for me when I patch the library's package.json. Node.js can now `require(esm)` for modules without top-level awaits in versions >= 20.17. However, Node will still only use `require` or `node` package.json `exports` conditions only (they also added `'module-sync'` for explicit "no-top-level-await ESM" but yeah no one's using that :laughing:). So the recommendation is for ESM-only libraries to use the `default` export condition. You've actually got it setup here https://github.com/garden-co/jazz/blob/main/packages/jazz-tools/package.json perfectly.

The patch that fixed it for me is essentially the same as the PR exactly.

```diff
diff --git a/package.json b/package.json
index ae179c26d1fee96d18ff9360caf38ad8397be246..d8aaa1c8ae7ec1d6473fc62902860d255bd491d8 100644
--- a/package.json
+++ b/package.json
@@ -6,12 +6,12 @@
   "version": "0.16.1",
   "exports": {
     "./startSyncServer": {
-      "import": "./dist/startSyncServer.js",
-      "types": "./dist/startSyncServer.d.ts"
+      "types": "./dist/startSyncServer.d.ts",
+      "default": "./dist/startSyncServer.js"
     },
     "./createWorkerAccount": {
-      "import": "./dist/createWorkerAccount.js",
-      "types": "./dist/createWorkerAccount.d.ts"
+      "types": "./dist/createWorkerAccount.d.ts",
+      "default": "./dist/createWorkerAccount.js"
     }
   },
   "dependencies": {
```

I actually don't recommend publishing CJS now, especially if a library never did in the first place. Practically all tooling, bundlers and runtimes support ESM now. The only reason I would say old and popular libraries might want to keep publishing CJS, is that even if a minority 1% of their users are still using legacy software, 1% of millions is still quite significant.

Gabrola:
@gdorsi no worries! It actually works for me when I patch the library's package.json. Node.js can now `require(esm)` for modules without top-level awaits in versions >= 20.17. However, Node will still only use `require` or `node` package.json `exports` conditions (they also added `'module-sync'` for explicit "no-top-level-await ESM" but yeah no one's using that :laughing:). So the recommendation is for ESM-only libraries to use the `default` export condition. You've actually got it setup here https://github.com/garden-co/jazz/blob/main/packages/jazz-tools/package.json perfectly.

The patch that fixed it for me is essentially the same as the PR exactly.

```diff
diff --git a/package.json b/package.json
index ae179c26d1fee96d18ff9360caf38ad8397be246..d8aaa1c8ae7ec1d6473fc62902860d255bd491d8 100644
--- a/package.json
+++ b/package.json
@@ -6,12 +6,12 @@
   "version": "0.16.1",
   "exports": {
     "./startSyncServer": {
-      "import": "./dist/startSyncServer.js",
-      "types": "./dist/startSyncServer.d.ts"
+      "types": "./dist/startSyncServer.d.ts",
+      "default": "./dist/startSyncServer.js"
     },
     "./createWorkerAccount": {
-      "import": "./dist/createWorkerAccount.js",
-      "types": "./dist/createWorkerAccount.d.ts"
+      "types": "./dist/createWorkerAccount.d.ts",
+      "default": "./dist/createWorkerAccount.js"
     }
   },
   "dependencies": {
```

I actually don't recommend publishing CJS now, especially if a library never did in the first place. Practically all tooling, bundlers and runtimes support ESM now. The only reason I would say old and popular libraries might want to keep publishing CJS, is that even if a minority 1% of their users are still using legacy software, 1% of millions is still quite significant.

PR #2677, opened 2025-07-28
feat: Add support for nullable non-collaborative fields

nrainhart:
# Description

Add support for nullable non-collaborative fields. I also took the opportunity to simplify `CoMapSchema.create`'s input and return types.

Motivation: an adopter was asking about it: https://discord.com/channels/1139617727565271160/1139621689882321009/1399159852416372826, and I think it's an useful feature to have. Plus, it's easy to implement now that we distinguish between collaborative and non-collaborative fields.

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/implementation/zodSchema/runtimeConverters/schemaFieldToCoFieldDef.ts
+++ b/packages/jazz-tools/src/tools/implementation/zodSchema/runtimeConverters/schemaFieldToCoFieldDef.ts
@@ -22,6 +22,7 @@ export type SchemaField =
   | CoValueClass
   | ZodPrimitiveSchema
   | z.core.$ZodOptional<z.core.$ZodType>
+  | z.core.$ZodNullable<z.core.$ZodType>
   | z.core.$ZodUnion<z.core.$ZodType[]>
   | z.core.$ZodDiscriminatedUnion<z.core.$ZodType[]>
   | z.core.$ZodObject<z.core.$ZodLooseShape>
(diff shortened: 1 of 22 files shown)


packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
```diff
@@ -16,37 +17,27 @@ import {
...
-    init: Simplify<CoMapInitZod<Shape>>,
+    init: Simplify<CoMapSchemaInit<Shape>>,
     options?:
       | {
           owner: Owner;
           unique?: CoValueUniqueness["uniqueness"];
         }
       | Owner,
-  ) => (Shape extends Record<string, never>
-    ? {}
```
nrainhart, line 39: I could see no reason for having this check, so I removed it

packages/jazz-tools/src/tools/implementation/zodSchema/runtimeConverters/schemaFieldToCoFieldDef.ts
```diff
@@ -49,9 +50,22 @@ export function schemaFieldToCoFieldDef(schema: SchemaField) {
...
+        zodSchemaDef.type === "nullable"
+      ) {
         const inner = zodSchemaDef.innerType as ZodPrimitiveSchema;
-        return schemaFieldToCoFieldDef(inner);
+        const coFieldDef: any = schemaFieldToCoFieldDef(inner);
+        if (
+          zodSchemaDef.type === "nullable" &&
+          coFieldDef === coField.optional.Date
+        ) {
+          throw new Error("Nullable z.date() is not supported");
```
nrainhart, line 65: If you think there's a way to support nullable date coFields let me know, but I think this is good enough

gdorsi: Could you keep documented why we don't support nullable z.date?

gdorsi (commented):
LGTM!

Could you add a type test?
Just to help me understand which type we expect on a Loaded nullable field.

packages/jazz-tools/src/tools/implementation/zodSchema/runtimeConverters/schemaFieldToCoFieldDef.ts
nrainhart: In the code, you mean? Or in https://jazz.tools/docs?

nrainhart:
> Could you add a type test?
Just to help me understand which type we expect on a Loaded nullable field.

Added here [`9f654a2` (#2677)](https://github.com/garden-co/jazz/pull/2677/commits/9f654a260366c573525fb0769c5b08983b89d1a9)

packages/jazz-tools/src/tools/implementation/zodSchema/runtimeConverters/schemaFieldToCoFieldDef.ts
gdorsi: In the code

gdorsi (approved):

PR #2678, opened 2025-07-28
Homepage/button link fix

joeinnes:
# Description
Currently, the Button component wraps a Next Link. That means the Button's padding is applied around the link, so only the *text* in the button is clickable. This changes the order of the components so that the Button is fully wrapped by the Link, meaning the whole of the button is now clickable.

1. Build and load the homepage
2. Attempt to click the 'Get started' button on the text
    1. You should be navigated to the appropriate docs landing page
3. Attempt to click the 'Get started' button *outside* the text (between the text and the outer edge of the element).
    1. You should be navigated to the appropriate docs landing page

## Tests

## Checklist

- [ ] ~~I've updated the part of the docs that are affected the PR changes~~
- [ ] ~~I've generated a changeset, if a version bump is required~~
- [ ] ~~I've updated the jsDoc comments to the public APIs I've modified, or added them when missing~~

### DIFF
--- a/homepage/homepage/components/home/GetStartedSnippetSelect.tsx
+++ b/homepage/homepage/components/home/GetStartedSnippetSelect.tsx
@@ -1,4 +1,4 @@
-'use client'
+"use client";
 
 import { Framework } from "@/content/framework";
 import { useFramework } from "@/lib/use-framework";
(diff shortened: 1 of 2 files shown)


joeinnes:
New homepage already has fix integrated.

PR #2681, opened 2025-07-29
Jazz database adapter for Better Auth

Takeno:
# Description
Implemented a Jazz adapter for Better Auth. It is to save Better Auth's users, accounts, and sessions on Jazz.

Check the new example

## Tests

## Checklist

## Missing
- [x] New docs


artalar:
Hi! Very excited about this feature! What is the timing, how do you think, when it will be in production?

examples/better-auth-adapter/README.md
```diff
@@ -0,0 +1,5 @@
+# Example of using Jazz as database adapter for Better Auth
```
gdorsi: Would it be possible to merge this example app with `examples/betterauth` (deleting `examples/betterauth` is also ok)

packages/jazz-better-auth/src/database-adapter/index.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  await runAdapterTest({
+    disableTests: {
+      SHOULD_PREFER_GENERATE_ID_IF_PROVIDED: true,
+    },
+    getAdapter: async (betterAuthOptions = {}) => {
+      return adapter(betterAuthOptions);
+    },
+  });
```
gdorsi: Nice 👌

packages/jazz-better-auth/src/database-adapter/schema.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    const coMap = co.map(coMapSchema);
+    dbSchema[key] = coMap;
+    rootSchema[key] = co.list(coMap);
+  }
+
+  const WorkerAccount = co
+    .account({
+      profile: co.profile(),
+      root: co.map(rootSchema),
```
gdorsi: **question:** why did you use a `co.map` instead of a `co.record`?

packages/jazz-better-auth/src/database-adapter/schema.ts
```diff
@@ -0,0 +1,109 @@
...
+            root: true,
+          },
+        });
+      }
+
+      for (const [key, value] of Object.entries(rootSchema)) {
+        if (account.root?.[key] === undefined) {
+          account.root![key] = value.create([], account);
+        }
+      }
```
gdorsi: To make this more type-safe:

```suggestion
      }
      const { root } =  await account.ensureLoaded({
        resolve: {
          root: true,
        },
      });

      for (const [key, value] of Object.entries(rootSchema)) {
        if (root[key] === undefined) {
          root[key] = value.create([], account);
        }
      }
```

packages/jazz-better-auth/src/database-adapter/schema.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export function createJazzSchema(schema: BetterAuthDbSchema): JazzSchema {
+  const dbSchema: DbSchema = {};
+  const rootSchema: RootSchema = {};
+
+  for (const [key, value] of Object.entries(schema)) {
+    const coMapSchema: Record<
+      string,
+      ZodPrimitiveSchema | ZodOptionalPrimitiveSchema
+    > = {};
```
gdorsi: **nit:** I would name this `modelShape` rather than `coMapSchema`

Model because it's the table representation of the BetterAuth model
Shape because it's the shape of the schema, not a schema (the return type of `co.map` is a schema)

packages/jazz-better-auth/src/database-adapter/schema.ts
```diff
@@ -0,0 +1,109 @@
...
+  root: co.Map<RootSchema>;
+}>;
+
+type JazzSchema = {
+  WorkerAccount: WorkerAccount;
+  dbSchema: DbSchema;
+  rootSchema: RootSchema;
+};
+
+export function createJazzSchema(schema: BetterAuthDbSchema): JazzSchema {
```
gdorsi: We handle new models, but do we need to handle migrations on the fields?

What's the better auth policy on the schema?

Takeno:
> Hi! Very excited about this feature! What is the timing, how do you think, when it will be in production?

@artalar we don't have a definite date for publishing it yet, but after some structural changes emerged during the review, I think I will publish a pre-release tag in the middle of this upcoming week.

examples/better-auth-adapter/README.md
Takeno: The example `betterauth` shows how to use Better Auth as an authentication provider for Jazz. The new example shows how to use Jazz as a database adapter. I wonder if keeping them separated would help people understand they are independent, or if we will explain it inside the BA integration docs. wdyt?

packages/jazz-better-auth/src/database-adapter/schema.ts
Takeno: Schemas are unpredictable: users can also manipulate the column definitions for the main account table. There is a hook for the adapter, which is called when the user requests migration generation based on new settings. However, the other adapters are based on schema-full relational databases and compare the new schema provided by BA with the database's schema.
In our case, if the user changes the BA's schema, we cannot determine its previous state and they must create their migrations by themself.

Got the point for the docs.

packages/jazz-better-auth/src/database-adapter/schema.ts
Takeno: Due to the `root` property definition in [AccountSchema.ts:26-29](https://github.com/garden-co/jazz/blob/main/packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/AccountSchema.ts#L26-L29):

```ts
export type BaseAccountShape = {
  profile: CoreCoMapSchema<BaseProfileShape>;
  root: CoreCoMapSchema;
};
```

aeplay (approved):
Great work!

PR #2690, opened 2025-08-01
docs: adds jsDocs to useCoState and useAccount react hooks

gdorsi:
Adding jsDocs to useCoState and useAccount react hooks to make them self-documented and give users and AIs the context to understand them without reading the docs.

### DIFF
--- a/packages/jazz-tools/src/react-core/hooks.ts
+++ b/packages/jazz-tools/src/react-core/hooks.ts
@@ -145,7 +145,7 @@ function useCoValueSubscription<
  * React hook for subscribing to CoValues and handling loading states.
  *
  * This hook provides a convenient way to subscribe to CoValues and automatically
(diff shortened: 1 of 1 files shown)


packages/jazz-tools/src/react-core/hooks.ts
```diff
@@ -141,13 +141,100 @@ function useCoValueSubscription<
   return subscription.subscription;
 }
 
+/**
+ * React hook for subscribing to CoValues and handling loading states.
+ *
+ * This hook provides a convenient way to subscribe to CoValues and automatically
+ * handles subscription lifecycle (subscribe on mount, unsubscribe on unmount).
```
nrainhart:
```suggestion
 * handles the subscription lifecycle (subscribe on mount, unsubscribe on unmount).
```

nrainhart (approved):
Thanks for adding this! 💅

Takeno (approved):

PR #2692, opened 2025-08-01
feat: added a TTL-based garbage collection

gdorsi:
# Description

Added a TTL-based garbage collection for CoValues.

To play safe enabled only on the sync server, and doesn't collect groups or accounts.

## Tests

## Checklist


packages/cojson/src/GarbageCollector.ts
```diff
@@ -0,0 +1,45 @@
...
+    }, GARBAGE_COLLECTOR_CONFIG.INTERVAL);
+  }
+
+  getCurrentTime() {
+    return performance.now();
+  }
+
+  markCoValueAccess(coValue: CoValueCore) {
+    if (coValue.verified && coValue.verified.header.ruleset.type !== "group") {
+      this.coValueAccess.set(coValue.id, this.getCurrentTime());
```
gdorsi: We could store the access as property of the coValue, for performance should be better than accessing a Map.

packages/cojson/src/GarbageCollector.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.interval = setInterval(() => {
+      this.collect();
+    }, GARBAGE_COLLECTOR_CONFIG.INTERVAL);
+  }
+
+  getCurrentTime() {
+    return performance.now();
+  }
+
+  markCoValueAccess(coValue: CoValueCore) {
```
gdorsi: this should be `trackCoValueAccess`

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -81,7 +81,14 @@ export class CoValueCore {
...
   private readonly peers = new Map<
     PeerID,
-    | { type: "unknown" | "pending" | "available" | "unavailable" }
+    | {
+        type:
+          | "unknown"
+          | "pending"
+          | "available"
+          | "unavailable"
+          | "garbageCollected";
```
nrainhart: Is this type being used?

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -213,6 +219,16 @@ export class CoValueCore {
...
   }
 
+  unmount() {
+    this.counter.add(-1, { state: this.loadingState });
+
+    if (this.groupInvalidationSubscription) {
+      this.groupInvalidationSubscription();
+      this.groupInvalidationSubscription = undefined;
+    }
+    this.listeners.clear();
```
nrainhart: Isn't it contradictory that we're only calling unmount if listeners is empty, but also clear it here?

packages/cojson/src/tests/GarbageCollector.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    const map = group.createMap();
+    map.set("hello", "world", "trusting");
+
+    await new Promise((resolve) => setTimeout(resolve, 10));
+
+    client.node.garbageCollector?.collect();
+
+    const coValue = client.node.getCoValue(map.id);
+
+    expect(coValue.isAvailable()).toBe(false);
```
nrainhart: How is this test different from the previous one?

packages/cojson/src/GarbageCollector.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (coValue.verified && coValue.verified.header.ruleset.type !== "group") {
+      this.coValueAccess.set(coValue.id, this.getCurrentTime());
+    }
+  }
+
+  collect() {
+    const currentTime = this.getCurrentTime();
+    for (const [id, accessTime] of this.coValueAccess) {
```
nrainhart: Do you have a sense of how many covalues we'll be iterating over (like, order of magnitude)? I'm trying to get a sense of how long the garbage collection pause could take

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: Nope, going to clean it!

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: Well, yes.
Going to remove it.

packages/cojson/src/GarbageCollector.ts
gdorsi: In the core server, without GC we go around 1 million of CoValues.

Can make the iteration allocation-free by storing the access timestamp as property on CoValueCore and iterating the coValues map values.

The operation should be fairly quick, will give you numbers, so we can also see if optimizing the iteration makes sense.

Also this is not a long-term solution, but more like a way to quickly introduce a garbage collection in order to see how it affects sync.

The final implementation will take into consideration more variables (such as the memory pressure and the size of the coValues) and will have to be more sophisticated.

packages/cojson/src/tests/GarbageCollector.test.ts
gdorsi: Yes, it's a duplicate.

Going to remove it.

aeplay (approved):

PR #2693, opened 2025-08-01
feat: add `$jazz` property to CoValues

nrainhart:
# Description

## Notes for reviewers

The changes in this PR are straightforward, as I'm mostly just moving code around. I'm interested in higher-level feedback about things you think should be moved into `$jazz` (or things I moved that you think I shouldn't have). I also left several comments in this PR I want to get input on. Relevant changes are in spin-off PRs.

## Motivation

We want to move all CoValue methods and internal properties into a single `$jazz` field, to avoid conflicts/shadowing with user-defined properties in CoMaps. This field is not enumerable, to allow CoValues to behave similarly to JSON objects.

CoValues are now intended to be read-only, so all mutation methods have been moved to `$jazz`. This affects mostly CoLists and CoFeeds.

In cases where there's no risk of shadowing, Jazz methods are preserved in CoValues to avoid unnecessary boilerplate. This is the case for most Account, Group, Plain/rich text and FileStream methods.

## Changes

- Moved all CoMap methods into `$jazz`
	- Made all CoMap fields readonly to prevent setting properties directly
	- ⚠️ This works only for CoMaps created with `co.map`. Those created with class schemas won't get type errors on direct property assignments (but they will get a runtime error prompting them to use `$jazz.set`)
	- the `delete` operator can no longer be used to delete CoRecord properties. Added `$jazz.delete` in its place.
- Moved CoList's array-mutation methods into `$jazz`
	- CoLists are now readonly. Trying to use any mutation method yields a type error
	- ⚠️ The only operation that isn't caught by a type error (and throws a runtime error instead) is directly setting an array element (using `array[0] = value`)
	- `sort`, `reverse`, `fill` and `copyWithin` have been deprecated without replacements
- Added a `$jazz.set` method to CoMap and CoList
- Moved subscription and loading methods (ensureLoaded, subscribe, waitForSync) into `$jazz` in all CoValues
- Moved common properties (id, raw, owner, etc.) to `$jazz` for all CoValues
- Renamed `_type` to `$type` and kept it as a top-level property for CoValues
- Removed `id` and `_type` from `toJSON()`'s output in Account, CoMap, CoFeed & FileStream
- Removed `root` & `profile` fields from Group

## Spin-off changes 

Some changes are significant and isolated enough to justify having separate PRs. I'll open them against this branch and merge them once they're approved:
- https://github.com/garden-co/jazz/pull/2743
- https://github.com/garden-co/jazz/pull/2767
- https://github.com/garden-co/jazz/pull/2778
- https://github.com/garden-co/jazz/pull/2815

I recommend you play around with the new API in jazz-tools test files (e.g. `coMap.test.ts`).

## Tests

## Checklist


packages/jazz-tools/src/tools/coValues/account.ts
```diff
@@ -93,119 +93,40 @@ export class Account extends CoValueBase implements CoValue {
...
     });
 
-    if (this.isLocalNodeOwner) {
-      this.sessionID = options.fromRaw.core.node.currentSessionID;
-    }
-
     return new Proxy(this, AccountAndGroupProxyHandler as ProxyHandler<this>);
   }
 
+  // TODO Q: does it make sense to move Group methods into `$jazz` as well?
```
nrainhart: Given that Groups cannot be extended with user-defined properties, requiring people to prefix all operations with `$jazz` feels like unnecessary boilerplate. The same question applies to Accounts methods

gdorsi: Agree

gdorsi: Same for FileStream, and image

packages/jazz-tools/src/tools/coValues/coMap.ts
```diff
@@ -324,24 +179,24 @@ export class CoMap extends CoValueBase implements CoValue {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   toJSON(_key?: string, processedValues?: ID<CoValue>[]): any {
     const result = {
-      id: this.id,
+      id: this.$jazz.id,
       _type: this._type,
```
nrainhart: @aeplay @gdorsi should I go ahead and remove `id` and `_type` from `toJSON`? That'd be consistent with viewing CoMaps as plain JSON objects

gdorsi: It makes sense to me, yes.

packages/jazz-tools/src/tools/coValues/interfaces.ts
```diff
@@ -32,26 +32,28 @@ export interface CoValueClass<Value extends CoValue = CoValue> {
...
   /** @category Type Helpers */
-  _type: string;
-  /** @category Collaboration */
-  _owner: Account | Group;
-  /** @category Internals */
-  _raw: RawCoValue;
-
-  /** @internal */
-  _subscriptionScope?: SubscriptionScope<this>;
+  $type: string;
```
nrainhart: I initially planned to move `_type` inside `$jazz`, but that breaks Typescript's type narrowing for CoValues (since only literals are allowed as discriminators), so I decided to keep it as a top-level field (in line with `$jazz` and `$jazzState`).

`$type` is a fairly common field name though, so perhaps we should use something more specific?

packages/jazz-tools/src/tools/coValues/account.ts
```diff
@@ -343,10 +236,7 @@ export class Account extends CoValueBase implements CoValue {
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   toJSON(): object | any[] {
-    return {
-      id: this.id,
-      _type: this._type,
-    };
```
nrainhart, line 349: I wonder if we should include root and profile in `Account.toJSON`. @gdorsi what do you think?

packages/jazz-tools/src/react-core/tests/useCoState.test.ts
```diff
@@ -504,7 +504,8 @@ describe("useCoState", () => {
     );
 
     const { result } = renderHook(
-      () => useCoState(Dog, dog.id)?._owner.castAs(Group).members,
+      () =>
```
nrainhart: I reckon this isn't a typical use case, but I found this to be too boilerplate-ry

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -499,6 +302,301 @@ export class CoList<out Item = any> extends Array<Item> implements CoValue {
     return subscribeToCoValueWithoutMe<L, R>(this, id, options, listener);
   }
 
+  // Override mutation methods defined on Array, as CoLists aren't meant to be mutated directly
+  override push(...items: Item[]): number {
+    throw new Error(
+      "Cannot mutate a CoList directly. Use .$jazz.push(...) instead.",
+    );
```
nrainhart: Overriding mutable Array methods to return an error will make the migration to `$jazz` more difficult to adopters, as they won't get type errors for unsupported methods.

I don't see a way to inherit only read-only array methods, though, so if we wanted CoList to only implement the `ReadonlyArray` interface, we'd need to reimplement all methods manually.

@gdorsi what do you think about this?

gdorsi: We can go with a warning, and leave things simple.
It would be nice to mark them as deprecated as well, if possible.

As for protect us from undesired mutations, maybe we should get rid of the `set` part of the proxy?
This way the only way to really change the CoValue will be `.$jazz`

How much complexity would this add?

nrainhart: > We can go with a warning, and leave things simple.

What do you mean by a warning? Log something to the console but still call the real method?

> As for protect us from undesired mutations, maybe we should get rid of the set part of the proxy?
This way the only way to really change the CoValue will be .$jazz
How much complexity would this add?

I think this is a similar case to the array mutation methods: throwing an error when calling set on the proxy is trivial, but will make the migration more difficult. I'd love direct assignments to throw a type error, but not sure how to do that without making big refactors on CoMap and CoList.

gdorsi: How about focus on making Typescript throw errors?

e.g. we could make push not take arguments anymore:
```ts

  /** @deprecated Use `.$jazz.push` instead */
  override push(value: never): number {
    throw new Error("Not implemented");
  }

  /** @deprecated Use `.$jazz.pop` instead */
  // @ts-expect-error
  override pop(value: never): Item | undefined {
    throw new Error("Not implemented");
  }
```

As for the assignements, how amout making the co.loaded type mark all the CoList and CoMap readonly?
This would simplify the migration alot, because users would only need to make Typescript pass to be sure that everything has been handled.

gdorsi: Seems to do the job
Screenshot 2025-08-13 at 22 55 12

packages/jazz-tools/src/tools/coValues/account.ts
gdorsi: It would make sense, but out of scope for now.

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -499,6 +297,301 @@ export class CoList<out Item = any> extends Array<Item> implements CoValue {
...
+      } else if ((value as CoValue)?.$jazz?.id) {
+        rawValue = (value as CoValue).$jazz.id;
+      } else {
+        throw new Error(
+          `Cannot set reference ${index} to a non-CoValue. Got ${value}`,
+        );
+      }
+    }
+    // @ts-expect-error TS incorrectly believes 'rawValue' is used before being assigned
+    this.raw.replace(index, rawValue);
```
gdorsi, line 536: Why aren't you using `toRawItems` here?

packages/jazz-tools/src/tools/coValues/CoValueBase.ts
```diff
@@ -21,29 +20,59 @@ import type {
...
 
 /** @internal */
+export abstract class CoValueBase implements CoValue {
+  declare $type: string;
 
-export class CoValueBase implements CoValue {
-  declare id: ID<this>;
-  declare _type: string;
-  declare _raw: RawCoValue;
+  declare abstract $jazz: CoValueJazzApi<this>;
```
gdorsi, line 25: Since the whoule point of `$jazz` is to reduce conflicts, how about giving to this a more internal name?

Something like `$type$` or `$$type`
This is meant for internal usage and hacks anyway, so it's ok to keep it ugly

nrainhart: [Done](https://github.com/garden-co/jazz/pull/2693/commits/44dc7bfd21c6272f2d93d2e5d355136c1954e9e5)! Renamed it to `$type$`, to align it with other fields like `$items$` and `$SchemaInit$`

packages/jazz-tools/src/tools/coValues/coList.ts
nrainhart: I copied this code verbatim from `CoListProxyHandler`, but you're right. I'll refactor CoList to use `toRawItems` wherever possible

packages/jazz-tools/src/tools/tests/coFeed.test.ts
```diff
@@ -825,9 +823,11 @@ describe("waitForSync", async () => {
...
     const stream = FileStream.create();
-    expect(stream._owner._type).toEqual("Group");
-    expect(stream._owner.castAs(Group)._raw.roleOf(account._raw.id)).toEqual(
-      "admin",
-    );
+    expect(stream.$jazz.owner[TypeSym]).toEqual("Group");
+    expect(
+      stream.$jazz.owner.$jazz
+        .castAs(Group)
+        .$jazz.raw.roleOf(account.$jazz.raw.id),
```
nrainhart: Another case that is pretty verbose

packages/jazz-tools/src/tools/coValues/coMap.ts
```diff
@@ -673,48 +524,88 @@ export class CoMap extends CoValueBase implements CoValue {
...
-    ...args: SubscribeRestArgs<M, R>
-  ): () => void {
-    const { options, listener } = parseSubscribeRestArgs(args);
-    return subscribeToExistingCoValue<M, R>(this, options, listener);
+   * @category Content
+   */
+  delete<
+    K extends OptionalCoKeys<M> | (string extends keyof M ? string : never),
+  >(key: K): void {
+    this.raw.delete(key);
```
nrainhart, line 627: I'm relying on the type system to prevent deletions of required properties. I don't think runtime checks are required as well, but leaving this comment here in case someone wants to have that discussion

starters/svelte-passkey-auth/src/lib/components/Form.svelte
```diff
@@ -28,7 +28,10 @@
         id="firstName"
         placeholder="Enter your first name here..."
         class="border border-stone-300 rounded shadow-xs py-1 px-2 flex-1"
-        bind:value={me.profile.firstName}
+        bind:value={
+          () => me.profile.firstName,
+          newValue => me.profile.$jazz.set("firstName", newValue)
+        }
```
nrainhart, line 34: [Function bindings](https://svelte.dev/docs/svelte/bind#Function-bindings) are now required for `bind:value`, as direct property assignments are no longer allowed for CoValues

gdorsi: This looks even better, Svelte probably won't complain about this.

I had to silence warnings before.

homepage/homepage/content/docs/groups/intro.mdx
```diff
@@ -11,6 +12,16 @@ Every CoValue has an owner, which can be a `Group` or an `Account`.
...
 
+CoValues owned by an Account can only be accessed by that Account. Additional collaborators cannot be added,
+and the ownership cannot be transferred to another Account. This makes account ownership very rigid.
+
+Creating a Group for every new CoValue is a best practice, even if the Group only has a single user in it
+(this is the default behavior when creating a CoValue with no explicit owner).
+
+<Alert variant="info" className="flex gap-2 items-center my-4">
+  Account ownership is kept for backwards compatibility, but will be removed in a future release.
+</Alert>
```
joeinnes: I personally would make it even more discouraging, WDYT?

```suggestion
<Alert variant="warning" className="flex gap-2 items-center my-4">
  While creating CoValues with Accounts as owners is still technically possible for backwards compatibility, it will be removed in a future release.
</Alert>
```

homepage/homepage/content/docs/upgrade/0-9-0/react-native.mdx
joeinnes, line 1: These are old upgrade docs, no longer visible. I wouldn't correct these, and leave as is.

homepage/homepage/content/docs/upgrade/0-9-0/svelte.mdx
joeinnes, line 1: These are old upgrade docs, no longer visible. I wouldn't correct these, and leave as is. If you do decide to correct them, then there's a typo (should be `$jazz.id`)

homepage/homepage/content/docs/upgrade/0-9-0/vue.mdx
joeinnes, line 1: These are old upgrade docs, no longer visible. I wouldn't correct these, and leave as is.

homepage/homepage/content/docs/upgrade/0-18-0.mdx
```diff
@@ -0,0 +1,196 @@
...
+Jazz 0.18.0 introduces two new utility methods for easier CoList editing:
+
+<CodeGroup>
+```ts
+// Remove items
+list.$jazz.remove(1); // By index
+list.$jazz.remove(item => item.completed); // By predicate
+
+// Keep only items matching the predicate
+list.$jazz.retain(item => !item.completed);
```
joeinnes, line 78: I love these utilities!

homepage/homepage/content/docs/upgrade/0-18-0.mdx
```diff
@@ -0,0 +1,196 @@
...
+
+CoValues owned by Accounts will now return a Group that represents that Account. It will continue to behave the same way it previously did:
+trying to add a new member to the Group will fail.
+
+Accounts as CoValue owners will be phased out in a future release.
+
+### Removed Features
+
+#### castAs Method
+The `castAs` method has been completely removed as it was an unsafe operation that bypassed type checking and enabled using CoValues in unsupported ways.
```
joeinnes:
```suggestion
#### castAs() Method
The `castAs()` method has been completely removed as it was an unsafe operation that bypassed type checking and enabled using CoValues in unsupported ways.
```

homepage/homepage/content/docs/using-covalues/colists.mdx
```diff
@@ -182,17 +182,17 @@ export function getCurrentTasks(list: co.loaded<ListOfTasks>) {
 
 async function main() {
   const myTaskList = ListOfTasks.create([]);
```
joeinnes, line 184: Is there a specific reason we're creating the Task CoValues manually here rather than just pushing JSON to the list? Might make things a bit more concise otherwise.

nrainhart: There isn't. I'll make the change, thanks for spotting it!

gdorsi (approved):
Amazing job 🚀

PR #2696, opened 2025-08-03
feat(chat): implement lazy-loading

gdorsi:
Replaced the deep loading with a lazy load solution paired with a preload hook to avoid the initial flicker.

Also fixed:
- The initial navigation on the chat was broken when running locally, added a fix on hash-slash to use a custom event for the imperative navigation
- npm installation of the template was broken due to peer dependencies failing with lucide-react. Fixed by updating lucide to the latest version

- Load https://jazz-chat-git-feat-chat-pagination-gcmp.vercel.app/#/chat/co_zYi5HsqwE4PbWRprAoHUQZzjK52
- Compare the perf with https://chat.jazz.tools/#/chat/co_zYi5HsqwE4PbWRprAoHUQZzjK52

The loading should be faster than the production deploy, and pagination should work.

### DIFF
--- a/.changeset/ten-cobras-fetch.md
+++ b/.changeset/ten-cobras-fetch.md
@@ -2,5 +2,5 @@
 "hash-slash": patch
 ---
 
-Fix router update when calling navitate\
+Fix router update when calling navigate\
 
--- a/.changeset/ten-cobras-fetch.md
+++ b/.changeset/ten-cobras-fetch.md
@@ -1,0 +1,6 @@
+---
+"hash-slash": patch
+---
+
+Fix router update when calling navitate\
+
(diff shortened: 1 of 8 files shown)


examples/chat/src/chatScreen.tsx
```diff
@@ -61,7 +64,10 @@ export function ChatScreen(props: { chatID: string }) {
           chat
             .slice(-showNLastMessages)
             .reverse() // this plus flex-col-reverse on ChatBody gives us scroll-to-bottom behavior
```
joeinnes: Might it be helpful here to add a bit to the comment clarifying that we can only reverse because we already sliced and so are not working on the value directly to avoid confusion, or is that too verbose? I guess TS would warn if someone tries to mutate, so maybe it’s overkill?

.changeset/ten-cobras-fetch.md
```diff
@@ -0,0 +1,6 @@
+---
+"hash-slash": patch
+---
+
+Fix router update when calling navitate\
```
joeinnes:
```suggestion
Fix router update when calling navigate\
```

PR #2698, opened 2025-08-04
feat: Add `.pick()` and `.partial()` methods to CoMapSchema

nrainhart:
# Description

These methods are currently available on Zod object schemas but not on CoMap schemas. Having these methods available makes it easier to work with "draft" CoMaps.

## Tests

## Checklist


gdorsi (approved):
Wonderful!

Added some docs, removed the lodash-es dependency and aligned pick to the Zod API.

[.partial](https://zod.dev/api?id=partial) also supports a Mask in Zod, but that's something we can add later.

PR #2699, opened 2025-08-04
fix: fixes error when extending a group without having child groups loaded

gdorsi:
# Description

The isSelfExtend check was using the child groups to check circular deps, but could fail since child groups are not loaded as dependencies.

Fixed the bug by using parent groups for the same check, which are always loaded because are considered as dependencies.

Found the bug while trying to add a track to a playlist in the music player example.

Now it should be working.

## Tests

## Checklist

### DIFF
--- a/packages/cojson/src/tests/testUtils.ts
+++ b/packages/cojson/src/tests/testUtils.ts
@@ -40,6 +40,14 @@ export function randomAgentAndSessionID(): [ControlledAgent, SessionID] {
   return [new ControlledAgent(agentSecret, Crypto), sessionID];
 }
 
+export function agentAndSessionIDFromSecret(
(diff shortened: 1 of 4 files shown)


packages/cojson/src/tests/testUtils.ts
```diff
@@ -551,6 +563,13 @@ export function setupTestNode(
 
       return node;
     },
+    spawnNewSession: () => {
+      return setupTestNode({
+        secret: node.agentSecret,
+        connected: opts.connected,
+        isSyncServer: opts.isSyncServer,
+      });
+    },
```
gdorsi, line 572: Added an utility to spawn a new session of the same agent to test sync on agents with the same permissions.

PR #2702, opened 2025-08-05
Feat/add input label

Sammii-HK:
# Description

Adding input and label components to quint

## Tests

## Checklist


packages/quint-ui/src/components/label.tsx
```diff
@@ -0,0 +1,34 @@
+import clsx from "clsx";
```
Elfo404: we should use `cn` instead so that tailwind classes are properly handled when they conflict

```suggestion
import { cn } from "../lib/utils.js";
```

packages/quint-ui/src/components/label.tsx
```diff
@@ -0,0 +1,34 @@
...
+      )}
+      {...props}
+    />
+  );
+}
+
+const label = tv({
+  base: "block text-sm font-medium text-stone-900 dark:text-white flex items-center",
+  variants: {
+    isHidden: {
```
Elfo404: what about `visuallyHidden` instead to better indicate that's only hidden visually but not to screenreaders?

Sammii-HK: so I wanted to keep the 'is' pattern prefix for the boolean and as 'hidden' is the main word in the const I wanted that to come first for DX/pre-fill (as typing 'v' for hiding something it's smooth and involves thought IMO), so came to 'isHiddenVisually', WDYT?

packages/quint-ui/src/components/label.tsx
```diff
@@ -0,0 +1,38 @@
...
+    isHiddenVisually: {
+      true: "sr-only",
+    },
+    size: {
+      sm: "text-sm",
+      md: "text-base",
+      lg: "text-lg",
+    },
+  },
+});
```
Elfo404: here we can do something like this instead:

```suggestion
import { tv, VariantProps } from "tailwind-variants";
// biome-ignore lint/correctness/useImportExtensions: <explanation>
import { cn } from "../lib/utils";
import { ComponentProps } from "react";

type LabelVariants = VariantProps<typeof label>;

interface LabelProps extends ComponentProps<"label">, LabelVariants {}

export default function Label({
  size,
  isHiddenVisually,
  ...props
}: LabelProps) {
  return (
    <label
      className={cn(
        label({
          isHiddenVisually,
          size,
        }),
        props.className,
      )}
      {...props}
    />
  );
}

const label = tv({
  base: "block text-sm font-medium text-stone-900 dark:text-white flex items-center",
  variants: {
    isHiddenVisually: {
      true: "sr-only",
    },
    size: {
      sm: "text-sm",
      md: "text-base",
      lg: "text-lg",
    },
  },
  defaultVariants: {
    size: "md",
    isHiddenVisually: false,
  },
});

```

there are a few key differences:

- we don't pass down to the underlying label element `size` and `isVisuallyHidden` props, as they are not valid props for the label element
- the `type LabelVariants = VariantProps<typeof label>;`  (can also be inlined) automatically extracts all the props need from the tv definition without you having to keep them in sync
- default variants are handled automatically by tv, so you don't have to add logic to the component render.


same things can be done for the input

packages/quint-ui/src/components/label.tsx
Elfo404: yep, makes sense!

packages/quint-ui/src/components/label.tsx
Sammii-HK: How's it looking now?

Elfo404 (approved):

PR #2704, opened 2025-08-06
fix: skip rotateKey when everyone has read access

gdorsi:
# Description

A user reported that was getting `Error: Can't extend group without parent read key secret` when extending a group

Digging into the issue I've found that in the key rotation we weren't revealing the read key to everyone/reader 

This would lead the Group to be in an inconsistent state, where everyone has a reader role but lack access to the read key.

Fixed by making rotateKey a no-op when everyone has reader access.

## Tests

### DIFF
--- a/packages/cojson/src/coValueCore/coValueCore.ts
+++ b/packages/cojson/src/coValueCore/coValueCore.ts
@@ -16,25 +16,15 @@ import {
   SignerID,
   StreamingHash,
 } from "../crypto/crypto.js";
-import {
-  RawCoID,
-  SessionID,
-  TransactionID,
-  getParentGroupId,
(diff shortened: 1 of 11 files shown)


packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -793,183 +781,43 @@ export class CoValueCore {
     }
   }
 
+  readKeyCache = new Map<KeyID, KeySecret>();
   getReadKey(keyID: KeyID): KeySecret | undefined {
```
gdorsi, line 772: Moved the `getReadKey` implementation inside group, to be able to access it from the group constructor without having to pass through `getCurrentContent`

PR #2706, opened 2025-08-07
New BetterAuth integration

Takeno:
~~Based on #2681~~

# Description
The previous BetterAuth plugin to store Jazz's credentials was based on async calls during the authentication flow, which caused multiple issues:
- If something happens in the middle of calls, a BA user is created, but without Jazz's credentials
- On the server side, devs couldn't get the account's data using the registration hooks, as Jazz's account ID is stored later
- The client plugin exposes signIn and signUp methods to manage credentials, and the BA client instance for the rest of the calls
- Most of the logic is inside the React implementation, making the porting to other frameworks harder
- The plugin system was spread into 4 different packages: `jazz-auth-betterauth`, `jazz-betterauth-client-plugin`, `jazz-betterauth-server-plugin`, `jazz-react-auth-betterauth`.

The new plugin system is based on Better Auth's plugin hooks. Now:
- Plugins and components are now exposed from the `jazz-tools/better-auth/auth` package
- Jazz's credentials are injected in sign-up calls and stored together with the account
- BetterAuth hooks now have the `accountID` property always accessible for logged accounts
- Jazz's authentication now happens synchronously with the response of signUp or signIn calls
- The Better Auth <-> Jazz communication happens inside the client/server plugins, and devs use the better auth native calls; the framework implementation is now trivial
- The credentials are still encrypted exactly like before, and the migration to the new plugin system should not require any migration on the database
- Logout and account deletion, too, are now managed on the plugin

Check the doc

## Tests

## Checklist


packages/jazz-tools/src/better-auth/auth/client.ts
```diff
@@ -0,0 +1,169 @@
+import type { BetterAuthClientPlugin } from "better-auth";
+import type {
+  AuthSecretStorage,
+  AuthSetPayload,
+  Account,
+  JazzContextType,
+} from "jazz-tools";
+import { jazzPlugin } from "./server";
```
Elfo404: i believe you want this
```suggestion
import { jazzPlugin } from "./server.js";
```

packages/jazz-tools/src/better-auth/auth/client.ts
```diff
@@ -0,0 +1,169 @@
...
+                return;
+              }
+
+              if (!context.data?.user) {
+                return;
+              }
+
+              if (authSecretStorage.isAuthenticated === false) {
+                console.info(
+                  "Jazz is not authenticated, using Better Auth stored credentials",
```
Elfo404, line 132: did you mean this?
```suggestion
                  "Jazz is authenticated, using Better Auth stored credentials",
```

packages/jazz-tools/src/better-auth/auth/react.tsx
```diff
@@ -0,0 +1,105 @@
...
+ *  }}
+ *  sync={{
+ *    peer: "ws://localhost:4200",
+ *  }}
+ * >
+ *   <App />
+ * </JazzReactProviderWithBetterAuth>
+ * ```
+ */
+export const JazzReactProviderWithBetterAuth = <
```
Elfo404, line 91: neat!

packages/jazz-tools/src/better-auth/auth/tests/client.test.ts
```diff
@@ -0,0 +1,247 @@
+import { createAuthClient } from "better-auth/client";
+import type { Account, AuthSecretStorage } from "jazz-tools";
+import {
+  TestJazzContextManager,
+  createJazzTestAccount,
```
Elfo404: leftover?
```suggestion
```

packages/jazz-tools/src/better-auth/auth/tests/client.test.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    authClient = createAuthClient({
+      baseURL: "http://localhost:3000",
+      plugins: [jazzPluginClient()],
+      fetchOptions: {
+        customFetchImpl,
+      },
+    });
+
+    authClient.jazz.setJazzContext(jazzContextManager.getCurrentValue()!);
```
Elfo404: nit: 
```suggestion
    const context = jazzContextManager.getCurrentValue();
    assert(context, "Jazz context is not available");
    authClient.jazz.setJazzContext(context);
```

packages/jazz-tools/src/better-auth/auth/tests/client.test.ts
```diff
@@ -0,0 +1,247 @@
...
+            updatedAt: new Date(),
+          },
+          jazzAuth: {
+            accountID: credentials!.accountID,
+            secretSeed: credentials!.secretSeed,
+            accountSecret: credentials!.accountSecret,
+          },
+        }),
+      ),
+    );
```
Elfo404: nit to avoid non-null assertions: (and for later access to `credentials`)
```suggestion
    const credentials = await authSecretStorage.get();
    expect(authSecretStorage.isAuthenticated).toBe(false);
    assert(credentials, "Jazz credentials are not available");

    customFetchImpl.mockResolvedValue(
      new Response(
        JSON.stringify({
          token: "6diDScDDcLJLl3sxAEestZz63mrw9Azy",
          user: {
            id: "S6SDKApdnh746gUnP3zujzsEY53tjuTm",
            email: "test@jazz.dev",
            name: "Matteo",
            image: null,
            emailVerified: false,
            createdAt: new Date(),
            updatedAt: new Date(),
          },
          jazzAuth: {
            accountID: credentials.accountID,
            secretSeed: credentials.secretSeed,
            accountSecret: credentials.accountSecret,
          },
        }),
      ),
    );
```

Elfo404 (approved):
LGTM, i'm just not a big fan of the non-null assertions, up to you if you want to add a few asserts to avoid them.

homepage/homepage/content/docs/authentication/better-auth.mdx
```diff
@@ -0,0 +1,255 @@
+export const metadata = {
+  description: "Integrate Better Auth with your Jazz app to authenticate users."
+};
+import { CodeGroup, ContentByFramework } from "@/components/forMdx";
+import { Alert } from "@garden-co/design-system/src/components/atoms/Alert";
+
+
+# Better Auth authentication
+
+[Better Auth](https://better-auth.com/) is a framework-agnostic authentication and authorization framework for TypeScript. Jazz can be integrated with Better Auth for the authentication phase, to store the [Jazz's Account keys](./overview) on Better Auth's users.
```
joeinnes:
```suggestion
[Better Auth](https://better-auth.com/) is a self-hosted, framework-agnostic authentication and authorisation framework for TypeScript. 

You can integrate Better Auth with your Jazz app, allowing your Jazz user's account keys to be saved with the corresponding Better Auth user.
```

homepage/homepage/content/docs/authentication/better-auth.mdx
```diff
@@ -0,0 +1,255 @@
...
+Despite other cloud solutions, Better Auth is self-hosted.
```
joeinnes:
```suggestion
```

homepage/homepage/content/docs/authentication/better-auth.mdx
```diff
@@ -0,0 +1,255 @@
...
+1. Users sign up or sign in through Better Auth's authentication system
+2. Jazz securely stores the user's account keys with Better Auth
+3. When logging in, Jazz retrieves these keys from Better Auth
+4. Once authenticated, users can work offline with full Jazz functionality
+
+This authentication method is not fully local-first, as login and signup need to be done online, but once authenticated, users can use all of Jazz's features without needing to be online.
+
+## Authentication methods and plugins
+
+Better Auth supports several authentication methods and plugins. The Jazz plugin has not been tested with all of them yet. Here is the compatibility matrix:
```
joeinnes, line 26: At some point I will try to work through testing these so we can at least remove the ❓

homepage/homepage/content/docs/authentication/better-auth.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+✅: tested and working
+❓: not tested yet
+❌: not supported
+
+## Getting started
+
+First of all, follow the [Better Auth documentation](https://www.better-auth.com/docs/installation) to install Better Auth:
+- Install the dependency and set env variables
+- Create the betterAuth instance in the common `auth.ts` file, using the database adapter you want. {/* here the assist for next Jazz database adapter */}
```
joeinnes, line 83:
```suggestion
- Create the betterAuth instance in the common `auth.ts` file, using the database adapter you want.
```

homepage/homepage/content/docs/authentication/better-auth.mdx
```diff
@@ -0,0 +1,255 @@
...
+  // like database, email/password authentication, social providers, etc.
+});
+```
+</CodeGroup>
+
+Now run [migrations](https://www.better-auth.com/docs/concepts/database#running-migrations) to add the new fields to the users table.
+
+<Alert variant="warning" className="mt-4">
+  **Note**: the server-side plugin intercepts the custom header `x-jazz-auth` sent by client-side plugin. If server is behind a proxy, the header must be forwarded. If the server runs on a different origin than the client, the header must be allowed for cross-origin requests.
+</Alert>
```
joeinnes:
```suggestion
<Alert variant="warning" className="mt-4" title="Note">
  The server-side plugin intercepts the custom header `x-jazz-auth` sent by client-side plugin. If server is behind a proxy, the header must be forwarded. If the server runs on a different origin than the client, the header must be allowed for cross-origin requests.
</Alert>
```

homepage/homepage/content/docs/authentication/better-auth.mdx
```diff
@@ -0,0 +1,255 @@
...
+// src/App.tsx
+"use client";
+
+import { AuthProvider } from "jazz-tools/better-auth/auth/react";
+import { betterAuthClient } from "@/lib/auth-client";
+
+
+export function App() {
+  return (
+    /* The other providers */
```
joeinnes:
```suggestion
    /* Other providers (e.g. your Jazz Provider) */
```

I'd rather show the provider in context, but that would mean a new CodeGroup for each framework, which is a bit of a headache.

homepage/homepage/content/docs/authentication/better-auth.mdx
```diff
@@ -0,0 +1,255 @@
...
+  );
+}
+```
+</CodeGroup>
+
+</ContentByFramework>
+
+<ContentByFramework framework={["vanilla", "svelte"]}>
+
+Set JazzContext and AuthSecretStorage in the better auth client:
```
joeinnes:
```suggestion
Set JazzContext and AuthSecretStorage in the Better Auth client:
```

homepage/homepage/content/docs/authentication/better-auth.mdx
```diff
@@ -0,0 +1,255 @@
...
+```
+</CodeGroup>
+
+</ContentByFramework>
+
+## Authentication methods
+
+The Jazz plugin intercepts the Better Auth client's calls, so you can use the Better Auth [methods](https://www.better-auth.com/docs/basic-usage) as usual.
+
+Here is how to signUp with email and password, and transform the Jazz's anonymous account into a logged in user:
```
joeinnes:
```suggestion
Here is how to sign up with email and password, and transform an anonymous Jazz account into a logged in user authenticated by Better Auth:
```

homepage/homepage/content/docs/authentication/better-auth.mdx
```diff
@@ -0,0 +1,255 @@
...
+      if (account?.me?.profile) {
+        account.me.profile.name = "John Doe";
+      }
+    },
+  }
+);
+```
+</CodeGroup>
+
+And here is how to signIn with email and password and signOut:
```
joeinnes:
```suggestion
You can then use the `signIn` and `signOut` methods on the `betterAuthClient`:
```

homepage/homepage/content/docs/authentication/better-auth.mdx
```diff
@@ -0,0 +1,255 @@
...
+  password: "password",
+});
+
+await betterAuthClient.signOut();
+```
+</CodeGroup>
+
+## Authentication states
+
+Although Better Auth is not fully local-first, the Jazz's client plugin tries to keep the Jazz's authentication state in sync with Better Auth's. The best practice to check if the user is authenticated is using the Jazz's methods [described here](./authentication-states#detecting-authentication-state).
```
joeinnes:
```suggestion
Although Better Auth is not fully local-first, the Jazz client plugin tries to keep Jazz's authentication state in sync with Better Auth's. The best practice to check if the user is authenticated is using Jazz's methods [as described here](./authentication-states#detecting-authentication-state).
```

homepage/homepage/content/docs/authentication/better-auth.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+If you want to check the Better Auth's state, you can use its [native methods](https://www.better-auth.com/docs/basic-usage#session).
```
joeinnes:
```suggestion
You can use Better Auth's [native methods](https://www.better-auth.com/docs/basic-usage#session) if you need to check the Better Auth state directly.
```

homepage/homepage/content/docs/authentication/better-auth.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+
+## Server-side hooks
+
+Better Auth provides [database hooks](https://www.better-auth.com/docs/reference/options#databasehooks) to run code when things happen. When using Jazz's plugin, the Jazz's AccountID is always available in the `user` object. This means you can access it in any part of Better Auth's hooks.
```
joeinnes:
```suggestion
Better Auth provides [database hooks](https://www.better-auth.com/docs/reference/options#databasehooks) to run code when things happen. When using the Jazz, the user's Jazz account ID is always available in the `user` object. This means you can access it anywhere in Better Auth hooks.
```

joeinnes (commented):
Super detailed and helpful guide, few wording nits.

packages/jazz-tools/src/better-auth/auth/client.ts
Takeno: No? If `authSecretStorage.isAuthenticated` is false, then Jazz context is _not_ authenticated.

homepage/homepage/content/docs/authentication/better-auth.mdx
Takeno: It's also on my roadmap. We decided to postpone deep tests with the auth methods and plugins after the publication.

homepage/homepage/content/docs/authentication/better-auth.mdx
Takeno: This was intentional :D #2681

homepage/homepage/content/docs/authentication/better-auth.mdx
Takeno: Yes, I thought the same. That's why I used a more general form.

packages/jazz-tools/src/better-auth/auth/client.ts
Elfo404: 🤦 i misread it, disregard.

homepage/homepage/content/docs/authentication/better-auth.mdx
joeinnes: Ah, so it's just foreshadowing that alternative database adapters are Coming Soon™? In that case ignore my comment!

gdorsi (approved):
Let's go! 🚀

kriszlazar:
super excited! 🥳

aeplay (approved):
Nice!!

PR #2710, opened 2025-08-08
fix: handle missing child groups when rotating key

gdorsi:
# Description

Makes the key rotation not fail when child groups are unavailable or their readkey is not accessible.

Also changes the Group.removeMember method to not return a Promise, because:
- All the locally available child groups are rotated immediately
- All the remote child groups are rotated in background, but since they are not locally available the user won't need the new key immediately

## Tests

## Checklist

### DIFF
--- a/packages/cojson/src/coValues/group.ts
+++ b/packages/cojson/src/coValues/group.ts
@@ -65,18 +65,56 @@ export type GroupShape = {
   [child: ChildGroupReference]: "revoked" | "extend";
 };
 
+// We had a bug on key rotation, where the new read key was not revealed to everyone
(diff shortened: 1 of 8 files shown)


packages/cojson/src/coValues/group.ts
```diff
@@ -827,16 +798,26 @@ export class RawGroup<
...
       }
 
-      child.rotateReadKey(removedMemberKey);
-    }
+      try {
+        child.rotateReadKey(removedMemberKey);
+      } catch (error) {
+        if (error instanceof NoReadKeyAccessError) {
+          logger.warn(
+            `Can't rotate read key on child ${child.id} because we don't have access to the read key`,
```
gdorsi, line 858:
```suggestion
            `Can't rotate read key on child ${child.id} because we don't have access to the child read key`,
```

PR #2716, opened 2025-08-09
Fix return type on deep loaded co.record() when using single keys

Takeno:
# Description
Deep loading on the co record using single keys returned `never` even if the data is returned properly.
This PR fix that never.

```ts
const Dog = co.map({
  name: z.string(),
  breed: z.string(),
});

const Person = co.record(z.string(), Dog);

Person.create({
  pet1: Dog.create({ name: "Rex", breed: "Labrador" }),
  pet2: Dog.create({ name: "Fido", breed: "Poodle" }),
});


// BEFORE
const p = await Person.load("id", {
  resolve: {
    pet1: true
  }
});

p.pet1 // <- never


// NOW
const p = await Person.load("id", {
  resolve: {
    pet1: true
  }
});

p.pet1 // <- {name: string; breed: string;} & CoMap

```

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/deepLoading.ts
+++ b/packages/jazz-tools/src/tools/coValues/deepLoading.ts
@@ -97,6 +97,27 @@ type onErrorNullEnabled<Depth> = Depth extends { $onError: null }
   ? null
   : never;
 
+type CoMapLikeLoaded<
+  V extends object,
+  Depth,
+  DepthLimit extends number,
+  CurrentDepth extends number[],
+> = {
+  -readonly [Key in keyof Depth]-?: Key extends CoKeys<V>
+    ? NonNullable<V[Key]> extends CoValue
+      ?
+          | DeeplyLoaded<
+              NonNullable<V[Key]>,
+              Depth[Key],
+              DepthLimit,
+              [0, ...CurrentDepth]
+            >
+          | (undefined extends V[Key] ? undefined : never)
+          | onErrorNullEnabled<Depth[Key]>
+      : never
+    : never;
+} & V;
+
 export type DeeplyLoaded<
   V,
   Depth,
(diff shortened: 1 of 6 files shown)


packages/jazz-tools/src/tools/tests/coMap.record.test-d.ts
```diff
@@ -173,6 +173,103 @@ describe("CoMap.Record", () => {
...
+        : "ok";
+
+      expectTypeOf("ok" as const).toEqualTypeOf<Expect>();
+
+      expectTypeOf(loadedPerson?.pet1).toEqualTypeOf<
+        Loaded<typeof Dog> | undefined
+      >();
+      expectTypeOf(loadedPerson?.pet3).toEqualTypeOf<
+        Loaded<typeof Dog> | undefined
+      >();
```
gdorsi, line 249: Added a new test, to cover the type resolution when the resolve query has a `string` property

In the CoMap this case is resolved as `Loaded<typeof Dog> | null`
It's the correct resolution, because since we don't know which field is loaded, we should consider as all of them as not directly loaded.
It's a minor issue, but in this specific case `loadedPerson.pet3` will be null at runtime so the type is not fully correct.

Takeno: Fixed in [packages/jazz-tools/src/tools/coValues/deepLoading.ts:167-170](https://github.com/garden-co/jazz/pull/2716/files#diff-d2501375775e11bd359cd17bd61c10a6976c6503cb8576785951bea4b91755afR167-R170) and [updated tests](https://github.com/garden-co/jazz/pull/2716/files#diff-7b77a7c45a5bbfce3ead3411ad61c377fda5b210c30f3be99dcbaaedf5e88cfcR225-R249)

packages/jazz-tools/src/tools/coValues/deepLoading.ts
```diff
@@ -126,38 +147,31 @@ export type DeeplyLoaded<
...
+                        ItemDepth,
+                        DepthLimit,
+                        [0, ...CurrentDepth]
+                      >
+                    | onErrorNullEnabled<Depth["$each"]>;
+                } & V // same reason as in CoList
+              : // 1.2. Deeply loaded Record-like CoMap with { [key: string]: true }
+                string extends keyof Depth
+                ? // if at least one key is `string`, then we treat the resolve as it was empty
+                  DeeplyLoaded<V, {}, DepthLimit, [0, ...CurrentDepth]> & V
```
gdorsi, line 170: We should continue to navigate the depth because the user might do something like this

```ts
  const userId: string = "pet1";
  const loadedPerson = await Person.load(person.id, {
        resolve: {
          [userId]: { avatar: true } ,
        },
      });
```

```suggestion
                  DeeplyLoaded<V, Depth[string], DepthLimit, [0, ...CurrentDepth]> & V
```

PR #2724, opened 2025-08-11
feat(jazz-tools/media): `createImage()` for servers

Gabrola:
# Description
- Changed `createImageFactory` to use generics in place of the image input type and the `resize()` function return value. Leaving full freedom to to be able to pass types other than `SourceType`. In this case used to be able to pass Buffers. This is fully backwards compatible.
- Added the a new `createImage()` function that uses the "sharp" library for server environments

# Exports

- Since the types for browser + server vis-a-vis react native are not overlapping, package.json exports have been adjusted such that users with correctly setup tsconfig.json (`react-native`, etc. in `customConditions`) shall see the correct platform-specific types
- Otherwise, all overloads are exposed to typescript and we fallback onto the runtime check which has been refactored into `imageTypeGuard`
- The server `createImage()` was exposed on `index.server.ts` for SSR environments to consume it. `import { createImage } from 'jazz-tools/media';` should now work isomorphically between browser and server environments.
- Exposed each platform in separate exports so apps that are not cross-platform can consume them directly and guarantee having the correct types.

I don't have super strong feelings towards the exports changes. If it feels like it's too much, I can just revert this part; just need to determine where and under what export name to expose the server `createImage()`.

I honestly feel since the native `createImage` and browser/node `createImage` do not share their call signatures or even overlap, there's little benefit to exposing them through the same export. A user would still need to write extra logic to handle the different platforms.

## Tests

## Checklist


Takeno:
Thank you for this contribution @Gabrola!

The reason why `createImage` was designed with the same import for both browser and React Native, keeping in mind a future integration of https://github.com/mrousavy/react-native-nitro-image. Using the lib, it will be possible to accept Blobs also in the React Native platform, overlapping the signatures.

But with wider options from server libraries like Sharp, the common signature is actually overly restrictive.

As long as we're still defining the proper organization for server-side stuff (like requests, workers, and so on), I would keep the `import {createImage} from "jazz-tools/media/server` for now and ship this useful feature.

For browser and react-native, I would keep them as they are for now (without the additional export), following the current docs, but I keep the point on the roadmap to find the best DX on this.

I'm going to merge main into this branch to get some fixes done on the React Native implementation, and minor changes to the export as described above.

gdorsi (approved):

PR #2725, opened 2025-08-12
feat(jazz-tools/react): add useCoStateWithSelector for granular reactivity

Gabrola:
# Description
Half of this functionality could have been implemented by piping the existing `useSyncExternalStore`'s `getSnapshot` return value through the selector function, but still would not have the ability to use a custom equality function which can be pretty powerful when combined fast shallow or deep equality libs.

## Tests

## Checklist


aeplay (changes requested):
Hey @Gabrola this looks like a powerful API and thanks so much for the high quality contribution.

~~I think I would prefer to call it `useMappedCoState` and call the callback `map` since it can potentially do more than selecting. What do you think?~~

TIL about useSyncExternalStoreWithSelector - happy to align ourselves with that nomenclature, so you can keep things as they are! What about `useCoStateWithSelector` for exact correspondence? (Sounds more natural to me)

Gabrola:
@aeplay yeah `useCoStateWithSelector` makes more sense to me as well. Happy to change it into that.

What do you think actually of just merging the implementation into `useCoState` and do something like this?

```ts
export function useCoState<
  S extends CoValueClassOrSchema,
  const R extends ResolveQuery<S> = true,
  TReturn = Loaded<S, R> | undefined | null,
>(
  /** The CoValue schema or class constructor */
  Schema: S,
  /** The ID of the CoValue to subscribe to. If `undefined`, returns `null` */
  id: string | undefined,
  /** Optional configuration for the subscription */
  options?: {
    /** Resolve query to specify which nested CoValues to load */
    resolve?: ResolveQueryStrict<S, R>;
    /** Transform or select part of the loaded CoValue */
    select?: (value: Loaded<S, R> | undefined | null) => TReturn;
    /** Custom equality function, if omitted `Object.is` will be used */
    equalityFn?: (a: TReturn, b: TReturn) => boolean;
  },
): TReturn {
  const subscription = useCoValueSubscription(Schema, id, options);

  return useSyncExternalStoreWithSelector<
    Loaded<S, R> | undefined | null,
    TReturn
  >(
    React.useCallback(
      (callback) => {
        if (!subscription) {
          return () => {};
        }

        return subscription.subscribe(callback);
      },
      [subscription],
    ),
    () => (subscription ? subscription.getCurrentValue() : null),
    () => (subscription ? subscription.getCurrentValue() : null),
    options?.select ?? ((value) => value as unknown as TReturn),
    options?.equalityFn ?? Object.is,
  );
}
```

This should be fully backwards compatible (`useSyncExternalStore` internally uses `Object.is` so this should be a drop-in replacement), and no need to use a hook name as verbose as `useCoStateWithSelector`. Very similar actually to tanstack react-query's `useQuery` with the `select` option.

I can see the benefit of possibly wanting to use a custom equality function without a selector.

aeplay:
@Gabrola I was also wondering if it makes sense to just roll it into the main `useCoState`. It would be more discoverable, but would also make useCoState look more complicated for the normal use-case. For now I think I'd prefer a separate hook to make it clear this is a more advanced use-case. Thanks so much!

Gabrola:
@aeplay sorry, was away for a bit. But yeah, that makes sense to me. I'll change it now.

gdorsi (approved):
Sweet, thanks!

PR #2733, opened 2025-08-13
chore: extract the sessions management into a SessionMap class

gdorsi:
Extracting the new Session logic from verifiedState into a separate class in order to simplify the alignement with main.


packages/cojson/src/coValueCore/SessionMap.ts
```diff
@@ -0,0 +1,231 @@
...
+    if (
+      sessionLog.txSizeSinceLastInbetweenSignature > MAX_RECOMMENDED_TX_SIZE
+    ) {
+      sessionLog.signatureAfter[sessionLog.transactions.length - 1] = signature;
+      sessionLog.txSizeSinceLastInbetweenSignature = 0;
+    }
+  }
+
+  // TODO: Shall we get rid of this?
+  testExpectedHashAfter(
```
gdorsi, line 162: This method is used only on some tests.

I think it would be good to get rid of it, if you agree I can do the work directly on main.

PR #2735, opened 2025-08-13
feat: use multiple catalogs for easier version bumps

boorad:
# Description
There are groups of dependencies with React, React Native, Expo, etc. that should and/or have to be in sync.  Bumping the versions of these dependencies in the monorepo isn't a difficult exercise with find/replace, but it is still error-prone to have the compatible versions agree across all projects using them.

This PR uses expanded `pnpm` catalogs that make these groups of dependencies more manageable.

## Open questions:
- ~~What is the `scripts/check-catalog-deps.js` file supposed to do?  If we get these catalogs correct, does this go away?~~
- Did I clobber some `react` and `react-dom` versions that are required for certain dependencies in the `homepage` package?  If so, we can have `catalog:homepage` with those versions.

`pnpm i`

## Checklist


Gabrola:
In my experience a lot of tools do not support pnpm catalogs; even some pnpm commands like `pnpm deploy` (this might have changed in newer pnpm versions) do not work with pnpm catalogs. The `examples` projects as well would now be kinda unusable without using `create-jazz-app`.

You can use the `syncpack` CLI tool to solve this problem and it can group dependencies such that something like `react` and `react-dom`, for example, are always in sync.

gdorsi:
Thanks @Gabrola for your feedback!

>  The examples projects as well would now be kinda unusable without using create-jazz-app.
It kinda already is, given that we use `workspace:*` deps

For us `catalog:` is good enough for now, given the way we use `pnpm` and deploy stuff.
We are already using it outside the example apps, and the main blocker of using it there was create-jazz-app support for catalog replacement.

If things go wrong with catalog `syncpack` is surely going to be evaluated by us.

PR #2743, opened 2025-08-15
feat: add `CoList.$jazz.remove` and `CoList.$jazz.retain`

nrainhart:
# Description

A lot of people aren’t very familiar/good with `splice`, so we're adding this two alternative methods that are easier to use.

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/coList.ts
+++ b/packages/jazz-tools/src/tools/coValues/coList.ts
@@ -614,15 +614,9 @@ export class CoListJazzApi<L extends CoList>
         if (predicate(this.coList[i], i, this.coList)) indices.push(i);
       }
     } else {
-      indices = args as number[];
-      const outOfBoundIndices = indices.filter(
-        (index) => index < 0 || index >= this.coList.length,
-      );
-      if (outOfBoundIndices.length > 0) {
-        throw new Error(
-          `Indices [${outOfBoundIndices}] out of bounds for length ${this.coList.length}`,
-        );
-      }
+      indices = (args as number[])
+        .filter((index) => index >= 0 && index < this.coList.length)
+        .sort();
     }
     const deletedItems = indices.map((index) => this.coList[index]);
     for (const index of indices.reverse()) {
--- a/packages/jazz-tools/src/tools/coValues/coList.ts
+++ b/packages/jazz-tools/src/tools/coValues/coList.ts
@@ -584,6 +584,54 @@ export class CoListJazzApi<L extends CoList>
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -579,6 +579,67 @@ export class CoListJazzApi<L extends CoList>
...
+    } else {
+      indices = args as number[];
+      const outOfBoundIndices = indices.filter(
+        (index) => index < 0 || index >= this.coList.length,
+      );
+      if (outOfBoundIndices.length > 0) {
+        throw new Error(
+          `Indices [${outOfBoundIndices}] out of bounds for length ${this.coList.length}`,
+        );
+      }
```
gdorsi: Unsure about having this throwing an error.

Would it work to skip the invalid indexes and log a warning?

gdorsi (approved):
LGTM

Left a comment on the throw Error

packages/jazz-tools/src/tools/coValues/coList.ts
nrainhart: Hmmmm I kind of went on autopilot mode when writing this method, but now that I think of it... should we even allow removing elements by index? When editing CoLists collaboratively, if you got the index by reading from a stale copy of the list, it's possible that the index is now out-of-bounds (or worse, it could refer to a different element that the one you meant to delete).

Not sure if this will be a problem in practice, but maybe it'd be safe to just accept elements to remove instead of indices? (and do nothing if it's no longer in the CoList)

gdorsi: Good point.

We can start by accepting a predicate, and add more complexity on the future.

nrainhart: On the other hand... we already have an index-based CoList-mutation method (splice). And it never throws an out-of-bounds exception. So I think it's fine to just:
1. keep the index-based remove method
2. do nothing if an index is out of bounds (not even show a warning)

aeplay: deleting by index is fine *if you do it on the same stale view of the CoList that you got the index from*<br><br>My assumption was this was almost always the case?

nrainhart: Yeah, I agree. My concern was mainly theoretical, given that the moment where you get the index may not be synchronous with the moment where the removal happens. But I'm pretty sure this won't be a problem in practice (and fwiw, other CRDTs like Yjs support index-based deletions, so that seems to confirm it)

PR #2751, opened 2025-08-15
fix: remove unnecessary content sent as dependency

gdorsi:
# Description

While checking the test I've noticed some redundant messages sent around involving groups.

Digging deeper I noticed that the `knownStateFromContent` utility didn't fill the header value, always tracking the `header` as false in the optimistic known state after a local transaction.

## Tests

## Checklist


packages/cojson/src/tests/sync.garbageCollection.test.ts
```diff
@@ -163,13 +163,10 @@ describe("sync after the garbage collector has run", () => {
         "edge -> storage | LOAD Map sessions: empty",
         "storage -> edge | CONTENT Group header: true new: After: 0 New: 5",
         "storage -> edge | CONTENT Map header: true new: After: 0 New: 1",
-        "edge -> server | CONTENT Map header: true new: ",
```
gdorsi, line 166: This is the extra message we don't need.

packages/cojson/src/tests/sync.load.test.ts
```diff
@@ -973,7 +973,6 @@ describe("loading coValues from server", () => {
         "server -> client | KNOWN Group sessions: header/6",
         "server -> client | KNOWN ParentGroup sessions: header/8",
         "server -> client | KNOWN Map sessions: header/1",
-        "client -> server | CONTENT ParentGroup header: true new: ",
```
gdorsi, line 976: Same here

aeplay (approved):

nrainhart (approved):

PR #2753, opened 2025-08-15
Closes #GCO-642 - Clarify docs for adding group members by id

mculotta120:
# Description
Updates docs for `adding-group-members-by-id` to confirm the equivalence between co.account().load(...) and Account.load(...)

Run locally or test deployed version, navigate to `http://localhost:3000/docs/react/groups/intro#adding-group-members-by-id`

## Tests

## Checklist

### DIFF
--- a/homepage/homepage/content/docs/groups/intro.mdx
+++ b/homepage/homepage/content/docs/groups/intro.mdx
@@ -49,18 +49,22 @@ if (bob) {
 ```
 </CodeGroup>
 
-**Note:** if the account ID is of type `string`, because it comes from a URL parameter or something similar, you need to cast it to `ID<Account>` first:
+**Note:** 
+- Both `Account.load(id)` and `co.account().load(id)` do the same thing — they load an account from its ID.
+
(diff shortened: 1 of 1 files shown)


homepage/homepage/content/docs/groups/intro.mdx
```diff
@@ -49,18 +49,22 @@ if (bob) {
 ```
 </CodeGroup>
 
-**Note:** if the account ID is of type `string`, because it comes from a URL parameter or something similar, you need to cast it to `ID<Account>` first:
+**Note:** 
+- Both `Account.load(id)` and `co.account().load(id)` do the same thing — they load an account from its ID.
+
+- If your account ID is a string (e.g., from a URL parameter), you need to cast it to `ID<Account>` before loading:
```
gdorsi: Users shouldn't need to do that.

We should simplify the `ID<Account>` type to be a string, so no casting would be required.

mculotta120: Thanks for catching that! I added stricter casting earlier when I ran into some Twoslash errors, but that ended up overcomplicating things. I’ve simplified it back to the cleaner version, which matches the original example and runs fine now.

PR #2755, opened 2025-08-16
fix: catalog replacement for `create-jazz-app`

boorad:
# Description
The locating of `pnpm-workspaces.yml` file worked perfectly in local dev, because it's in the monorepo.  However, upon distributing the npm module, the file was no longer present, and replacement of `catalog:` with proper versions failed.  

regression in #2735 

This PR adds the workspaces file to the distribution, and adds CI tests for `pre-release` to catch this in the future.

## Tests

## Checklist


.github/workflows/pre-release.yml
```diff
@@ -107,3 +107,67 @@ jobs:
                 await logPublishInfo();
               }
             }
+
+  test-create-jazz-app-distribution:
+    if: contains(github.event.pull_request.labels.*.name, 'pre-release')
+    runs-on: ubuntu-latest
```
gdorsi: Nice one, thanks!

Why is this located in the `pre-release` workflow?

How about running it when there is a change inside the `create-jazz-app` directory?

.changeset/config.json
```diff
@@ -9,6 +9,7 @@
       "cojson-storage-indexeddb",
       "cojson-storage-sqlite",
       "cojson-transport-ws",
+      "create-jazz-app",
```
gdorsi: I'd like to keep `create-jazz-app` on his own version, because it doesn't depend on any other package on the monorepo

.github/workflows/create-jazz-app.yml
```diff
@@ -0,0 +1,78 @@
...
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'packages/create-jazz-app/**'
+
+jobs:
+  test-create-jazz-app-distribution:
+    if: contains(github.event.pull_request.labels.*.name, 'pre-release')
```
gdorsi:
```suggestion
```

.changeset/config.json
gdorsi:
```suggestion
```

gdorsi (approved):

PR #2759, opened 2025-08-18
Add host option to the jazz-run sync command

0x100101:
# Description

Added a `--host` option to the jazz-run sync command. This makes it possible to bind to 0.0.0.0, which can be important when running inside Docker containers. Binding to 0.0.0.0 exposes the server on the container’s external interface, making it reachable from other containers on the same network.

Made a slight modification to `startSyncServer` so that the returned promise now resolves once the server is listening. This allows the command to display the true host/ip of the newly created server (instead of just the arguments passed to the command). This should offer a more bullet proof status message, and could also be helpful if  `--port=0` was passed and a random port was selected.

```
npx jazz-run sync --host=0.0.0.0

 ./packages/jazz-run/dist/index.js sync --host=0.0.0.0 --in-memory
```

## Tests

## Checklist

### DIFF
--- a/.changeset/tiny-panthers-check.md
+++ b/.changeset/tiny-panthers-check.md
@@ -1,0 +1,5 @@
+---
+"jazz-run": patch
+---
+
+Add host option to the jazz-run sync command
--- a/homepage/homepage/content/docs/sync-and-storage.mdx
+++ b/homepage/homepage/content/docs/sync-and-storage.mdx
@@ -46,6 +46,7 @@ In this case, provide the WebSocket endpoint your proxy exposes as the sync serv
 
 ### Command line options:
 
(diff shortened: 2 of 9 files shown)


aeplay (changes requested):
Looks great, thanks so much for the contribution! Please may you add a patch changeset for jazz-run using `pnpm changeset` on the root and add the changeset files to this PR? This will make it easy for us to release your improvement,

0x100101:
> Looks great, thanks so much for the contribution! Please may you add a patch changeset for jazz-run using `pnpm changeset` on the root and add the changeset files to this PR? This will make it easy for us to release your improvement,

No prob! I've added the changeset file. Let me know if you need anything else.

gdorsi (approved):

PR #2761, opened 2025-08-18
fix(jazz-tools/media): get resized file's id without triggering shallow load

Takeno:
# Description
Follow up of #2726


boorad (approved):
LGTM!

gdorsi (approved):

PR #2762, opened 2025-08-18
perf: switch to decryptNextTransactionChangesJson

gdorsi:
Changes:
- Switches to the new decryptNextTransactionChangesJson SessionLog method to decrypt a transaction with less data passing
- Removed expectedNewHashAfter which was used only on some tests
- Updated and fixed all the failing tests

Got a little better in the import + decrypt bench
Screenshot 2025-08-18 at 20 16 34


aeplay (approved):
😍

PR #2683, opened 2025-07-29
feat: create CoValues using plain JSON objects

nrainhart:
# Description

Allows creating CoValues using plain JSON objects.

For example:
```typescript
const Track = co.plainText();
const ListOfTracks = co.list(Track);
const Playlist = co.map({
  title: z.string(),
  tracks: ListOfTracks,
});

// Instead of
const playlist = Playlist.create({
  title: "My playlist",
  tracks: ListOfTracks.create([Track.create("Track 1")]),
});

// People can now do
const playlist = Playlist.create({ title: "My playlist", tracks: ["Track 1"] });
```

This PR also makes fields that can be `undefined` optional for create methods' inputs, instead of requiring an explicit `undefined` value.

I also took the opportunity to reduce duplication across generic types used to derive instance types from CoValue schemas.

## Tests

- [x] Tests have been added and/or updated

## Checklist

- [x] I've updated the part of the docs that are affected the PR changes
- [x] I've generated a changeset, if a version bump is required
- [ ] I've updated the jsDoc comments to the public APIs I've modified, or added them when missing

PR #2767, opened 2025-08-18
feat: support updating CoValues using JSON objects

nrainhart:
# Description

Follow-up to https://github.com/garden-co/jazz/pull/2683. CoValues can now be updated using JSON objects.

For example:
```typescript
const Todo = co.plainText();
const TodoList = co.list(Todo);
const todos = TodoList.create([]);

// Instead of
todos.$jazz.push(Todo.create("Buy groceries"));
todos.$jazz.push(Todo.create("Walk the dog"));

// People can now use
todos.$jazz.push("Buy groceries");
todos.$jazz.push("Walk the dog");
```

CoValues are created automatically, and permissions are handled in the same way than when creating CoValues from JSON (see [docs](https://jazz.tools/docs/react/groups/inheritance#group-hierarchy-on-covalue-creation)).

The update methods that support JSON inputs are:
- `CoMap.$jazz.set`
- `CoMap.$jazz.applyDiff`
- `CoList.$jazz.set`
- `CoList.$jazz.push`
- `CoList.$jazz.unshift`
- `CoList.$jazz.splice`
- `CoList.$jazz.applyDiff`
- `CoFeed.$jazz.push`

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/CoFieldInit.ts
+++ b/packages/jazz-tools/src/tools/coValues/CoFieldInit.ts
@@ -3,7 +3,6 @@ import { CoList } from "./coList";
 import { CoMap, CoMapInit } from "./coMap";
 import { CoPlainText } from "./coPlainText";
 import { CoRichText } from "./coRichText";
-import { CoValue } from "./interfaces";
 
 /**
  * Returns the type of values that can be used to initialize a field of the provided type.
@@ -11,14 +10,11 @@ import { CoValue } from "./interfaces";
  * For CoValue references, either a CoValue of the same type, or a plain JSON value that can be
  * converted to the CoValue type are allowed.
  */
-export type CoFieldInit<V> =
-  | V
-  | (V extends CoValue
-      ? V extends CoMap
-        ? CoMapInit<V>
-        : V extends CoList<infer T> | CoFeed<infer T>
-          ? ReadonlyArray<CoFieldInit<T>>
-          : V extends CoPlainText | CoRichText
-            ? string
-            : never
-      : never);
+// Note: we don't define this type as V | ... because it prevents TS from inlining the type
+export type CoFieldInit<V> = V extends CoMap
+  ? V | CoMapInit<V>
+  : V extends CoList<infer T> | CoFeed<infer T>
+    ? V | ReadonlyArray<CoFieldInit<T>>
+    : V extends CoPlainText | CoRichText
+      ? V | string
+      : V;
--- a/packages/jazz-tools/src/tools/coValues/CoFieldInit.ts
+++ b/packages/jazz-tools/src/tools/coValues/CoFieldInit.ts
@@ -1,0 +1,24 @@
+import { CoFeed } from "./coFeed";
+import { CoList } from "./coList";
+import { CoMap, CoMapInit } from "./coMap";
+import { CoPlainText } from "./coPlainText";
+import { CoRichText } from "./coRichText";
+import { CoValue } from "./interfaces";
+
+/**
+ * Returns the type of values that can be used to initialize a field of the provided type.
+ *
+ * For CoValue references, either a CoValue of the same type, or a plain JSON value that can be
+ * converted to the CoValue type are allowed.
+ */
+export type CoFieldInit<V> =
+  | V
+  | (V extends CoValue
+      ? V extends CoMap
+        ? CoMapInit<V>
+        : V extends CoList<infer T> | CoFeed<infer T>
+          ? ReadonlyArray<CoFieldInit<T>>
+          : V extends CoPlainText | CoRichText
+            ? string
+            : never
+      : never);
--- a/packages/jazz-tools/src/tools/coValues/coMap.ts
+++ b/packages/jazz-tools/src/tools/coValues/coMap.ts
@@ -639,13 +639,13 @@ class CoMapJazzApi<M extends CoMap> extends CoValueJazzApi<M> {
 
         if (descriptor === "json" || "encoded" in descriptor) {
           if (currentValue !== newValue) {
-            this.set(tKey as any, newValue);
+            this.set(tKey as any, newValue as CoFieldInit<M[keyof M]>);
           }
         } else if (isRefEncoded(descriptor)) {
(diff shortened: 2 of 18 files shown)


packages/jazz-tools/src/tools/coValues/coMap.ts
```diff
@@ -156,10 +161,12 @@ export class CoMap extends CoValueBase implements CoValue {
    * ```
    *
    * @category Creation
+   *
+   * @deprecated Use `co.map(...).create`.
```
nrainhart, line 165: I marked the static methods in the `CoMap` class as deprecated, as users should be using `co.map` schema methods instead.

packages/jazz-tools/src/tools/coValues/CoFieldInit.ts
```diff
@@ -0,0 +1,20 @@
...
+import { CoPlainText } from "./coPlainText";
+import { CoRichText } from "./coRichText";
+
+/**
+ * Returns the type of values that can be used to initialize a field of the provided type.
+ *
+ * For CoValue references, either a CoValue of the same type, or a plain JSON value that can be
+ * converted to the CoValue type are allowed.
+ */
+// Note: we don't define this type as V | ... because it prevents TS from inlining the type
```
gdorsi, line 13: 👍 

Thanks for documenting the design choice

packages/jazz-tools/src/tools/tests/coMap.test.ts
```diff
@@ -560,6 +560,92 @@ describe("CoMap", async () => {
...
+
+        expect(person.name.toString()).toEqual("Jack");
+        expect(person.bio!.toString()).toEqual("I am a lawyer");
+        expect(person.dog.name).toEqual("Fido");
+        expect(person.friends[0]!.name.toString()).toEqual("Jane");
+        expect(person.friends[0]!.dog.name).toEqual("Firulais");
+        expect(person.friends[0]!.pet.name).toEqual("Nala");
+        expect(person.reactions.byMe?.value?.toString()).toEqual("🧑‍🔬");
+        expect(person.pet.name).toEqual("Salem");
+      });
```
gdorsi, line 644: **request**: Could you split this in multiple tests?

It will help with debugging when one of the expectations fail.

packages/jazz-tools/src/tools/tests/coList.test.ts
```diff
@@ -423,6 +454,38 @@ describe("CoList applyDiff operations", async () => {
     expect(list.$jazz.raw.asArray()).toEqual([]);
   });
 
+  test("applyDiff with reference values using JSON", () => {
+    const NestedItem = co.list(z.string());
+    const RefList = co.list(NestedItem);
```
gdorsi: **suggestion:** how about making this test a bit more fun and easier to follow by using a TicTacToe grid as example?

packages/jazz-tools/src/tools/tests/coMap.test.ts
```diff
@@ -1401,6 +1487,24 @@ describe("CoMap applyDiff", async () => {
...
+      nested: NestedMap.create({ value: "original" }),
+    });
+
+    const newValues = {
+      nested: { value: "updated" },
+    };
+
+    map.$jazz.applyDiff(newValues);
+
+    expect(map.nested?.value).toEqual("updated");
```
gdorsi, line 1527: **observation:** an important detail here is how nested values are updated

Since we check equality via ids, we should add an expect checking that `applyDiff` does create new nested values.

packages/jazz-tools/src/tools/tests/coList.test.ts
```diff
@@ -423,6 +454,38 @@ describe("CoList applyDiff operations", async () => {
...
+    // Test adding reference items
+    list.$jazz.applyDiff([item1, item2, item3]);
+    expect(list.length).toBe(3);
+    expect(list[2]?.[0]).toBe("item3");
+
+    // Test removing reference items
+    list.$jazz.applyDiff([item1, item3]);
+    expect(list.length).toBe(2);
+    expect(list[0]?.[0]).toBe("item1");
+    expect(list[1]?.[0]).toBe("item3");
```
gdorsi: **observation:** an important detail here is how nested values are updated

Since we check equality via ids, we should add an expect checking that `applyDiff` does create new nested values.

packages/jazz-tools/src/tools/tests/coList.test.ts
```diff
@@ -190,32 +190,80 @@ describe("Simple CoList operations", async () => {
...
+
+      const Recipe = co.list(Ingredient);
+
+      const recipe = Recipe.create(
+        [{ name: "bread" }, { name: "butter" }, { name: "onion" }],
+        { owner: me },
+      );
+
+      recipe.$jazz.set(1, { name: "margarine" });
+      expect(recipe[1]?.name).toBe("margarine");
```
gdorsi, line 209: **suggestion:** to ensure that the spec is fully followed, how about checking that this has created a new CoValue?

gdorsi (approved):
LGTM

The only detail to define is `applyDiff`
I find his behavior hard to guess.

Some might expect a deep merge, instead of the current algorithm.
I think that this should be documented both on the docs and with a jsDoc in the API.

I think the algorithm implemented here is quite consistent, because we can't do a deep merge on co.list which would make us lose the ability of handling move operations on values.

I can see the need for a deepMerge API arise in the future, but we don't need to make things more complicated now.

aeplay:
> LGTM
> 
> The only detail to define is `applyDiff` I find his behavior hard to guess.
> 
> Some might expect a deep merge, instead of the current algorithm. I think that this should be documented both on the docs and with a jsDoc in the API.
> 
> I think the algorithm implemented here is quite consistent, because we can't do a deep merge on co.list which would make us lose the ability of handling move operations on values.
> 
> I can see the need for a deepMerge API arise in the future, but we don't need to make things more complicated now.

ApplyDiff containing CoValues should always operate shallowly (treat the references as primitive values) IMO

cc @nrainhart

gdorsi (approved):

PR #2777, opened 2025-08-19
Explicit loadAs in upsertUnique to use it without loaded context

Takeno:
# Description
Using `upsertUnique()` throws `Error: No active account` if no global active user is set.

Like in `loadCoValueWithoutMe()`, we force the loadAs using `owner._loadedAs`.


nrainhart (approved):

gdorsi (approved):

gdorsi:
@Takeno 

Merging, could you add a test in a follow-up PR?

PR #2779, opened 2025-08-19
Add Fallback properties to JazzProvider and ClerkProvider

### LINKED ISSUE #2758: Clerk Provider: Custom Loading Component
Image

the clerk auth provider renders null while loading.

this creates a blank flicker when opening my jazz site.

i would love to keep my custom loading component (looks like a splash screen) visible until i can render the final UI.

an optional custom fallback property would be great here.

carlassmann:
# Description

fixes https://github.com/garden-co/jazz/issues/2758

This PR adds a fallback property to both the JazzProvider and the JazzReactProviderWithClerk.
I need this to avoid an empty content flash while my app is loading.

You can add a loading placeholder to the HTML that is returned by the server and use the same placeholder as the fallback.
Then the app will effectively have a consistent Splash-Screen like start.

I've added a loading fallback to the index.html and JazzReactProviderWithClerk in the clerk example.
I wasn't able to execute them due to vite failing to resolve imports when i try to run the dev server in the example directory?

## Tests

- [x] I need help with writing tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/react/provider.tsx
+++ b/packages/jazz-tools/src/react/provider.tsx
@@ -20,7 +20,7 @@ export type JazzProviderProps<
 > = {
   children: React.ReactNode;
   enableSSR?: boolean;
-  fallback?: React.ReactNode;
+  fallback?: React.ReactNode | null;
 } & JazzContextManagerProps<S>;
 
 /** @category Context & Hooks */
@@ -39,7 +39,7 @@ export function JazzReactProvider<
   logOutReplacement,
   onAnonymousAccountDiscarded,
   enableSSR,
-  fallback,
+  fallback = null,
 }: JazzProviderProps<S>) {
   const [contextManager] = React.useState(
     () =>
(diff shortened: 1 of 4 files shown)


packages/jazz-tools/src/react/provider.tsx
```diff
@@ -38,6 +39,7 @@ export function JazzReactProvider<
   logOutReplacement,
   onAnonymousAccountDiscarded,
   enableSSR,
+  fallback,
```
gdorsi:
```suggestion
  fallback = null,
```

packages/jazz-tools/src/react/provider.tsx
```diff
@@ -20,6 +20,7 @@ export type JazzProviderProps<
 > = {
   children: React.ReactNode;
   enableSSR?: boolean;
+  fallback?: React.ReactNode;
```
gdorsi:
```suggestion
  fallback?: React.ReactNode | null;
```

packages/jazz-tools/src/react/provider.tsx
```diff
@@ -100,7 +102,7 @@ export function JazzReactProvider<
   return (
     <JazzContext.Provider value={value}>
       <JazzContextManagerContext.Provider value={contextManager}>
-        {value && children}
+        {!fallback ? value && children : !value ? fallback : children}
```
gdorsi:
```suggestion
        {value ? children : fallback}
```

gdorsi (approved):
Thanks!

Committed some suggestions to simplify the fallback in the React provider

PR #2785, opened 2025-08-20
fix: RN Expo E2E dep fixes w catalog

boorad:
# Description

React Native (Expo) End-to-End Tests were failing after the catalog changes.  This is partly due to the state of Expo trying to get SDK 54 out.  The `-preview.X` builds are better than `canary-` so I'm trying those, and patching up some other deps so we are consistent across the monorepo.  The `catalog:` versioning helps with this, but needs further integration.

Something freaked out the `svelte` dependencies, and they ended up getting their own `catalog:svelte`

Something also exposed a missing dev dependency for `community-jazz-vue`

## Tests

## Checklist


examples/chat-rn-expo/package.json
```diff
@@ -11,22 +12,22 @@
...
     "react-native-get-random-values": "^1.11.0",
+    "react-native-nitro-modules": "0.26.4",
     "readable-stream": "^4.7.0"
   },
   "devDependencies": {
     "@babel/core": "^7.25.2",
-    "@types/react": "~19.0.10",
-    "typescript": "~5.8.3"
+    "@types/react": "19.1.10",
+    "typescript": "~5.9.2"
```
gdorsi, line 30: **question:** why did you remove the catalog from here?

gdorsi (approved):
Thanks!

examples/chat-rn-expo/package.json
boorad: Because `pnpx expo install --fix` fetches the proper versions.  This seems preferable to us maintaining it in a catalog.

PR #2786, opened 2025-08-20
fix: make it possible for an admin to self-downgrade to writeOnly

gdorsi:
# Description

- Allow admin to self-downgrade to `writeOnly`
- Prevent admin from downgrading other admins to `writeOnly`

## Tests

## Checklist


gdorsi (commented):
Wonderful, thanks!

PR #2788, opened 2025-08-20
feat: skip the agent resolve

gdorsi:
# Description

## Tests

## Checklist


packages/cojson/src/coValueCore/SessionMap.ts
```diff
@@ -58,22 +58,34 @@ export class SessionMap {
...
+    }
+
     try {
       sessionLog.impl.tryAdd(newTransactions, newSignature, skipVerify);
 
       this.addTransactionsToJsLog(sessionLog, newTransactions, newSignature);
 
       return ok(true as const);
     } catch (e) {
+      console.log("error", e);
```
justinrosenthal: Did you mean to leave this here?

packages/cojson/src/coValueCore/SessionMap.ts
```diff
@@ -58,15 +58,26 @@ export class SessionMap {
...
-    signerID: SignerID,
+    signerID: SignerID | undefined,
     newTransactions: Transaction[],
     newSignature: Signature,
     skipVerify: boolean = false,
   ): Result<true, TryAddTransactionsError> {
     const sessionLog = this.getOrCreateSessionLog(sessionID, signerID);
 
+    if (!skipVerify && !sessionLog.signerID) {
+      if (!signerID) {
```
justinrosenthal: Is this check necessary/desired? If we get here but `signerID` is set, it's going to fail during `tryAdd()` below and end up returning `InvalidSignature` instead of the new more specific error.

gdorsi: Yep, it's not required

justinrosenthal: Ok. Seems better to remove the `if (!signerID)` to me, but not a huge deal either way.

justinrosenthal (approved):

PR #2793, opened 2025-08-21
 Explicit loadAs in CoList.upsertUnique to use it without loaded context

Takeno:
# Description
As a follow-up to #2777, I found the same behaviour in CoList.upsertUnique.

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/tests/coList.test.ts
+++ b/packages/jazz-tools/src/tools/tests/coList.test.ts
@@ -3,6 +3,7 @@ import { assert, beforeEach, describe, expect, test, vi } from "vitest";
 import { Account, Group, subscribeToCoValue, z } from "../index.js";
 import {
   Loaded,
+  activeAccountContext,
   co,
   coValueClassFromCoValueClassOrSchema,
 } from "../internal.js";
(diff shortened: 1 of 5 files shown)


packages/jazz-tools/src/tools/tests/coList.test.ts
```diff
@@ -906,6 +907,29 @@ describe("CoList unique methods", () => {
     expect(result?.[2]).toBe("item3");
   });
 
+  test("upsertUnique without an active account", async () => {
+    const account = activeAccountContext.get();
+    activeAccountContext.set(null);
```
nrainhart: I recently added a `runWithoutActiveAccount` helper that sets the active account to null and restores it afterward. It'd be nice to use it here

nrainhart (approved):

gdorsi (approved):

packages/jazz-tools/src/tools/tests/coList.test.ts
Takeno: Done in 65332631d2c4b0f4e2ce2b105da2461b919f39a7

PR #2802, opened 2025-08-22
feat: add meta field in transactions

gdorsi:
# Description

Introducing a new `meta` field in the transactions data.

We are going to use it to implement version control features (such as tracking the branch start and store merge info) in a backward compatible way (old clients won't break when loading a CoValue with merge info).

Also, since meta is a `JsonObject` we will be able to scan for meta information before parsing the transactions in a fast loop.

Changes:
- Add optional `meta` param in `makeTransaction` which like for `changes` follows the privacy option
- Add `decryptNextTransactionMetaJson` to the crypto APIs to decrypt private meta info

## Tests


aeplay (approved):
LGTM

PR #2811, opened 2025-08-25
feat: branching and merging on cojson

gdorsi:
# Description

Implement the branch creation and merge in `cojson` at CoValue level.

Branch loading will be implemented in a follow-up PR.

## Tests


packages/cojson/src/ids.ts
```diff
@@ -20,7 +20,11 @@ export function rawCoIDfromBytes(bytes: Uint8Array): RawCoID {
   return `co_z${base58.encode(bytes.slice(0, shortHashLength))}` as RawCoID;
 }
 
-export type TransactionID = { sessionID: SessionID; txIndex: number };
+export type TransactionID = {
+  sessionID: SessionID;
+  txIndex: number;
+  source?: RawCoID;
```
nrainhart: Mind adding TS docs to explain what `source` means?

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -646,86 +645,235 @@ export class CoValueCore {
     return newContent;
   }
 
-  getValidTransactions(options?: {
-    ignorePrivateTransactions: boolean;
-    knownTransactions?: CoValueKnownState["sessions"];
-  }): DecryptedTransaction[] {
```
gdorsi, line 652: Changed approach here, instead of using the knownState for the incremental processing I've moved to using sets to track the parsed transactions.

This way we can track both the parsed transactions from the branch and from the source.

Also refactored the `getValidTransactions` in two steps:
- transaction parsing, where the progressive computations is shared between all the content views
- `getValidTransactions` that returns the transactions based on the filters in input

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -646,86 +645,243 @@ export class CoValueCore {
...
+    return matchingTransactions;
+  }
+
+  createBranch(name: string, ownerId: RawCoID) {
+    return createBranch(this, name, ownerId);
+  }
+
+  mergeBranch() {
+    return mergeBranch(this);
+  }
```
nrainhart, line 895: A branch can only be merged against its source? I.e. is it not possible to merge it into any other arbitrary branch?

packages/cojson/src/tests/branching.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      // Add transactions to original map
+      originalMap.set("key1", "value1", "trusting");
+      originalMap.set("key2", "value2", "trusting");
+
+      // Create branch
+      const branch = expectMap(
+        originalMap.core.createBranch(branchName, group.id).getCurrentContent(),
+      );
+
+      // Merge branch twice
```
nrainhart, line 105:
```suggestion
```

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -646,86 +645,243 @@ export class CoValueCore {
...
+
+  getBranchId(name: string, ownerId: RawCoID) {
+    return getBranchId(this, name, ownerId);
   }
 
   getValidSortedTransactions(options?: {
     ignorePrivateTransactions: boolean;
-    knownTransactions: CoValueKnownState["sessions"];
+    knownTransactions?: Set<Transaction>;
```
nrainhart, line 905: Is this parameter necessary?

packages/cojson/src/coValueCore/branching.ts
```diff
@@ -0,0 +1,198 @@
...
+    return target;
+  }
+
+  // Create a merge commit to identify the merge point
+  target.makeTransaction([], "private", {
+    // The point where the branch was merged
+    merge: { ...branch.knownState().sessions },
+    // The id of the branch
+    id: branch.id,
+    // The number of transactions to merge, will be used in the future to handle the edits history properly
```
nrainhart: Nit: it'd be good to add these comments in the `MergeCommit` type

packages/cojson/src/permissions.ts
```diff
@@ -79,7 +79,7 @@ function logPermissionError(
 
 export function determineValidTransactions(
   coValue: CoValueCore,
-  knownTransactions?: CoValueKnownState["sessions"],
+  validatedTransactionsSet?: Set<Transaction>,
```
nrainhart: Some indication that this is an inout parameter would be nice

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -646,86 +645,243 @@ export class CoValueCore {
...
+   * Returns the valid transactions matching the criteria specified in the options
+   */
+  getValidTransactions(options?: {
+    ignorePrivateTransactions: boolean;
+    // The range, described as knownState sessions, to filter the transactions returned
+    from?: CoValueKnownState["sessions"];
+    to?: CoValueKnownState["sessions"];
+
+    // The transactions that have already been processed, used for the incremental updates
+    knownTransactions?: Set<Transaction>;
```
nrainhart: Ditto

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -646,86 +645,243 @@ export class CoValueCore {
...
+    }
+  }
 
-        changes = decryptedChanges;
-      } else {
-        try {
-          changes = parseJSON(tx.changes);
-        } catch (e) {
-          logger.error("Failed to parse trusting transaction on " + this.id, {
+  getTxData({
```
nrainhart: Should this method be private?

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -646,86 +645,243 @@ export class CoValueCore {
...
+    this.branchStart = undefined;
+    this.mergeCommits = [];
+    this.validatedTransactionsSet = new Set();
+    this.parsedTransactionsSet = new Set();
+  }
+
+  /**
+   * Goes through the CoValue transactions, and load the new changes & meta in memory
+   */
+  parseNewTransactions(ignorePrivateTransactions: boolean) {
```
nrainhart: Ditto

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -646,86 +645,243 @@ export class CoValueCore {
...
+  // The starting point of the branch, in case this CoValue is a branch
+  branchStart:
+    | { branch: CoValueKnownState["sessions"]; madeAt: number }
+    | undefined;
+
+  // The list of merge commits that have been made
+  mergeCommits: { commit: MergeCommit; madeAt: number }[] = [];
+
+  // Reset the parsed transactions and branches, to validate them again from scratch when the group is updated
+  resetParsedTransactions() {
```
nrainhart, line 686: This one could be private as well

nrainhart (approved):
Changes look good from my end

packages/cojson/src/ids.ts
gdorsi: It's actually unused, removed

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: No it's not, for now.

I prefer to keep the options limited, to have more control on what could happen.
In the future we could support merging vs arbitrary coValues (the design supports that), but it's out of scope for now.

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: Yes, it's used to calculate the incremental builds of the CoList view.

Added a comment

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: Yes, it's used for the incemental builds of CoMap and CoFeed

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: Removed, refactored the code after discussing with Anselm the approach to adopt.

gdorsi:
@nrainhart 

Changed significantly the way that `getValidTransactions` works, to be more future-proof with what we want to do on CoValueCore.

Now the transactions are parsed and validated in a multi-step process, and their state is stored in a `ValidatedTransaction` object.

This way we can do these future improvements more easily:
- Be smarter about invalidating on group updates, and rebuild only the affected transactions
- Make group membership part of the CoValueCore logic, so it can be self-contained there and easier to use when building authenticated sync

It also makes the state self-contained in `ValidatedTransaction` removing the cache management for decryption and simplifying the tracking of the validation status of a transaction.

PR #2812, opened 2025-08-25
feat: make use of the  improvements on music-player

gdorsi:


examples/music-player/src/4_actions.ts
```diff
@@ -84,61 +78,40 @@ export async function addTrackToPlaylist(
...
+
+  const isPartOfThePlaylist = tracks.some((t) => t.$jazz.id === track.$jazz.id);
+
+  if (!isPartOfThePlaylist) return;
+
+  const trackGroup = track.$jazz.owner;
+  trackGroup.removeMember(playlist.$jazz.owner);
+
+  // @ts-expect-error The t is nullable, but it shouldn't be
+  tracks.$jazz.remove((t) => t.$jazz.id === track.$jazz.id);
```
gdorsi: @nrainhart 

I think this is a bug on the types.

PR #2813, opened 2025-08-25
Add remove action for CoList table in the inspector

gdorsi:
# Description

While doing user-testing for the $jazz upgrade, I've broken my production data for the Music player.

The easiest way to recover it was to build the remove in CoList

Load a CoList in the inspector:
- With write access the Remove button should be visible
- Without it shouldn't be

## Tests

- [x] Tests have not been updated, because: unsure we want to keep this

### DIFF
--- a/packages/jazz-tools/src/inspector/viewer/page.tsx
+++ b/packages/jazz-tools/src/inspector/viewer/page.tsx
@@ -90,7 +90,7 @@ function canEdit(value: RawCoValue) {
   try {
     const myRole = value.group.myRole();
 
-    return myRole === "writer" || myRole === "admin" || myRole === "writeOnly";
+    return myRole === "writer" || myRole === "admin";
   } catch (e) {
     return false;
   }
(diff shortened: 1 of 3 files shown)


packages/jazz-tools/src/inspector/viewer/page.tsx
```diff
@@ -86,6 +86,16 @@ type PageProps = {
   className?: string;
 };
 
+function canEdit(value: RawCoValue) {
+  try {
+    const myRole = value.group.myRole();
+
+    return myRole === "writer" || myRole === "admin" || myRole === "writeOnly";
```
nrainhart: Is it ok that people with writeOnly access are able to remove items from the CoList as well?

nrainhart (approved):
Approving with one question

packages/jazz-tools/src/inspector/viewer/page.tsx
```diff
@@ -119,7 +129,23 @@ function View(
...
+              const list = value as RawCoList;
+              list.delete(index);
+            }
+          }
+        : undefined;
+    return (
+      <TableView
+        data={snapshot}
+        node={node}
+        onNavigate={onNavigate}
```
Takeno, line 143: I would split into two different if-statements for records and colist, to have one nested condition less (twice `type === "colist"`)

packages/jazz-tools/src/inspector/viewer/page.tsx
gdorsi: Good point, they could actually, but it would be pretty messy.

PR #2815, opened 2025-08-25
feat: `Account.$jazz.set`

nrainhart:
# Description

Add `Account.$jazz.set`, to make `root` and `profile` editable the same way as CoMaps. This includes being able to use plain JSON objects as input.

## Tests

## Checklist

### DIFF
--- a/examples/file-share-svelte/src/lib/schema.ts
+++ b/examples/file-share-svelte/src/lib/schema.ts
@@ -23,9 +23,15 @@ export const FileShareAccount = co.account({
     const publicGroup = Group.create({ owner: account });
     publicGroup.addMember('everyone', 'reader');
 
-    account.root = FileShareAccountRoot.create({
-      type: 'file-share-account',
-      sharedFiles: co.list(SharedFile).create([], publicGroup),
-    }, publicGroup);
+    account.$jazz.set(
+      'root',
+      FileShareAccountRoot.create(
+        {
+          type: 'file-share-account',
+          sharedFiles: co.list(SharedFile).create([], publicGroup)
+        },
+        publicGroup
+      )
+    );
   }
 });
--- a/packages/jazz-tools/src/tools/tests/ContextManager.test.ts
+++ b/packages/jazz-tools/src/tools/tests/ContextManager.test.ts
@@ -256,11 +256,14 @@ describe("ContextManager", () => {
         profile: co.profile(),
       })
       .withMigration(async (account) => {
-        account.root = AccountRoot.create(
-          {
-            value: "Hello",
-          },
-          Group.create(this).makePublic(),
+        account.$jazz.set(
+          "root",
+          AccountRoot.create(
+            {
+              value: "Hello",
+            },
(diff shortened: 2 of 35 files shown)


tests/jazz-svelte/src/routes/virtual-list/schema.ts
```diff
@@ -27,48 +27,30 @@ export const TestAccount = co
...
-                  publicGroup,
-                ),
-              ),
-            ),
-          }),
-        },
-        publicGroup,
+      account.$jazz.set(
+        "root",
+        Root.create(
```
nrainhart, line 32: Removing all those creates felt really nice ngl

packages/jazz-tools/src/tools/tests/ContextManager.test.ts
```diff
@@ -323,11 +326,11 @@ describe("ContextManager", () => {
...
         profile: co.profile(),
       })
-      .withMigration(async (account) => {
-        account.root = AccountRoot.create({
+      .withMigration(async (account: Account) => {
+        account.$jazz.set("root", {
           value: "Hello",
         });
-        lastRootId = account.root.$jazz.id;
+        lastRootId = account.root!.$jazz.id;
```
nrainhart, line 333: The caveat from using `$jazz.set` instead of direct assignments is that TS can't know that the property has been set. I tried to get this working by adding an [assert condition](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) to `Account.$jazz.set`, but I don't think it's possible to narrow the type of `account` when calling a method in `account.$jazz`

examples/file-share-svelte/src/lib/schema.ts
```diff
@@ -23,9 +23,15 @@ export const FileShareAccount = co.account({
...
+    account.$jazz.set(
+      'root',
+      FileShareAccountRoot.create(
+        {
+          type: 'file-share-account',
+          sharedFiles: co.list(SharedFile).create([], publicGroup)
+        },
+        publicGroup
+      )
+    );
```
gdorsi, line 35: This one is a breaking change.

The root isn't public like before

gdorsi: Ah, my bad! The code is ok.

packages/jazz-tools/src/tools/tests/ContextManager.test.ts
gdorsi: Let's keep things simple and handle this problem when we are going to work on the Proxy killing.

examples/file-share-svelte/src/lib/schema.ts
nrainhart: How come? We're explicitly creating the FileShareAccountRoot with the public group, and the sharedFiles list will have that public group as member, so it should be readable by everyone as well

Edit: i wish Github's comments were updated in real time... we should try to sell them Jazz 😛

gdorsi (approved):
Looks good!

PR #2816, opened 2025-08-25
fix sign out Jazz account when signing out from Clerk

Takeno:
# Description
When signing out from Clerk, the Jazz plugin cleaned the authSecretStorage, but it didn't wipe the in-memory context. The result was that the user was logged out of Clerk, had no more of their credentials stored locally, but they were still seeing the account's data until the page was refreshed.

Now, when the clerk's user is signed out, we log out the Jazz context too.

To make it work, `logOutReplacement` has been replaced with `onLogOut` to avoid a short-circuit. The original behaviour has been preserved: calling `jazzContext.logOut()` still calls Clerk's signOut.

Use the clerk's example and call `clerk.user.delete()`. Without any page refreshing, the Jazz AccountID should change to an anonymous user.

No BC intended.

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/auth/clerk/index.ts
+++ b/packages/jazz-tools/src/tools/auth/clerk/index.ts
@@ -18,6 +18,7 @@ export { isClerkCredentials };
 export class JazzClerkAuth {
   constructor(
     private authenticate: AuthenticateAccountFunction,
+    private logOut: () => Promise<void> | void,
     private authSecretStorage: AuthSecretStorage,
   ) {}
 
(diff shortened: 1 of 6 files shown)


packages/jazz-tools/src/tools/auth/clerk/index.ts
```diff
@@ -77,6 +78,7 @@ export class JazzClerkAuth {
     if (!clerkClient.user) {
       if (isAuthenticated) {
         this.authSecretStorage.clear();
+        await this.logOut();
```
gdorsi, line 81: If we are calling logOut we don't need to call `authSecretStorage.clear`

```suggestion
        await this.logOut();
```

gdorsi (approved):
LGTM

packages/jazz-tools/src/tools/auth/clerk/index.ts
gdorsi: Nope, looks necessary.

PR #2818, opened 2025-08-26
fix: prevent adding and removing members from Accounts

nrainhart:
# Description

Account members should not be modifiable. So we're overriding group-membership-related methods in `RawAccount`, to prevent accounts from getting into an inconsistent state even when using `account.castAs(Group)`.

This change was backported from https://github.com/garden-co/jazz/pull/2778 (which is targeted for `0.18.0`).

## Tests

## Checklist


gdorsi (approved):
Sweet, thanks!

PR #2820, opened 2025-08-26
codemod for jazz 0.18

gdorsi:
Codemod to simplify the upgrade to Jazz v0.18

It uses `ts-morph` which gives us some information on the resolved types.

Since we do some type-aware filtering, the codemod needs to run on top of a Jazz v0.18 installation.

The transform is executed more than one time to iterate over fixes on the types (e.g. when we do fix a `ensureLoaded` the returned type is no more any and we can check that the values are related to Jazz)

### DIFF
--- a/packages/codemod-upgrade-0-18/README.md
+++ b/packages/codemod-upgrade-0-18/README.md
@@ -1,0 +1,71 @@
+# Jazz Codemods
+
+This repository contains codemods for upgrading Jazz projects to newer versions. Codemods are automated tools that help migrate your codebase to be compatible with newer versions of Jazz.
+
+## What are Codemods?
+
+Codemods are automated refactoring tools that help you upgrade your codebase. 
+
+They rewrite your code using AST modification.
+
+We use ts-morph under the hood, so we do also some typechecks to target Jazz specific values in our migrations.
+
+Codemods modify your code, not always in the correct way.
+Be sure that everything important is committed in your repo and there aren't unstaged changes that might get lost in the process.
+
+## Available Codemods
+
+### Jazz 0.18 Upgrade
+
+The main codemod available in this repository is for upgrading to Jazz 0.18.
+
+## Usage
+
+### Running the Jazz 0.18 Codemod
+
+To run the Jazz 0.18 codemod upgrade the jazz-tools version to 0.18 and then execute:
+
+```bash
+npx jazz-tools-codemod-0-18
+```
+
+### How it Works
+
+By default, the codemod will:
+
+1. **Try to use the tsconfig.json in the current working directory** to resolve project files
+2. **Fall back to glob search** if no tsconfig.json is found or if it can't resolve the project structure
+
+## Prerequisites
+
+- Node.js (version 14 or higher)
+- npm, yarn, or pnpm
+
+## Installation
+
+The codemod is available as an npm package and can be run directly with npx, so no local installation is required.
+
+## Running on Specific Files
+
+If you want to run the codemod on specific files or directories, you can pass them as arguments:
+
+```bash
+npx jazz-tools-codemod-0-18 src/
+```
+
+## Contributing
+
+If you find issues with the codemod or want to contribute improvements, please:
+
+1. Open an issue describing the problem
+2. Fork the repository
+3. Make your changes
+4. Submit a pull request
+
+## License
+
+MIT
+
+## Support
+
+For issues related to the codemod itself, please open an issue in this repository. For general Jazz support, please refer to the official Jazz documentation.
--- a/packages/codemod-upgrade-0-18/README.md
+++ b/packages/codemod-upgrade-0-18/README.md
@@ -8,7 +8,7 @@ Codemods are automated refactoring tools that help you upgrade your codebase.
 
 They rewrite your code using AST modification.
 
-We use ts-morph under the hood, so we do also some typechecks to target Jazz specific values in our migrations.
+We use `ts-morph` under the hood, to target specific Jazz types in our migrations.
 
 Codemods modify your code, not always in the correct way.
 Be sure that everything important is committed in your repo and there aren't unstaged changes that might get lost in the process.
--- a/packages/codemod-upgrade-0-18/package.json
+++ b/packages/codemod-upgrade-0-18/package.json
@@ -1,0 +1,36 @@
+{
+  "name": "jazz-tools-codemod-0-18",
+  "version": "0.1.0",
+  "description": "Codemod to migrate from Jazz Tools 0.17 to 0.18",
+  "type": "module",
+  "license": "MIT",
+  "files": [
+    "dist"
+  ],
+  "main": "./dist/index.js",
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": "./dist/index.js",
+    "./package.json": "./package.json"
+  },
+  "bin": {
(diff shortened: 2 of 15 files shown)


codemod/fixtures/test-example.ts
```diff
@@ -0,0 +1,131 @@
...
+import { Account, Chat, Message, Group } from "jazz-tools";
+
+// Example 1: Property access patterns
+function examplePropertyAccess() {
+  const chat = Chat.create([], Group.create());
+  const chatId = chat.id; // Will become: chat.$jazz.id
+  const owner = chat._owner; // Will become: chat.$jazz.owner
+  const refs = chat._refs.messages; // Will become: chat.$jazz.refs.messages
+  const edits = chat._edits.title; // Will become: chat.$jazz.getEdits().title
+  const type = chat._type; // Will become: chat.$jazz.$type$
```
nrainhart:
```suggestion
  const type = chat._type; // Will become: chat.$type$
```

codemod/fixtures/test-example.ts
```diff
@@ -0,0 +1,131 @@
...
+  const createdAt = chat._createdAt; // Will become: chat.$jazz.createdAt
```
nrainhart:
```suggestion
  const createdAt = chat._createdAt; // Will become: chat.$jazz.createdAt
  const lastUpdatedAt = chat._lastUpdatedAt; // Will become: chat.$jazz.lastUpdatedAt
```

codemod/transform.ts
```diff
@@ -0,0 +1,305 @@
...
+  options: Options,
+) {
+  const j = api.jscodeshift;
+  const root = j(file.source);
+
+  // 1. Transform property access patterns
+  root
+    .find(j.MemberExpression, {
+      property: { type: "Identifier", name: "id" },
+    })
```
nrainhart: This will transform any `.id` access, right? Is it possible to constraint it to CoValue objects?

codemod/transform.ts
```diff
@@ -0,0 +1,305 @@
...
+            j.memberExpression(object, j.identifier("$jazz")),
+            j.identifier("delete"),
+          ),
+          [j.stringLiteral(propertyName)],
+        );
+        j(path).replaceWith(deleteCall);
+      }
+    });
+
+  // 9. Transform splice operations to remove
```
nrainhart: 🔥

nrainhart (commented):
Thanks for writing this codemod!! Left a few comments, excited to see how well it works

codemod/transform.ts
gdorsi: Added some type checks to guess that the member might be a CoValue

packages/codemod-upgrade-0-18/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "jazz-tools-codemod-0-18",
```
nrainhart, line 2: Having this specific codemod for the 0.18 upgrade as a package means we'll need to add new packages for future codemods. Should we just make this jazz-tools-codemod and parameterize the jazz version? Not sure how much we want to future-prove migration scripts

packages/codemod-upgrade-0-18/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+This repository contains codemods for upgrading Jazz projects to newer versions. Codemods are automated tools that help migrate your codebase to be compatible with newer versions of Jazz.
+
+## What are Codemods?
+
+Codemods are automated refactoring tools that help you upgrade your codebase. 
+
+They rewrite your code using AST modification.
+
+We use ts-morph under the hood, so we do also some typechecks to target Jazz specific values in our migrations.
```
nrainhart:
```suggestion
We use `ts-morph` under the hood, to target specific Jazz types in our migrations.
```

packages/codemod-upgrade-0-18/src/transform.ts
```diff
@@ -0,0 +1,641 @@
...
+
+  if (fs.existsSync(`${projectPath}/tsconfig.json`)) {
+    project = new Project({
+      tsConfigFilePath: `${projectPath}/tsconfig.json`,
+    });
+  } else {
+    project = new Project();
+
+    // Check if projectPath ends with a supported file extension
+    const supportedExtensions = [".ts", ".tsx", ".js", ".jsx"];
```
nrainhart: Trying to migrate .js & .jsx files simply omits all checks that depend on Jazz types, right?

Also, should we support .svelte files?

packages/codemod-upgrade-0-18/src/transform.ts
```diff
@@ -0,0 +1,641 @@
...
+    if (node.getKind() === SyntaxKind.CallExpression) {
+      const callExpression = node as CallExpression;
+      const callee = callExpression.getExpression();
+
+      if (callee.getKind() === SyntaxKind.PropertyAccessExpression) {
+        const property = (callee as PropertyAccessExpression).getNameNode();
+        if (property.getText() === "castAs") {
+          const baseObject = (
+            callee as PropertyAccessExpression
+          ).getExpression();
```
nrainhart, line 803: We don't want to check that the receiver is a CoValue here?

packages/codemod-upgrade-0-18/src/transform.ts
```diff
@@ -0,0 +1,641 @@
...
+
+function transformArrayOperations(sourceFile: SourceFile) {
+  sourceFile.forEachDescendant((node) => {
+    if (node.getKind() === SyntaxKind.CallExpression) {
+      const callExpression = node as CallExpression;
+      const callee = callExpression.getExpression();
+
+      if (callee.getKind() === SyntaxKind.PropertyAccessExpression) {
+        const property = (callee as PropertyAccessExpression).getNameNode();
+        if (property.getText() === "push") {
```
nrainhart: We're missing migrations for unshift, pop and shift

packages/codemod-upgrade-0-18/src/transform.ts
```diff
@@ -0,0 +1,641 @@
...
+          "$jazz.owner",
+          hasOptional,
+        );
+
+        propertyAccess.replaceWithText(newText);
+      }
+    }
+  });
+
+  // Transform obj._type to obj.$jazz.$type$
```
nrainhart: The comment is wrong, but the code is right

packages/codemod-upgrade-0-18/package.json
gdorsi: I see the upgrade codemods as self-contained throwaway packages.

This helps on keeping things easy, and makes simple to remove the package when we unlink the upgrade guide.

packages/codemod-upgrade-0-18/src/transform.ts
gdorsi: Makes sense to skip js files.

Supporting .svelte would be hard, unless `ts-morph` supports them natively.

nrainhart: Yeah, it doesn't. I guess it makes sense to skip svelte files as well

nrainhart (approved):
Let's gooooo!! 🚀

packages/codemod-upgrade-0-18/package.json
nrainhart: Makes sense to me! I was thinking we'd keep supporting these codemods forever, but deleting them when we unlink the upgrade guide sounds good

PR #2824, opened 2025-08-27
Getting Started Docs

joeinnes:
# Description
We don't currently have a great 'quickstart' guide to hook people into Jazz. This PR aims to add it.

Go through the doc at docs/getting-started/quickstart. Estimated time taken should be around 10 minutes.

Key things to check:

1. Is it clear and understandable?
2. Are there improvements that can be made to the language used to make it more accessible (removing jargon, adding explanations/clarifications, etc.)?
3. Are the instructions easy to follow?
4. Do the instructions work? (Note: there are a couple of points which are ready for v0.18 and the dashboard launch, these are not working yet, but should be by the time this body of work is complete).
5. Does it take ~10 minutes?

## Tests

- [x] Tests have not been updated, because: Docs update

## Checklist


homepage/homepage/content/docs/getting-started/quickstart.mdx
```diff
@@ -0,0 +1,395 @@
...
+</TabbedCodeGroup>
+
+<ContentByFramework framework="react">
+  <Alert variant="info" className="mt-4 flex gap-2 items-center">
+    `"use client";` tells Next.js that this component can't be server-side rendered. If you're not using Next.js, you can remove it.
+  </Alert>
+</ContentByFramework>
+
+## Start your app
+Moment of truth&thinsp;—&thinsp;time to start your app and see if it works.
```
mculotta120: When I got to this step, I didn't actually see the login, just the next.js default project page. I'm thinking this is because I'm already logged in? I wasn't actually able to test because when I try to logout in dashboard, I get a 400 error. With the message "Failed to get Session" 

CC: @Elfo404 - I wasn't sure the the dashboard error for logout was a known issue yet or not (I didn't find any corresponding issues), I can create an issue if it's new.

mculotta120 (commented):
@joeinnes - Looks great! I followed the guide exactly and it worked. I just had one hiccup on the "Start your app" section, I believe because I was already logged in . We may want to add a disclaimer there or something? I'll leave it up to you. I don't think it's a blocker.

homepage/homepage/content/docs/getting-started/quickstart.mdx
joeinnes: No, there's still a slight mismatch between what's written and the code samples because there's still a decision to be made on whether we include auth or not (https://vercel.live/link/jazz-homepage-git-docs-getting-started-gcmp.vercel.app?page=%2Fdocs%2Freact%2Fgetting-started%2Fquickstart%3FvercelThreadId%3DlyfqI&via=in-app-copy-link&p=1, not sure if @aeplay or @gdorsi have a strong opinion). If we keep auth in the example, then I'll extend the end section with some guidance on how you can sign in with an incognito tab and put the windows side by side to see your data updating live, otherwise I'll update this section to say that you should see the default Next.js/SvelteKit project page.

This is interesting: https://github.com/garden-co/jazz/pull/2860

It means that potentially actually we *could* do a multidevice demo with minimal complexity, just by telling the user to access using the same URL on both devices (Next.js hosts on 0.0.0.0 by default, SvelteKit hides it behind a flag, but it might make sense).

For now, I'll update without auth and can plug it back in later. Thanks!

packages/jazz-run/src/index.ts
```diff
@@ -27,7 +27,8 @@ const createAccountCommand = Command.make(
       if (json) {
         yield* Console.log(JSON.stringify({ accountID, agentSecret }));
       } else {
-        yield* Console.log(`# Credentials for Jazz account "${name}":
+        yield* Console.log(`
```
joeinnes, line 30: Enables `npx jazz-run account create --name 'My Worker' >> .env`

homepage/homepage/content/docs/getting-started/quickstart.mdx
gdorsi: My personal take is that auth needs a dedicated getting started.

What we can show here could be some bare-minumum auth, but consider that passkeys only work on HTTPS.

We can try to make it easier to use passphrase auth, by providing a default wordlist, so you can make people add a "Copy passphrase" button to play with multiple devices in the same network.

Something like 

```ts
import { wordlist } from `jazz-tools/wordlist`;
import { usePassphraseAuth } from `jazz-tools/react`

export function CopyPassphrase() {
    const auth = usePassphraseAuth({
      wordlist,
    });
    const [isCopied, setIsCopied] = useState(false);
    const textarea = useRef()
    
      const handleCopy = async () => {
        textarea.current.select()
        document.execCommand('copy');
        setIsCopied(true);
      };
      
    return <> <textarea
              ref={textarea}
              readOnly
              value={auth.passphrase}
              rows={5}
            />
              <button onClick={handleCopy}>
                {isCopied ? "Copied!" : "Copy Passphrase"}
              </button><>
}
```

gdorsi: Then we can add a login form, that takes a passphrase input and calls `auth.login(passphrase)`

joeinnes: OK, I think I'm going to keep it without auth, and build out a separate auth guide. 

This guide is already on the long side for a 'quick start', and trying to shoehorn more in seems like a bad idea.

In future, I'd like to have some kind of embedded 'mini-project' type thing (à la Sandpack https://sandpack.codesandbox.io/) where adding extra files etc. as readonly resources wouldn't take up 'column inches' and so might not distract too much from the flow of the actual guide, so I'll maybe come back to this idea then.

mculotta120 (commented):
✅ Tested both react and svelte versions. Working as expected!
QA is complete

homepage/homepage/content/docs/getting-started/quickstart.mdx
```diff
@@ -0,0 +1,428 @@
...
+<script lang="ts">
+  import { AccountCoState } from "jazz-tools/svelte";
+  import { JazzFestAccount } from "$lib/schema";
+  const me = new AccountCoState(JazzFestAccount, { 
+    resolve: { root: { myFestival: true } } 
+  });
+</script>
+
+<ul>
+  {#each me.current?.root?.myFestival || [] as band}
```
gdorsi:
```suggestion
  {#each me.current?.root.myFestival || [] as band}
```

gdorsi (approved):
LGTM

Well done!

gdorsi:
**nit:** changing framework resets the page position

We could fix the jump by attaching an anchor to the code snippets

joeinnes:
> **nit:** changing framework resets the page position
> 
> We could fix the jump by attaching an anchor to the code snippets

Unfortunately it wasn't quite as simple as that, but added some scroll preservation stuff and should be working now.

gdorsi (approved):
Wonderful work!

PR #2829, opened 2025-08-27
feat: change the migration checks to use .$jazz.has

gdorsi:
Moving away from `account.root === undefined` based checks, because they rely on a subtle detail on how in Jazz `undefined` has a different meaning compared to `null`.

With the new `$jazz` APIs accessing `raw` doesn't look like a sin anymore (before was `_raw` which made it look like accessing an internal prop) so I'm suggesting to use `!account.$jazz.has("root")` to check if a root is setted, regardless of the loading statuses.

We could even start pushing for using `raw` instead of `refs` for accessing references ids without triggering the autoload.

### DIFF
--- a/.changeset/nervous-rules-heal.md
+++ b/.changeset/nervous-rules-heal.md
@@ -2,4 +2,4 @@
 "jazz-tools": patch
 ---
 
-Add `CoMap.$jazz.has` method to check for property existance without loading referenced CoValues or checking permissions
+Add `CoMap.$jazz.has` and `Account.$jazz.has` method to check for property existance without loading referenced CoValues or checking permissions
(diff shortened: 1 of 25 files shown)


nrainhart:
> We could even start pushing for using raw instead of refs for accessing references ids without triggering the autoload.

Conceptually I prefer `!account.$jazz.refs.root` over `!account.$jazz.raw.get("root")`. Accessing the raw covalue feels like a last resort to me. What will `refs` behavior be like when we remove autoload? Maybe it makes sense to wait until then to change this migration pattern?

gdorsi:
> What will refs behavior be like when we remove autoload? 

The idea was to remove it when we introduce `$jazzState` and remove the autoload.

With `$jazzState` it would be possible to access a reference id even when it's not loaded, replacing the use cases for refs.

At that point `=== undefined` would still work, but I don't like we need to explain why this works and `!account.root` does not when people take the first steps on Jazz.

gdorsi:
@nrainhart Updated the PR to use .has

nrainhart (approved):

PR #2830, opened 2025-08-27
feat: add `CoMap.$jazz.has`

nrainhart:
# Description

Related to https://github.com/garden-co/jazz/pull/2829.

`CoMap.$jazz.has` allows to check for property existence without loading referenced CoValues or checking permissions.

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/coMap.ts
+++ b/packages/jazz-tools/src/tools/coValues/coMap.ts
@@ -581,7 +581,8 @@ class CoMapJazzApi<M extends CoMap> extends CoValueJazzApi<M> {
    * @category Content
    */
   has(key: CoKeys<M>): boolean {
-    return this.raw.get(key) !== undefined;
+    const entry = this.raw.getRaw(key);
+    return entry?.change !== undefined && entry.change.op !== "del";
   }
 
   /**
@@ -1017,7 +1018,7 @@ const CoMapProxyHandler: ProxyHandler<CoMap> = {
     const descriptor = target.$jazz?.getDescriptor(key as string);
 
     if (target.$jazz?.raw && typeof key === "string" && descriptor) {
-      return target.$jazz.has(key as CoKeys<typeof target>);
+      return target.$jazz.raw.get(key) !== undefined;
     } else {
       return Reflect.has(target, key);
     }
--- a/packages/jazz-tools/src/tools/coValues/coMap.ts
+++ b/packages/jazz-tools/src/tools/coValues/coMap.ts
@@ -572,6 +572,19 @@ class CoMapJazzApi<M extends CoMap> extends CoValueJazzApi<M> {
   }
 
   /**
+   * Check if a key is defined in the CoMap.
+   *
+   * This check does not load the referenced value or validate permissions.
+   *
+   * @param key The key to check
+   * @returns True if the key is defined, false otherwise
+   * @category Content
+   */
+  has(key: CoKeys<M>): boolean {
+    return this.raw.get(key) !== undefined;
+  }
+
+  /**
    * Set a value on the CoMap
    *
    * @param key The key to set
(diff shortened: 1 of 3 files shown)


packages/jazz-tools/src/tools/tests/coMap.test.ts
```diff
@@ -744,6 +744,81 @@ describe("CoMap", async () => {
...
+      );
+
+      await person.$jazz.waitForSync({ timeout: 1000 });
+
+      const loadedPerson = await Person.load(person.$jazz.id, {
+        resolve: true,
+        loadAs: serverAccount,
+      });
+
+      assert(loadedPerson);
```
nrainhart, line 816: Still debugging why the person is not being loaded here

packages/jazz-tools/src/tools/coValues/coMap.ts
```diff
@@ -571,6 +571,19 @@ class CoMapJazzApi<M extends CoMap> extends CoValueJazzApi<M> {
...
+   * Check if a key is defined in the CoMap.
+   *
+   * This check does not load the referenced value or validate permissions.
+   *
+   * @param key The key to check
+   * @returns True if the key is defined, false otherwise
+   * @category Content
+   */
+  has(key: CoKeys<M>): boolean {
+    return this.raw.get(key) !== undefined;
```
gdorsi: If we want to mimic the behavoir of `Map.prototype.has` we should return false for 

```ts
map.$jazz.set(key, undefined);
map.$jazz.has(key)
```

An implementation of this would be:
```suggestion
   const entry = this.raw.getRaw(key);

    if (entry?.change === undefined || entry.change.op === "del") {
      return false;
    }

    return true;
```

packages/jazz-tools/src/tools/tests/coMap.test.ts
```diff
@@ -744,6 +744,81 @@ describe("CoMap", async () => {
...
+        name: co.plainText(),
+      });
+
+      const { clientAccount, serverAccount } = await setupTwoNodes();
+
+      const person = Person.create(
+        {
+          name: "John",
+        },
+        { owner: Group.create().makePublic() },
```
gdorsi:
```suggestion
        { owner: Group.create(clientAccount).makePublic() },
```

packages/jazz-tools/src/tools/tests/coMap.test.ts
```diff
@@ -744,6 +744,81 @@ describe("CoMap", async () => {
...
+      await person.$jazz.waitForSync({ timeout: 1000 });
+
+      const loadedPerson = await Person.load(person.$jazz.id, {
+        resolve: true,
+        loadAs: serverAccount,
+      });
+
+      assert(loadedPerson);
+      expect(loadedPerson.$jazz.has("name")).toBe(true);
+      expect(loadedPerson.name).toBeNull();
```
gdorsi, line 818: This check won't work, because the property access is able to load the value syncronously in the test environment.

The loadedPerson is also on a server peer, so the CoPlainText is sent directly when created, so it's already in the server node after you call `waitForSync`

packages/jazz-tools/src/tools/coValues/coMap.ts
nrainhart: This actually returns true:
```typescript
const map = new Map();
map.set('key', undefined);
map.has('key'); // Returns true
```

The only way for it to return false is if the key was never defined, or if `delete` was used

nrainhart: I guess that's what you meant? Because your proposed implementation returns true in this case

packages/jazz-tools/src/tools/tests/coMap.test.ts
nrainhart: > The loadedPerson is also on a server peer, so the CoPlainText is sent directly when created, so it's already in the server node after you call waitForSync

Makes sense, I hadn't thought of that. I made the test pass by creating the map in the server peer and then loading it on the client peer.

I'm confused by why the last assertion is working though, that's not consistent with this comment you made:

> This check won't work, because the property access is able to load the value syncronously in the test environment.

gdorsi: Well, my bad.

When the value is not available in memory we call:

```ts
this.node
        .loadCoValueCore(this.id as any, undefined, skipRetry)
        .then((value) => {
```

Which waits a promise, so even if the sync is syncronous, the update is delayed by a microtask.

This might change in the future thou, because we want to change the approach into load + subscribe which might change the timing.

But let's not care about that now.

gdorsi (approved):

PR #2832, opened 2025-08-28
feat: implement checkoutBranch method for loading and creating branches in LocalNode

gdorsi:
# Description

Added `checkoutBranch` method to `LocalNode` for loading a branch from a coValue, creating it if it doesn't exist.

## Tests

## Checklist

- [x] I've updated the part of the docs that are affected the PR changes (not public yet)

### DIFF
--- a/packages/cojson/src/localNode.ts
+++ b/packages/cojson/src/localNode.ts
@@ -457,6 +457,46 @@ export class LocalNode {
     return core.getCurrentContent() as T;
   }
 
+  /**
+   * Loads a branch from a group coValue, creating a new one if it doesn't exist.
+   *
+   * Returns "unavailable" in case of errors or missing source.
+   */
+  async checkoutBranch<T extends RawCoValue>(
+    id: CoID<T>,
+    branch: string,
+    branchOwnerID?: RawCoID,
+  ): Promise<T | "unavailable"> {
+    const source = await this.loadCoValueCore(id);
+
+    if (!source.isAvailable()) {
+      return "unavailable";
+    }
+
+    const header = source.verified.header;
+
+    // Group and account coValues can't have branches
+    if (header.ruleset.type !== "ownedByGroup") {
+      logger.error("Can't load a branch from a group coValue", {
+        id,
+      });
+      return "unavailable";
+    }
+
+    const ownerID = branchOwnerID || header.ruleset.group;
+
+    const branchID = source.getBranchId(branch, ownerID);
+
+    // Passing skipRetry to true because otherwise creating a new branch would always take 1 retry delay
+    let branchCoValue = await this.loadCoValueCore(branchID, undefined, true);
+
+    if (!branchCoValue.isAvailable()) {
+      branchCoValue = source.createBranch(branch, ownerID);
+    }
+
+    return branchCoValue.getCurrentContent() as T;
+  }
+
   getLoaded<T extends RawCoValue>(id: CoID<T>): T | undefined {
     const coValue = this.getCoValue(id);
 
--- a/packages/cojson/src/localNode.ts
+++ b/packages/cojson/src/localNode.ts
@@ -477,7 +477,7 @@ export class LocalNode {
 
     // Group and account coValues can't have branches
     if (header.ruleset.type !== "ownedByGroup") {
-      logger.error("Can't load a branch from a group coValue", {
+      logger.error("Can't checkout a branch from a group coValue", {
         id,
       });
       return "unavailable";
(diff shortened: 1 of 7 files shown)


packages/cojson/src/tests/branching.test.ts
```diff
@@ -138,4 +148,256 @@ describe("Branching Logic", () => {
...
+      originalMap.set("key1", "value1");
+      originalMap.set("key2", "value2");
+
+      const originalBranch = expectMap(
+        originalMap.core.createBranch(branchName, group.id).getCurrentContent(),
+      );
+
+      originalBranch.set("branchKey", "branchValue");
+
+      // Load the branch - it should create a new one since it doesn't exist
```
nrainhart:
```suggestion
      // Load the branch - it should return the already created branch
```

packages/cojson/src/tests/branching.test.ts
```diff
@@ -138,4 +148,256 @@ describe("Branching Logic", () => {
...
```
nrainhart:
```suggestion
      // Load the branch - it should return the already created branch
```

packages/cojson/src/tests/branching.test.ts
```diff
@@ -138,4 +148,256 @@ describe("Branching Logic", () => {
...
+      const branch = await client.node.load(originalBranch.id);
+
+      expect(branch).not.toBe("unavailable");
+
+      if (branch !== "unavailable") {
+        // The branch should inherit the original data
```
nrainhart:
```suggestion
```

packages/cojson/src/tests/branching.test.ts
```diff
@@ -138,4 +148,256 @@ describe("Branching Logic", () => {
...
+        expect(branch.get("key1")).toBe("value1");
+        expect(branch.get("key2")).toBe("value2");
+        expect(branch.get("branchKey")).toBe("branchValue");
+      }
+    });
+
+    test("should create a branch with a different group when passing the group was ownerId", async () => {
```
nrainhart:
```suggestion
    test("should create a branch with a different group when passing the group as ownerId", async () => {
```

packages/cojson/src/tests/branching.test.ts
```diff
@@ -138,4 +148,256 @@ describe("Branching Logic", () => {
...
+      expect(branchOnTheServer).not.toBe("unavailable");
+
+      if (branchOnTheServer !== "unavailable") {
+        expect(branchOnTheServer.get("branchKey")).toBe(undefined);
+        expect(branchOnTheServer.get("key1")).toBe(undefined);
+        expect(branchOnTheServer.get("key2")).toBe(undefined);
+      }
+    });
+
+    test("should return unavailable if the user tries to load a group branch", async () => {
```
nrainhart:
```suggestion
    test("should return unavailable if the user tries to checkout a branch from a group", async () => {
```

packages/cojson/src/localNode.ts
```diff
@@ -457,6 +457,46 @@ export class LocalNode {
...
+
+    if (!source.isAvailable()) {
+      return "unavailable";
+    }
+
+    const header = source.verified.header;
+
+    // Group and account coValues can't have branches
+    if (header.ruleset.type !== "ownedByGroup") {
+      logger.error("Can't load a branch from a group coValue", {
```
nrainhart:
```suggestion
      logger.error("Can't checkout a branch from a group coValue", {
```

nrainhart: I feel like there's a lot of overlap between checkout and load. Longer term it may make sense to keep only checkout? But for now I feel like we should be careful about which term we use on each context to avoid confusion

packages/cojson/src/tests/branching.test.ts
```diff
@@ -138,4 +148,256 @@ describe("Branching Logic", () => {
...
+
+      // Create a map but don't create any branches
+      const originalMap = group.createMap();
+      originalMap.set("key1", "value1");
+
+      // Try to load a branch that doesn't exist
+      const branch = await client.node.checkoutBranch(
+        originalMap.id,
+        "non-existent-branch",
+      );
```
nrainhart: I'm confused 😕. Shouldn't this create a branch if it doesn't exist?

nrainhart (approved):

packages/cojson/src/tests/branching.test.ts
gdorsi: The client isn't connected to the cloud, so the checkout fails.

Updating the comment

PR #2834, opened 2025-08-29
chore: Update deps and rm unused components for server-worker-http example

andreibudoi:
# Description
There is one dependency missing from package.json and 3 unused components which cause failed builds

- Install deps (e.g. `pnpm i`)
- Do a build (e.g. `pnpm build`)

## Tests

- [x] Run tests
- [x] Build runs successfully


gdorsi (approved):
Thanks for the fix!

PR #2837, opened 2025-08-29
feat: remove CoMapSchema.catchall

nrainhart:
# Description

_Keeping this PR as a draft until we've defined how to migrate existing CoMaps with index signatures._

This PR removes the `catchall` method from `co.map` schemas. It prevents building CoMaps that have both known properties and a catchall schema for unknown properties. `co.map` can now only be used to define struct-like CoMaps with known fields. Record-like CoMaps can still be defined using `co.record`.

The biggest impact this change has is on `ImageDefinition`: images relied on catchall to provide different-sized versions of the same image. These images have now been moved to `image.resolutions`.

## Motivation

Our support for CoMaps with both known & unknown properties was very limited:
- `co.map(...).create` does not allow initializing catchall properties
- `co.map(...).load` (and all other load methods) do not allow accessing catchall properties on the loaded CoMap
- Typescript has limitations for working with objects that contain both known properties + an index signature. The index signature's type often conflicts with the one defined for known fields.

Since struct-like CoMaps can contain record-like CoMaps, there's no motivation for providing a single data structure that allows both known and unknown properties. Removing `catchall` makes our CoValue data model simpler.

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/react-native-core/media/image.tsx
+++ b/packages/jazz-tools/src/react-native-core/media/image.tsx
@@ -68,7 +68,9 @@ export const Image = forwardRef<RNImage, ImageProps>(function Image(
   { imageId, width, height, ...props },
   ref,
 ) {
-  const image = useCoState(ImageDefinition, imageId);
+  const image = useCoState(ImageDefinition, imageId, {
+    resolve: { resolutions: true },
+  });
   const [src, setSrc] = useState<string | undefined>(image?.placeholderDataURL);
 
   const dimensions: { width: number | undefined; height: number | undefined } =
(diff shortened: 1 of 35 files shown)


packages/jazz-tools/src/react-native-core/media/image.tsx
```diff
@@ -68,7 +68,9 @@ export const Image = forwardRef<RNImage, ImageProps>(function Image(
   { imageId, width, height, ...props },
   ref,
 ) {
-  const image = useCoState(ImageDefinition, imageId);
+  const image = useCoState(ImageDefinition, imageId, {
+    resolve: { resolutions: true },
```
nrainhart, line 72: Now there's an additional covalue to load, which makes DX a bit worse. Not sure how much a problem this is in practice (do end users usually handle loading themselves? Or do they rely on higher-level utilities that load the image definition for them?), but we could look into making this resolve query a default if necessary.

cc @Takeno

gdorsi:
Closing because there are too many conflicts, and if we want to go for this we would need to start over anyway.

PR #2838, opened 2025-08-29
feat: deprecate `co.map().catchall`

nrainhart:
# Description

Marks the `co.map().catchall` method as deprecated.

See https://github.com/garden-co/jazz/pull/2837 for the motivation behind this change.

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
+++ b/packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
@@ -131,6 +131,11 @@ export interface CoMapSchema<
     R
   > | null>;
 
+  /**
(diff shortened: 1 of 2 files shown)


packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
```diff
@@ -131,6 +131,11 @@ export interface CoMapSchema<
     R
   > | null>;
 
+  /**
+   * @deprecated Use `co.map().catchall` will be removed in an upcoming version.
+   *
+   * Use a `co.record` nested inside a `co.map` if you need to store key-value properties.
```
gdorsi, line 137: Could you add a migration example in the doc comment?

So that LLMs can handle it more easily.

gdorsi (approved):

PR #2852, opened 2025-09-01
fix: fix merge operation for co.list and co.richText

gdorsi:
# Description

Found that the merge operation is breaking co.list and co.richText, because the transactions store pointers to the session log.

When merging we append all the transactions of the branch to the current session log of the source CoValue, making the pointers invalid.

Fixed by mapping the colist pointers to the new session.

Since I had to learn deeply the inner workings of colist, I've used the opportunity to simplify the code there and managed to get some performance improvmenents in the process

Screenshot 2025-09-02 at 15 40 55

## Tests

### DIFF
--- a/packages/cojson/src/coValueCore/coValueCore.ts
+++ b/packages/cojson/src/coValueCore/coValueCore.ts
@@ -717,6 +717,8 @@ export class CoValueCore {
       return;
     }
 
+    const isBranch = this.isBranch();
+
     for (const [sessionID, sessionLog] of this.verified.sessions.entries()) {
       const count = this.verifiedTransactionsKnownSessions[sessionID] ?? 0;
 
@@ -725,12 +727,20 @@ export class CoValueCore {
           return;
         }
 
+        const txID = isBranch
+          ? {
+              sessionID,
+              txIndex,
+              branch: this.id,
+            }
+          : {
+              sessionID,
+              txIndex,
+            };
+
         this.verifiedTransactions.push({
           author: accountOrAgentIDfromSessionID(sessionID),
-          txID: {
-            sessionID,
-            txIndex,
-          },
+          txID,
           madeAt: tx.madeAt,
           isValidated: false,
           isValid: false,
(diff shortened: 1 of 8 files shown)


packages/cojson/src/coValues/coList.ts
```diff
@@ -133,87 +208,51 @@ export class RawCoList<
...
+
               if (!beforeEntry) {
                 continue;
               }
-              beforeEntry.predecessors.splice(0, 0, {
-                ...txID,
-                changeIdx,
-              });
+
+              beforeEntry.predecessors.push(opID);
```
gdorsi, line 236: Changed the operation from inserting the value at the start of the array to a push operation.

This should be a little faster, and I had to simply change the predecessors traversal to not be in reverse order, so it also simplified the code.

packages/cojson/src/coValues/coList.ts
```diff
@@ -338,22 +374,19 @@ export class RawCoList<
 
       // We navigate the predecessors before processing the current opID in the list
       if (shouldTraversePredecessors) {
-        for (let i = entry.predecessors.length - 1; i >= 0; i--) {
-          todo.push(entry.predecessors[i]!);
+        for (const predecessor of entry.predecessors) {
+          todo.push(predecessor);
```
gdorsi, line 378: Changed the iteration order, to align with the splice -> push change done above.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -857,11 +859,23 @@ export class CoValueCore {
...
       }
 
-      matchingTransactions.push(transaction);
+      // Are we in a branch?
+      if (source) {
+        matchingTransactions.push({
+          txID: {
+            sessionID: txID.sessionID,
+            txIndex: txID.txIndex,
+            branch: this.id,
```
gdorsi: Tracking the tx coming from the branches, so it's easier to translate them when merging.

It also keep the door open for handing the creation of a branch of a branch.

packages/cojson/src/coValues/coList.ts
```diff
@@ -109,6 +107,83 @@ export class RawCoList<
     this.processNewTransactions();
   }
 
+  private getInsertionsEntry(opID: OpID) {
```
gdorsi, line 110: Extracted these methods from the processNewTransactions code, to reduce the cognitive load of understanding how a RawCoList works.

packages/cojson/src/coValues/coList.ts
```diff
@@ -39,12 +35,14 @@ type InsertionEntry<T extends JsonValue> = {
...
-} & InsertionOpPayload<T>;
+  change: InsertionOpPayload<T>;
+};
 
 type DeletionEntry = {
   madeAt: number;
   deletionID: OpID;
-} & DeletionOpPayload;
+  change: DeletionOpPayload;
+};
```
gdorsi, line 45: Changed to have a `change` property, instead of have everything on the first level.

This way we can reuse the same change object coming from the transactions, making the memory footprint smaller.

It also simplifies the code, making it more readable.

nrainhart (approved):
Thanks for the explanation! Reviewing the PR was really easy

PR #2853, opened 2025-09-01
fix: upgrade design system to jazz-tools 0.18

nrainhart:
# Description

Noticed some CI errors when building the design system in another PR. These errors are due to the 0.18 upgrade.

Not sure why they were not caught by CI, but I'm upgrading the design system to get the build passing again.


gdorsi (approved):
Thanks!

PR #2856, opened 2025-09-02
feat: add `$jazz.id` to `toJSON` in Account, CoMap, CoFeed & FileStream

nrainhart:
# Description

In https://github.com/garden-co/jazz/pull/2693 we removed the `id` property from `toJSON`'s output to be consistent with viewing CoValues as plain JSON objects. However, many adopters rely on the `id` being present, and adding it manually is not trivial for nested CoValues. We decided to add the field back as `$jazz.id`, to avoid conflicts with user-defined properties in CoMaps.

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/account.ts
+++ b/packages/jazz-tools/src/tools/coValues/account.ts
@@ -274,7 +274,9 @@ export class Account extends CoValueBase implements CoValue {
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   toJSON(): object | any[] {
-    return {};
+    return {
+      "$jazz.id": this.$jazz.id,
+    };
   }
 
   [inspect]() {
--- a/packages/jazz-tools/src/tools/coValues/account.ts
+++ b/packages/jazz-tools/src/tools/coValues/account.ts
@@ -275,7 +275,7 @@ export class Account extends CoValueBase implements CoValue {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   toJSON(): object | any[] {
     return {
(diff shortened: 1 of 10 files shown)


packages/jazz-tools/src/tools/coValues/account.ts
```diff
@@ -274,7 +274,9 @@ export class Account extends CoValueBase implements CoValue {
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   toJSON(): object | any[] {
-    return {};
+    return {
+      "$jazz.id": this.$jazz.id,
```
gdorsi: I think we should keep it straightforward and go for:

```ts
$jazz: { id: this.$jazz.id }
```

This way you can do `value.$jazz.id` instead of having to do `value["$jazz.id"]`

nrainhart: Done!

packages/jazz-tools/src/tools/tests/coList.test.ts
```diff
@@ -243,6 +243,26 @@ describe("Simple CoList operations", async () => {
         list.$jazz.push("cheese");
         expect(list[3]?.toString()).toBe("cheese");
       });
+
+      test("cannot push a shallowly-loaded CoValue into a deeply-loaded CoList", async () => {
```
nrainhart, line 247: Squeezed in this test to document this discussion: https://discord.com/channels/1139617727565271160/1139621689882321009/1412404126003892354

gdorsi (approved):

PR #2857, opened 2025-09-02
feat: auth token APIs

gdorsi:
# Description

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/interfaces.ts
+++ b/packages/jazz-tools/src/tools/coValues/interfaces.ts
@@ -587,3 +587,73 @@ export function importContentPieces(
     node.syncManager.handleNewContent(piece, "import");
   }
 }
+
+export function generateAuthToken(owner?: Account) {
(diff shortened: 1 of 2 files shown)


packages/jazz-tools/src/tools/coValues/interfaces.ts
```diff
@@ -587,3 +587,73 @@ export function importContentPieces(
...
+    } as const;
+  }
+
+  const node = account.$jazz.localNode;
+  const crypto = node.crypto;
+
+  // Verify the signature of the message to prevent tampering
+  const signPayload = crypto.secureHash({
+    id: account.$jazz.id,
+    createdAt,
```
joeinnes:
```suggestion
    createdAt: Number(createdAt),
```

joeinnes:
https://linear.app/garden-co/issue/GCO-820/rfc-high-level-api-for-auth-token-generation-and-consumption

PR #2858, opened 2025-09-02
Version Control APIs on Jazz-Tools

gdorsi:
Introduces version control APIs on jazz-tools

More info here: https://jazz-homepage-git-feat-merge-api-gcmp.vercel.app/docs/react/using-covalues/version-control

TBD:
- [ ] Do we want to keep the `unstable_` prefix?


homepage/homepage/content/docs/using-covalues/version-control.mdx
```diff
@@ -0,0 +1,240 @@
...
+    owner: privateGroup 
+  }
+});
+
+// Only members of privateGroup can see the branch content
+// The sync server cannot read the branch content
+```
+</CodeGroup>
+
+When a private branch is merged, the changes are merged into the main values, which keeps the privacy of the branch.
```
joeinnes: I'm not sure I understand this. When the branch gets merged, all of the branch's history also gets merged onto `main`, right?

The above implies I could do something like
```ts
const privateBranch = useCoState(Project, projectId, {
  unstable_branch: { 
    name: "private-edit", 
    owner: privateGroup 
  }
});

privateBranch.$jazz.set('myPassword', 'hunter2');
privateBranch.$jazz.delete('myPassword')
privateBranch.$jazz.unstable_merge();
```

and my password would not be visible in the history of the `main` branch.

Maybe better just to remove this all together?

gdorsi: You mean removing:
> When a private branch is merged, the changes are merged into the main values, which keeps the privacy of the branch.

Agree, can be misleading.

gdorsi:
```suggestion
```

homepage/homepage/content/docs/using-covalues/version-control.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+<CodeGroup>
+```ts
+const branch = new CoState(Project, projectId, {
+  unstable_branch: { name: "feature-branch" }
+});
+```
+</CodeGroup>
+</ContentByFramework>
+
```
joeinnes, line 48:
```suggestion

You can also include nested CoValues in your branch by using a [`resolve` query](/docs/using-covalues/subscription-and-loading#resolve-queries). All of the nested CoValues will have their own branched versions created, and you can access them in the same way you would normally access a deeply loaded property.

<Alert variant="info" className="mt-4 flex gap-2 items-center">
  In case you create a separate reference to a nested CoValue (for example by loading it by its ID), you will need to specify the branch you wish to load.
</Alert>
```

packages/jazz-tools/src/svelte/jazz.class.svelte.ts
```diff
@@ -42,13 +45,16 @@ export class CoState<
...
+      const options = this.#options;
 
       return untrack(() => {
         if (!ctx || !id) {
           return this.update(undefined);
         }
         const agent = "me" in ctx ? ctx.me : ctx.guest;
 
+        console.log("options?.unstable_branch", options);
+
```
gdorsi:
```suggestion
```

packages/jazz-tools/src/react-core/hooks.ts
```diff
@@ -275,7 +354,7 @@ function useAccountSubscription<
   const createSubscription = () => {
     const agent = getCurrentAccountFromContextManager(contextManager);
 
-    if (agent[TypeSym] === "Anonymous") {
```
gdorsi, line 440: It was causing a circular import issue

packages/jazz-tools/src/prosemirror/lib/plugin.ts
```diff
@@ -57,7 +57,13 @@ export function createJazzPlugin(
...
 
       if (coRichText) {
-        coRichText.$jazz.subscribe(handleCoRichTextChange);
+        const unsubscribe = coRichText.$jazz.subscribe(handleCoRichTextChange);
+        return {
+          destroy() {
+            setView(undefined);
+            unsubscribe();
+          },
+        };
```
gdorsi, line 66: Found the unsubscribe missing here, so the coValue was never unsubscribed

gdorsi:
Moved into https://github.com/garden-co/jazz/pull/2858

PR #2864, opened 2025-09-03
feat: add skipInboxLoad option to startWorker

gdorsi:
Add the option of skipping the inbox load when calling `startWorker`

This will become the default in the future, and the inbox will require an esplicit load to be subscribed.

Also fixed the return type to be `co.loaded` instead of `InstanceOfSchema`

### DIFF
--- a/.changeset/chilled-bats-run.md
+++ b/.changeset/chilled-bats-run.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Fixed the startWorker type to return a shallow loaded worker account.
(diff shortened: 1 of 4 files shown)


.changeset/chilled-bats-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Fixed the startWorker type to return a shallow loaded worker account.
```
nrainhart:
```suggestion
Fixed the startWorker type to return a shallowly-loaded worker account.
```

nrainhart (approved):

PR #2867, opened 2025-09-04
Fix: BetterAuth provider on React Native

Takeno:
# Description
Initially, there was a trivial `JazzReactProviderWithBetterAuth` utility component, but it caused a runtime error in react-native due to [this function](https://github.com/garden-co/jazz/blob/main/packages/jazz-tools/src/browser/index.ts#L12C1-L12C18) invocation.

The export has been simplified to have a single implementation for both React and React Native.

### DIFF
--- a/.changeset/yellow-crabs-travel.md
+++ b/.changeset/yellow-crabs-travel.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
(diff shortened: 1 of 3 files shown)


Elfo404 (approved):

boorad (approved):
LGTM 🚀

PR #2870, opened 2025-09-04
fix: coList merge

gdorsi:


packages/cojson/src/coValues/coList.ts
```diff
@@ -216,13 +221,17 @@ export class RawCoList<
...
+          const created = this.createInsertionsEntry(opID, {
             madeAt,
             predecessors: [],
             successors: [],
             change,
           });
 
+          if (!created) {
+            continue;
+          }
```
gdorsi, line 233: Skips duplicates when double merge is applied

PR #2871, opened 2025-09-04
Fixes GCO-821: Don't greedily send dependencies to server peers when handling new content

justinrosenthal:
# Description

This is similar to #2795.

When handling new content messages, one of the final steps is to forward that new content on to other connected peers. Previously we would forward the covalue and its dependencies to every peer. This changes that behavior in the case of a server peer -- to only send the covalue itself.

N/A

## Tests

## Checklist


packages/cojson/src/sync.ts
```diff
@@ -777,7 +777,11 @@ export class SyncManager {
 
       // We directly forward the new content to peers that have an active subscription
       if (peer.optimisticKnownStates.has(coValue.id)) {
-        this.sendNewContentIncludingDependencies(coValue.id, peer);
+        if (peer.role === "server") {
+          this.sendNewContentWithoutDependencies(coValue.id, peer);
+        } else {
+          this.sendNewContentIncludingDependencies(coValue.id, peer);
+        }
```
gdorsi, line 784: This change unexpectedly breaks server-to-server communication.

Investigating

gdorsi: Fixed the failing test.

The problem is that, by turning off the eager sync, a bug on the toldKnownState tracking surfaced, showing that when connecting multiple servers to each other the knownState might not always be propagated.

Created an issue: https://github.com/garden-co/jazz/issues/2874

This means that this change doesn't bring a real regression, it just made it easier to reproduce this bug.

gdorsi (approved):
LGTM

justinrosenthal:
Closing in favor of #2872

PR #2872, opened 2025-09-04
Fixes GCO-821: Never include dependencies in sendNewContent for server peers

justinrosenthal:
# Description

This is an alternative for #2871.

Instead of choosing whether to use `sendNewContentIncludingDependencies` or `sendNewContentWithoutDependencies` based on `peer.role` in multiple spots (see #2795), we instead just never include dependencies when it's a `server` peer.

N/A

## Tests

## Checklist

### DIFF
--- a/.changeset/afraid-kiwis-camp.md
+++ b/.changeset/afraid-kiwis-camp.md
@@ -1,0 +1,5 @@
+---
+"cojson": patch
+---
+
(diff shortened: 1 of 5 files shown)


gdorsi (approved):
LGTM, thanks for the cleanup.

Investigated the failing test in https://github.com/garden-co/jazz/pull/2871/files#r2324832776

PR #2877, opened 2025-09-05
Make latency bar red if there's an ongoing outage

joeinnes:
# Description
It was quite frustrating to see during today's outage a bright green latency indicator next to a 'down' indicator. This update cascades the `isUp` property down to the latency chart, and in case `isUp` is false, the last latency indicator will be red.

Technically this is not accurate. It may be the case that we've got high levels of availability during the interval, and the outage does not meaningfully change the percentage, but nevertheless, as a user, I'd expect to see a red bar.

Needs a Grafana key to test, and then change one of the cascaded `isUp` props to false.

## Tests

- [x] Tests have not been updated, because: not testable

## Checklist


joeinnes:
Tested locally, all working.

Elfo404 (approved):

PR #2880, opened 2025-09-05
Export hwrServerPeerSelector

justinrosenthal:
# Description
This exports `hwrServerPeerSelector` because we'll need access to it in `jazz-sync`. Alternatively we could move it there if we don't want `jazz` to be prescriptive in any way if that's preferred.

N/A

## Tests

- [x] Tests have not been updated, because: it's just an export

## Checklist

### DIFF
--- a/.changeset/warm-drinks-invite.md
+++ b/.changeset/warm-drinks-invite.md
@@ -1,0 +1,5 @@
+---
+"cojson": patch
+---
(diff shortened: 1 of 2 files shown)


gdorsi (approved):
LGTM

How about exporting the API directly? Do you see any issues in that?

PR #2882, opened 2025-09-08
feat: improve merge to track the source txID and keep the views consistent

gdorsi:
# Description

Refactored the merge commits to:
- carry the txID of merged commits on the source branch. This fixes the content processing on co.list and co.feed
- store the merged checkpoints in the branch. This way the cost of keeping track of the merged sessions is paid on the branches, which should be more disposable.
- fixed the madeAt value to be the same as the original transaction before the merge. This affects the conflicts resolution, aligning them to our spec

## Tests

### DIFF
--- a/packages/cojson/src/coValueCore/branching.ts
+++ b/packages/cojson/src/coValueCore/branching.ts
@@ -254,10 +254,6 @@ export function mergeBranch(branch: CoValueCore): CoValueCore {
     target.makeTransaction(tx.changes, tx.tx.privacy, mergeMeta, tx.madeAt);
     lastSessionId = tx.txID.sessionID;
     lastBranchId = tx.txID.branch;
-
-    if (tx.txID.txIndex === 0) {
-      mergeMeta.b = branch.id;
-    }
   }
 
   // Track the merged transactions for the branch, so future merges will know which transactions have already been merged
--- a/packages/cojson/src/coValueCore/branching.ts
+++ b/packages/cojson/src/coValueCore/branching.ts
@@ -49,21 +49,16 @@ export function getBranchId(
     );
   }
 
-  if (!ownerId) {
-    const header = coValue.verified.header;
+  const currentOwnerId = ownerId ?? getBranchOwnerId(coValue);
 
-    // Group and account coValues can't have branches, so we return the source id
-    if (header.ruleset.type !== "ownedByGroup") {
-      return coValue.id;
-    }
-
-    ownerId = header.ruleset.group;
+  if (!currentOwnerId) {
+    return coValue.id;
   }
 
   const header = getBranchHeader({
     type: coValue.verified.header.type,
     branchName: name,
-    ownerId,
+    ownerId: currentOwnerId,
     sourceId: coValue.id,
   });
 
@@ -71,8 +66,29 @@ export function getBranchId(
 }
 
 export type BranchCommit = {
-  branch: CoValueKnownState["sessions"];
+  from: CoValueKnownState["sessions"];
 };
+
+export type BranchPointerCommit = {
+  branch: string;
+  ownerId?: RawCoID;
+};
+
+export function getBranchOwnerId(coValue: CoValueCore) {
+  if (!coValue.verified) {
+    throw new Error(
+      "CoValueCore: getBranchOwnerId called on coValue without verified state",
+    );
+  }
+
+  const header = coValue.verified.header;
+
+  if (header.ruleset.type !== "ownedByGroup") {
+    return undefined;
+  }
+
+  return header.ruleset.group;
+}
 
 /**
  * Given a coValue, a branch name and an owner id, creates a new branch CoValue
@@ -88,32 +104,34 @@ export function createBranch(
     );
   }
 
-  if (!ownerId) {
-    const header = coValue.verified.header;
+  const branchOwnerId = ownerId ?? getBranchOwnerId(coValue);
 
-    // Group and account coValues can't have branches, so we return the source coValue
-    if (header.ruleset.type !== "ownedByGroup") {
-      return coValue;
-    }
-
-    ownerId = header.ruleset.group;
+  if (!branchOwnerId) {
+    return coValue;
   }
 
   const header = getBranchHeader({
     type: coValue.verified.header.type,
     branchName: name,
-    ownerId,
+    ownerId: branchOwnerId,
     sourceId: coValue.id,
   });
 
-  const value = coValue.node.createCoValue(header);
+  const branch = coValue.node.createCoValue(header);
+  const sessions = { ...coValue.knownState().sessions };
 
   // Create a branch commit to identify the starting point of the branch
-  value.makeTransaction([], "private", {
-    branch: coValue.knownState().sessions,
+  branch.makeTransaction([], "private", {
+    from: sessions,
   } satisfies BranchCommit);
 
-  return value;
+  // Create a branch pointer, to identify that we created a branch
+  coValue.makeTransaction([], "private", {
+    branch: name,
+    ownerId,
+  } satisfies BranchPointerCommit);
+
+  return branch;
 }
 
 /**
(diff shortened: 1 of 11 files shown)


packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -915,15 +979,46 @@ export class CoValueCore {
...
+    return Boolean(this.verified?.branchSourceId);
+  }
+
+  hasBranch(name: string, ownerId?: RawCoID) {
+    // This function requires the meta information to be parsed, which might not be the case
+    // if the value content hasn't been loaded yet
+    this.parseNewTransactions(false);
+
+    const currentOwnerId = getBranchOwnerId(this);
+    return this.branches.some((item) => {
```
gdorsi, line 984: The branch commits are optimized to contain only the information for future loadings.

For this reason they require multiple checks to fill the meaning of undefined values

packages/cojson/src/tests/branching.test.ts
```diff
@@ -233,18 +236,20 @@ describe("Branching Logic", () => {
 
       expect(list.toJSON()).toEqual([
         "bread",
-        "cheese",
         "apples",
         "oranges",
         "carrots",
         "tomatoes",
         "cucumber",
+        "cheese",
```
nrainhart, line 244: Cheese ends up last, even though it was added later? I thought this it would end up in index 1 now that we're preserving the original madeAt

packages/cojson/src/coValueCore/branching.ts
```diff
@@ -49,31 +49,47 @@ export function getBranchId(
...
 
   return idforHeader(header, coValue.node.crypto);
 }
 
 export type BranchCommit = {
-  branch: CoValueKnownState["sessions"];
+  from: CoValueKnownState["sessions"];
 };
 
+export type BranchPointerCommit = {
```
nrainhart: Mind adding ts docs to each commit type explaining what they are used for?

packages/cojson/src/coValueCore/branching.ts
```diff
@@ -142,12 +160,21 @@ export function getBranchSource(
 }
 
 export type MergeCommit = {
```
nrainhart: Same thing with merge commit types

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -819,7 +883,7 @@ export class CoValueCore {
    * - Decodes the changes & meta for each transaction
    * - Parses the meta information of the transaction
    */
-  private parseNewTransactions(ignorePrivateTransactions: boolean) {
+  parseNewTransactions(ignorePrivateTransactions: boolean) {
```
nrainhart: The private modifier can be kept, right?

packages/cojson/src/tests/sync.upload.test.ts
```diff
@@ -92,11 +92,11 @@ describe("client to server upload", () => {
...
         "client -> server | KNOWN Group sessions: header/5",
         "client -> server | KNOWN Map sessions: header/2",
-        "client -> server | LOAD Branch sessions: empty",
-        "server -> client | KNOWN Branch sessions: empty",
         "client -> server | CONTENT Branch header: true new: After: 0 New: 1",
+        "client -> server | CONTENT Map header: false new: After: 0 New: 1",
         "client -> server | CONTENT Branch header: false new: After: 1 New: 1",
         "server -> client | KNOWN Branch sessions: header/1",
+        "server -> client | KNOWN Map sessions: header/3",
         "server -> client | KNOWN Branch sessions: header/2",
```
nrainhart, line 100: I don't understand why this changed 🤔. Mind explaining?

packages/cojson/src/localNode.ts
```diff
@@ -479,8 +479,14 @@ export class LocalNode {
       return branch.getCurrentContent() as T;
     }
 
-    // Passing skipRetry to true because otherwise creating a new branch would always take 1 retry delay
-    await this.loadCoValueCore(branch.id, undefined, true);
+    // Do a synchronous check to see if the branch exists, if not we don't need to try to load the branch
+    if (!source.hasBranch(branchName, branchOwnerID)) {
```
nrainhart, line 483: Isn't it possible that the branch exists but is not being tracked yet by the source in this node?

packages/cojson/src/tests/branching.test.ts
```diff
@@ -527,4 +548,132 @@ describe("Branching Logic", () => {
...
+
+      await map.core.waitForSync();
+
+      const newSession = client.spawnNewSession();
+      const loadedMapCore = await newSession.node.loadCoValueCore(map.core.id);
+
+      expect(loadedMapCore.hasBranch("feature-branch", group.id)).toBe(true);
+    });
+
+    test("should alias the txID when a transaction comes from a merge", async () => {
```
nrainhart: Should this test be inside the hasBranch describe block?

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -876,7 +940,7 @@ export class CoValueCore {
       const { txID } = transaction;
 
       const from = options?.from?.[txID.sessionID] ?? -1;
-      const to = options?.to?.[txID.sessionID] ?? Infinity;
+      const to = options?.to ? (options.to[txID.sessionID] ?? -1) : Infinity;
```
nrainhart, line 928: May be worth a comment to highlight we ignore any sessions that are not requested

nrainhart (approved):

packages/cojson/src/localNode.ts
gdorsi: Yes, that's the case of when someone created the branch offline and synced it later.

The only case we need to be wary of thou is avoid re-creating the same branch with the same session.
That's the only case that would bring troubles, and I think we should set up some safety net for it anyway.

This check should keep us safe for that because:
- on browsers/mobile the sessions is re-used on reload, but if a branch had been created the info should be here
- on workers sessions are re-created on reload, so no issues

But I'm planning to cover the case for when a unique value is created twice on the same session, because it could now corrupt the signatures, and being a public API we should provide some protection for that.

packages/cojson/src/tests/branching.test.ts
gdorsi: The other append are linked to bread, and as being processed after chees they are placed before it.

It's related to the CoList algorithm, and this is how it deals with this kind of conflicts.

packages/cojson/src/tests/sync.upload.test.ts
gdorsi: It's related to the changes on the checkout method

Now we don't try to load the branch anymore and create it directly if not tracked in the source CoValue.

There is also an additional transaction in the source, which is the branch link

packages/cojson/src/coValueCore/branching.ts
```diff
@@ -232,37 +237,33 @@ export function mergeBranch(branch: CoValueCore): CoValueCore {
...
+      mi: tx.txID.txIndex,
+    };
 
     if (lastSessionId !== tx.txID.sessionID) {
       mergeMeta.s = tx.txID.sessionID;
     }
 
     if (lastBranchId !== tx.txID.branch) {
-      mergeMeta.b = branch.id;
+      mergeMeta.b = tx.txID.branch;
```
gdorsi, line 251: In the case of a branch of a branch we want to preserve the original branch pointer, instead of using the current branch id

packages/cojson/src/coValueCore/branching.ts
```diff
@@ -232,37 +237,33 @@ export function mergeBranch(branch: CoValueCore): CoValueCore {
...
     }
 
     target.makeTransaction(tx.changes, tx.tx.privacy, mergeMeta, tx.madeAt);
     lastSessionId = tx.txID.sessionID;
     lastBranchId = tx.txID.branch;
-  });
+
+    if (tx.txID.txIndex === 0) {
+      mergeMeta.b = branch.id;
+    }
```
nrainhart: Any reason why we're doing this after makeTransaction?

gdorsi: This was test code, going to remove it.

Thanks for catching it.

PR #2888, opened 2025-09-08
GCO-825 - Removes external dependencies for fonts

mculotta120:
# Description
- Removes external dependencies for fonts (builds were failing in vercel when google fonts was down)

1. Start locally, everything should look the same as production site.
2. Run the build for the homepage, it should succeed.

## Tests

## Checklist


joeinnes:
Ran a build with this, homepage fonts and OG Images all looked correct 👍

gdorsi (approved):
Thanks!

PR #2893, opened 2025-09-09
feat: make CoLists structurally equal to arrays

nrainhart:
# Description

I implemented `ownKeys` & `getOwnPropertyDescriptor` in the `CoListProxyHandler` to make the iteration of a CoList's keys the same as an array's.

This means CoLists should be considered structurally equal to arrays in most cases. E.g. this test now passes:
```typescript
const TestList = co.list(z.string());
const list = ["a", "b", "c"];
const coList = TestList.create(list);
expect(coList).toEqual(list);
```

While before it'd fail with a very cryptic `expected ['a', 'b', 'c'] to deeply equal ['a', 'b', 'c']` error.

This was already true for CoMaps and plain JS objects, now it works for CoList as well. We'll need to revisit the approach for both when we remove proxies.

## Tests

## Checklist

### DIFF
--- a/.changeset/metal-apples-shave.md
+++ b/.changeset/metal-apples-shave.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Make CoLists structurally equal to arrays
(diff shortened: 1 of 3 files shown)


gdorsi (approved):
Nice one, thanks!

PR #2895, opened 2025-09-09
feat: add createdAt & lastUpdatedAt property to $jazz

gdorsi:
# Description

Adds lastUpdatedAt property to $jazz

Haven't added `createdAt` because we don't have a stable way to get it from every coValue.

CoMap are an exception because empty CoMap always start with an empty transaction.
Tried to do the same with every other type, but made the whole test suite blow up so I've settled for adding only lastUpdatedAt for now

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/CoValueBase.ts
+++ b/packages/jazz-tools/src/tools/coValues/CoValueBase.ts
@@ -77,4 +77,36 @@ export abstract class CoValueJazzApi<V extends CoValue> {
 
     return new AnonymousJazzAgent(this.localNode);
   }
+
+  /**
+   * The timestamp of the creation time of the CoValue
+   *
+   * @category Content
+   */
+  get createdAt(): number {
+    const createdAt = this.raw.core.verified.header.meta?.createdAt;
+
+    if (typeof createdAt === "string") {
+      return new Date(createdAt).getTime();
+    }
+
+    return this.raw.core.earliestTxMadeAt;
+  }
+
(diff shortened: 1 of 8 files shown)


packages/jazz-tools/src/tools/coValues/CoValueBase.ts
```diff
@@ -77,4 +77,15 @@ export abstract class CoValueJazzApi<V extends CoValue> {
 
     return new AnonymousJazzAgent(this.localNode);
   }
+
+  /**
+   * The timestamp of the last updated time of the CoValue
+   *
+   * Returns 0 if there are no updates.
```
nrainhart: In the future (when all CoValues have a createdAt timestamp) I think it'd make more sense for lastUpdatedAt to match createdAt if there are no edits. But this makes sense for now

nrainhart (approved):
🚀

nrainhart:
PS don't forget to add a changeset 🤓

packages/jazz-tools/src/tools/coValues/CoValueBase.ts
gdorsi: In CoMap, where we expose `.$jazz.createdAt`, this is true.

But agree that making this now, might be a breaking change painful to deal with in the future.
Going to give another try at initializing the creation time tomorrow.

gdorsi: Done!

PR #2898, opened 2025-09-10
fix: better-auth client plugin authenticating jazz multiple times simultaneously

andreibudoi:
# Description
Calling `await betterAuthClient.getSession()` at the root of my Jazz app led to the better-auth plugin repeatedly triggering Jazz authentication during OAuth logins. This, in turn, caused onAnonymousAccountDiscarded to fire multiple times, resulting in duplicate account entries after migration. I observed that the better-auth plugin and ContextManager are missing a state for authentication or anonymous account migration. I just added some checks for that.

The main fix is in the plugin. This is the piece of code that stops my issue.
```
if (
  pendingAuthentication &&
  lastAuthenticatedAccountID === sessionAccountID
) {
  console.log(
    "Authentication with better-auth already in progress for account",
    sessionAccountID,
  );
  return;
}
```
The fix in ContextManager may be overkill but could be useful in the future.

Add the following in the home page of the better-auth example. This should trigger `onAnonymousAccountDiscarded` once with the fix and three times without the fix, when logging in with oauth. 
```
useEffect(() => {
    const loadInitialData = async () => {
      await betterAuthClient.getSession();
    };
    void loadInitialData();
}, []);
```

## Tests

- [x] Tests have been added to cover race conditions

## Checklist


gdorsi (approved):
Nice job, thanks!

packages/jazz-tools/src/tools/implementation/ContextManager.ts
```diff
@@ -68,6 +68,8 @@ export class JazzContextManager<
   protected authSecretStorage = new AuthSecretStorage();
   protected keepContextOpen = false;
   contextPromise: Promise<void> | undefined;
+  private authenticationInProgress = false;
+  private migrationCompleted = false;
```
Takeno: Can you help me understand better what cases are covered by this variable? Preventing concurrent authentications using `authenticationInProgress` should be enough. I tried to remove it, and the tests pass.

packages/jazz-tools/src/tools/implementation/ContextManager.ts
```diff
@@ -68,6 +68,8 @@ export class JazzContextManager<
   protected authSecretStorage = new AuthSecretStorage();
   protected keepContextOpen = false;
   contextPromise: Promise<void> | undefined;
+  private authenticationInProgress = false;
```
Takeno: We could expose this flag in a getter so that anyone can check if an authentication is in progress before starting a new one. What do you think?

packages/jazz-tools/src/better-auth/auth/client.ts
```diff
@@ -26,17 +26,51 @@ export const jazzPluginClient = () => {
...
+      return;
+    }
+
+    if (pendingAuthentication) {
+      console.info(
+        "Different authentication in progress, waiting for completion before starting new auth for",
+        jazzAuth.accountID,
+      );
+      await pendingAuthentication;
+    }
```
Takeno: I like this approach! Just a quick thought—what happens if authentication is called three times with different accounts (due to some bugs)? The first call sets the pendingAuthentication promise, and then the next two will wait for that same promise before they proceed with their concurrent `jazzContext.authenticate()`.

On the other hand, stacking multiple calls might not be the best approach for consistency.

What if we move the `lastAuthenticatedAccountID === jazzAuth.accountID` condition inside the ContextManager to make it noop? And just make this fail in case of a double authentication with different accounts?

We expect getSession to return the same account quickly, so skipping with the above condition should prevent races. WDYT?

Takeno (changes requested):
Hey @andreibudoi , thank you for your contribution!

I have some considerations on the code to make the implementation more resilient in the long term. Waiting for your thoughts!

packages/jazz-tools/src/tools/implementation/ContextManager.ts
andreibudoi: Yeah you're right, I was thinking we might need it to keep track of long running migrations but I don't think that will ever be the case

packages/jazz-tools/src/better-auth/auth/client.ts
andreibudoi: Hmm, you're right, that would simplify things a whole lot. I'll add test cases to cover that too. Should we throw for `lastAuthenticatedAccountID !== jazzAuth.accountID` in both client plugin and ContextManager?

packages/jazz-tools/src/tools/implementation/ContextManager.ts
andreibudoi: Sounds good, should we also expose `lastAuthenticatedAccountID` based on your other comment?

andreibudoi (commented):
Thanks for the comments! Updated the logic. Renamed `lastAuthenticatedAccountID` to `authenticatingAccountID`. I also expose it in ContextManager. I wanted to compute `authenticationInProgress` from `authenticatingAccountID` but I cannot do that because of the `register` case, where we don't have any ID yet. What do you think about the changes?

packages/jazz-tools/src/better-auth/auth/client.ts
Takeno: I would say to keep the control only on the ContextManager, to have it consistent with any other authentication system in the future.

packages/jazz-tools/src/better-auth/auth/tests/client.test.ts
```diff
@@ -334,4 +334,232 @@ describe("Better-Auth client plugin", () => {
...
+      expect(authSecretStorage.isAuthenticated).toBe(true);
+    });
+
+    it("should fail fast when trying to authenticate different accounts concurrently", async () => {
+      const originalCredentials = await authSecretStorage.get();
+      assert(originalCredentials, "Jazz credentials are not available");
+
+      const testAccount1 = await setupJazzTestSync();
+      const testAccount2 = await setupJazzTestSync();
+      const testAccount3 = await setupJazzTestSync();
```
Takeno: Use `setupJazzTestSync` only once per test. For additional accounts, use `createJazzTestAccount()`

packages/jazz-tools/src/better-auth/auth/client.ts
```diff
@@ -26,8 +26,20 @@ export const jazzPluginClient = () => {
...
+      authenticatingAccountID &&
+      authenticatingAccountID !== jazzAuth.accountID
+    ) {
+      throw new Error(
+        `Authentication already in progress for different account (${authenticatingAccountID}), cannot authenticate ${jazzAuth.accountID}`,
+      );
+    }
+
+    authenticatingAccountID = jazzAuth.accountID;
+
```
Takeno: As discussed in the other comment, I would keep this logic (throw error or noop, based on accountID) only inside the ContextManager. The result should be the same behaviour, but the logic is in a single file.
This way, it will be more resilient to any unwanted multiple authentication calls.
Am I missing some cases?

packages/jazz-tools/src/tools/implementation/ContextManager.ts
```diff
@@ -68,6 +68,8 @@ export class JazzContextManager<
   protected authSecretStorage = new AuthSecretStorage();
   protected keepContextOpen = false;
   contextPromise: Promise<void> | undefined;
+  protected authenticationInProgress = false;
+  protected authenticatingAccountID: string | null = null;
```
Takeno: Can we collapse these if we consider a string like "register" as AccountID in the `register` function? It would simply be an identifier to prevent multiple `register` calls anyway.

packages/jazz-tools/src/tools/implementation/ContextManager.ts
```diff
@@ -228,29 +270,45 @@ export class JazzContextManager<
...
-    }).finally(() => {
-      this.keepContextOpen = false;
-    });
-
-    if (migratingAnonymousAccount) {
-      await this.handleAnonymousAccountMigration(prevContext);
+    if (this.authenticationInProgress) {
+      console.warn(
+        "Authentication already in progress, skipping duplicate registration request",
+      );
```
Takeno:
```suggestion

```

As we throw the error, I don't think there's a need for this log here.

packages/jazz-tools/src/tools/implementation/ContextManager.ts
```diff
@@ -228,29 +270,45 @@ export class JazzContextManager<
...
+      throw new Error("Authentication already in progress");
     }
 
-    if (this.context && "me" in this.context) {
-      return this.context.me.$jazz.id;
-    }
+    this.authenticationInProgress = true;
+    // For registration, we don't know the account ID yet, so we'll set it to null
+    this.authenticatingAccountID = null;
```
Takeno: as mentioned above, a string like `"register"` should work for our cases

packages/jazz-tools/src/tools/tests/ContextManager.test.ts
```diff
@@ -540,4 +540,215 @@ describe("ContextManager", () => {
...
+      for (let i = 0; i < 5; i++) {
+        promises.push(manager.authenticate(credentials));
+      }
+
+      await Promise.all(promises);
+
+      // onAnonymousAccountDiscarded should only be called once despite multiple authenticate calls
+      expect(onAnonymousAccountDiscarded).toHaveBeenCalledTimes(1);
+    });
+
```
Takeno, line 569: I would add a test that calls authenticate with two different credentials simultaneously, expecting an error to be thrown.

Takeno (changes requested):
Good work! Thank you very much for your effort here.
The more I see the implementation, the more I think the main logic should be only in the context manager.
Check out my comments and let me know.

packages/jazz-tools/src/tools/tests/ContextManager.test.ts
andreibudoi: Would the test I just added fit that description? I already had the `fails fast when trying to authenticate different accounts concurrently` one which I thought covered that case

packages/jazz-tools/src/better-auth/auth/client.ts
andreibudoi: Can't think of any other scenarios, but I'll quickly look at the other auth methods besides better-auth to see if we missed something, prob not.

packages/jazz-tools/src/tools/implementation/ContextManager.ts
andreibudoi: Wanna export a const string for the "register"? Maybe it's too much

Takeno: Nah, I would keep it only as an internal identifier to know if the authentication is ongoing or not.

Takeno (approved):
LGTM 🚀

gdorsi (approved):
Wonderful job @Takeno and @andreibudoi

PR #2901, opened 2025-09-10
feat: add branch & merge APIs for CoValues

gdorsi:
Introduces version control APIs on jazz-tools

More info here: https://jazz-homepage-git-feat-version-control-gcmp.vercel.app/docs/react/using-covalues/version-control

Design choices:
- Added unstable_branch option in all the loading/subscription APIs (Group & Account excluded)
- Added unstable_branch option in useCoState, useCoStateWithSelector and Svelte CoState
- The results from value.subscribe or value.ensureLoaded inherit the branch configuration from the value subscription
- The `value.$jazz.id` property points to the main branch id, this way the only valid path to load a branch is to use the branch option.

TBD:
- [ ] Do we want to keep the `unstable_` prefix?

Examples updated in a separate PR: https://github.com/garden-co/jazz/pull/2902


packages/jazz-tools/src/prosemirror/lib/plugin.ts
```diff
@@ -57,7 +57,13 @@ export function createJazzPlugin(
...
 
       if (coRichText) {
-        coRichText.$jazz.subscribe(handleCoRichTextChange);
+        const unsubscribe = coRichText.$jazz.subscribe(handleCoRichTextChange);
+        return {
+          destroy() {
+            setView(undefined);
+            unsubscribe();
+          },
+        };
```
gdorsi, line 66: While working on refining the experience of richtext editing I've found that the coRichText subscription is never unsubscribed.

packages/jazz-tools/src/tools/coValues/coFeed.ts
```diff
@@ -334,9 +335,27 @@ export class CoFeedJazzApi<F extends CoFeed> extends CoValueJazzApi<F> {
...
+    if (sourceId) {
+      return sourceId as ID<F>;
+    }
+
     return this.raw.id;
   }
 
+  get branchId(): ID<F> {
+    return this.raw.id;
+  }
```
gdorsi: Unsure that we need to expose this at all.

My idea is that the underlying mechanism of having a different value acting as branch should never be exposed to the user, and I don't see a real use case for this ATM.

homepage/homepage/content/docs/using-covalues/version-control.mdx
```diff
@@ -0,0 +1,246 @@
...
+<CodeGroup>
+```ts
+const branch = new CoState(Project, projectId, {
+  unstable_branch: { name: "feature-branch" }
+});
+```
+</CodeGroup>
+</ContentByFramework>
+
+You can also include nested CoValues in your branch by using a [`resolve` query](/docs/using-covalues/subscription-and-loading#resolve-queries). All of the nested CoValues will have their own branched versions created, and you can access them in the same way you would normally access a deeply loaded property.
```
nrainhart:
```suggestion
You can include nested CoValues in your branch by using a [`resolve` query](/docs/using-covalues/subscription-and-loading#resolve-queries). All of the nested CoValues will have their own branched versions created, and you can access them in the same way you would normally access a deeply loaded property.
```

homepage/homepage/content/docs/using-covalues/version-control.mdx
```diff
@@ -0,0 +1,246 @@
...
+
+<Alert variant="info" className="mt-4 flex gap-2 items-center">
+  In case you create a separate reference to a nested CoValue (for example by loading it by its ID), you will need to specify the branch you wish to load.
+</Alert>
+
+### Making Changes
+
+Once you have a branch, you can make changes using the same methods as the original CoValue:
```
nrainhart:
```suggestion
Once you have a branch, you can make changes just as you would with the original CoValue:
```

homepage/homepage/content/docs/using-covalues/version-control.mdx
```diff
@@ -0,0 +1,246 @@
...
+// Branch modifies priority to "high"
+branch.$jazz.applyDiff({ priority: "high" });
+
+// Meanwhile, main modifies priority to "urgent"
+originalProject.$jazz.applyDiff({ priority: "urgent" });
+
+// Merge the branch
+branch.$jazz.unstable_merge();
+
+// Main's value wins ("urgent") because it was written later
```
nrainhart:
```suggestion
// Main's value ("urgent") wins because it was written later
```

homepage/homepage/content/docs/using-covalues/version-control.mdx
```diff
@@ -0,0 +1,246 @@
...
+```
+</CodeGroup>
+
+All the changes made to the branch will be merged into the main CoValue, preserving both author and timestamp.
+
+Calling `unstable_merge()` will merge every change done in the project and the loaded tasks.
+
+The merge is idempotent, so you can merge the same branch multiple times without changing the result.
+
+The merge operation is scoped, so if you call `unstable_merge()` on a task, only the changes to the task will be merged
```
nrainhart: On the first read this was quite confusing to me. I thought "wait, so merging the project won't merge the tasks either?", even though you explained that 2 paragraphs above 😝

Maybe we should rephrase this to say the merge operation is cascaded down to the covalue's children, but not to its parents.

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -873,6 +888,15 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
...
   }
+
+  /**
+   * Deeply merge the current branch into the main CoValues.
+   *
+   * Doesn't have any effect when there are no changes to merge, or the current CoValue is not a branch
+   */
+  unstable_merge() {
+    unstable_mergeBranch(this.coList);
+  }
```
nrainhart: Can we move this up to `CoValueJazzApi` to avoid repeating it on every CoValue class?

packages/jazz-tools/src/tools/coValues/coFeed.ts
```diff
@@ -1026,9 +1050,23 @@ export class FileStreamJazzApi<F extends FileStream> extends CoValueJazzApi<F> {
    * @category Content
    */
   get id(): ID<F> {
```
nrainhart: Same thing with id, isBranch, and branchName

homepage/homepage/content/docs/using-covalues/version-control.mdx
```diff
@@ -0,0 +1,246 @@
...
+
+<CodeGroup>
+```ts 
+const branch = await Project.load(projectId, {
+  unstable_branch: { name: "feature-branch" }
+});
+
+// Branch ID is the same as source
+console.log(branch.$jazz.branchName); // "feature-branch"
+console.log(branch.$jazz.isBranch); // true
```
nrainhart: I feel like at some point we'll want to make a distinction between the idea of a branch and a "branched value". How do you feel about renaming this method to `isBranched`?

packages/jazz-tools/src/tools/exports.ts
```diff
@@ -94,6 +94,8 @@ export {
   type AuthResult,
   type Credentials,
   type JazzContextWithAccount,
+  type BranchDefinition,
+  unstable_mergeBranch,
```
nrainhart: We want to export the `mergeBranch` function?

packages/jazz-tools/src/tools/subscribe/CoValueCoreSubscription.ts
```diff
@@ -64,19 +62,26 @@ export class CoValueCoreSubscription {
    * Handles the case where the CoValue source is immediately available.
    * Either subscribes directly or attempts to get the requested branch.
    */
-  private handleAvailableSource(source: CoValueCore): void {
-    if (!this.branchName) {
-      this.subscribe(source.getCurrentContent());
+  private handleAvailableSource(): void {
+    if (
+      !this.branchName ||
+      this.source.verified?.header.ruleset.type !== "ownedByGroup"
```
nrainhart: We should have a `canBeBranched` function/method

packages/jazz-tools/src/tools/subscribe/CoValueCoreSubscription.ts
```diff
@@ -64,19 +62,26 @@ export class CoValueCoreSubscription {
...
+    const branch = this.source.getBranch(this.branchName, this.branchOwnerId);
 
     if (branch.isAvailable()) {
       // Branch is available, subscribe to it
       this.subscribe(branch.getCurrentContent());
       return;
+      // If the branch hasn't been created, we create it directly so we can syncronously subscribe to it
+    } else if (!this.source.hasBranch(this.branchName, this.branchOwnerId)) {
+      this.source.createBranch(this.branchName, this.branchOwnerId);
+      this.subscribe(branch.getCurrentContent());
```
nrainhart, line 87: This increases the risk of two users creating the same branch, right? I assume you decided to still do this optimization because that situation should be somewhat rare, and being able to synchronously subscribe to a new branch is pretty useful?

packages/jazz-tools/src/tools/coValues/interfaces.ts
```diff
@@ -587,3 +606,23 @@ export function importContentPieces(
     node.syncManager.handleNewContent(piece, "import");
   }
 }
+
+export function unstable_mergeBranch<V extends CoValue>(existing: V) {
+  if (!existing.$jazz.raw.core.isBranch()) {
```
nrainhart:
```suggestion
  if (!existing.$jazz.isBranch()) {
```

homepage/homepage/content/docs/using-covalues/version-control.mdx
```diff
@@ -0,0 +1,246 @@
...
+
+You can get the current branch information from the `$jazz` field.
+
+<CodeGroup>
+```ts 
+const branch = await Project.load(projectId, {
+  unstable_branch: { name: "feature-branch" }
+});
+
+// Branch ID is the same as source
```
nrainhart:
```suggestion
console.log(branch.$jazz.id); // Branch ID is the same as source
```

homepage/homepage/content/docs/using-covalues/version-control.mdx
nrainhart: Should we clarify that autoloaded CoValues are also branched?

packages/jazz-tools/src/react-core/hooks.ts
```diff
@@ -240,6 +266,59 @@ export function useCoState<
   options?: {
     /** Resolve query to specify which nested CoValues to load */
     resolve?: ResolveQueryStrict<S, R>;
+    /**
+     * Create or load a branch for isolated editing.
+     *
+     * Branches allow you to work on CoValues in isolation before merging changes back.
+     * This is useful for implementing features like drafts, isolated editing,
```
nrainhart:
```suggestion
     * This is useful for implementing features like drafts, private editing,
```
Just to avoid repetition with the previous sentence.

homepage/homepage/content/docs/using-covalues/version-control.mdx
```diff
@@ -0,0 +1,246 @@
...
+// Main's value wins ("urgent") because it was written later
+console.log(originalProject.priority); // "urgent"
+```
+</CodeGroup>
+
+## Private branches
+
+When the owner is not specified, the branch has the same permissions as the main values.
+
+You can also create a private branch by providing a group owner.
```
nrainhart, line 280: Maybe we should have a more explicit callout saying that same branch name + different owner actually create two different branches

nrainhart (commented):
It looks great!! Left several comments, but nothing major. I still need to review the test changes, I'll do it tomorrow morning

homepage/homepage/content/docs/using-covalues/version-control.mdx
gdorsi: I also had a thought about it, but don't want to complicate things too much.

Maybe I should change the explaination to say that:
- Every sub-property that comes from `branch` is going to be branched
- When using `branch.$jazz.ensureLoaded` and `branch.$jazz.subscribe` the branch checkout is applied to their results

This way we give a more broad idea on how the branch selection is propagated to childs & sub-subscritions.

gdorsi: Done

homepage/homepage/content/docs/using-covalues/version-control.mdx
gdorsi: Refactored following your suggestion

homepage/homepage/content/docs/using-covalues/version-control.mdx
gdorsi: Added an alert below

homepage/homepage/content/docs/using-covalues/version-control.mdx
gdorsi: Good point, I like isBranched!

packages/jazz-tools/src/react-core/hooks.ts
gdorsi: private !== isolated

But agree that this explaination needs some improvements, changed it into

```
    /**
     * Create or load a branch for isolated editing.
     *
     * Branching lets you take a snapshot of the current state and start modifying it without affecting the canonical/shared version.
     * It's a fork of your data graph: the same schema, but with diverging values.
     *
     */
```

packages/jazz-tools/src/tools/exports.ts
gdorsi: No, we don't need it now.

packages/jazz-tools/src/tools/subscribe/CoValueCoreSubscription.ts
gdorsi: > This increases the risk of two users creating the same branch, right?

That should be ok, just landed an improvement on how to handle such conflicts. https://github.com/garden-co/jazz/pull/2901/commits/1d7f5c8dcc57f9a27e7d23c1424785b06b0f7f78

This also prepares the groundwork for `value.$jazz.rebase`

gdorsi: The only case that breaks stuff now is when a user creates the same branch in the same session.

This shouldn't happen, but I want to get this handled anyway because it's related to `uniqueness` more than with branching

homepage/homepage/content/docs/using-covalues/version-control.mdx
```diff
@@ -0,0 +1,255 @@
...
+const branch = new CoState(Project, projectId, {
+  unstable_branch: { name: "feature-branch" }
+});
+```
+</CodeGroup>
+</ContentByFramework>
+
+You can also include nested CoValues in your branch by using a [`resolve` query](/docs/using-covalues/subscription-and-loading#resolve-queries). When you create a branch, the branch selection automatically propagates to all sub-properties and nested CoValues:
+
+- **Every sub-property** that comes from the branched CoValue will also be branched
```
nrainhart: Can we use a different term than _sub-property_?

nrainhart (approved):
Tests look good! The only outstanding thing I'd love to get in is moving id, merge, branchName & isBranch into CoValueJazzApi if possible.

But other than that, this looks good to go!

gdorsi:
> The only outstanding thing I'd love to get in is moving id, merge, branchName & isBranch into CoValueJazzApi if possible.

Haven't done that because Group & Account can't have branches.

homepage/homepage/content/docs/using-covalues/version-control.mdx
```diff
@@ -0,0 +1,319 @@
...
+
+When you create a branch, the branch selection automatically propagates to nested CoValues:
+- **Nested CoValues** will have their own branched versions created automatically when using a `resolve` query
+- `branch.$jazz.ensureLoaded()` and `branch.$jazz.subscribe()` won't have the branch checkout applied to their results, unless specified otherwise
+
+This means you can access branched nested CoValues in the same way you would normally access a deeply loaded property, and all operations will work within the branch context.
+
+<Alert variant="info" className="mt-4 flex gap-2 items-center">
+  In case you create a separate reference to a nested CoValue (for example by loading it by its ID), you will need to specify the branch you wish to load.
+</Alert>
```
joeinnes:
```suggestion
You are in control of how nested CoValues are included in your branch. When you specify the CoValue to branch, any nested CoValues specified in a `resolve` query will also be branched. Nested CoValues *not* specified in your resolve query will not be branched.

In order to access branched nested CoValues, you should access them in the same way you would normally access a deeply loaded property, and all operations will work within the branch context.

<Alert variant="info" className="mt-4 flex gap-2 items-center">
  In case you create a separate reference to a nested CoValue (for example by loading it by its ID), or you use `.$jazz.ensureLoaded()` or `.$jazz.subscribe()`, you will need to specify the branch you wish to load.
</Alert>
```

We might need to clarify this a bit with an example.

gdorsi: Already way better, thanks!

PR #2907, opened 2025-09-11
feat: useAccountWithSelector hook

gdorsi:
# Description

Adds a new `useAccountWithSelector` hook, which simplifies reaching nested values the user account.

Tested in the music player, and it's more useful than I would have expected.

## Tests

## Checklist

### DIFF
--- a/examples/music-player/src/components/MusicTrackRow.tsx
+++ b/examples/music-player/src/components/MusicTrackRow.tsx
@@ -46,7 +46,7 @@ export function MusicTrackRow({
     resolve: {
       root: { playlists: { $onError: null, $each: { tracks: true } } },
     },
-    select: (account) => account?.root.playlists ?? [],
+    select: (account) => account?.root.playlists,
   });
 
   const isActiveTrack = useAccountWithSelector(MusicaAccount, {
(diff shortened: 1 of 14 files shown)


examples/music-player/src/components/MusicTrackRow.tsx
```diff
@@ -42,14 +42,21 @@ export function MusicTrackRow({
   const [isDropdownOpen, setIsDropdownOpen] = useState(false);
   const [isHovered, setIsHovered] = useState(false);
 
-  const { me } = useAccount(MusicaAccount, {
+  const playlists = useAccountWithSelector(MusicaAccount, {
     resolve: {
       root: { playlists: { $onError: null, $each: { tracks: true } } },
     },
+    select: (account) => account?.root.playlists ?? [],
```
nrainhart: Returning a new array each time the select fn is called will cause the component to re-render more often than it should

nrainhart (approved):

examples/music-player/src/components/MusicTrackRow.tsx
gdorsi: Ha! Nasty footgun!

PR #2911, opened 2025-09-12
feat: enhance invite listener to handle hash changes

joeinnes:
# Description
By default, the `createInviteLink` generates a hash link. This does not trigger a navigation or the listener to be reinstantiated. As a result, if you're currently viewing `localhost:5173` and you paste the link `http://localhost:5173/#/invite/co_z<val>/inviteSecret_z<secret>`, the invite is not accepted, and the `onAccept` callback doesn't run.

Implements the TODO, by adding an event listener to hash-change events.

Check that you can create and accept invites, without route changes (i.e.: hash change only)

## Tests

## Checklist

### DIFF
--- a/.changeset/calm-cups-compare.md
+++ b/.changeset/calm-cups-compare.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Update the Svelte InviteListener to listen to hash change events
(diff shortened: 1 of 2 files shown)


gdorsi (approved):
LGTM!

Could you add a changeset?

PR #2916, opened 2025-09-12
fix: improve header validation in CoValueCore and SyncManager

gdorsi:
# Description

Added a validation to provideHeader and syncManager to accept only headers which hash matches the given id.

## Tests

### DIFF
--- a/.changeset/dirty-weeks-itch.md
+++ b/.changeset/dirty-weeks-itch.md
@@ -1,0 +1,5 @@
+---
+"cojson": patch
+---
+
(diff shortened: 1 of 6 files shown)


nrainhart (approved):
The new check makes sense. Was there a real-world case that prompted you to add it now?

gdorsi:
> The new check makes sense. Was there a real-world case that prompted you to add it now?

The branching is based on uniqueness, so while thinking for possible breaking situations this one did come in my mind.

PR #2918, opened 2025-09-15
feat: WASMcrypto on cloudflare

antoniomuso:
# Description

experiment [WASM Crypto on Cloudflare & v0](https://linear.app/garden-co/project/wasmcrypto-support-for-cloudflare-and-v0-99cef0569acc/overview)

This PR enables WasmCrypto to work even in edge runtimes like Cloudflare Workers and Vercel functions in `edge` mode.
```ts
import "jazz-tools/load-edge-wasm"

export default {
  async fetch(req) {
     // start to use jazz with WasmCrypto
  },
};

```

Vercel function tests don't work for the moment with our test suite; you have to call them independently. I want to convert it to Vitest.

## Tests

## Checklist


homepage/homepage/content/docs/server-side/runtime.mdx
```diff
@@ -0,0 +1,59 @@
+
+import { CodeGroup } from "@/components/forMdx";
+export const metadata = {
+    description: "How to support different runtimes as Cloudflare and Vercel Edge"
+};
+
+# Wasm on Edge runtimes
+
+On some edge platforms, such as Cloudflare Workers or Vercel Edge Functions, WASM may fail to load due to environment security restrictions. To prevent this, you should import the WASM loader before using Jazz. Otherwise, Jazz will fall back to the slower JavaScript implementation. For example:
```
gdorsi: I think this could go on https://jazz.tools/docs/react/server-side/setup.

We should specify that the import should be placed at the start of the entry module, like a polyfill.

The fallback on Javascript could be explained after the example, mentioning the log they would see in the case of fallback.

gdorsi (approved):
Wonderful job!

Left a comment on the docs, the code looks good and the solution quite clean.

homepage/homepage/content/docs/server-side/runtime.mdx
```diff
@@ -0,0 +1,59 @@
+
+import { CodeGroup } from "@/components/forMdx";
+export const metadata = {
+    description: "How to support different runtimes as Cloudflare and Vercel Edge"
```
joeinnes:
```suggestion
    description: "How to use Jazz on edge runtimes."
```

homepage/homepage/content/docs/server-side/runtime.mdx
joeinnes:
```suggestion
To maximise compatibility, Jazz falls back to a slower, JavaScript crypto implementation if the faster WASM implementation is not available.

On some edge platforms, such as Cloudflare Workers or Vercel Edge Functions, environment security restrictions may trigger this fallback unnecessarily.

You can ensure that Jazz uses the faster WASM implementation by importing the WASM loader before using Jazz. For example:
```

We can focus on a positive story here with a few minor wording changes :)

antoniomuso: I love it :+1:

.changeset/fluffy-kiwis-sell.md
```diff
@@ -0,0 +1,10 @@
+---
+"cojson-core-wasm": minor
+"jazz-tools": minor
+"cojson": minor
```
gdorsi:
```suggestion
"cojson-core-wasm": patch
"jazz-tools": patch
"cojson": patch
```

We ship all non-breaking changes as patch, since we are still in v0

antoniomuso: Ok :+1:

homepage/homepage/content/docs/server-side/runtime.mdx
antoniomuso: Done

PR #2924, opened 2025-09-16
refactor: remove unused crypto APIs and clean up agents caching

gdorsi:
Been working on trying to optimize PureJSCrypto and found somethings that were unused and needed some cleanup


packages/cojson/src/coValues/account.ts
```diff
@@ -117,11 +117,6 @@ export interface ControlledAccountOrAgent {
 export class ControlledAccount implements ControlledAccountOrAgent {
   account: RawAccount<AccountMeta>;
   agentSecret: AgentSecret;
-  _cachedCurrentAgentID: AgentID | undefined;
-  _cachedCurrentSignerID: SignerID | undefined;
-  _cachedCurrentSignerSecret: SignerSecret | undefined;
-  _cachedCurrentSealerID: SealerID | undefined;
-  _cachedCurrentSealerSecret: SealerSecret | undefined;
```
gdorsi, line 124: Moved the cache inside `crypto` because it makes more sense there

PR #2926, opened 2025-09-17
perf: optimize browser storage with meta preload and limiting the concurrency on load

gdorsi:
Loading 1k tasks.

Before:
Screenshot 2025-09-17 at 18 34 20

After:
Screenshot 2025-09-17 at 18 33 47

### DIFF
--- a/packages/cojson-storage-indexeddb/src/idbClient.ts
+++ b/packages/cojson-storage-indexeddb/src/idbClient.ts
@@ -13,7 +13,6 @@ import {
   putIndexedDbStore,
   queryIndexedDbStore,
 } from "./CoJsonIDBTransaction.js";
-import { StoreName } from "./CoJsonIDBTransaction.js";
 
 export class IDBClient implements DBClientInterfaceAsync {
   private db;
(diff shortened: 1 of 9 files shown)


packages/cojson/src/storage/storageAsync.ts
```diff
@@ -44,12 +46,19 @@ export class StorageApiAsync implements StorageAPI {
...
     id: string,
     callback: (data: NewContentMessage) => void,
     done: (found: boolean) => void,
   ) {
-    await this.loadCoValue(id, callback, done);
+    // Limit the parallel loading of coValues to balance the load on the database
+    // and the data processing
+    this.semaphore.acquire(() => {
+      this.loadCoValue(id, callback, (found) => {
+        this.semaphore.release();
```
nrainhart, line 58: How important is it to allow releasing the semaphore before calling the done callback? The API would be simpler and less error-prone if the release happened automatically once acquire's callback finishes executing

packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -13,16 +13,84 @@ import {
...
     this.db = db;
+
+    /**
+     * Preloads the coValues, sessions and signatureAfter into memory
+     *
+     * In the browsers context, this take a few milliseconds an it makes the coValue loads
+     * shaving off around the 25% of the load time
+     *
+     * The memory allocated is limited, because the header, sessions and signatureAfter sizes are somewhat fixed
+     * because they don't contain any user data
```
nrainhart, line 37: Just to double-check: this is fine even if the DB has millions of covalues?

nrainhart (approved):
Left two minor questions, but this looks great!

PR #2927, opened 2025-09-17
feat: optimize loading CoLists with lots of elements

nrainhart:
# Description

Loading a CoList with 10k elements currently takes 8s on my Macbook M4. A lot of that time is spent on the SubscriptionScope. In particular, checking if updates should be fired while loading all CoList elements takes 2.5s:

before

Added a separate counter to keep track of unloaded children, in order to brush off that time  from the load:

after

_Note:_ the first commit has a different optimization using a counter for keeping track of pending explicitly-loaded children. I'm not a fan of it, but if anyone's interested, feel free to take a look.

## Tests

- [x] Tests have not been updated, because: this is an optimization, it doesn't affect any functionality

## Checklist


gdorsi (approved):
Awesome work!

PR #2936, opened 2025-09-19
feat(jazz-tools): Add CoVector schema, covalue, definer, tests

kysely:
# Description
This PR adds the new **CoVector** CoValue to enable storing vectors.

CoVector type extends _Float32Array_ which is ideal to store vectors because:
- it's JS's chosen way to store arrays of 32-bit floating point numbers
- it's a view of an underlying binary data buffer so it can be easily synced using `BinaryCoStream`

CoVector is meant to be immutable, so setters & mutation methods are disallowed.

_This PR only adds the CoValue, the vector search and other features are to be added on top of this change_

```sh
pnpm --filter jazz-tools test:watch covector --reporter verbose
```

## Tests

```sh
 ✓  jazz-tools  src/tools/tests/coVector.test.ts (74 tests) 6936ms
   ✓ Defining a co.vector() (3)
     ✓ given a non-positive dimensions count • throws an error 63ms
     ✓ given a non-integer dimensions count • throws an error 55ms
     ✓ given a valid dimensions count • creates a CoVector schema 57ms
   ✓ Creating a CoVector (5)
     ✓ given a vector of mismatched dimensions count • throws an error 57ms
     ✓ given a Array<number> with valid dimensions count • creates a CoVector 62ms
     ✓ given a Float32Array with valid dimensions count • creates a CoVector 63ms
     ✓ given Account as owner • creates a CoVector 56ms
     ✓ given Group as owner • creates a CoVector 56ms
   ✓ CoVector structure (3)
     ✓ CoVector keys can be iterated over just like an Float32Array's 54ms
     ✓ Float32Array can be constructed from a CoVector 55ms
     ✓ a CoVector is NOT structurally equal to a Float32Array 53ms
   ✓ CoVector methods & properties (Float32Array-like) (30)
     ✓ has .length 55ms
     ✓ has .buffer 53ms
     ✓ has .byteLength 53ms
     ✓ has .byteOffset 155ms
     ✓ has index access 74ms
     ✓ supports .at 54ms
     ✓ supports .entries 54ms
     ✓ supports .every 55ms
     ✓ supports .filter 54ms
     ✓ supports .find 53ms
     ✓ supports .findIndex 52ms
     ✓ supports .findLast 52ms
     ✓ supports .findLastIndex 53ms
     ✓ supports .forEach 55ms
     ✓ supports .includes 52ms
     ✓ supports .indexOf 52ms
     ✓ supports .join 51ms
     ✓ supports .keys 51ms
     ✓ supports .lastIndexOf 60ms
     ✓ supports .map 54ms
     ✓ supports .reduce 56ms
     ✓ supports .reduceRight 54ms
     ✓ supports .slice 54ms
     ✓ supports .some 53ms
     ✓ supports .toJSON (JSON.stringify) 53ms
     ✓ supports .toLocaleString 66ms
     ✓ supports .toReversed 52ms
     ✓ supports .toSorted 58ms
     ✓ supports .toString 51ms
     ✓ supports .values 52ms
   ✓ CoVector setters & mutators (8)
     ✓ updating value at index • throws an error 54ms
     ✓ calling .copyWithin • throws an error 52ms
     ✓ calling .fill • throws an error 53ms
     ✓ calling .reverse • throws an error 52ms
     ✓ calling .set • throws an error 52ms
     ✓ calling .sort • throws an error 54ms
     ✓ calling .subarray • throws an error 55ms
     ✓ calling .with • throws an error 54ms
   ✓ Vector calculations (13)
     ✓ magnitude (3)
       ✓ schema method 54ms
       ✓ static method 54ms
       ✓ instance method 54ms
     ✓ normalize (3)
       ✓ schema method 53ms
       ✓ static method 53ms
       ✓ instance method 53ms
     ✓ dot product (3)
       ✓ schema method 53ms
       ✓ static method 52ms
       ✓ instance method 54ms
     ✓ cosine similarity (3)
       ✓ schema method 53ms
       ✓ static method 55ms
       ✓ instance method 55ms
     ✓ equals 53ms
   ✓ CoVector loading & availability (4)
     ✓ when .waitForSync is called • the entire vector is uploaded 82ms
     ✓ when .waitForSync is called on 1024 kB vector • the entire vector is uploaded 133ms
     ✓ when calling .load from different account • syncs across peers 139ms
     ✓ when calling .load on a 1024 kB vector from different account • syncs across peers  449ms
   ✓ CoVector in subscription (6)
     ✓ standalone CoVector (2)
       ✓ subscribing to locally available CoVector • sends update holding the CoVector 167ms
       ✓ subscribing to remotely available CoVector • sends update holding the CoVector 202ms
     ✓ CoVector in CoList[CoMap] (4)
       ✓ locally available list (2)
         ✓ subscribing with deep resolve • sends update holding the CoVectors 272ms
         ✓ subscribing with autoload • sends update holding the CoVectors 272ms
       ✓ remotely available list (2)
         ✓ subscribing with deep resolve • sends update holding the CoVectors  614ms
         ✓ subscribing with autoload • sends update holding the CoVectors  817ms
   ✓ CoVector dimension mismatch after loading (2)
     ✓ accessing CoVector loaded with an incompatible schema • throws an error 97ms
     ✓ accessing CoVector from subscription, loaded with an incompatible schema • throws an error 200ms
 ✓  TS   jazz-tools  src/tools/tests/coVector.test-d.ts (3 tests)
   ✓ CoVector types (3)
     ✓ co.vector() • defines a correct CoVectorSchema
     ✓ co.vector().create() • creates a CoVector with Float32Array-like typing
     ✓ CoVector instance • has the owner property

 Test Files  2 passed (2)
      Tests  77 passed (77)
Type Errors  no errors
   Start at  13:27:25
   Duration  9.21s (transform 617ms, setup 521ms, collect 905ms, tests 6.94s, environment 0ms, prepare 50ms, typecheck 9.16s)
```

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/coVector.ts
+++ b/packages/jazz-tools/src/tools/coValues/coVector.ts
@@ -41,6 +41,7 @@ export class CoVector extends Float32Array implements CoValue {
 
   protected static requiredDimensionsCount: number | undefined = undefined;
   private declare _loadedVector: Float32Array | null;
+  private declare _requiredDimensionsCount: number;
 
   constructor(
     options:
@@ -52,6 +53,15 @@ export class CoVector extends Float32Array implements CoValue {
         },
   ) {
     super();
+
+    const dimensionsCount = (this.constructor as typeof CoVector)
+      .requiredDimensionsCount;
+
+    if (dimensionsCount === undefined) {
+      throw new Error(
+        "Instantiating CoVector without a dimensions count is not allowed. Use co.vector(...).create() instead.",
+      );
+    }
 
     const proxy = new Proxy(this, CoVectorProxyHandler as ProxyHandler<this>);
 
@@ -71,6 +81,11 @@ export class CoVector extends Float32Array implements CoValue {
         enumerable: false,
       },
       _loadedVector: { value: null, enumerable: false, writable: true },
+      _requiredDimensionsCount: {
+        value: dimensionsCount,
+        enumerable: false,
+        writable: false,
+      },
     });
 
     return proxy;
@@ -160,10 +175,6 @@ export class CoVector extends Float32Array implements CoValue {
     return new Float32Array(u8.buffer, u8.byteOffset, total / 4);
   }
 
-  private get requiredDimensionsCount(): number | undefined {
-    return (this.constructor as typeof CoVector).requiredDimensionsCount;
-  }
-
   get vector(): Float32Array {
     if (this._loadedVector !== null) {
       return this._loadedVector;
@@ -178,12 +189,9 @@ export class CoVector extends Float32Array implements CoValue {
 
     const vector = CoVector.fromByteArray(chunks.chunks);
 
-    if (
-      this.requiredDimensionsCount !== undefined &&
-      vector.length !== this.requiredDimensionsCount
-    ) {
+    if (vector.length !== this._requiredDimensionsCount) {
       throw new Error(
-        `Vector dimension mismatch! CoVector '${this.$jazz.raw.id}' loaded with ${vector.length} dimensions, but the schema requires ${this.requiredDimensionsCount} dimensions`,
+        `Vector dimension mismatch! CoVector '${this.$jazz.raw.id}' loaded with ${vector.length} dimensions, but the schema requires ${this._requiredDimensionsCount} dimensions`,
       );
     }
 
--- a/packages/jazz-tools/src/tools/coValues/coVector.ts
+++ b/packages/jazz-tools/src/tools/coValues/coVector.ts
@@ -611,14 +611,6 @@ export class CoVectorJazzApi<V extends CoVector> extends CoValueJazzApi<V> {
     super(coVector);
   }
 
-  /**
-   * The ID of this `CoVector`
-   * @category Content
-   */
-  get id(): ID<V> {
-    return this.raw.id;
-  }
-
   get owner(): Group {
     return getCoValueOwner(this.coVector);
   }
--- a/packages/jazz-tools/src/tools/coValues/coVector.ts
+++ b/packages/jazz-tools/src/tools/coValues/coVector.ts
@@ -1,0 +1,696 @@
+import type { RawBinaryCoStream } from "cojson";
+import { cojsonInternals } from "cojson";
+import {
+  AnonymousJazzAgent,
+  CoValue,
+  CoValueClass,
+  getCoValueOwner,
+  Group,
+  ID,
+  RefsToResolve,
+  Resolved,
+  SubscribeListenerOptions,
+  SubscribeRestArgs,
+  TypeSym,
+} from "../internal.js";
+import {
+  Account,
+  CoValueJazzApi,
+  inspect,
+  loadCoValueWithoutMe,
+  parseCoValueCreateOptions,
+  parseSubscribeRestArgs,
+  subscribeToCoValueWithoutMe,
+  subscribeToExistingCoValue,
+} from "../internal.js";
+
+/**
+ * CoVectors are collaborative storages of vectors (floating point arrays).
+ *
+ * @category CoValues
+ */
+export class CoVector extends Float32Array implements CoValue {
+  declare $jazz: CoVectorJazzApi<this>;
+
+  /** @category Type Helpers */
+  declare [TypeSym]: "BinaryCoStream";
+
+  static get [Symbol.species]() {
+    return Float32Array;
+  }
+
+  static requiredDimensionsCount: number | undefined = undefined;
+  declare _loadedVector: Float32Array | null;
+
+  constructor(
+    options:
+      | {
+          owner: Account | Group;
+        }
+      | {
+          fromRaw: RawBinaryCoStream;
+        },
+  ) {
+    super();
+
+    const proxy = new Proxy(this, CoVectorProxyHandler as ProxyHandler<this>);
+
+    let raw: RawBinaryCoStream;
+
+    if ("fromRaw" in options) {
+      raw = options.fromRaw;
+    } else {
+      const rawOwner = options.owner.$jazz.raw;
+      raw = rawOwner.createBinaryStream();
+    }
+
+    Object.defineProperties(this, {
+      [TypeSym]: { value: "BinaryCoStream", enumerable: false },
+      $jazz: {
+        value: new CoVectorJazzApi(proxy, raw),
+        enumerable: false,
+      },
+      _loadedVector: { value: null, enumerable: false, writable: true },
+    });
+
+    return proxy;
+  }
+
+  /** @category Internals */
+  static fromRaw<V extends CoVector>(
+    this: CoValueClass<V> & typeof CoVector,
+    raw: RawBinaryCoStream,
+  ) {
+    return new this({ fromRaw: raw });
+  }
+
+  /**
+   * Create a new `CoVector` instance with the given vector.
+   *
+   * @category Creation
+   * @deprecated Use `co.vector(...).create` instead.
+   */
+  static create<S extends CoVector>(
+    this: CoValueClass<S> & { requiredDimensionsCount: number | undefined },
+    vector: number[] | Float32Array,
+    options?: { owner?: Account | Group } | Account | Group,
+  ) {
+    const vectorAsFloat32Array =
+      vector instanceof Float32Array ? vector : new Float32Array(vector);
+
+    const givenVectorDimensions =
+      vectorAsFloat32Array.byteLength / vectorAsFloat32Array.BYTES_PER_ELEMENT;
+
+    if (
+      this.requiredDimensionsCount !== undefined &&
+      givenVectorDimensions !== this.requiredDimensionsCount
+    ) {
+      throw new Error(
+        `Vector dimension mismatch! Expected ${this.requiredDimensionsCount} dimensions, got ${
+          givenVectorDimensions
+        }`,
+      );
+    }
+
+    const coVector = new this(parseCoValueCreateOptions(options));
+    coVector._loadedVector = vectorAsFloat32Array;
+
+    const byteArray = CoVector.toByteArray(vectorAsFloat32Array);
+
+    coVector.$jazz.raw.startBinaryStream({
+      mimeType: "application/vector+octet-stream",
+      totalSizeBytes: byteArray.byteLength,
+    });
+
+    const chunkSize =
+      cojsonInternals.TRANSACTION_CONFIG.MAX_RECOMMENDED_TX_SIZE;
+
+    // Although most embedding vectors are small
+    // (3072-dimensional vector is only 12,288 bytes),
+    // we should still chunk the data to avoid transaction size limits
+    for (let idx = 0; idx < byteArray.length; idx += chunkSize) {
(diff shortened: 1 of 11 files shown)


packages/jazz-tools/src/tools/coValues/coVector.ts
```diff
@@ -0,0 +1,696 @@
...
+
+  /** @category Type Helpers */
+  declare [TypeSym]: "BinaryCoStream";
+
+  static get [Symbol.species]() {
+    return Float32Array;
+  }
+
+  static requiredDimensionsCount: number | undefined = undefined;
+  declare _loadedVector: Float32Array | null;
```
nrainhart: Could these two properties be private?

packages/jazz-tools/src/tools/coValues/coVector.ts
```diff
@@ -0,0 +1,696 @@
...
+      u8.set(c, off);
+      off += c.byteLength;
+    }
+
+    return new Float32Array(u8.buffer, u8.byteOffset, total / 4);
+  }
+
+  get requiredDimensionsCount(): number | undefined {
+    return (this.constructor as typeof CoVector).requiredDimensionsCount;
+  }
```
nrainhart: In practice all concrete CoVector classes will have a `requiredDimensionCount`, right? Should this getter include undefined?

packages/jazz-tools/src/tools/coValues/coVector.ts
```diff
@@ -0,0 +1,696 @@
...
+    return subscribeToCoValueWithoutMe<V, R>(this, id, options, listener);
+  }
+
+  // Vector operations
+  /**
+   * Calculate the magnitude of a vector.
+   */
+  static magnitude(vector: Float32Array | CoVector): number {
+    return Math.sqrt(vector.reduce((s, x) => s + x * x, 0));
+  }
```
nrainhart, line 295: Any reason to provide both static and instance methods? Except for loading/subscriptions, CoValue classes implement all operations as instance methods.

packages/jazz-tools/src/tools/coValues/coVector.ts
```diff
@@ -0,0 +1,696 @@
...
+  subscribeToCoValueWithoutMe,
+  subscribeToExistingCoValue,
+} from "../internal.js";
+
+/**
+ * CoVectors are collaborative storages of vectors (floating point arrays).
+ *
+ * @category CoValues
+ */
+export class CoVector extends Float32Array implements CoValue {
```
nrainhart, line 32: Are we overriding all `Float32Array` methods? If so, should we be implementing Float32Array as an interface instead, so that we don't accidentally leak Float32Array instance methods if we don't override them?
```suggestion
export class CoVector implements Float32Array, CoValue {
```

packages/jazz-tools/src/tools/coValues/coVector.ts
```diff
@@ -0,0 +1,696 @@
...
+    super(coVector);
+  }
+
+  /**
+   * The ID of this `CoVector`
+   * @category Content
+   */
+  get id(): ID<V> {
+    return this.raw.id;
+  }
```
nrainhart: Nit: this is no longer needed, Guido moved it to `CoValueJazzApi`

packages/jazz-tools/src/tools/tests/coVector.test.ts
```diff
@@ -0,0 +1,928 @@
...
+
+  test("Float32Array can be constructed from a CoVector", () => {
+    expect(new Float32Array(coVector)).toEqual(vector);
+  });
+
+  test("a CoVector is NOT structurally equal to a Float32Array", () => {
+    // Since CoVector is a proxy, it will never pass the equality check with a Float32Array (TypedArray)
+    expect(coVector).not.toEqual(vector);
+    // but
+    expect(coVector.vector).toEqual(vector);
```
nrainhart, line 141: Maybe we can get around this, see https://github.com/garden-co/jazz/pull/2893

packages/jazz-tools/src/tools/tests/coVector.test.ts
```diff
@@ -0,0 +1,928 @@
...
+
+  test("given a vector of mismatched dimensions count • throws an error", () => {
+    expect(() => EmbeddingSchema.create([1, 2, 3, 4, 5])).toThrow(
+      "Vector dimension mismatch! Expected 3 dimensions, got 5",
+    );
+  });
+
+  test("given a Array<number> with valid dimensions count • creates a CoVector", () => {
+    const embedding = EmbeddingSchema.create([1, 2, 3]);
+    expect(embedding).toBeInstanceOf(CoVector);
```
nrainhart, line 90: Should we assert the CoVector contains the same elements?

nrainhart (approved):
This looks great! Left a few comments, but nothing major.

As a follow-up, it'd be great to support automatically creating CoVectors from plain JSON objects when the CoVectors are nested inside other CoValues (see https://github.com/garden-co/jazz/pull/2683 and https://github.com/garden-co/jazz/pull/2767)

packages/jazz-tools/src/tools/coValues/coVector.ts
```diff
@@ -0,0 +1,696 @@
...
+      | {
+          owner: Account | Group;
+        }
+      | {
+          fromRaw: RawBinaryCoStream;
+        },
+  ) {
+    super();
+
+    const proxy = new Proxy(this, CoVectorProxyHandler as ProxyHandler<this>);
```
gdorsi, line 66: **question:** how much we do use the proxy?

Index access would be a little more expensive this way, compared to a vanilla `Float32Array`

gdorsi (commented):
Good job!

I like the approach taken on creating a new schema type.

Not a big fan of primitive values extensions, as it makes certain usages unpredictable, but aligns with the other CoValue types so it makes sense.

We are going to drop proxies in favor of vanilla values in the near future, so starting from this would make the users migration easier.

Before merging, I'd like to have at least one example app, to be able to get a "feeling-check" on the API ergonomy in a realistic use case.

packages/jazz-tools/src/tools/coValues/coVector.ts
kysely: So the `(static) requiredDimensionCount` is only set in the schema hydration.

If somehow an instance was constructed from pure `CoVector` (through `CoVector.create`), it would remain `undefined`.

Because my lack of knowledge of the entire internals, I don't know whether that can happen so I simply wanted to _bypass_ the dimension check altogether when undefined.

But alternatively I could throw an error in the constructor (when undefined) that'd ensure the instance always has that value set (and effectively forbids the pure CoVector.create call).

packages/jazz-tools/src/tools/coValues/coVector.ts
kysely: I assume the instance methods will be the most used (as they're easily available).

The static ones are present because I wanted the calculation functions to:
1. be pure with explicit inputs
2. be available as vector utility functions beyond CoVector instances (e.g. _give me magnitude of this random vector which doesn't need to be synced so it's not a CoVector instance_)

Does that go against the design principles here? I admit the (2) might be over-assuming use cases though

packages/jazz-tools/src/tools/tests/coVector.test.ts
kysely: I tried doing that before (originally copied a lot of CoList code).

Spent quite some time on this. These proxy traps help to make it array-like but the proxy strips the `[[TypedArray]]` internal slot so it never passes as _Float32Array_ anywhere.

Screenshot 2025-09-19 at 18 29 10

packages/jazz-tools/src/tools/coValues/coVector.ts
nrainhart: > But alternatively I could throw an error in the constructor (when undefined) that'd ensure the instance always has that value set (and effectively forbids the pure CoVector.create call).

Yeah, I'd prefer this alternative. Class-based CoValues have been deprecated for a while, so I'd expect everyone to be using Zod-like schemas

packages/jazz-tools/src/tools/coValues/coVector.ts
nrainhart: I have a slight preference towards keeping only the instance methods (to be consistent with other CoValue types and reduce the API surface) and only add the static ones later if we find the need, but it's not a strongly-held opinion. @gdorsi what do you think?

kysely:
> Not a big fan of primitive values extensions, as it makes certain usages unpredictable, but aligns with the other CoValue types so it makes sense.

Me neither but the alignment with other CoValues was my deciding factor :)

> We are going to drop proxies in favor of vanilla values in the near future, so starting from this would make the users migration easier.

Hmm, so when you say _“drop proxies in favor of vanilla values”_ do you mean:
1) use the proper `super()` extension w/o proxies
2) make the actual values available on a property of CoValue, e.g. `.value` (as in CoFeed entry)
3) or dropping the CoValue instances altogether, keeping only raw data and syncing them via explicit fn calls?
4) ... or anything else? @gdorsi 

I'm trying to think about possible changes based on that comment as it seems the current design is to become obsolete.

kysely:
> Before merging, I'd like to have at least one example app, to be able to get a "feeling-check" on the API ergonomy in a realistic use case.

So the base branch for this is another feature branch where I intend to accumulate all parts of “vectors in Jazz” project before merging it all to main.

packages/jazz-tools/src/tools/tests/coVector.test.ts
nrainhart: Ahh got it!

packages/jazz-tools/src/tools/coValues/coVector.ts
kysely: Good point, I think it mostly applies to vector calculations as these are the most performance sensitive.

When a vector calculation method was called on an instance, we were always taking the internal F32Array of _this_ instance, but for the other vector (in dot product or cosine similarity) it used whatever was passed (CoVector | F32Array).

I just pushed a change that makes the vector calculations _always_ run on pure Float32Arrays (which has price of a  single `instanceof` before the calc starts).

So for the performance sensitive calculations, proxy shouldn't be touched now

packages/jazz-tools/src/tools/coValues/coVector.ts
kysely: FYI: As per Guido's original question about performance, I rewrote the base calculation functions to run directly on Float32Arrays, so my former point (1) is also eliminated and it really boils down to whether a) we want to provide the vector calculation functions beyond instances; b) whether those should be static methods on CoVector/schema.

gdorsi (approved):
Great job!

gdorsi:
> Hmm, so when you say “drop proxies in favor of vanilla values” do you mean

Use Javascript values and enrich them with `.$jazz` via `Object.defineProperty`
In this case, we are probably going to create a `Float32Array` as basis and do something like:

```ts
const instance = new Float32Array(data)
Object.defineProperties(this, {
  [TypeSym]: { value: "BinaryCoStream", enumerable: false },
  $jazz: {
    value: new CoVectorJazzApi(proxy, raw),
    enumerable: false,
  },
});
```

kysely:
Superseded by #2948

PR #2938, opened 2025-09-19
feat: add connection status hook

gdorsi:
# Description

Add connection status API for React and Svelte

- **React**: Added `useSyncConnectionStatus()` hook that returns the current connection status to the Jazz sync server
- **Svelte**: Added `SyncConnectionStatus` class that provides reactive connection status monitoring

Since it's tied with the connection status inside Jazz, offline state is detected after 5s of lack of response from the sync server.

Play with the cursor example https://jazz-multi-cursors-git-feat-use-connection-status-gcmp.vercel.app/

## Tests

## Checklist


packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -1198,11 +1198,11 @@ export class CoValueCore {
         node.syncManager.handleNewContent(data, "storage");
       },
       (found) => {
+        done?.(found);
+
```
gdorsi, line 1202: Found a bug while playing on the cursor example.

Moving this and the related tests in a separate PR

gdorsi: https://github.com/garden-co/jazz/pull/2953

tests/e2e/tests/ConnectionStatus.test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    await page.goto("/connection-status");
+
+    // Initially should be connected
+    await expect(page.getByTestId("connected")).toBeVisible();
+    await expect(page.getByTestId("connected")).toHaveText("true");
+
+    // Go offline and wait for the ping timeout
+    await context.setOffline(true);
+    await page.waitForTimeout(5000);
```
nrainhart: I wonder if setting the timeout to exactly 5s could make this test flaky. Do you trust the connection status change to always happen before this?

nrainhart (approved):
Thanks for adding this! It's going to be super useful

tests/e2e/tests/ConnectionStatus.test.ts
gdorsi: Yep, it was flaky.

Removed in favor of unit tests.

PR #2946, opened 2025-09-22
feat: support filtering, sorting & pagination when loading CoValues

nrainhart:
# Description

We want to extend the resolve query API to allow filtering, sorting and paginating CoList elements:
```typescript
const Ranking = co.list(co.map({
  name: z.string(),
  score: z.number(),
  date: z.date(),
}));

Ranking.load(ranking.id, {
  resolve: {
    $where: {
      'name': 'Darts',
    },
    $orderBy: { 'score': 'desc' },
    $limit: 20,
    $offset: 40,
});
```

This will provide a simpler way to perform these operations and pave the way for implementing indexes. For now, we're loading all the CoList elements and performing the filter/sort/pagination operations in memory, but longer term the goal is to use indexes to avoid fetching the CoList elements.

There are some pending things to do before merging this PR:
- [ ] add support for filtering, sorting & pagination in Svelte & Vue 
- [ ] migrate one or more examples to use the new APIs
- [ ] write docs

But this PR can already be reviewed.

## Key design decisions

- Extended resolve query API with new options:
	-	`$where` for filtering CoLists elements
		- Comparison operators: `$eq`, `$ne`, `$gt`, `$gte`, `$lt`, `$lte`
		- Logical operators: `$and`, `$or`, `$not`
	- `$orderBy` for sorting
	- `$limit` & `offset` for pagination
- CoLists now have an optional query view. The query view gives us a way to decouple the elements from a jazz-tools CoList from the elements in the underlying RawCoList. This means users will see only the elements returned by their resolve query, but will still be able to perform operations on the CoList that will be mapped to the underlying RawCoList.
	- The logic for creating and invalidating the query view is pretty naive, but has worked well enough on my tests. We can go with this simple implementation for now and optimize it later if needed (which may not be necessary if we end up using indexes most of the time).
- In order for pagination to work, I modified `useCoState` to re-create the subscription scope if the resolve query is updated

## Tests

## Checklist


packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -128,18 +128,10 @@ export class CoValueCore {
   private _cachedDependentOn?: Set<RawCoID>;
   private counter: UpDownCounter;
 
-  private constructor(
-    init: { header: CoValueHeader } | { id: RawCoID },
-    node: LocalNode,
```
nrainhart, line 133: We're never providing a header, so I removed that option

gdorsi: Ah, nice.

packages/jazz-tools/src/react-core/hooks.ts
```diff
@@ -535,7 +562,10 @@ function useAccountSubscription<
       subscription.contextManager !== contextManager ||
       subscription.Schema !== Schema ||
       subscription.branchName !== options?.unstable_branch?.name ||
-      subscription.branchOwnerId !== options?.unstable_branch?.owner?.$jazz.id
+      subscription.branchOwnerId !==
+        options?.unstable_branch?.owner?.$jazz.id ||
+      (subscription.subscription &&
+        subscription.subscription.resolve !== resolveQuery)
```
gdorsi, line 568: **question:** why are we relying on `useStructuralEquals` vs calling structuralEquals here?

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
```diff
@@ -350,7 +371,10 @@ export class SubscriptionScope<D extends CoValue> {
     this.triggerUpdate();
   }
 
-  subscribeToKey(key: string) {
+  subscribeToKey(key: string): void {
+    if (this.isSubscribedToId(key)) {
+      return;
+    }
```
gdorsi: **question:** chould you explain this change?

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
```diff
@@ -73,7 +89,11 @@ export class SubscriptionScope<D extends CoValue> {
           return;
         }
 
-        // Need all these checks because the migration can trigger new syncronous updates
+        if (queryFields.length > 0 && value !== "unavailable") {
+          this.requestCoListChildrenLoad();
+        }
```
gdorsi, line 94: Is it ok for ths to run also on non-coList values?

packages/jazz-tools/src/react-core/hooks.ts
nrainhart: This useLayoutEffect is re-subscribing to the context manager each time it runs, so I wanted to detect the change in the resolve query earlier. I see there's still an integration test failing due to an infinite render loop, so I need to look into this some more

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
nrainhart: Heh, `$where` and `$orderBy` can only be used with CoLists that contain CoMaps right now. I'll work to extend it to CoLists with plain JSON objects as well

gdorsi (commented):
What happens when a value that doesn't match the filter criteria is updated?

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
nrainhart: It's a botched refactor: I tried to move the check we were previously doing [here](https://github.com/garden-co/jazz/pull/2946/files#diff-bd89994888f681ae7feb34d194c7033e595d271ad465c3e9948647ea0279dc4bL45) but got the params mixed up. I'll revert this, thanks for catching it!

nrainhart:
> What happens when a value that doesn't match the filter criteria is updated?

Right now the subscription update will still be triggered. I'll modify this to prevent the update from happening if the query view didn't change.

PR #2947, opened 2025-09-22
fix: keep the subscription off when creating new values

gdorsi:
Optimized the inactive subscriptions (anything from .load and .create or unsubscribed subscriptions) to not listen to updates to their childs.

Should fix also a memory leak when autoloading on inactive subscriptions.

Filling 1k tasks list

Before:
Screenshot 2025-09-22 at 19 22 09

After:
Screenshot 2025-09-22 at 19 17 26

### DIFF
--- a/.changeset/neat-pigs-bake.md
+++ b/.changeset/neat-pigs-bake.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Optimized the inactive subscriptions, improving performance of updates on created and loaded values by 2.5x
--- a/packages/jazz-tools/src/tools/implementation/refs.ts
+++ b/packages/jazz-tools/src/tools/implementation/refs.ts
@@ -1,9 +1,10 @@
 import { type Account } from "../coValues/account.js";
-import type {
+import {
   AnonymousJazzAgent,
   CoValue,
   ID,
   RefEncoded,
+  SubscriptionScope,
 } from "../internal.js";
 import {
   accessChildById,
@@ -26,9 +27,28 @@ export class Ref<out V extends CoValue> {
   async load(): Promise<V | null> {
     const subscriptionScope = getSubscriptionScope(this.parent);
 
-    subscriptionScope.subscribeToId(this.id, this.schema);
+    let node: SubscriptionScope<CoValue> | undefined | null;
 
-    const node = subscriptionScope.childNodes.get(this.id);
+    /**
+     * If the parent subscription scope is closed, we can't use it
+     * to subscribe to the child id, so we create a detached subscription scope
+     * that is going to be destroyed immediately after the load
+     */
+    if (subscriptionScope.closed) {
+      node = new SubscriptionScope<CoValue>(
+        subscriptionScope.node,
+        true,
+        this.id,
+        this.schema,
+        subscriptionScope.skipRetry,
+        subscriptionScope.bestEffortResolution,
+        subscriptionScope.unstable_branch,
+      );
+    } else {
+      subscriptionScope.subscribeToId(this.id, this.schema);
+
+      node = subscriptionScope.childNodes.get(this.id);
+    }
 
     if (!node) {
       return null;
@@ -50,6 +70,10 @@ export class Ref<out V extends CoValue> {
           } else if (value?.type === "unauthorized") {
             unsubscribe();
             resolve(null);
+          }
+
+          if (subscriptionScope.closed) {
+            node.destroy();
           }
         });
       });
--- a/packages/jazz-tools/src/tools/subscribe/CoValueCoreSubscription.ts
+++ b/packages/jazz-tools/src/tools/subscribe/CoValueCoreSubscription.ts
@@ -41,6 +41,16 @@ export class CoValueCoreSubscription {
     this.initializeSubscription();
   }
 
+  rehydrate() {
+    if (!this.unsubscribed) {
+      return;
+    }
+
+    this.unsubscribed = false;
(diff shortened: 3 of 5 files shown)


.changeset/neat-pigs-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Optimized the inactive subscriptions
```
nrainhart: Mind adding a bit more detail?

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
```diff
@@ -69,7 +70,6 @@ export class SubscriptionScope<D extends CoValue> {
 
         if (skipRetry && value === "unavailable") {
           this.handleUpdate(value);
-          this.destroy();
```
nrainhart, line 72: We don't want to destroy the subscription scope in this case?

packages/jazz-tools/src/tools/implementation/refs.ts
```diff
@@ -26,9 +27,28 @@ export class Ref<out V extends CoValue> {
...
-    subscriptionScope.subscribeToId(this.id, this.schema);
+    let node: SubscriptionScope<CoValue> | undefined | null;
 
-    const node = subscriptionScope.childNodes.get(this.id);
+    /**
+     * If the parent subscription scope is closed, we can't use it
+     * to subscribe to the child id, so we create a detached subscription scope
+     * that is going to be destroyed immediately after the load
+     */
+    if (subscriptionScope.closed) {
```
nrainhart, line 37: It'd be nice to add a test case for this scenario

packages/jazz-tools/src/tools/subscribe/CoValueCoreSubscription.ts
```diff
@@ -41,6 +41,16 @@ export class CoValueCoreSubscription {
...
   }
 
+  rehydrate() {
+    if (!this.unsubscribed) {
+      return;
+    }
+
+    this.unsubscribed = false;
+    this.initializeSubscription();
+    this.unsubscribe();
```
nrainhart, line 55: Will this work? on `loadCoValue` we're bailing out early if `this.unsubscribed` is true

nrainhart (commented):
The improvement is amazing! But tbh I had to think real hard to understand these changes, which makes me fear this will be brittle/hard to maintain. This idea that subscriptions can continue to emit values after they are closed is a bit hard to grasp from just looking at the code, as the logic is spread across a lot of different places.

Do you think it'd be possible to create a different type of `SubscriptionScope` (say, a `ClosedSubscriptionScope`) and centralize all the logic there?

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
gdorsi: Yeah, that destroy shouldn't have never been there.

packages/jazz-tools/src/tools/implementation/refs.ts
gdorsi: It's already covered, I've found the code path thanks to a test.

Otherwise I would have never remembered an esoteric feature as ref.load 😆

Also, how about deprecating both load and value?

ensureLoaded and property access are way more valid alternatives.

packages/jazz-tools/src/tools/subscribe/CoValueCoreSubscription.ts
gdorsi: Yes, the rehydrate is meant to synchronously pull values from memory, so if the value isn't already there before unsubscribe we don't care about it.

Good question, going to add a comment.

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
```diff
@@ -398,8 +398,39 @@ export class SubscriptionScope<D extends CoValue> {
...
+    }
+
+    this.subscription.rehydrate();
+    this.setListener(listener);
+    this.destroy();
+  }
+
   subscribeToId(id: string, descriptor: RefEncoded<any>) {
     if (this.isSubscribedToId(id)) {
+      if (this.closed !== true) {
```
antoniomuso:
```suggestion
      if (!this.closed) {
```

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
```diff
@@ -398,8 +398,51 @@ export class SubscriptionScope<D extends CoValue> {
...
   }
 
+  /**
+   * Checks if the currently unloaded value has got some updates
+   *
+   * Used to make the autoload work on closed subscription scopes
+   */
+  pullValue(listener: (value: SubscriptionValue<D, any>) => void) {
+    if (!this.closed) {
+      throw new Error("Cannot rehydrate a non-closed subscription scope");
```
antoniomuso:
```suggestion
      throw new Error("Cannot pull a non-closed subscription scope");
```

antoniomuso (approved):

packages/jazz-tools/src/tools/implementation/refs.ts
nrainhart: > Also, how about deprecating both load and value? ensureLoaded and property access are way more valid alternatives.

I wonder if ref.load will be more useful once we get rid of autoloading. But yeah, so far I haven't seen the need to use it

PR #2948, opened 2025-09-22
feat(jazz-tools): CoVector properly extends Float32Array w/o Proxy

kysely:
# Description
Based on the comments [[1](https://github.com/garden-co/jazz/pull/2936#discussion_r2362764967)] [[2](https://github.com/garden-co/jazz/pull/2936#pullrequestreview-3244763956)] in the former related PR I explored the **`CoVector` that could be made without proxy**.

Because we know the expected vector length at construction (from the schema definer), we can `super()` call `Float32Array` properly with that expected length, then populate its data on load (this is important because TypedArrays cannot be resized like regular Arrays after init).

This allows CoVector to properly extend Float32Array with _all of its native behaviors, without a proxy and overrides_.

With 1 caveat: **losing the proxy it makes the CoVector instance locally mutable**. Although we could override the mutation methods, one could still mutate the contents via index access (ie `vec[6] = 4.5`) so I wouldn't go there.

## Questions

1. Is this better than the base branch implementation? ref #2936
2. Given the future plans as outlined in [[3](https://github.com/garden-co/jazz/pull/2936#issuecomment-3313678937)], should we remove the vector calculations from the CoVector altogether; or move inside `vec.$jazz`?

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/coVector.ts
+++ b/packages/jazz-tools/src/tools/coValues/coVector.ts
@@ -41,6 +41,7 @@ export class CoVector extends Float32Array implements CoValue {
 
   protected static requiredDimensionsCount: number | undefined = undefined;
   private declare _loadedVector: Float32Array | null;
+  private declare _requiredDimensionsCount: number;
 
   constructor(
     options:
@@ -52,6 +53,15 @@ export class CoVector extends Float32Array implements CoValue {
         },
   ) {
     super();
+
+    const dimensionsCount = (this.constructor as typeof CoVector)
+      .requiredDimensionsCount;
+
+    if (dimensionsCount === undefined) {
+      throw new Error(
+        "Instantiating CoVector without a dimensions count is not allowed. Use co.vector(...).create() instead.",
+      );
+    }
 
     const proxy = new Proxy(this, CoVectorProxyHandler as ProxyHandler<this>);
 
@@ -71,6 +81,11 @@ export class CoVector extends Float32Array implements CoValue {
         enumerable: false,
       },
       _loadedVector: { value: null, enumerable: false, writable: true },
+      _requiredDimensionsCount: {
+        value: dimensionsCount,
+        enumerable: false,
+        writable: false,
+      },
     });
 
     return proxy;
@@ -160,10 +175,6 @@ export class CoVector extends Float32Array implements CoValue {
     return new Float32Array(u8.buffer, u8.byteOffset, total / 4);
   }
 
-  private get requiredDimensionsCount(): number | undefined {
-    return (this.constructor as typeof CoVector).requiredDimensionsCount;
-  }
-
   get vector(): Float32Array {
     if (this._loadedVector !== null) {
       return this._loadedVector;
@@ -178,12 +189,9 @@ export class CoVector extends Float32Array implements CoValue {
 
     const vector = CoVector.fromByteArray(chunks.chunks);
 
-    if (
-      this.requiredDimensionsCount !== undefined &&
-      vector.length !== this.requiredDimensionsCount
-    ) {
+    if (vector.length !== this._requiredDimensionsCount) {
       throw new Error(
-        `Vector dimension mismatch! CoVector '${this.$jazz.raw.id}' loaded with ${vector.length} dimensions, but the schema requires ${this.requiredDimensionsCount} dimensions`,
+        `Vector dimension mismatch! CoVector '${this.$jazz.raw.id}' loaded with ${vector.length} dimensions, but the schema requires ${this._requiredDimensionsCount} dimensions`,
       );
     }
 
--- a/packages/jazz-tools/src/tools/coValues/coVector.ts
+++ b/packages/jazz-tools/src/tools/coValues/coVector.ts
@@ -40,7 +40,7 @@ export class CoVector extends Float32Array implements CoValue {
   }
 
   protected static requiredDimensionsCount: number | undefined = undefined;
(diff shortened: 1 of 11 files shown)


nrainhart:
> With 1 caveat: losing the proxy it makes the CoVector instance locally mutable. Although we could override the mutation methods, one could still mutate the contents via index access (ie vec[6] = 4.5) so I wouldn't go there.

Preventing direct mutations is important, so we should still override the mutation methods if we go with the proxyless route. Regarding mutating the contents via index access, we could at least prevent it at the type level by ensuring all creation/load methods that return a CoVector actually return a `Readonly<CoVector>`.

> 1. Is this better than the base branch implementation?

I like this version better, assuming we can prevent direct mutations.

> 2. Given the future plans as outlined in [https://github.com/garden-co/jazz/pull/2936#issuecomment-3313678937], should we remove the vector calculations from the CoVector altogether; or move inside vec.$jazz?

Why not keep them where they are? The rationale for moving Array mutation methods like push/splice/etc. into `$jazz` was to reinforce this idea that CoValues are "readonly JSON objects". All CoVector calculation methods generate new instances (in a similar way to map/filter/etc.) so I see no issues with keeping them inside CoVector.

kysely:
> Why not keep them where they are? The rationale for moving Array mutation methods like push/splice/etc. into $jazz was to reinforce this idea that CoValues are "readonly JSON objects". All CoVector calculation methods generate new instances (in a similar way to map/filter/etc.) so I see no issues with keeping them inside CoVector.

This was mostly hypothetical question based on [Guido's mention of future plans](https://github.com/garden-co/jazz/pull/2936#issuecomment-3313678937) to use plain JS objects, and mostly an API consideration. Assuming we'd put the vector calculations elsewhere now, it would make future migrations to plain values easier (and CoVector even simpler). 

But because:
1) a vector doesn't have 1:1 JS opposite (unlike `CoMap/Map` and `CoList/Array`), and
2) vector operations (math) are not even Jazz specific (unlike `applyDiff`, collaborative `push/splice/...`),

moving them into $jazz also doesn't make than much sense to me.

In case we'd go with this proxy-less version, I'd personally go `ImageDefinition`/`jazz-media` route where we'd keep `CoVector`/`Float32Array` pure and put the calculation functions that work on those in a separate module (e.g. `jazz/vector`). That module will exist anyways (at least internally) as it will hold the vector search logic.

**Would that work?**

---

> > With 1 caveat: losing the proxy it makes the CoVector instance locally mutable. Although we could override the mutation methods, one could still mutate the contents via index access (ie vec[6] = 4.5) so I wouldn't go there.

> Preventing direct mutations is important, so we should still override the mutation methods if we go with the proxyless route. Regarding mutating the contents via index access, we could at least prevent it at the type level by ensuring all creation/load methods that return a CoVector actually return a Readonly<CoVector>.

@gdorsi **Regarding the future plans on move to plain JS objects, did you think about mutability? Are we okay with going proxy-less where mutation via index access is only prohibited in types?**

gdorsi:
> Regarding the future plans on move to plain JS objects, did you think about mutability?

All the values are going to be returned as `readonly` and we will support reactive mutations only on `.$jazz` APIs (we already do).

Agree with @nrainhart about the readonly type, if it looks challenging to you (Typescript can be really really hard to tame!) you can get help from either me or @nrainhart.

> should we remove the vector calculations from the CoVector altogether; or move inside vec.$jazz?

Having them into `$jazz` will make moving to vanilla values easier, because we won't need to add more custom `defineProperty` which are slow.

Having them as static methods also could work, but it's both ugly and a pain to write so I'm more for `.$jazz` when possible.

gdorsi:
Nice improvement BTW!

kysely:
Thanks for both of your comments!

I will make it readonly (with mutation methods overriden for now), move out the custom calculation methods.

While I do that, can I attempt to change your mind once more about moving the vector calc methods completely outside, to own module?

>But because:
>- a vector doesn't have 1:1 JS opposite (unlike CoMap/Map and CoList/Array), and
>- vector operations (math) are not even Jazz specific (unlike applyDiff, collaborative push/splice/...),

> moving them into $jazz also doesn't make than much sense to me.

> In case we'd go with this proxy-less version, I'd personally go ImageDefinition/jazz-media route where we'd keep CoVector/Float32Array pure and put the calculation functions that work on those in a separate module (e.g. jazz/vector). That module will exist anyways (at least internally) as it will hold the vector search logic.

Isn't vector more similar to Image here, where the datatype only stores the underlying data but function operating on it are too separate from Jazz (db/sync)?

gdorsi:
> Isn't vector more similar to Image here, where the datatype only stores the underlying data but function operating on it are too separate from Jazz (db/sync)?

Image lives in a separate entry because it weights on the bundle size, so having it separate ensures that it won't get in the bundles of people that aren't using Jazz images.

Could make sense to do the same for vector if it requires some big dependencies, otherwise we can keep it inside the main entry for simplicity.

kysely:
Can you please review final version before merging? @gdorsi 

To sum everything up:
- no proxy
- mutation methods overridden for runtime immutability
- instances have `Readonly<CoVector>` typing
- vector calculations available on the instance via `.$jazz` api

I'd love to get a check especially on the [last commit that introduced Readonly typing](https://github.com/garden-co/jazz/pull/2948/commits/4af1cebc68f00fa4d524c6c12a2868c33bf575e9), I'm unsure whether it has correctly used “nullable” and Resolved types 🙏

gdorsi (approved):
LGTM

Well done

PR #2950, opened 2025-09-22
feat(examples): Add example app for vector search

kysely:
# Description
Adding the [example app for vector search](https://github.com/garden-co/jazz/tree/feat/vector-example-app/examples/vector-search).

It is based on the former library [example](https://jazz-vector-demo.playdeux.com), and works the same.

As the testing dataset is being loaded from a static JSON file; and we're deploying examples on Vercel, I added a caching directive for that file to avoid bandwidth consumption for data transfers.

After `pnpm install` and `pnpm build:packages`

```sh
cd examples/vector-search && npm run dev
```

## Tests

- [x] Tests have not been updated, because: unnecessary?

## Checklist

---

> [!NOTE]
> Adds a new example app that performs local vector similarity search over a journal using Jazz for sync and Transformers.js for embeddings, with dataset seeding and caching.
> 
> - **Example app (`examples/vector-search`)**:
>   - React + Vite + Tailwind UI; components for auth, model status, embeddings, and search results.
>   - Jazz integration: `JazzReactProvider`, passkey auth, account/schema with 384‑dim `Embedding`, list of `JournalEntry`.
>   - Local embeddings via `@huggingface/transformers`; model load/progress tracking; cosine similarity ranking of entries.
>   - Seed from static dataset (`public/datasets/journal/*.json`); create/delete entries utilities; environment config (`VITE_JAZZ_API_KEY`, `VITE_APP_NAME`).
> - **Build/Config**:
>   - Vite config/tsconfigs; Tailwind PostCSS; `.env.example`.
>   - Vercel headers to cache dataset JSON.
>   - Dependencies added (e.g., `@huggingface/transformers`) and lockfile updates.
> - **Docs**:
>   - README with setup, monorepo/template usage, and sync server notes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f9cab755371ca3f1a1f86b3a9ec0d69e8bf11081. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


examples/vector-search/src/schema.ts
```diff
@@ -0,0 +1,50 @@
...
+      account.$jazz.set(
+        "profile",
+        JazzProfile.create(
+          {
+            name: "Anonymous",
+          },
+          profileGroup,
+        ),
+      );
+    }
```
joeinnes: Nit: You don't really need this, profiles should be created automatically.

examples/vector-search/src/components/AuthButton.tsx
joeinnes, line 1: There's a bit of a bug here where if you log in while you're loading the dataset, everything breaks, and you have to clear your local storage and create a new account. Can we just disable these buttons when the dataset is being created?

joeinnes (approved):
Nice :) a couple of small comments but a super cool demo app!

examples/vector-search/src/App.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+  // 1) Prepare local embeddings model (for query text embedding)
+  const { createEmbedding, modelStatus } = useLocalEmbeddings({
+    modelName: DEFAULT_MODEL,
+  });
+  const { queryEmbedding, isCreatingEmbedding, createQueryEmbedding } =
+    useCreateEmbedding({ createEmbedding });
+
+  // 2) Load a list of elements containing embeddings
+  const journalEntries = useCoStateWithSelector(
```
nrainhart, line 31: Why not just use useAccountWithSelector instead of useAccount + useCoStateWithSelector?

examples/vector-search/src/helpers/use-delete-entries.ts
```diff
@@ -0,0 +1,22 @@
...
+  const deleteEntries = useCallback(async () => {
+    const confirmed = confirm("Are you sure you want to delete all entries?");
+
+    if (!confirmed) {
+      return;
+    }
+
+    try {
+      if (owner?.root?.journalEntries) {
+        owner.root.$jazz.set("journalEntries", JournalEntryList.create([]));
```
nrainhart: No need to create the JournalEntryList manually, set already does it under the hood
```suggestion
        owner.root.$jazz.set("journalEntries", []);
```

examples/vector-search/src/helpers/use-delete-entries.ts
```diff
@@ -0,0 +1,22 @@
+import { useCallback } from "react";
+import { JazzAccount, JournalEntryList } from "../schema";
+
+export function useDeleteEntries({ owner }: { owner: JazzAccount }) {
```
nrainhart: Nit: I'd pass the journalEntries colist to all hooks that currently expect  an owner, since that's all you really need

examples/vector-search/src/helpers/use-journal-seed.ts
```diff
@@ -0,0 +1,70 @@
...
+
+        const journalEntry = JournalEntry.create({
+          text: entry.c,
+          feelings: entry.f,
+          topics: entry.t,
+          embedding: Embedding.create(embedding),
+        });
+
+        if (owner?.root?.journalEntries) {
+          owner.root.journalEntries.$jazz.push(journalEntry);
```
nrainhart: Seeding the journal was kinda slow (it took around 30s) in my Macbook M4. Not sure if the slowness is coming from creating the embeddings or the CoValues (or both), but if it's CoValues here are a few optimizations you can apply:
- creating all embeddings and then pushing them all into journalEntries in a single push operation should be considerably faster than doing it one by one. If the process is still kind of slow and you want to create results incrementally, maybe you can do batches of 50/100 elements?
- consider also creating each JournalEntry using the same `journalEntries.$jazz.owner` to avoid creating one group per entry.

examples/vector-search/src/App.tsx
```diff
@@ -0,0 +1,233 @@
...
+        // If no query embedding, return all entries
+        if (!queryEmbedding) return journalEntries.map((value) => ({ value }));
+
+        return journalEntries
+          .map((value) => ({
+            value,
+            similarity: value.embedding.$jazz.cosineSimilarity(queryEmbedding),
+          }))
+          .sort((a, b) => b.similarity - a.similarity)
+          .slice(0, 5);
```
nrainhart, line 48: You should never create objects/arrays inside a selector, as it will cause the component to re-render every time. I'd simply return the journal entries colist, and then do the rest of the processing in the component's body.

This is more of a conceptual than an optimization comment in this case, since we're recomputing the top 5 entries whenever anything changes, but I want to avoid having anti-patterns in our example apps.

examples/vector-search/src/App.tsx
```diff
@@ -0,0 +1,233 @@
...
+  const { deleteEntries } = useDeleteEntries({ owner: me as JazzAccount });
+
+  return (
+    <>
+      <Header />
+
+      <main className="max-w-2xl mx-auto px-3 pt-8 pb-10 flex flex-col gap-2">
+        <h2 className="text-4xl font-bold text-zinc-800 mb-8">Journal</h2>
+
+        {/* Search UI */}
```
nrainhart: Nit: mind extracting the search UI and the status line into separate React components?

nrainhart (commented):
Love the example! Left a few comments, the most important one being around seed performance.

It'd also be good to add at least one E2E test to confirm the example doesn't break as we make changes to Jazz. Maybe we can have just a few journal entries for this case to avoid needing to seed the whole journal?

examples/vector-search/src/components/AuthButton.tsx
kysely: Thanks, I disabled the auth buttons during the seeding

examples/vector-search/src/helpers/use-journal-seed.ts
kysely: Good point, unfortunately the vast majority of the time comes from embeddings generation. However I managed to save some time on CoValues:
- I'm reusing the owner on all colist entries; and referenced covectors inside them
- I'm now pushing the entries to the colist in batches of 10

On M1 Safari, I was able to go from ~60 seconds do ~42 seconds.

The batching helped a lot as well, I'm assuming for 2 reasons:
- every $jazz.push creates a new transaction, so it's much fewer txs?
- React now also doesn't need to update on each entry

Tried spawning all embedding creation in parallel, but that doesn't help inside a single thread. Larger batches didn't come with any noticeable performance boost, but felt weird UI-wise. So batches of 10 entries feel real-time, natural, and add a little performance boost.

examples/vector-search/src/App.tsx
kysely: Hmm, I honestly can't see how this would be an “anti-pattern”. [Selectors](https://jazz.tools/docs/react/using-covalues/subscription-and-loading#selectors) seem to be a way to show parts or a different view of your state, and that obviously re-renders when the data changes?

This was an alternative to `useCoState` with `useMemo` suggested by Guido, and we chose to use this as the primary way of running the similarity sort, [even in the docs](https://github.com/garden-co/jazz/pull/2996).

Can you help me understand more how this is anti pattern?

examples/vector-search/src/App.tsx
kysely: Honestly a personal taste decision.

The CoList in this specific example _is_ placed in the account root so it'd be more efficient, but as an example I was trying to show the common use of running the similarity sort over any state. So it feels more general (and personally a bit easier to understand)

kysely:
> It'd also be good to add at least one E2E test to confirm the example doesn't break as we make changes to Jazz. Maybe we can have just a few journal entries for this case to avoid needing to seed the whole journal?

@nrainhart Yes, I added a basic spec that manually adds handful of entries, then runs one search and asserts the results in the expected order (and their similarity score)

examples/vector-search/tests/basic.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    await expect(entryContent).toHaveText(expectedText);
+    await expect(entrySimilarity).toHaveText(expectedSimilarity);
+  }
+});
+
+async function createEntry(page: Page, text: string) {
+  const createEntryButton = page.getByRole("button", { name: "+ New entry" });
+
+  createEntryButton.click();
```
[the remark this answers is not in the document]
kysely: expected

examples/vector-search/src/App.tsx
nrainhart: > Selectors seem to be a way to show parts or a different view of your state, and that obviously re-renders when the data changes?

That's correct, but selectors are also a way to prevent unnecessary re-renders when data you're not interested in changes.

Let's say you're using `useAccountWithSelector` instead of `useCoStateWithSelector`. If the profile data changed, you wouldn't want this component to re-render. And if you're returning new objects on each selector call, it will.

It's true that re-rendering in this case is not a real issue, because the result depends on the whole journalEntries CoList, but I'd like to find a way to highlight the two "roles" that the selector fn has (show a different view on your state, and prevent unnecessary re-renders).

Since the top 5 results may still be the same even if the list changes, we could provide a `equalityFn` to this `useCoStateWithSelector` that makes a structural comparison between the selector's outputs and only re-renders if the result actually changed. What do you think? I'm also wondering if this is something we'd like to add to the docs

nrainhart (approved):
Thanks for addressing my comments! And specially for adding the E2E test. Left only one comment, but it's not blocking

examples/vector-search/src/App.tsx
kysely: I see your point now, the issue is with `Object.is` vs new custom object instances. I didn't even get that about `useCoStateWithSelector`.

So I added [the `equalityFn` to this example app](https://github.com/garden-co/jazz/pull/2950/commits/c9bbc98a05468806393fbfb1f9ded8367bf915f4) because 1) the search bar lets you re-submit the query multiple times; and 2) it should also act as a more complete example of good practices.

However it seems like adding `equalityFn` didn't help with the case when the search results remain same after the underlying CoList changes. When I had the data loaded in a second window, added a new entry in different peer, it still re-rendered twice due to the state becoming `undefined` (loading) before loading up with the new entry added (where it actually performed the sorting yielding the same results).

So I'm leaning towards only leaving it here and not the docs for the following reasons:
- brevity
- the pattern shown for similarity sort is exactly this simple, where the results depends on the entire CoList; and is presented with `useCoStateWithSelector`

Is that ok with you?

PR #2954, opened 2025-09-23
feat(jazz-tools): allow usage of z.codec with custom encoders

Gabrola:
# Description
I've been wanting to switch to `Temporal` objects for a while now, but didn't want to keep converting between `Date` and `Temporal.PlainDate`, etc. every time we create or load CoValues. I think it's still too early to add first-class support for `Temporal` in Jazz, but that spurred the idea to allow any class instances really to be stored in Jazz and it ended up being a simple change.

I'm not in love with the idea that encoder registration becomes a library side-effect, but the only alternative I can think of is wrapping `z.instanceof()` to include the encoder in the params, but then it becomes unergonomic and we'd lose the ability of re-using existing zod schemas that already include `z.instanceof`

## Tests

## Checklist

---

> [!NOTE]
> Adds support for Zod z.codec with custom encoders in CoValues, updating runtime/type handling, docs, and tests.
> 
> - **jazz-tools**:
>   - **Runtime**: Extend `schemaFieldToCoFieldDef` to support `z.codec()` (via `pipe`) with JSON-compatible encode/decode, handling optional/nullable/nullish and rejecting unsupported input schemas; import `JsonValue`.
>   - **Types**: Update `TypeOfZodSchema` to infer codec output types; include `$ZodCodec` in `SchemaField`.
>   - **Zod re-exports**: Expose `z.codec`.
> - **Docs**:
>   - Add `docs/using-covalues/codecs` with usage guidance; add "Codecs" to docs navigation.
> - **Tests**:
>   - Add codec tests (e.g., `DateRange`, `RegExp`, optional/nullable/nullish, unsupported inner schema).
> - **Examples/Starters**:
>   - Add `zod` devDependency to Svelte example and starter packages.
> - **Changeset**:
>   - Patch release for `jazz-tools`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d47e02b60e5faa271cd4c5491107cb7ba394435d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/jazz-tools/src/tools/implementation/zodSchema/runtimeConverters/schemaFieldToCoFieldDef.ts
+++ b/packages/jazz-tools/src/tools/implementation/zodSchema/runtimeConverters/schemaFieldToCoFieldDef.ts
@@ -173,7 +173,7 @@ export function schemaFieldToCoFieldDef(schema: SchemaField) {
           schemaFieldToCoFieldDef(zodSchemaDef.in as SchemaField);
         } catch (error) {
           if (error instanceof Error) {
-            error.message = `z.codec() is only supported if the input schema is already supported: ${error.message}`;
+            error.message = `z.codec() is only supported if the input schema is already supported. ${error.message}`;
           }
 
           throw error;
--- a/packages/jazz-tools/src/tools/implementation/zodSchema/runtimeConverters/schemaFieldToCoFieldDef.ts
+++ b/packages/jazz-tools/src/tools/implementation/zodSchema/runtimeConverters/schemaFieldToCoFieldDef.ts
@@ -1,3 +1,4 @@
+import type { JsonValue } from "cojson";
 import { CoValueClass, isCoValueClass } from "../../../internal.js";
 import { coField } from "../../schema.js";
 import { CoreCoValueSchema } from "../schemaTypes/CoValueSchema.js";
@@ -32,10 +33,30 @@ export type SchemaField =
   | z.core.$ZodLazy<z.core.$ZodType>
   | z.core.$ZodTemplateLiteral<any>
   | z.core.$ZodLiteral<any>
-  | z.core.$ZodCatch<z.core.$ZodType>
   | z.core.$ZodEnum<any>
+  | z.core.$ZodCodec<z.core.$ZodType, z.core.$ZodType>
   | z.core.$ZodDefault<z.core.$ZodType>
   | z.core.$ZodCatch<z.core.$ZodType>;
+
+function makeCodecCoField(
+  codec: z.core.$ZodCodec<z.core.$ZodType, z.core.$ZodType>,
+) {
+  return coField.optional.encoded({
+    encode: (value: any) => {
+      if (value === undefined) return undefined as unknown as JsonValue;
+      if (value === null) return null;
+      return codec._zod.def.reverseTransform(value, {
+        value,
+        issues: [],
+      }) as JsonValue;
+    },
+    decode: (value) => {
+      if (value === null) return null;
+      if (value === undefined) return undefined;
+      return codec._zod.def.transform(value, { value, issues: [] });
+    },
+  });
+}
 
 export function schemaFieldToCoFieldDef(schema: SchemaField) {
   if (isCoValueClass(schema)) {
@@ -54,7 +75,7 @@ export function schemaFieldToCoFieldDef(schema: SchemaField) {
         zodSchemaDef.type === "optional" ||
         zodSchemaDef.type === "nullable"
       ) {
-        const inner = zodSchemaDef.innerType as ZodPrimitiveSchema;
+        const inner = zodSchemaDef.innerType as SchemaField;
         const coFieldDef: any = schemaFieldToCoFieldDef(inner);
         if (
           zodSchemaDef.type === "nullable" &&
@@ -137,6 +158,30 @@ export function schemaFieldToCoFieldDef(schema: SchemaField) {
             "z.union()/z.discriminatedUnion() of collaborative types is not supported. Use co.discriminatedUnion() instead.",
           );
         }
+      } else if (zodSchemaDef.type === "pipe") {
+        const isCodec =
+          zodSchemaDef.transform !== undefined &&
+          zodSchemaDef.reverseTransform !== undefined;
+
+        if (!isCodec) {
+          throw new Error(
+            "z.pipe() is not supported. Only z.codec() is supported.",
+          );
+        }
+
+        try {
+          schemaFieldToCoFieldDef(zodSchemaDef.in as SchemaField);
+        } catch (error) {
+          if (error instanceof Error) {
+            error.message = `z.codec() is only supported if the input schema is already supported: ${error.message}`;
+          }
+
+          throw error;
+        }
+
+        return makeCodecCoField(
+          schema as z.core.$ZodCodec<z.core.$ZodType, z.core.$ZodType>,
+        );
       } else {
         throw new Error(
           `Unsupported zod type: ${(schema._zod?.def as any)?.type || JSON.stringify(schema)}`,
--- a/packages/jazz-tools/src/tools/implementation/zodSchema/typeConverters/TypeOfZodSchema.ts
+++ b/packages/jazz-tools/src/tools/implementation/zodSchema/typeConverters/TypeOfZodSchema.ts
@@ -71,8 +71,13 @@ export type TypeOfZodSchema<S extends z.core.$ZodType> =
                                       infer Default extends z.core.$ZodType
                                     >
                                   ? TypeOfZodSchema<Default>
-                                  : S extends z.core.$ZodCatch<
-                                        infer Catch extends z.core.$ZodType
+                                  : S extends z.core.$ZodCodec<
+                                        any,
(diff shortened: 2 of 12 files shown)


packages/jazz-tools/src/tools/implementation/registeredEncoders.ts
```diff
@@ -0,0 +1,78 @@
...
+export const registerInstanceEncoder = <T>(
+  cls: new (...args: any) => T,
+  encoder: InstanceEncoder<T>,
+) => {
+  registeredInstanceEncoders.set(
+    cls,
+    coField.optional.encoded({
+      encode: (value: T | null | undefined) => {
+        if (value === undefined) return undefined as unknown as JsonValue;
+        if (value === null) return null;
```
Gabrola: I used the `Date` encoder as a reference, but I didn't understand why `z.nullable()` of dates are not supported? I understand that `.optional()` dates are stored as `null` so then there's no way to distinguish between `z.optional()` and `z.nullable()` in this case.

However, in this case, I just made it return `null` or `undefined` so we can distinguish between optional and nullable and it just worked just fine despite undefined not existing in JSON? I honestly didn't look into how `undefined` is internally stored, but I'm guessing it must be the same as when storing primitive values.

packages/jazz-tools/src/tools/tests/zod.test.ts
```diff
@@ -511,6 +520,131 @@ describe("co.map and Zod schema compatibility", () => {
...
+        range: z.instanceof(DateRange).nullish(),
+      });
+
+      const map = schema.create({});
+      expect(map.range).toBeUndefined();
+      expect(map.$jazz.has("range")).toEqual(false);
+
+      map.$jazz.set("range", undefined);
+      expect(map.range).toBeUndefined();
+      expect(map.$jazz.has("range")).toEqual(true);
```
Gabrola, line 609: Jazz seems to even be able to distinguish between an optional field that is not passed, and an explicitly set `undefined`, so I'm surprised this works too despite having to "force" passing `undefined` in place of a `JsonValue`

gdorsi (commented):
Wow, thanks for the contribution!

For a feature like this one I'd like to give a try to support Zod's [codecs](https://zod.dev/api#codecs).

It would remove the need of registering the encoders globally, and it's already something supported by Zod.

Haven't checked those internally, so I don't know how complex the task would be.
Do you think it would be doable?

Gabrola:
@gdorsi Zod codecs would actually be perfect as then we'd not be limited to encoding just class instances. I'm still not sure how complex of a task it would be, but there are two points I can think of right now.

- Codecs are part of Zod 4.1, which is not shipped under a v3 subpath anymore, so the Zod package would need a major version bump
- I don't think Jazz actually runs any data through the zod parser at all at the moment, ~~using codecs would need to though. So a decision would need to be made if that's something the Jazz team is open to doing, and also decide on what the approach would for dealing, or lack thereof, with parsing or validation errors~~

I'll give it a quick shot in the meantime, I'd be happy to go with either direction

Gabrola:
This actually ended up being simpler than the z.instanceof implementation. I was actually able to get around calling the Zod parser by just calling the inner transformation functions, rather than `encode`/`decode` on the Zod schema itself. If this looks good, I'll go ahead and clean up the commits, changeset, etc.

packages/jazz-tools/src/tools/implementation/zodSchema/zodReExport.ts
```diff
@@ -35,6 +35,7 @@ export {
   // record,
   // intersection,
   int,
+  codec,
```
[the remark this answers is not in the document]
Gabrola: Yes, it's intentional because Jazz does not re-export schemas that it does not support directly like `instanceof` or `map`.

gdorsi (commented):
Thanks, looks very good!

I didn't take into consideration the upgrade to Zod v4, last time we had some troubles with libraries depending on Zod v3.

Going to play with it a bit to see if we can manage to do the upgrade this time.

packages/jazz-tools/src/tools/implementation/zodSchema/typeConverters/TypeOfZodSchema.ts
```diff
@@ -71,8 +71,13 @@ export type TypeOfZodSchema<S extends z.core.$ZodType> =
                                       infer Default extends z.core.$ZodType
                                     >
                                   ? TypeOfZodSchema<Default>
-                                  : S extends z.core.$ZodCatch<
-                                        infer Catch extends z.core.$ZodType
+                                  : S extends z.core.$ZodCodec<
+                                        any,
```
nrainhart, line 75: Nit: I think `unknown` should work here as well

packages/jazz-tools/src/tools/tests/zod.test.ts
```diff
@@ -511,6 +511,118 @@ describe("co.map and Zod schema compatibility", () => {
...
+
+    const dateRangeCodec = z.codec(
+      z.tuple([z.string(), z.string()]),
+      z.z.instanceof(DateRange),
+      {
+        encode: (value) =>
+          [value.start.toISOString(), value.end.toISOString()] as [
+            string,
+            string,
+          ],
```
nrainhart, line 535:
```suggestion
          [value.start.toISOString(), value.end.toISOString()] as const,
```

packages/jazz-tools/src/tools/tests/zod.test.ts
nrainhart: Yeah, this is expected behaviour. A key explicitly set as `undefined` is considered as "set". This is in line to how `Map.has()` works

packages/jazz-tools/src/tools/implementation/zodSchema/runtimeConverters/schemaFieldToCoFieldDef.ts
```diff
@@ -137,6 +158,30 @@ export function schemaFieldToCoFieldDef(schema: SchemaField) {
...
+          throw new Error(
+            "z.pipe() is not supported. Only z.codec() is supported.",
+          );
+        }
+
+        try {
+          schemaFieldToCoFieldDef(zodSchemaDef.in as SchemaField);
+        } catch (error) {
+          if (error instanceof Error) {
+            error.message = `z.codec() is only supported if the input schema is already supported: ${error.message}`;
```
nrainhart:
```suggestion
            error.message = `z.codec() is only supported if the input schema is already supported. ${error.message}`;
```

nrainhart (commented):
This is amazing! Thanks for the contribution @Gabrola!!

I already upgraded to Zod 4.1 in https://github.com/garden-co/jazz/pull/2983, so you should be good to rebase on top of main.

Mind updating the PR's title, description and changeset?

Gabrola:
Really nice @nrainhart, thank you! I'll fix up everything now

packages/jazz-tools/src/tools/tests/zod.test.ts
Gabrola: I tried this first but it complains of `The type readonly [string, string] is readonly and cannot be assigned to the mutable type [string, string]`. I'm pretty well-versed in Typescript and I think I've never seen this error before 😅

Gabrola:
@nrainhart I was just updating the docs now, and remembered that all of the non-supported zod schemas such as `z.instanceof()`, `z.map()`, etc. are not re-exported, but they could potentially be used inside codecs now. Do you think it should be kept this way and document that these schemas are exported from `z.z.(...)` or re-export everything now?

packages/jazz-tools/src/tools/tests/zod.test.ts
nrainhart: 🫠 nvm then

nrainhart:
> Do you think it should be kept this way and document that these schemas are exported from z.z.(...) or re-export everything now?

The problem with re-exporting everything is that we don't really support these schemas inside Jazz schemas, so it could get confusing for folks. That's why I'm in favour of going with `z.z.(...)` for now. Don't love it, but given that codecs are somewhat of an advanced use case I think people doing this won't have any problems with it.

cc @gdorsi in case you have other thoughts

packages/jazz-tools/src/tools/implementation/zodSchema/typeConverters/TypeOfZodSchema.ts
Gabrola: It doesn't in this case because it expects `z.core.$ZodType`. I'd generally just pass `z.core.$ZodType` instead of `any`, but there isn't any real type-safety lost here by using it. Using `any` would have a slight performance advantage because it skips type checking this generic which is useful on recursive types like this which often get slow or error with `Type instantiation is excessively deep...` for large types, so every little optimization counts

Gabrola:
I agree, that was my gut feeling as well. I'll add an alert box mentioning the fact in the docs.

gdorsi:
@Gabrola @nrainhart 

Agree. Also the limited re-export is just temporary until we fully support all the Zod APIs, so that's ok for now.

gdorsi (approved):

PR #2878, opened 2025-09-05
GCO-817 - Add app submission form to showcase

mculotta120:
# Description
Added form to `/showcase` page for jazz application submission. 
When a user submits the form, the content goes to a notion db for tracking and pings a channel in discord. 
[Notion DB](https://www.notion.so/gcmp/Jazz-App-Submissions-2633ad00c39580c3a8aadc75cf7dbec7)

## Manual testing instructions
Try submitting the form in the deployed version from PR, the Notion DB should contain your form data, the discord channel should update with a copy of your form information.

To run locally, You can copy the following variables from Vercel and add to your .env.local file: 
`NOTION_TOKEN`
`NOTION_DATABASE_ID`
`DISCORD_WEBHOOK_URL`

Running locally should also update both the Notion DB and the Slack channel.
## Tests

- [ ] Tests have been added and/or updated
- [x] Tests have not been updated, because: there were no existing unit tests for homepage functionality beyond the route. The route was not changed.
- [ ] I need help with writing tests

## Checklist

- [ ] I've updated the part of the docs that are affected the PR changes
- [ ] I've generated a changeset, if a version bump is required
- [ ] I've updated the jsDoc comments to the public APIs I've modified, or added them when missing

## Screenshots

Screenshot 2025-09-05 at 9 00 38 AM
Screenshot 2025-09-05 at 10 00 10 AM
Screenshot 2025-09-05 at 8 51 04 AM
Screenshot 2025-09-05 at 9 20 13 AM

PR #2956, opened 2025-09-24
fix: downgrade pagefind & fix search

Elfo404:
My suspicion is that the pagefind compiled file is cached (because it's name is not hashed) and the update in https://github.com/garden-co/jazz/pull/2878/files#diff-b11f680cd67169518f8eb6f971e3159cb67a5bc5f1f7d1301c456db213a3f509 make it break because of a version mismatch. this does not happen in preview branches because there the file is not cached.


joeinnes (approved):
Even if you're wrong, this won't make it worse

PR #2960, opened 2025-09-24
fix: batch remote updates and wait for streaming before migrations

### LINKED ISSUE #2117: Bug: Empty coValue update before content loading
## Description
We're experiencing an issue where we emit an update with an empty coValue before emitting the content. This happens because we call `markAvailable` before loading the content.

## Context
While working on the new subscription system, I've found this bug where we emit an update with an empty coValue before emitting the content because we call `markAvailable` before loading the content.

With the old version of the code, the fix was simple, but now it requires more careful consideration.

## Proposed Solution
@aeplay suggests:

> I think the ideal way to resolve this problem is by pulling precise peer state into CoValueCore and letting subscribers skip all updates until local state matches peer known state (either storage, or server or max(storage, server)).
>
> I can try to do that in a follow up PR

gdorsi:
Some users reported that their migrations where sometimes running on empty covalues.

I haven't been able to reproduce the issue, but my guess is that it could happen for two reasons:
1. The value is large, and we don't wait for full download before triggering the migration
2. https://github.com/garden-co/jazz/issues/2117

It's more likely to be 2, but fixing 1 as well.

This PR also brings a small improvement on subscriptions, and prepares the ground for batching multi-value updates in the same subscription.

Fixes #2117

### DIFF
--- a/packages/cojson/src/coValueCore/coValueCore.ts
+++ b/packages/cojson/src/coValueCore/coValueCore.ts
@@ -536,7 +536,7 @@ export class CoValueCore {
     });
   }
 
-  updateCurrentContent() {
+  private updateCurrentContent() {
     if (
       this._cachedContent &&
       "processNewTransactions" in this._cachedContent &&
@@ -550,8 +550,9 @@ export class CoValueCore {
     this._cachedDependentOn = undefined;
   }
 
-  #batching = false;
+  #isNotificationScheduled = false;
   #batchedUpdates = false;
+
   private scheduleNotifyUpdate() {
     if (this.listeners.size === 0) {
       return;
@@ -559,11 +560,11 @@ export class CoValueCore {
 
     this.#batchedUpdates = true;
 
-    if (!this.#batching) {
-      this.#batching = true;
+    if (!this.#isNotificationScheduled) {
+      this.#isNotificationScheduled = true;
 
       queueMicrotask(() => {
-        this.#batching = false;
+        this.#isNotificationScheduled = false;
 
         // Check if an immediate update has been notified
         if (this.#batchedUpdates) {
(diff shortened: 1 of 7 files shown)


packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -674,7 +671,11 @@ export class CoValueCore {
     const session = this.verified.sessions.get(sessionID);
     const txIdx = session ? session.transactions.length - 1 : 0;
 
-    this.updateContentAndNotifyUpdate("immediate");
+    this.updateCurrentContent();
+
+    // force immediate notification because local updates may come from the UI
+    // where we need syncronous updates
```
```suggestion
    // where we need synchronous updates
```

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -528,17 +528,15 @@ export class CoValueCore {
...
 
       return result;
     });
   }
 
-  deferredUpdates = 0;
-  nextDeferredNotify: Promise<void> | undefined;
-
-  updateContentAndNotifyUpdate(notifyMode: "immediate" | "deferred") {
+  updateCurrentContent() {
```
nrainhart: Nit: could be private

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -550,47 +548,46 @@ export class CoValueCore {
     }
 
     this._cachedDependentOn = undefined;
-
-    this.notifyUpdate(notifyMode);
   }
 
-  notifyUpdate(notifyMode: "immediate" | "deferred") {
+  #batching = false;
+  #batchedUpdates = false;
```
nrainhart: Flag names are really similar. Mind adding a small ts doc on each flag explaining what they are used for?

nrainhart: Maybe `batching` can be renamed? Wdyt about `isNotificationScheduled`?

nrainhart (approved):
Excited to get batched updates! Left only a few minor comments

PR #2963, opened 2025-09-24
feat: webhook

gdorsi:
This PR introduces an experimental API for registering WebHooks on CoValue updates.


justinrosenthal (approved):

packages/jazz-webhook/src/webhook.ts
```diff
@@ -0,0 +1,373 @@
...
+    { nRetries: number; timeout?: NodeJS.Timeout }
+  >();
+  unsubscribe: () => void;
+
+  constructor(
+    private webhook: WebhookRegistration,
+    private options: JazzWebhookOptions = {},
+  ) {
+    this.successMap = webhook.$jazz
+      .ensureLoaded({ resolve: { successMap: { $each: true } } })
```
gdorsi: We actually don't need `$each` here because `successMap` doesn't have child coValues (the value is just a boolean)

Typescript should've catched this, it's probably a bug in our resolve

packages/jazz-webhook/src/webhook.ts
```diff
@@ -0,0 +1,373 @@
...
+        }
+        if (!this.webhook.active) {
+          return;
+        }
+        const todo = Array.from(
+          getTransactionsToRetry(
+            await this.successMap,
+            value.core.knownState(),
+          ),
+        );
```
gdorsi, line 213: **nit:** the `Array.from` is unnecessary here

packages/jazz-webhook/src/webhook.ts
```diff
@@ -0,0 +1,373 @@
...
+      this.webhook.coValueId as CoID<RawCoValue>,
+      async (value) => {
+        if (value === "unavailable") {
+          return;
+        }
+        if (!this.webhook.active) {
+          return;
+        }
+        const todo = Array.from(
+          getTransactionsToRetry(
```
gdorsi: nit: this method should be named "getTransactionsToEmit" or something similar

packages/jazz-webhook/src/successMap.ts
```diff
@@ -0,0 +1,42 @@
...
+) {
+  // TODO: optimisation: we can likely avoid even constructing a CoMap view
+  // and just get/set raw transactions from the CoValueCore of SuccessMap
+  for (const [sessionID, knownTxCount] of Object.entries(knownState.sessions)) {
+    for (let txIndex = 0; txIndex < knownTxCount; txIndex++) {
+      if (!successMap[`${sessionID}:${txIndex}`]) {
+        yield {
+          sessionID: sessionID as SessionID,
+          txIndex: txIndex,
+        } satisfies CojsonInternalTypes.TransactionID;
```
gdorsi: question: why using a generator here?

PR #2964, opened 2025-09-25
feat(jazz-tools): subscription scope re-use in react and coValue provider/consumer helpers

Gabrola:
# Description
More granular reactivity changes, but this isn't the main goal actually. We've realized that every single `useCoState` call can incur a very heavy cost without actually needing to have a ton of them if you're doing some deep resolves. For example, we've have a use case where needed to do some medium complexity logic on a co-state with a couple of nested resolves, so we wrapped up the `useCoState` hook inside a custom hook that was used in about 5 to 10 components, and we realized that we'd often get 100ms+ render times for every single hook call, meaning a single re-render could take 500ms to 1000ms+ at times.

We were able to internally fix this issue by making sure these expensive hooks are called only once in the entire tree with the result served through a context. This by itself obviously wouldn't be ideal because then that would mean EVERYTHING would re-render on every single change. So we had to store the results in a non-reactive store that can then be read using reactive hooks wherever they're needed further down the tree. This wasn't a clean or issue-free solution, but it worked beautifully in terms of performance. Here are the results I'm getting from the benchmark I added.

```
 ✓  jazz-tools  src/react-core/tests/subscription.bench.tsx > 1000 value loads 2481ms
     name                                                   hz      min      max     mean      p75      p99     p995     p999     rme  samples
   · useAccountSubscription + useSubscriptionSelector  90.3840   8.1881  16.7962  11.0639  10.7508  16.7962  16.7962  16.7962  ±3.78%       50
   · useAccountWithSelector                            34.1230  25.1204  40.1154  29.3057  29.7515  40.1154  40.1154  40.1154  ±4.12%       50

 ✓  jazz-tools  src/react-core/tests/subscription.bench.tsx > deeply resolved coMaps 8756ms
     name                                                   hz      min      max     mean      p75      p99     p995     p999      rme  samples
   · useAccountSubscription + useSubscriptionSelector  97.7141   6.2516  15.5710  10.2339  11.7479  13.3393  13.5106  15.5710   ±3.00%      200
   · useAccountWithSelector                            31.1867  20.0294   413.61  32.0650  29.1083   131.56   195.34   413.61  ±14.21%      200

 BENCH  Summary

   jazz-tools  useAccountSubscription + useSubscriptionSelector - src/react-core/tests/subscription.bench.tsx > 1000 value loads
    2.65x faster than useAccountWithSelector

   jazz-tools  useAccountSubscription + useSubscriptionSelector - src/react-core/tests/subscription.bench.tsx > deeply resolved coMaps
    3.13x faster than useAccountWithSelector
```

2x - 3x faster is nice, but wouldn't really matter much if we're talking about single digit millisecond times, but you can see how especially in the deeply resolved benchmark the p99+ times are more than 10x slower.

It looks like the creation of SubscriptionScopes is what's really very computationally intensive, but adding listeners to the same subscription scope is basically negligible in terms of performance impact. When you think of it, there are a lot of examples ,`useAccount` being the main one, where it's just a waste to keep subscribing to the same data multiple times.

The solution in this PR is really nice, because now we can share the same subscription scope with all child components, while still maintaining granular reactivity. The DX enabled by the `createCoValueSubscriptionContext` and `createAccountSubscriptionContext` helper functions is pretty nice because it really simplifies the common pattern of sharing a "global" coValue with other components without having to explicitly add types to the props of all components (or pass an ID an keep creating new subscriptions for the same data).

## Tests

## Checklist

### DIFF
--- a/.changeset/popular-ears-remain.md
+++ b/.changeset/popular-ears-remain.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Expose subscription scope react hooks and new `createCoValueSubscriptionContext` and `createAccountSubscriptionContext` helper functions
--- a/examples/music-player/src/2_main.tsx
+++ b/examples/music-player/src/2_main.tsx
@@ -13,10 +13,14 @@ import "./index.css";
 import { MusicaAccount } from "@/1_schema";
 import { apiKey } from "@/apiKey.ts";
 import { SidebarProvider } from "@/components/ui/sidebar";
-import { JazzReactProvider, useAccountWithSelector } from "jazz-tools/react";
+import { JazzReactProvider } from "jazz-tools/react";
 import { onAnonymousAccountDiscarded } from "./4_actions";
 import { KeyboardListener } from "./components/PlayerControls";
 import { usePrepareAppState } from "./lib/usePrepareAppState";
+import {
+  AccountProvider,
+  useAccountSelector,
+} from "@/components/AccountProvider.tsx";
 
 /**
  * Walkthrough: The top-level provider `<JazzReactProvider/>`
@@ -33,9 +37,8 @@ function AppContent({
 }: {
   mediaPlayer: ReturnType<typeof useMediaPlayer>;
 }) {
-  const showWelcomeScreen = useAccountWithSelector(MusicaAccount, {
-    resolve: { root: true },
-    select: (me) => Boolean(me && !me.root.accountSetupCompleted),
+  const showWelcomeScreen = useAccountSelector({
+    select: (me) => !me.root.accountSetupCompleted,
   });
 
   const isReady = usePrepareAppState(mediaPlayer);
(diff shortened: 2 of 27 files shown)


gdorsi:
Man you are a magican! ✨ 

Besides the performance gap, I like the DX that a provider + a selector brings.

Some people wants to just load everything on load, and this makes it way easier.
It also plays nice with app-level branching, which is something I'd like to try (thinking about having a dev branch for production data!)

Thanks for another amazing contribution!

gdorsi (approved):

Gabrola:
@gdorsi thank you! 🙇 Yeah, we've been using a hacked up version of this provider pattern in our code internally, which was very useful, but the code was really ugly compared to having it as part of the Jazz library like this.

I didn't think of any app-level branching use cases, just didn't want to lose any functionality that already exists, but the branching for environments sounds like a genius idea!

PR #2967, opened 2025-09-25
fix: update CoValue _subscriptionScope when using getSubscriptionScope

nrainhart:
# Description

We had a small bug in `getSubscriptionScope` that led us to not cache created Subscription Scopes. Don't think it was causing any issues, but caching them prevents generating multiple subscription scopes for the same CoValue.

## Tests

## Checklist


gdorsi (approved):
Nice catch! Thanks

PR #2968, opened 2025-09-25
feat(jazz-tools): Add pure vector search library + tests

kysely:
# Description

This PR implements the library for performing vector similarity search on a CoList instance, and its hook bindings for React.

### Comments on API

Similarly to the original [jazz-vector lib](https://github.com/kysely/jazz-vector?tab=readme-ov-file#use-semantic-search), the semantic search runs on any `CoList<CoMap>` that includes a vector parameter. Same as the former lib, it's a separate function call/hook after loading the list.

However, as [query modifiers are coming to Jazz](https://github.com/garden-co/jazz/pull/2946), I was wondering whether we should start on top of those. And although I would love to move the vector similarity search to `.load` resolver, my suggestion is to postpone to after indexing is released.

Now the API proposed in this PR is the following (practically same for both the lib function & React hook):

```js
import { useVectorSearch } from "jazz-tools/react";

useVectorSearch(list, {
  $orderBy: {
    'embeddingParam': { $similarity: [ -1.217787, -0.293539, ... ] },
  },
  $limit: 10, // or `$similarityThreshold` or `$similarityTopPercent`
})'
```

**Why “order by”?**

Technically speaking, “similarity search” is a _ranking task_, not a boolean filter (as in “where field > 40”).

Despite its non-intuitiveness, I chose “order by” API with the following assumptions:
1) We will move similarity search later to the native [`.load` query resolver](https://github.com/garden-co/jazz/pull/2946), incl. the indexing

2) With bigger amounts of data, it's not uncommon that you want to “filter out” sections of your dataset to perform the search on (e.g. [Pinecone docs for search w/ filtering](https://docs.pinecone.io/guides/search/filter-by-metadata#search-with-a-metadata-filter)). Using “order by” will enable this naturally with query modifiers, for example:
```js
Diagnoses.load(diagnoses.id, {
  resolve: {
    $where: {
      'category': 'digestive system'
    },
    $orderBy: {
      'embeddingParam': { $similarity: [ -1.217787, -0.293539, ... ] }
    },
    $limit: 20
});
```

The [`pgvector` also does similarity search in ORDER BY](https://github.com/pgvector/pgvector?tab=readme-ov-file#getting-started) as it naturally extends SQL (although it's not a reason why this would be a good API decision).

On the other hand, the vector search fn/hook are separate for now and the API on its own is unnecessarily complex and quite non intuitive (which are both huge cons).

**Please let's discuss...**

```sh
pnpm --filter jazz-tools test:watch search --reporter verbose
```

```
 ✓  jazz-tools  src/vector/search.test.ts (12 tests) 518ms
   ✓ Vector search (rank CoList items by similarity) (12)
     ✓ when input list is `undefined` • returns undefined 1ms
     ✓ when input list is `null` • returns null 0ms
     ✓ when search query is `null` • returns result object with original list without ranking 108ms
     ✓ when search query is provided • computes similarity and sorts results descending by similarity 49ms
     ✓ results filtering (3)
       ✓ supports filtering by count `limit` 46ms
       ✓ supports filtering by `similarityThreshold` 45ms
       ✓ supports filtering by `similarityTopPercent` 46ms
     ✓ when AbortSignal is engaged • returns unsorted items without similarity 45ms
     ✓ when no `orderBy` key is provided • throws an error 45ms
     ✓ when multiple `orderBy` keys are provided • throws an error 44ms
     ✓ when `orderBy` key does not reference a vector field • throws an error 44ms
     ✓ when query vector has dimensions mismatch • throws an error 43ms
 ✓  jazz-tools  src/react/tests/vector/useVectorSearch.test.ts (12 tests) 1332ms
   ✓ useVectorSearch (React hook) (12)
     ✓ basic functionality (4)
       ✓ should perform vector search with valid data 171ms
       ✓ should handle empty list 102ms
       ✓ should handle null list 98ms
       ✓ should handle undefined list 101ms
     ✓ loading states (1)
       ✓ should show loading state during search 99ms
     ✓ error handling (1)
       ✓ should handle invalid orderBy key 103ms
     ✓ search options (3)
       ✓ should respect limit option 101ms
       ✓ should respect similarity threshold 100ms
       ✓ should respect top percent option 100ms
     ✓ search ID memoization (2)
       ✓ should not trigger new search when list and options are the same 101ms
       ✓ should trigger new search when list changes 152ms
     ✓ edge cases (1)
       ✓ should handle null query vector 103ms

 Test Files  2 passed (2)
      Tests  24 passed (24)
Type Errors  no errors
   Duration  2.41s
```

## Tests

## Checklist


kysely:
Abandoning custom search hook in favor of user's own similarity sort.

PR #2976, opened 2025-09-29
fix: stop new content processing until all the dependencies are available

gdorsi:
# Description

Refactoring the dependency management to pause the new content processing until all the dependencies are available.

This is required to work correctly with sharing where the cores won't push the dependencies.

Ensures that all the dependencies are available before sending the content, instead of the fragmented approach we currently have on sessions.

This would prevent partial loads of covalues in the client, which could lead into incosistent statuses.

## Tests

## Checklist


packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -283,25 +292,31 @@ export class CoValueCore {
...
+    this.peers.set(peerId, { type: "available" });
+    this.updateCounter(previousState);
+    this.scheduleNotifyUpdate();
+  }
+
   missingDependencies = new Set<RawCoID>();
 
-  // Checks if the current CoValueCore is already a missing dependency of the given CoValueCore
-  checkCircularDependencies(dependency: CoValueCore) {
+  isCircularMissingDependency(dependency: CoValueCore) {
```
gdorsi, line 303: Renamed and inverted the return value, just to make more clear the intent

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -310,43 +325,50 @@ export class CoValueCore {
...
+  }
 
-        if (this.isAvailable()) {
-          this.scheduleNotifyUpdate();
-        }
-      });
+  delayHandleNewContent(
+    msg: NewContentMessage,
+    from: PeerState | "storage" | "import",
+  ) {
```
gdorsi, line 348: Going to rename the method and make it work more like a queue, to make the status easier to inspect.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -360,11 +382,6 @@ export class CoValueCore {
       streamingKnownState,
     );
 
-    this.peers.set(fromPeerId, { type: "available" });
-
-    this.updateCounter(previousState);
-    this.scheduleNotifyUpdate();
```
gdorsi, line 363: Moved this logic in `markFoundInPeer` to reuse this method on CoValue creation

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -989,43 +1044,36 @@ export class CoValueCore {
     return matchingTransactions;
   }
 
-  dependenciesProcessed = false;
   dependencies: Set<RawCoID> = new Set();
-  processCoValueDependencies() {
-    if (!this.verified || this.dependenciesProcessed) {
-      return;
+  private addDependency(dependency: RawCoID, skipVerify: boolean) {
```
gdorsi: Made this the only place when dependencies are added, to ensure we handle correctly the missingDependencies tracking.

packages/cojson/src/localNode.ts
```diff
@@ -119,16 +120,6 @@ export class LocalNode {
...
 
-  private putCoValue(
-    id: RawCoID,
-    verified: VerifiedState,
-    { forceOverwrite = false }: { forceOverwrite?: boolean } = {},
-  ): AvailableCoValueCore {
-    const entry = this.getCoValue(id);
-    entry.internalMarkMagicallyAvailable(verified, { forceOverwrite });
-    return entry as AvailableCoValueCore;
-  }
```
gdorsi, line 130: Removed to make `provideHeader` the only way to set the covalues header.

packages/cojson/src/localNode.ts
```diff
@@ -755,7 +751,7 @@ export class LocalNode {
     return group;
   }
 
-  loadCoValueAsDifferentAgent(
+  async loadCoValueAsDifferentAgent(
```
gdorsi, line 760: Refactored to use the sync as a way to load a value with a temp agent instead of manually cloning.

This gives us a single point of entry, and reduce the risk of getting into incosistent states.

packages/cojson/src/sync.ts
```diff
@@ -659,57 +667,8 @@ export class SyncManager {
...
       }
 
-      // If we'll be performing transaction verification, ensure the account is available.
-      if (!this.skipVerify) {
-        const accountId = accountOrAgentIDfromSessionID(sessionID);
-
-        if (isAccountID(accountId)) {
-          const account = this.local.getCoValue(accountId);
-
-          // We can't verify the transaction without the account, so we delay the session content handling until the account is available
```
gdorsi, line 669: Delaying a single session could lead to inconsistent statuses on the client, where some of the transactions could be missing.

This can be quite a problem if there are some migrations that check the presence of a property to set default values.

justinrosenthal (approved):
LGTM

PR #2980, opened 2025-09-29
feat: add `skipErrorsInLists` option to subscriptions

nrainhart:
# Description

Adds a new global `skipErrorsInLists` configuration that allows skipping inaccessible elements from CoLists when loading them. In order to enable this feature, use:

```typescript
skipErrorsInLists(true)
```
anywhere in your app, before you start loading CoValues.

Before this feature, the only alternative for loading CoLists with inaccessible items was using `$each: { $onError: null }` on each resolve query, and manually filtering out the null references.

While working on this I found some edge cases that were not being properly handled by pop, shift and splice, so those are fixed as well.

## Key changes

Now that inaccessible items are not shown as part of the CoList, we've introduced a difference between the elements in the RawCoValue and the elements in the jazz-tools CoList.

This PR introduces the concept of **query views** to support this new feature. The query view gives us a way to decouple the elements from a jazz-tools CoList from the elements in the underlying RawCoList, by mapping the keys in the CoList to the keys in the RawCoList. This means users will see only the elements returned by their resolve query, but will still be able to perform operations on the CoList that will be mapped to the underlying RawCoList.

The logic for creating and invalidating the query view is pretty naive at the moment (we recompute the whole view on each change). We'll be implementing a more efficient strategy for incrementally updating the view as a follow-up.

In the future, query views will also enable filtering, sorting and pagination.

## Benchmark

Tested query view generation with CoLists containing 10k - 50k elements and got times between 4ms and 20ms. Considering the generation time will grow linearly, I think performance is good enough for this initial release. I'll start looking into Incremental View Maintenance as a follow-up. 

## Tests

## Checklist

---

> [!NOTE]
> Introduce global skipErrorsInLists and query view mapping so CoLists hide inaccessible items and all mutations/read ops respect the filtered view; add related subscribe utilities and tests.
> 
> - **CoList / Query Views**:
>   - Introduce query views to map visible indices to raw indices (`queryView`, `computeQueryView`).
>   - Update `CoListJazzApi` to respect query views: `get`, `set`, `push`, `unshift`, `pop`, `shift`, `splice`, `remove`, `applyDiff`, `length`, `asArray`, and `refs` now translate via `toRawIndex` and keep the view in sync.
>   - Adjust `CoList` proxy (`get`, `length`, `has`, `ownKeys`, `getOwnPropertyDescriptor`) to use query views.
> - **Subscriptions**:
>   - Add global `skipErrorsInLists(skip: boolean)`; when enabled, inaccessible list items are skipped by default.
>   - Support `$onError: "skip"` internally and cache per-scope query views in `SubscriptionScope`.
>   - Refine resolve handling, autoloading, and child access with `accessChildLoadingStateByKey` (new) and reuse in `accessChildByKey`.
> - **Exports**:
>   - Re-export `skipErrorsInLists` and new query view utilities.
> - **Bug fixes / Behavior**:
>   - Handle empty `pop`/`shift`, ignore out-of-bounds in `splice`, and ensure mutations operate on filtered views.
> - **Tests**:
>   - Add/expand tests for skipErrorsInLists, `$onError: null` behaviors, query views (reads/mutations/array methods), and edge cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1b7a70426bcb9d964434af257757643dabeee5e6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/coList.ts
+++ b/packages/jazz-tools/src/tools/coValues/coList.ts
@@ -11,6 +11,7 @@ import {
   coField,
   CoFieldInit,
   CoKeys,
+  computeQueryView,
   CoValue,
   CoValueClass,
   CoValueJazzApi,
@@ -509,6 +510,8 @@ export class CoList<out Item = any>
 type CoListItem<L> = L extends CoList<unknown> ? L[number] : never;
 
 export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
+  private _queryView: QueryView | null = null;
+
   constructor(
     private coList: L,
     private getRaw: () => RawCoList,
@@ -574,6 +577,7 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
       return;
     }
     this.raw.replace(rawIdx, rawValue);
+    this.syncQueryView();
   }
 
   /**
@@ -590,6 +594,7 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
       undefined,
       "private",
     );
+    this.syncQueryView();
 
     return this.length;
   }
@@ -608,6 +613,7 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
     )) {
       this.raw.prepend(item);
     }
+    this.syncQueryView();
 
     return this.length;
   }
@@ -626,6 +632,7 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
     const last = this.coList[this.coList.length - 1];
 
     this.raw.delete(this.toRawIndex(this.coList.length - 1)!);
+    this.syncQueryView();
 
     return last;
   }
@@ -644,6 +651,7 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
     const first = this.coList[0];
 
     this.raw.delete(this.toRawIndex(0)!);
+    this.syncQueryView();
 
     return first;
   }
@@ -670,10 +678,11 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
       idxToDelete--
     ) {
       const rawIdx = this.toRawIndex(idxToDelete);
-      if (rawIdx === undefined || rawIdx >= this.coList.length) {
+      if (idxToDelete >= this.coList.length || rawIdx === undefined) {
         continue;
       }
       this.raw.delete(rawIdx);
+      this.syncQueryView();
     }
 
     const rawItems = toRawItems(
@@ -696,6 +705,7 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
       } else {
         this.raw.append(item, this.toRawIndex(Math.max(start - 1, 0)));
       }
+      this.syncQueryView();
       return deleted;
     }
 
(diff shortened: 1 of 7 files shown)


packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
```diff
@@ -58,7 +72,16 @@ export class SubscriptionScope<D extends CoValue> {
...
   ) {
-    this.resolve = resolve;
+    this.resolve = resolve === true || !resolve ? {} : resolve;
+    if (SKIP_ERRORS_IN_COLISTS) {
+      if (this.resolve.$each === true) {
+        this.resolve.$each = {};
+      }
+      if (this.resolve.$each && this.resolve.$each.$onError !== null) {
+        // @ts-expect-error 'skip' is only used internally, it's not part of the resolve query API
+        this.resolve.$each.$onError = "skip";
```
nrainhart, line 82: I actually implemented `skipErrorsInLists` using `$onError: 'skip'`, so if we want we can switch to that API

packages/jazz-tools/src/tools/tests/coList.test.ts
```diff
@@ -1320,6 +1537,114 @@ describe("CoList unique methods", () => {
...
+      element.startsWith("b"),
+    );
+    expect(removedItems.length).toEqual(2);
+    expect(removedItems[0]?.toString()).toEqual("onion");
+    expect(removedItems[1]?.toString()).toEqual("pasta");
+  });
+
+  test.skip("splice takes the query view into account", () => {
+    // TODO we need to update the query view synchronously for this to work
+    const removedItems = list.$jazz.splice(1, 2);
```
nrainhart: Right now synchronous mutations break the query view (until the subscription scope recreates it). I need to ensure the view is updated synchronously on local mutations

nrainhart: [`1b7a704` (#2980)](https://github.com/garden-co/jazz/pull/2980/commits/1b7a70426bcb9d964434af257757643dabeee5e6)

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -808,6 +885,44 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
...
+  private get queryView(): QueryView | null {
+    // The query view is obtained from the subscription scope when the CoList is loaded,
+    // and then updated synchronously when the CoList is modified locally.
+    this._queryView ||= this._subscriptionScope?.queryView ?? null;
+    return this._queryView;
+  }
+
+  private syncQueryView() {
+    if (this._queryView && this._subscriptionScope?.queryView) {
+      this._queryView = computeQueryView(this.coList);
```
gdorsi, line 904: **question:** this mean that we potentially end up computing the queryView twice per local update?

nrainhart: Yeah. It's not ideal, but for skipping inaccessible items the cost of computing the query view isn't really high (generating the view for a 10k colist takes ~4ms), so I think it's ok for the time being. For more costly operations (like sorting) this won't work, unless our Incremental View Maintenance strategy is efficient enough

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -521,13 +524,60 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
...
     if (rawValue === null && !itemDescriptor.optional) {
       throw new Error(`Cannot set required reference ${index} to undefined`);
     }
-    this.raw.replace(index, rawValue);
+    const rawIdx = this.toRawIndex(index);
+    if (rawIdx === undefined) {
+      return;
+    }
+    this.raw.replace(rawIdx, rawValue);
+    this.syncQueryView();
```
gdorsi, line 580: Do we really need to invalidate the queryView here?

We do it already in SubscriptionScope.updateValue and that should be enough in theory to cover also syncronous local updates.

PR #2988, opened 2025-09-29
perf: batch notification updates in applyDiff

gdorsi:


nrainhart (approved):
Would love to get to a more principled solution when we refactor the subscription scope, but this is ok for the time being

gdorsi:
> Would love to get to a more principled solution when we refactor the subscription scope, but this is ok for the time being

It would be even better to batch the transactions, but I failed on the first try: https://github.com/garden-co/jazz/pull/2978

PR #2989, opened 2025-09-30
Refactor SSR agent to be reusable across frameworks

joeinnes:
# Description
As I was looking at the SSR agent, I realised that there's no specific reason this needs to be limited to react. Made some changes to export this from `jazz-tools/ssr` rather than `jazz-tools/react/ssr`, and added an example SvelteKit app where the load function runs on the server.

Spin up the `jazz-sveltekit` example app.
Check the returned response from the server to validate that data is returned from the load function.
Note: because of how SvelteKit does SSR/hydration, you won't see the pure HTML, you'll see an empty page with a script tag. The script tag should look something like this:

```
<script>
{
  __sveltekit_dev = {
    base: new URL("..", location).pathname.slice(0, -1),
    env: {}
  };
  const element = document.currentScript.parentElement;
  Promise.all([
    import("/@fs/Users/joe/WebDev/jazz/node_modules/@sveltejs/kit/src/runtime/client/entry.js"),
    import("/@fs/Users/joe/WebDev/jazz/examples/jazz-sveltekit/.svelte-kit/generated/client/app.js")
  ]).then(([kit, app]) => {
    kit.start(app, element, {
      node_ids: [0, 3],
      data: [null, {
        type: "data",
        data: {
          profile: {
            name: "Joe Innes" // This will be the server-loaded profile.name
          }
        },
        uses: {
          params: ["profileId"]
        }
      }],
      form: null,
      error: null
    });
  });
} 
</script>
```

## Tests

- [x] Tests have not been updated, because: Example app shows behaviour. No existing test for the SSR agent

## Checklist


homepage/homepage/content/docs/project-setup.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+# Installation and Setup
+
+Add Jazz to your application in minutes. This setup covers standard front-end set up, and gives an overview of experimental SSR approaches.
+
+Integrating Jazz with your existing app is straightforward. You'll define data schemas that describe your application's structure, then wrap your app with a provider that handles sync and storage. The whole process takes just three steps:
+
+1. [Install dependencies](#install-dependencies)
+2. [Write your schema](#write-your-schema)
+3. [Wrap your app in a provider`](#standard-setup)
```
nrainhart:
```suggestion
3. [Wrap your app in a provider](#standard-setup)
```

homepage/homepage/content/docs/project-setup.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+<JazzSvelteProvider {sync} enableSSR>
+  {@render children?.()}
+</JazzSvelteProvider>
+```
+</TabbedCodeGroupItem>
+</TabbedCodeGroup>
+
+<ContentByFramework framework="react">
+Since the provider is running without an authenticated user, all the `useCoState` and `useAccount` will return null.
```
nrainhart:
```suggestion
Since the provider is running without an authenticated user, all the `useCoState` and `useAccount` hooks will return null.
```

homepage/homepage/content/docs/project-setup/ssr.mdx
```diff
@@ -0,0 +1,307 @@
...
+
+export const jazzSSR = createSSRJazzAgent({
+  peer: "wss://cloud.jazz.tools/",
+});
+```
+</CodeGroup>
+
+## Telling Jazz to use the SSR agent
+
+Normally, Jazz expects a logged in user (or an anonymous user) to be accessing data. We can use the `enableSSR` setting to tell Jazz that this may not be the case, and the data on the page may be being accessed by an agent.
```
nrainhart, line 43: Is there a page in the docs where we explain the difference between accounts & agents in more detail? If we do, maybe we could link it here

examples/jazz-sveltekit/README.md
nrainhart, line 1: Should we replace this README with something specific for this example?

examples/jazz-sveltekit/src/routes/+layout.svelte
```diff
@@ -0,0 +1,13 @@
+<script lang="ts">
+  import { JazzSvelteProvider } from "jazz-tools/svelte";
+  import { apiKey } from "$lib/apiKey";
+
+  let { children } = $props();
+  const sync = { peer: `wss://cloud.jazz.tools/?key=${apiKey}` };
+</script>
+
+<JazzSvelteProvider {sync} enableSSR>
```
nrainhart, line 10: Maybe add a comment to highlight we're setting the `enableSSR` flag?

examples/jazz-sveltekit/src/routes/+page.svelte
```diff
@@ -0,0 +1,42 @@
...
+
+  const account = new AccountCoState(Account, {
+    resolve: {
+      profile: true,
+    },
+  });
+  const me = $derived(account.current);
+  const navigate = async (route: string) => {
+    // Flush the name to avoid SSR picking up the previous name
+    await me?.profile.$jazz.waitForSync();
```
nrainhart, line 14: Ugh, this is annoying. Can you explain the error you saw? I feel like this is something that we should address on our end

nrainhart (approved):
I don't have any context around Svelte SSR, but all changes look reasonable!

Left only a few minor comments around documentation

homepage/homepage/content/docs/project-setup/ssr.mdx
joeinnes: No, but I'll add that to my to do list.

examples/jazz-sveltekit/src/routes/+page.svelte
joeinnes: GCO-867 to potentially pick up in cooldown as discussed

Issue #1235, opened 2025-01-22
Feature: add a new super-admin role with the authority of revoking access to admin members


gdorsi:
## Background

Currently, the role management system in Jazz has limitations around admin permissions where admins cannot be downgraded except by themselves. This creates potential issues for organization management and differs from common patterns seen in platforms like GitHub. A new "Owner" role is proposed to provide additional control while maintaining security.

### New Role Definition

* Create a new "Owner" role that sits above Admin in the permission hierarchy
* Owner should have all Admin permissions plus:
  * Ability to downgrade/remove Admin users
  * Ability to manage all member roles regardless of their current role level
* The system will not enforce owner succession rules (when an owner leaves) - this will be left to individual applications to implement

## Additional Context

This change aims to provide better control for organization management while maintaining the secure-by-default approach of Jazz. The goal is to allow members to be given invitation powers (Admin role) without risking unauthorized removal of original owners or other members.

gdorsi:
From anselm feedback:

> Can we name it something else to distinguish it from the CoValue "owner" concept?

raymonddaikon:
This would be great, perhaps call it `creator`?

gdorsi:
> This would be great, perhaps call it `creator`?

Yeah, maybe it's a better way of naming it.

Elfo404:
from some other discussion (rephrased):

> i think we need a super-admin role in jazz that is allowed to remove other admins from groups - or maybe a "writer with invite permissions".
To implement account rotation for worker groups, where workers should be able to invite other members, they currently need to be admin. however this makes it impossible to remove those worker accounts from either a "god" account or other worker accounts because admin cannot kick/demote other admins.
So for this use case we would either need a role that is a writer + invite permission, so that a worker can add other accounts to a group - most likely up to this new "writer+invite" role, and at the same time be removed by regular admins, or have "super-admins" as per the issue description.
This is important because at this time, if one creates admin worker accounts, once one of them is compromised (ie. by leaking its secret) there's no way to remove it.

Takeno:
Introduced the new role _manager_ to achieve this: #3050

PR #2995, opened 2025-10-01
New SuperAdmin role

Takeno:
Ref #1235 

`superAdmin` is a new role available when creating a Group. 
Despite the `admin` role, the new role has unrestricted power in managing other roles. It can add new admins and super admins, but it can also remove them. 

Due to its dangerous behaviour, the default role is still `admin`.

```ts
const group = Group.create();
expect(group.getRoleOf("me")).toBe("admin");

const group = Group.create({role: "superAdmin"});
expect(group.getRoleOf("me")).toBe("superAdmin");
```

---

This PR also introduces a refactoring on addMember/removeMember to validate permissions at a lower level (inside permission.ts) 

Previously, cojson/group.ts was trying to understand why an operation failed. Now, if the role setting fails, the error message is unified and enriched by useful information for debugging: `Failed to set role ${role} to ${memberKey} (role of current account is ${this.myRole()})`.

To be consistent and to give appropriate feedback, `group.removeMember` now throws an error if it fails (it was failing silently before). Due to this change, this might be considered a minor update instead of a patch.

### DIFF
--- a/packages/cojson/src/coValues/group.ts
+++ b/packages/cojson/src/coValues/group.ts
@@ -51,7 +51,8 @@ export type ParentGroupReferenceRole =
   | "extend"
   | "reader"
   | "writer"
-  | "admin";
+  | "admin"
+  | "superAdmin";
 
 export type GroupShape = {
   profile: CoID<RawCoMap> | null;
@@ -311,7 +312,15 @@ export class RawGroup<
     account: RawAccount | ControlledAccountOrAgent | AgentID | Everyone,
     role: Role,
   ) {
-    if (account === EVERYONE) {
+    const memberKey = typeof account === "string" ? account : account.id;
+    const previousRole = this.get(memberKey);
+
+    // if the role is the same, we don't need to do anything
+    if (previousRole === role) {
+      return;
+    }
+
+    if (memberKey === EVERYONE) {
       if (!(role === "reader" || role === "writer" || role === "writeOnly")) {
         throw new Error(
           "Can't make everyone something other than reader, writer or writeOnly",
@@ -323,22 +332,13 @@ export class RawGroup<
         throw new Error("Can't add member without read key secret");
       }
 
-      const previousRole = this.get(account);
+      const previousRole = this.get(memberKey);
 
-      this.set(account, role, "trusting");
+      this.set(memberKey, role, "trusting");
 
-      if (this.get(account) !== role) {
-        // The role was not set correctly; this presents three scenarios:
-        // 1. The current user is an administrator trying to set another administrator to a lower role
-        // 2. The current user is an administrator but something has gone wrong with the role assignment
-        // 3. The current user is not an administrator and does not have sufficient permissions to set the role
-        const myRole = this.myRole();
+      if (this.get(memberKey) !== role) {
         throw new Error(
-          myRole === "admin"
-            ? this.get(account) === "admin"
-              ? "Administrators cannot demote other administrators in a group"
-              : "Failed to set role"
-            : `Failed to set role due to insufficient permissions (role of current account is ${myRole})`,
+          `Failed to set role ${role} to ${memberKey} (role of current account is ${this.myRole()})`,
         );
       }
 
@@ -359,9 +359,12 @@ export class RawGroup<
       return;
     }
 
-    const memberKey = typeof account === "string" ? account : account.id;
     const agent =
       typeof account === "string" ? account : account.currentAgentID();
+
+    if (agent === EVERYONE) {
+      throw new Error("Agent should not be everyone");
+    }
 
     /**
      * WriteOnly members can only see their own changes.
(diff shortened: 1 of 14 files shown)


packages/cojson/src/coValues/group.ts
```diff
@@ -1254,23 +1256,31 @@ export class RawGroup<
...
 function isMorePermissiveAndShouldInherit(
-  roleInParent: "revoked" | "admin" | "writer" | "reader",
+  roleInParent: "revoked" | "admin" | "superAdmin" | "writer" | "reader",
   roleInChild: Role | undefined,
 ) {
   if (roleInParent === "revoked") {
     return true;
   }
 
+  // TODO: verify
```
aeplay, line 1277: this seems inconsistent - superAdmin is more permissive and should be inherited if we are only an admin in the child group? cc @gdorsi

packages/cojson/src/tests/group.addMember.test.ts
```diff
@@ -1,326 +1,550 @@
...
-    });
-
-    const group = admin.node.createGroup();
-    const person = group.createMap({
-      name: "John Doe",
-    });
-
-    const personOnReaderNode = await loadCoValueOrFail(reader.node, person.id);
+// [sourceRole, from, to, canGrant]
+const GRANT_MATRIX: [Role, Role | undefined, Role, boolean][] = [
```
aeplay, line 16: beautiful

packages/cojson/src/tests/group.addMember.test.ts
```diff
@@ -1,326 +1,550 @@
...
+  ["admin", "reader", "reader", true],
+  ["admin", "reader", "writeOnly", true],
+
+  ["admin", "writeOnly", "superAdmin", false],
+  ["admin", "writeOnly", "admin", true],
+  ["admin", "writeOnly", "writer", true],
+  ["admin", "writeOnly", "reader", true],
+  ["admin", "writeOnly", "writeOnly", true],
+
+  ...["superAdmin", "admin", "writer", "reader", "writeOnly"].map(
```
aeplay, line 92: these might be worth spelling out for readability (the formatted type assertion is using a lot of lines anyways)

packages/cojson/src/tests/group.invite.test.ts
```diff
@@ -8,13 +8,67 @@ import {
...
   waitFor,
 } from "./testUtils.js";
+import { AccountRole } from "../permissions.js";
 
 beforeEach(async () => {
   SyncMessagesLog.clear();
   setupTestNode({ isSyncServer: true });
 });
 
+const CAN_CREATE_INVITE_MATRIX: [AccountRole, AccountRole, boolean][] = [
```
aeplay, line 18: Beautiful

packages/cojson/src/tests/group.test.ts
```diff
@@ -25,6 +25,37 @@ function expectGroup(content: RawCoValue): RawGroup {
   return content as RawGroup;
 }
 
+test("User is admin of a group he created", () => {
```
aeplay, line 28: Nitpick: let's use the gender neutral singular "they" for users, hopefully a quick find/replace

packages/cojson/src/permissions.ts
```diff
@@ -434,62 +450,125 @@ function determineValidTransactionsForGroup(
...
+
+    function markTransactionAsInvalid(message: string) {
+      logPermissionError(message);
+      transaction.isValid = false;
+    }
+
+    // is first self promotion to admin or super-admin
+    if (
+      transactorRole === undefined &&
       transactor === initialAdmin &&
```
aeplay, line 472: **Important:** I feel like we need to anchor superAdmin-ness in the CoValue header: otherwise someone who is just the initial admin could at a later date decide they maliciously want to become superAdmin and create a *back-dated* transaction (looks older than the initial self promotion to normal admin) that makes them superAdmin.

So we potentially need to add a boolean initialAdminIsSuperAdmin to the CoValue header or something equivalent. cc @gdorsi

packages/cojson/src/permissions.ts
```diff
@@ -434,62 +450,125 @@ function determineValidTransactionsForGroup(
...
+
+    /**
+     * Super admins can do anything
+     */
+    if (transactorRole === "superAdmin") {
+      markTransactionSetRoleAsValid(change);
+      continue;
+    }
+
+    /**
```
aeplay, line 501: these comments really help!

aeplay (changes requested):
Fantastic implementation of the feature and simultaneous big refactor that makes the most security relevant parts of the codebase more readable!

packages/cojson/src/coValues/group.ts
gdorsi: I think that we can't safely allow `superAdmin` inheritance.

Admins could exploit this to gain `superAdmin` access:
```ts
tatget.myRole() // admin
const exploit = Group.create({ role: "superAdmin" })
target.extend(exploit);
target.myRole() // superAdmin
``` 

Even with the checks below, it's just more complicated but doable.

packages/cojson/src/permissions.ts
gdorsi: Agree, we need an immutable flag in the header.

Takeno:
Closed in favor of a new role below the admin https://github.com/garden-co/jazz/pull/3050

Issue #2994, opened 2025-10-01
Feature request: `getOrCreateUnique` (or option on `upsertUnique`)


joeinnes:
## Problem

Right now `upsertUnique` always mutates, meaning user modifications will be overwritten if the record exists.

There are common cases where the intended semantics are instead: **create the record if it doesn’t exist, but if it does exist, just return it unchanged.**

---

## Option 1: Separate method

```ts
const learnJazzTask = await Task.getOrCreateUnique({
  value: {
    text: "Let's learn some Jazz!",
  },
  unique: "learning-jazz",
  owner: project.$jazz.owner,
});
```

**Pros**

* Very explicit about semantics.
* Matches naming conventions in other ORMs (`getOrCreate`, `findOrCreate`).
* Keeps `upsertUnique` narrowly focused.

**Cons**

* Duplicates a lot of `upsertUnique` implementation.
* Naming is a bit unwieldy (`getOrCreateUnique`).

---

## Option 2: Extend `upsertUnique` with a flag

```ts
const learnJazzTask = await Task.upsertUnique({
  value: {
    text: "Let's learn some Jazz!",
  },
  unique: "learning-jazz",
  owner: project.$jazz.owner,
  ifExists: "overwrite", // "overwrite" | "return" - defautlt 'overwrite'
});
```

**Pros**

* Reuses existing API surface — no new top-level method.
* Reduces code duplication internally.
* Easier to document and maintain one concept instead of two.

**Cons**

* Semantics are less immediately clear: `upsertUnique` is now doing two quite different things depending on an option.
* Risk of confusion between “update-or-insert” vs. “insert-or-return.”

---

## Option 3: Provide a thin alias (`getOrCreateUnique`)

```ts
const learnJazzTask = await Task.getOrCreate({
  value: {
    text: "Let's learn some Jazz!",
  },
  unique: "learning-jazz",
  owner: project.$jazz.owner,
});
```

Under the hood this would simply call:

```ts
Task.upsertUnique({ ..., ifExists: "return" });
```

**Pros**

* Clearer semantics than Option 2.
* Minimal implementation cost (just syntactic sugar).
* Avoids code duplication (unlike Option 1).

**Cons**

* Adds another method name to the API surface, which can cause slight cognitive overhead.
* Still means implementing Option 2, but 'hiding it' behind a different API.

itsneeku:
How about `ensureExists` to align with `ensureLoaded`

gdorsi:
[...]

PR #2997, opened 2025-10-02
Implement loadOrCreateUnique

joeinnes:
# Description
Implements a getOrCreateUnique to complement upsertUnique

N/A

## Tests

## Checklist


joeinnes:
I don't really understand why the E2E tests are failing, unit tests are good.

packages/jazz-tools/src/tools/tests/coMap.test.ts
```diff
@@ -2007,7 +2007,10 @@ describe("Creating and finding unique CoMaps", async () => {
         identifier: oldSourceData.identifier,
         external_id: oldSourceData._id,
       },
-      workspace,
+      {
```
joeinnes, line 2010: Drive-by fix: the old test doesn't seem to actually test the uniqueness behaviour at all.

packages/jazz-tools/src/tools/tests/coMap.test.ts
```diff
@@ -2025,7 +2028,6 @@ describe("Creating and finding unique CoMaps", async () => {
       identifier: newSourceData.identifier,
       external_id: newSourceData._id,
     });
-    expect(activeEvent).not.toEqual(oldActiveEvent);
```
joeinnes, line 2028: This should fail. `oldActiveEvent` should be the same as `activeEvent` because we upserted over it, so when we access `oldActiveEvent` here, we get the upserted event, `activeEvent`. The only reason this was working is because uniqueness wasn't added in this test. The earlier check ensuring that activeEvent equals the newSourceData should be adequate, but if we wanted to, we could potentially serialise the `oldActiveEvent` before the upsert and compare.

joeinnes:
Updated based on context in https://github.com/garden-co/jazz/issues/2994#issuecomment-3416485076

- name changed
- added deduplication map (similar to the [MessageQueue `processing` set](https://github.com/garden-co/jazz/blob/a0f64fe97b8fd3092a1231a42b71ce0c46a3ba7a/packages/jazz-tools/src/tools/coValues/inbox.ts#L122), but kept at module level like [coValueCache](https://github.com/garden-co/jazz/blob/a0f64fe97b8fd3092a1231a42b71ce0c46a3ba7a/packages/jazz-tools/src/tools/lib/cache.ts#L4) to avoid exposing to subclasses)
- adjust methods to check for in-flight operations
- add tests

Note: I didn't fix `upsertUnique`, but it has the same race condition concerns. I can apply the same fix if the approach is good (if so, I guess we should use the same map, right?)

gdorsi:
Let's pause this PR until we find a final pattern for unique

joeinnes:
Closing: https://github.com/garden-co/jazz/pull/3396

PR #2999, opened 2025-10-03
chore(inspector): show coID in listview + handle unavailable items

Takeno:
# Description
Unavailable items were not represented correctly inside the TableView. In addition, to help with debugging, the CoID of the list items is shown as the first column of the table. 

Before:
immagine

After:
immagine

### DIFF
--- a/.changeset/orange-ghosts-hug.md
+++ b/.changeset/orange-ghosts-hug.md
@@ -2,4 +2,4 @@
 "jazz-tools": patch
 ---
 
-Inspector now show the CoID of list items and handle correctly unavailable items
+Inspector now shows the CoID of list items and handles unavailable items correctly
--- a/.changeset/orange-ghosts-hug.md
+++ b/.changeset/orange-ghosts-hug.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Inspector now show the CoID of list items and handle correctly unavailable items
--- a/packages/jazz-tools/src/inspector/viewer/table-viewer.tsx
+++ b/packages/jazz-tools/src/inspector/viewer/table-viewer.tsx
@@ -16,6 +16,7 @@ import {
   TableRow,
 } from "../ui/table.js";
 import { Text } from "../ui/text.js";
+import { Icon } from "../ui/icon.js";
 
 const PaginationContainer = styled("div")`
   padding: 1rem 0;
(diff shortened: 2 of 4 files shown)


.changeset/orange-ghosts-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Inspector now show the CoID of list items and handle correctly unavailable items
```
nrainhart:
```suggestion
Inspector now shows the CoID of list items and handles unavailable items correctly
```

packages/jazz-tools/src/inspector/viewer/table-viewer.tsx
```diff
@@ -82,29 +87,47 @@ function CoValuesTableView({
...
-                    }}
-                  />
+              <TableCell>
+                <Text mono>
+                  {item.snapshot === "unavailable" && (
+                    <Icon
+                      name="caution"
+                      color="red"
+                      style={{ display: "inline-block", marginRight: "0.5rem" }}
+                    />
```
nrainhart: Would it be possible to show a tooltip on hover that explains the value cannot be loaded? Having the `Unavailable` legend as part of a column with an unrelated name is kind of confusing

packages/jazz-tools/src/inspector/viewer/table-viewer.tsx
```diff
@@ -82,29 +87,47 @@ function CoValuesTableView({
...
+                  )}
+                  {visibleRows[index]}
```
nrainhart: Right now we show nothing if the colist contains non-collaborative values, right? I wonder if that could be confusing

nrainhart (approved):
Thanks!! This bug made debugging super confusing.

Left a few comments that I think would be nice to address, but the current solution is already a big improvement over what we have, so feel free to disregard

packages/jazz-tools/src/inspector/viewer/table-viewer.tsx
Takeno: Added as hover tooltip
immagine

packages/jazz-tools/src/inspector/viewer/table-viewer.tsx
Takeno: We display the Index/Value for non-collaborative values, such as z.string()

immagine

packages/jazz-tools/src/inspector/viewer/table-viewer.tsx
```diff
@@ -25,6 +26,38 @@ const PaginationContainer = styled("div")`
   gap: 0.5rem;
 `;
 
+const RedTooltip = styled("span")`
+  position:relative; /* making the .tooltip span a container for the tooltip text */
+  border-bottom:1px dashed #000; /* little indicater to indicate it's hoverable */
+
+  &:before {
+    content: attr(data-text);
+    background-color: red;
```
nrainhart, line 35: Design nit: the combination of color and shape make this button look VERY similar to the button on the top right.

I'm not really good at making nice UIs though, so I don't have a better suggestion 😅

I'm ok with just merging as-is

nrainhart (approved):
Thanks for addressing my feedback!

PR #3001, opened 2025-10-03
feat: explicit loading states with `$isLoading` 

nrainhart:
# Description

Until now, we've used `undefined` to represent values that are being loaded, and `null` for values that cannot be loaded.

This has several problems:
- no information about why a CoValue is not available (e.g. is there a network connection issue or an authorization issue?)
- it's easy to confuse unset and unloaded CoValues, since both can be `undefined`
- unloaded values are usually handled implicitly using null checks/optional chaining, which conflate multiple loading states into one. This usually ends up causing unexpected behaviour once an app is deployed and used collaboratively.

This PR introduces a new way to represent loading states:
```typescript
// APIs that load a CoValue now return a "maybe loaded" CoValue:
type MaybeLoaded<C extends CoValue> = Loaded<CoValue> | Unloaded<CoValue>;

type Unloaded<C extends CoValue> = { 
  $isLoaded: false;
  $jazz: {    
    id: ID<C>;  
    loadingState: "unloaded" | "unauthorized" | "unavailable";  
  };
};

type Loaded<C extends CoValue> = {
  $isLoaded: true;
  $jazz: {
    id: ID<C>;
    loadingState: "loaded" ;
  } & CoValueAPI<C> // the whole $jazz API
  ... // all other properties from that CoValue
};
```

The goal with explicit loading states is to enable users to think more carefully about CoValues' resolution depth and loading states when building their apps, and make sure all cases are properly handled:

```typescript
const Person = co.map({
  name: z.string(),
  address: co.map({
    street: z.string(),
  }),
});

const john = await loadCoValue(Person, id, {
  resolve: { address: true },
});

// Before consuming the value we need a check that asserts it's loaded
assert(john.$isLoaded) 

// Thanks to the resolve query, we know that if john is loaded, so is address
console.log(john.address); // Fake Street 123 
```

## Key changes

- Added `$isLoaded` field to discriminate between loaded and unloaded CoValues
- Added `$jazz.loadingState` field to provide additional info about the loading state
- All methods and functions that load CoValues now return a `MaybeLoaded<CoValue>` instead of `CoValue | null | undefined`.
- Renamed `$onError: null` to `$onError: "catch"`
- Split the `useAccount` hook into three separate hooks:
	- `useAccount`: now only returns an Account CoValue
    - `useLogOut`: returns a function for logging out of the current account
    - `useAgent`: returns the current agent
- Added a `select` option (and an optional `equalityFn`) to `useAccount` and `useCoState`, and removed `useAccountWithSelector` and `useCoStateWithSelector`.
- Allowed specifying resolve queries at the schema level

## Follow-up work

I'll open separate PRs for the following pending tasks, and merge them against this PR before merging it in main:
- https://github.com/garden-co/jazz/pull/3061
- https://github.com/garden-co/jazz/pull/3068
- https://github.com/garden-co/jazz/pull/3085
- https://github.com/garden-co/jazz/pull/3147

Tagging this PR with `pre-release` in case you want to try it out in your apps!

## Tests

## Checklist

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit e11ffd002460d8cda16c9ce1df645721fca820db. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


examples/richtext-prosemirror/src/Editor.tsx
```diff
@@ -16,17 +17,26 @@ export function Editor() {
   });
   const editorRef = useRef<HTMLDivElement>(null);
 
+  const bioId =
```
nrainhart: Loading states becomes cumbersome if you want to use other React hooks in the same component, as you cannot use early returns to handle loading states until you've called all hooks

tests/jazz-svelte/src/routes/costate/+page.svelte
```diff
@@ -119,38 +120,39 @@
   </div>
 </div>
 
-{#if person.current}
+{#if person.current.$isLoaded}
+  {@const loadedPerson = person.current}
```
nrainhart, line 124: Type narrowings do not carry over inside lambdas, unless you assign the narrowed value to a variable (see e.g. https://github.com/garden-co/jazz/pull/3001/files#diff-73e7d86882108e9f1e8e526a6c599016dcaca4b09abff76958dd239916c5465bR136-R138). This is especially annoying when doing `$isLoading` assertions in JSX or Svelte templates.

Not sure how obvious the variable assignment solution would be for folks. We may need to add an `assertLoaded` function (or a `$jazz.assertLoaded` method) to get around this Typescript limitation, but I prefer to wait until this becomes a problem instead of preemptively extending our API surface.

packages/community-jazz-vue/src/composables.ts
```diff
@@ -140,14 +157,15 @@ export function useAccount<
 
 export function useCoState<
   S extends CoValueClassOrSchema,
-  const R extends RefsToResolve<S> = true,
+  const R extends ResolveQuery<S> = true,
```
nrainhart, line 157: Resolve queries were not being inferred correctly in Vue's `useCoState`. This is now fixed

examples/chat/src/chatScreen.tsx
```diff
@@ -71,11 +75,10 @@ export function ChatScreen(props: { chatID: string }) {
...
             .reverse()
-            .map(
-              (msg) =>
-                msg?.text && (
-                  <ChatBubble me={me} msg={msg} key={msg.$jazz.id} />
-                ),
+            .map((msg) =>
+              msg.text.$isLoaded ? (
+                <ChatBubble me={me} msg={msg} key={msg.$jazz.id} />
+              ) : null,
```
nrainhart, line 81: Type narrowing on children does not affect the parent CoValue 👎 

As far as I can remember, this is the only case where I modified a component to accept an unloaded nested CoValue (`msg.text`) to keep TS happy, so not sure how much of a problem this will be for adopters.

But we may need to add a way to assert a CoValue complies with a resolve query

packages/jazz-tools/src/tools/coValues/deepLoading.ts
```diff
@@ -31,60 +70,59 @@ export type RefsToResolve<
...
                         [0, ...CurrentDepth]
                       >;
-                      $onError?: null;
-                    }
+                    } & OnError)
                   | boolean
               : boolean);
 
-export type RefsToResolveStrict<T, V> = V extends RefsToResolve<T>
+export type RefsToResolveStrict<T, V> = [V] extends [RefsToResolve<T>]
```
nrainhart, line 126: This fixes a long-standing issue we had that prevented us from using string literals in resolve queries (e.g. `$onError: 'catch'`).

The problem was that when using `RefsToResolveStrict`, the type of a query using `$onError: 'catch'` was being incorrectly inferred as `$onError: string`, which is not a valid resolve query.

Making the conditional type [non-distributive](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types) solves the type inference issue. Now the resolve query type is correctly inferred as `$onError: 'catch'`.

packages/jazz-tools/src/tools/coValues/deepLoading.ts
```diff
@@ -22,6 +17,53 @@ type IsUnion<T, U = T> = (T extends any ? (x: T) => void : never) extends (
...
+
+/**
+ * By default, if a nested CoValue is not loaded, the parent CoValue will not be loaded either.
+ * When `$onError: "catch"` is used, the parent CoValue will always be loaded, and an {@link Unloaded}
+ * value will be returned for the nested CoValue if it cannot be loaded.
+ *
+ * Use `$onError` to handle cases where some data you have requested is inaccessible,
+ * similar to a `try...catch` block in your query.
+ */
+type OnError = { $onError?: "catch" };
```
nrainhart, line 59: I ended up going with `$onError: "catch"` instead of `$onError: "keep"` because I think the try-catch parallelism is easy enough to grasp.

`$onError: "keep"` could work, but we'd necessarily need to merge https://github.com/garden-co/jazz/pull/2980 first, and make skipping unaccessible items the default behavior (which I'm not convinced is the right approach).

@aeplay if you still prefer to go with `$onError: "keep"` let me know and we can discuss this further.

packages/jazz-tools/src/react-core/subscription-provider.tsx
```diff
@@ -42,24 +44,25 @@ export function createCoValueSubscriptionContext<
         resolve: resolve,
       });
 
-      const loadState = useSubscriptionSelector(subscription, {
-        select: (value) => (!value ? value : true),
-      });
+      const loadState = useSubscriptionSelector(subscription);
```
Gabrola:
```suggestion
      const loadState = useSubscriptionSelector(subscription, {
        select: (value) => value.$jazz.loadingState,
      });
```

Otherwise, the entire tree will re-render on changes for already loaded values

packages/jazz-tools/src/react-core/subscription-provider.tsx
```diff
@@ -42,24 +44,25 @@ export function createCoValueSubscriptionContext<
...
       }
 
       return (
         <Context.Provider value={subscription}>{children}</Context.Provider>
       );
     },
-    useSelector: <TSelectorReturn = Loaded<S, R>>(options?: {
-      select?: (value: Loaded<S, R>) => TSelectorReturn;
+    useSelector: <TSelectorReturn = MaybeLoaded<Loaded<S, R>>>(options?: {
+      select?: (value: MaybeLoaded<Loaded<S, R>>) => TSelectorReturn;
```
Gabrola: The types for this excluded the unloaded state because the provider will not render the children unless the value is loaded. The `useSelector` hook will never run (or throw if used outside the provider) unless the value is loaded.

nrainhart: Thanks for calling this out @Gabrola! I reverted the change and added a test suite for `createCoValueSubscriptionContext` to avoid regressions going forward.

What prompted me to make this change is that while migrating the organization example (that uses `createCoValueSubscriptionContext`) I noticed the selector was being triggered with what seemed to be unloaded values. But the error could be due to some other change I made in this PR, so I'll double check 🕵️‍♂️

packages/jazz-tools/src/tools/coValues/deepLoading.ts
nrainhart, line 1: This file contains the "real" changes in this PR. All the rest are just to accommodate to the new types

packages/jazz-tools/src/better-auth/database-adapter/repository/user.ts
```diff
@@ -68,6 +68,10 @@ export class UserRepository extends JazzRepository {
   private async findByEmail(email: string): Promise<TableItem[]> {
     const emailIndex = await this.loadEmailIndex(email);
 
+    if (!emailIndex.$isLoaded) {
+      return [];
+    }
+
     const user = emailIndex?.user;
```
gdorsi: nit: this optional chaining is not needed anymore

packages/jazz-tools/src/react/media/image.tsx
```diff
@@ -10,6 +10,7 @@ import {
 } from "react";
 import { highestResAvailable } from "../../media/index.js";
 import { useCoState } from "../hooks.js";
+import e from "cors";
```
gdorsi: Do we need this?

packages/jazz-tools/src/tools/testing.ts
```diff
@@ -392,3 +395,11 @@ export function disableJazzTestSync() {
   }
   syncServer.current = null;
 }
+
+export function assertLoaded<T extends MaybeLoaded<CoValue>>(
```
gdorsi: I think it would be useful to have this utility in `jazz-tools`

I would use this in some functions where I expect the value to be loaded successfully.

packages/jazz-tools/src/react/media/image.tsx
nrainhart: Nope, removed

packages/jazz-tools/src/tools/testing.ts
nrainhart: Sounds good. [`f96ea9e` (#3001)](https://github.com/garden-co/jazz/pull/3001/commits/f96ea9e0f51bf48eb91b4994664c747f91910be4)

packages/jazz-tools/src/tools/coValues/deepLoading.ts
```diff
@@ -117,7 +117,9 @@ export type RefsToResolve<
                       >;
                     } & OnError)
                   | boolean
-              : boolean);
+              : V extends { [TypeSym]: "CoPlainText" | "BinaryCoStream" }
+                ? boolean | OnError
```
nrainhart, line 123: `$onError` wasn't previously supported on plain/rich text and fileStreams. This is now fixed

gdorsi (approved):
Let's gooo!

PR #2983, opened 2025-09-29
chore: upgrade Zod to 4.1

nrainhart:
# Description

We want to upgrade to Zod 4.1 to get support for [codecs](https://zod.dev/codecs), so that we can have arbitrary classes with custom encoders inside Jazz schemas: https://github.com/garden-co/jazz/pull/2954.

## Caveats

Starting from 3.25/4.0, Zod does not allow mixing up schemas from different versions. If a user tries to use a Zod schema from a version prior to 4.1 inside a Jazz schema, they'll get the following type error:
```
The types of '_zod.version.minor' are incompatible between these types.
      Type '0' is not assignable to type '1'.
```

## Manual testing instructions

Tested in a project using Zod 3.25 to confirm the aren't any issues with multiple Zod versions (as long as schemas from the user's Zod library aren't used in Jazz schemas).

## Tests

- [x] Tests have not been updated, because: we're updating a library version

## Checklist

- [ ] I've updated the part of the docs that are affected the PR changes
- [x] I've generated a changeset, if a version bump is required
- [ ] I've updated the jsDoc comments to the public APIs I've modified, or added them when missing

---

> [!NOTE]
> Upgrade Zod to v4.1.11 in jazz-tools, align lockfile/peer deps, and remove Zod from example apps.
> 
> - **Dependencies**:
>   - Upgrade `zod` to `4.1.11` in `packages/jazz-tools` and update lockfile references.
>   - Align `better-auth` resolution to use `zod@4.1.11`.
>   - Remove `zod` dependency from `examples/chat` and `examples/multi-cursors`.
> - **Release**:
>   - Add changeset: `jazz-tools` patch (Upgrade Zod to 4.1).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7b0facc575a47c1a3c07d853b59867a8f0f4017e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #3004, opened 2025-10-06
fix: create-jazz-app not working with RN after Zod v4 upgrade

nrainhart:
# Description
After upgrading Zod from `3.25.76` to `4.1.11` in https://github.com/garden-co/jazz/pull/2983, React Native projects generated with `create-jazz-app` were not working due to an error in import resolution:

> Metro has encountered an error: While trying to resolve module zod/v4 from file /.../node_modules/jazz-tools/dist/index.js, the package /.../node_modules/zod/v4/package.json was successfully found. However, this package itself specifies a main module field that could not be resolved (/.../node_modules/zod/v4/index.cjs.

The issue was caused by [Zod 4.1.6](https://github.com/colinhacks/zod/pull/5222/files#diff-af75921c98100c36b0ba10db16be2bf8abd330024b5e63a9c219ffe4d9be5cceR130)  changing its `package.json` structure from [a single package.json file](https://github.com/colinhacks/zod/blob/463f03eb8183dcdcdf735b180f2bf40883e66220/packages/zod/package.json#L46) to one package.json stub file per Zod version.

Now each version has this `package.json` shape:
```javascript
{ 
  "type": "module",
  "main": "./index.cjs",
  "module": "./index.js",
  "types": "./index.d.cts" 
}
```

Since Metro tries to resolve the `main` entry point (see [resolverMainFields](https://metrobundler.dev/docs/configuration#resolvermainfields)), it was finding `/v4/index.d.cts` but couldn't include it in the bundle.

The solution was to update Metro's [sourceExts](https://metrobundler.dev/docs/configuration#sourceexts) to allow including .cjs files in the bundle.

Will run the pre-release starter and confirm the generated app is working before merging.

## Checklist

### DIFF
--- a/packages/create-jazz-app/src/index.ts
+++ b/packages/create-jazz-app/src/index.ts
@@ -307,7 +307,7 @@ async function scaffoldProject({
         const metroConfig = `const { getDefaultConfig } = require("expo/metro-config");
 const config = getDefaultConfig(__dirname);
 
-config.resolver.sourceExts = ["mjs", "js", "json", "ts", "tsx"];
+config.resolver.sourceExts = ["mjs", "js", "json", "ts", "tsx", "cjs"];
 config.resolver.requireCycleIgnorePatterns = [/(^|\\/|\\\\)node_modules($|\\/|\\\\)/];
 
 module.exports = config;`;
--- a/.changeset/dirty-seals-provide.md
+++ b/.changeset/dirty-seals-provide.md
@@ -1,0 +1,5 @@
+---
+"create-jazz-app": patch
+---
+
+fix: create-jazz-app not working with RN after Zod v4 upgrade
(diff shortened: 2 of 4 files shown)


packages/create-jazz-app/src/index.ts
```diff
@@ -307,7 +307,7 @@ async function scaffoldProject({
         const metroConfig = `const { getDefaultConfig } = require("expo/metro-config");
 const config = getDefaultConfig(__dirname);
 
-config.resolver.sourceExts = ["mjs", "js", "json", "ts", "tsx"];
```
nrainhart, line 310: Adding cjs sounds reasonable to me given that Metro handles both ES and CJS modules, but I'm no RN expert

aeplay (approved):
LGTM

PR #3006, opened 2025-10-07
Diffs in subscriptions

aeplay:
# Description

## Tests

## Checklist

### DIFF
--- a/packages/cojson/src/coValueCore/coValueCore.ts
+++ b/packages/cojson/src/coValueCore/coValueCore.ts
@@ -95,6 +95,12 @@ export type DecryptedTransaction = {
   tx: Transaction;
 };
 
+export type CoValueCoreCallback = (
+  core: CoValueCore,
+  unsub: () => void,
+  diffs: JsonValue[] | "initial",
+) => void;
+
 export type AvailableCoValueCore = CoValueCore & { verified: VerifiedState };
 
 export class CoValueCore {
@@ -132,8 +138,10 @@ export class CoValueCore {
 
   // cached state and listeners
   private _cachedContent?: RawCoValue;
-  readonly listeners: Set<(core: CoValueCore, unsub: () => void) => void> =
-    new Set();
+  readonly listeners: Map<
+    CoValueCoreCallback,
+    { diffs: boolean; pendingDiffs: JsonValue[] }
+  > = new Map();
   private counter: UpDownCounter;
 
   private constructor(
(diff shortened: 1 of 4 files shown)


packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -208,43 +216,34 @@ export class CoValueCore {
 
   waitForAvailableOrUnavailable(): Promise<CoValueCore> {
     return new Promise<CoValueCore>((resolve) => {
-      const listener = (core: CoValueCore) => {
+      this.subscribe((core, unsub) => {
```
aeplay, line 219: replacing manual adding/removing of subscriptions with just calling this.subscribe to have less duplicated logic (now that adding/removing subs is more complex)

packages/cojson/src/storage/knownState.ts
```diff
@@ -59,10 +59,13 @@ export class StorageKnownState {
...
-      const unsubscribe = coValue.subscribe((coValue) => {
-        req.knownState = coValue.knownState();
-        this.handleUpdate(id, this.getKnownState(id));
-      }, false);
+      const unsubscribe = coValue.subscribe(
+        (coValue) => {
+          req.knownState = coValue.knownState();
+          this.handleUpdate(id, this.getKnownState(id));
+        },
+        { skipInitial: true },
```
aeplay, line 67: just using the more explicit new option here

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -439,11 +438,14 @@ export class CoValueCore {
...
-          this.resetParsedTransactions();
-          this.scheduleNotifyUpdate();
-        }, false);
+        this.groupInvalidationSubscription = entry.subscribe(
+          (_groupUpdate) => {
+            // When the group is updated, we need to reset the cached content because the transactions validity might have changed
+            this.resetParsedTransactions();
+            this.scheduleNotifyUpdate();
+          },
+          { skipInitial: true },
```
aeplay, line 447: Just using the more explicit new option here

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -562,7 +564,14 @@ export class CoValueCore {
         "processNewTransactions" in this._cachedContent &&
         typeof this._cachedContent.processNewTransactions === "function"
       ) {
-        this._cachedContent.processNewTransactions();
+        const newDiffs = this._cachedContent.processNewTransactions();
```
aeplay, line 567: this is the core idea of this PR: the incremental `processNewTransactions()` of a CoValue view now has a chance to return semantic diffs corresponding to the new transactions, which are then forwarded to all listeners who are interested in diffs

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -610,27 +619,39 @@ export class CoValueCore {
...
-      listener(this, () => {
-        this.listeners.delete(listener);
-      });
+    if (!options?.skipInitial) {
+      listener(
+        this,
+        () => {
+          this.listeners.delete(listener);
+        },
+        "initial",
```
aeplay, line 653: this is just a marker - it's up to the CoValue view to translate the initial state into a corresponding diff (see CoMap implementation)

packages/cojson/src/coValues/coMap.ts
```diff
@@ -150,7 +152,19 @@ export class RawCoMapView<
 
     for (const [key, entries] of changedEntries.entries()) {
       this.latest[key] = entries[entries.length - 1];
+      const latestChange = this.latest[key]!.change;
+      if (latestChange.op === "set") {
```
nrainhart, line 156: I guess we're fine with always paying this cost of generating diffs when processing transactions, right?

The alternative would be to only start doing it if someone's subscribed to the diffs

nrainhart (approved):
Looks good! It's nice that CoMap diffs are enough for now, since they're the simplest

gdorsi (changes requested):
This approach doesn't consider the case of getting out-of-order transactions in `processNewTransactions` and requires the diffs being tracked even when the feature is off.

How about just tracking the knownState between emit and add methods to views to process the diff between two possible known states?

This way the diffing feature would be zero-cost when off, and it would be possible to provide the feature when the subscription is "resumed" (thinking about webhooks)

PR #3008, opened 2025-10-07
Closes #GCO-212 - Updates all pages for accessibility audits

mculotta120:
# Description

- Download chrome expansion for lighthouse, run accessibility scans. 
Or see screen grabs from audit results, after: 

Screenshot 2025-10-06 at 10 48 47 PM
Screenshot 2025-10-06 at 10 48 27 PM
Screenshot 2025-10-06 at 10 48 00 PM
Screenshot 2025-10-06 at 10 46 56 PM
Screenshot 2025-10-06 at 10 46 34 PM

## Tests

- [X] Tests have not been updated, because: tests not effected (we could add some testing tools for accessibility in the future, probably not necessary at this point)

## Checklist


homepage/design-system/src/components/atoms/Heading.tsx
```diff
@@ -28,7 +28,7 @@ export function Heading({
       {...props}
       className={clsx(
         "text-stone-950 dark:text-white font-display",
-        classes[size],
+        className ? className : classes[size],
```
joeinnes:
```suggestion
"text-stone-950 dark:text-white font-display",
 classes[size],
className,
```
`clsx` should automatically override any default styles with whatever comes later in the order (e.g. if `classes[size]` includes `text-stone-700` but `className` includes `text-stone-500`, `clsx` should resolve these down from ~~`text-stone-950`~~ -> ~~`text-stone-700`~~ -> `text-stone-500`. Should be enough just to add `className` at the end here.

homepage/design-system/src/components/atoms/Button.tsx
```diff
@@ -92,6 +93,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
           target={newTab ? "_blank" : undefined}
           className={classNames}
           onClick={onClick}
+          aria-label={type}
```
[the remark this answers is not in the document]
joeinnes: Screen readers should normally infer labels based on the text content of both buttons and links. 

I would vote for instead adding an optional `ariaLabel` prop that we can use for buttons/links without text content.

homepage/homepage/app/(others)/cloud/page.tsx
```diff
@@ -46,23 +35,28 @@ export default function Cloud() {
         <LatencyMap />
         <GappedGrid>
           <GridCard>
-            <H3>Optimal cloud routing</H3>
-
+            <H2 className="text-stone-950 dark:text-white font-display text-xl md:text-2xl mb-2 font-semibold tracking-tight">
```
joeinnes: Not for this PR, but as we go through the design system, I think ideally, we'd be able to do something like `<Heading level={2} styleAs={3}>Content</Heading>` to avoid this class soup. One to check with Vova I think (couldn't find his GH ID)

homepage/homepage/components/FileDownloadLink.tsx
```diff
@@ -23,7 +23,7 @@ export function FileDownloadLink(
       <Icon name="file" size="sm" className="mr-2" />
       {children}
 
-      <a href={href} download className="ml-12">
+      <a href={href} download className="ml-12" aria-label="download">
```
joeinnes: This shouldn't be necessary, 'Download' is text, so should be picked up by screen readers.

homepage/homepage/components/Pricing.tsx
```diff
@@ -132,8 +132,9 @@ export function Pricing() {
                 <span className="tabular-nums">20</span> GB egress/mo incl.{" "}
                 <span className="text-sm">(then $0.1 per GB)</span>
               </ListItem>
-              <hr className="my-2 border" />
-              <ListItem icon={LucideChevronsUp}>High-priority sync</ListItem>
+              <div aria-hidden="true" className="my-2 border-t-2">
+                <ListItem icon={LucideChevronsUp}>High-priority sync</ListItem>
+              </div>
```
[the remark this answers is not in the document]
joeinnes: +1. We can hide the hr, but I don't think we should hide the ListItem itself.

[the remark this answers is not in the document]
joeinnes: (Or actually same as you're doing below, insert an empty list item)

joeinnes (approved):
Thank you for working on this, I think it's super important!

homepage/homepage/components/Pricing.tsx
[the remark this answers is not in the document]
xxvis: +1

homepage/homepage/components/FileDownloadLink.tsx
xxvis: this `aria-label` is probably not needed since it simply repeats tame information

homepage/homepage/components/Pricing.tsx
```diff
@@ -176,7 +177,7 @@ export function Pricing() {
               </ListItem>
               <ListItem icon={LucideDatabase}>Custom storage</ListItem>
               <ListItem icon={LucideCloudDownload}>Custom egress/mo</ListItem>
-              <hr className="my-2 border" />
+              <li aria-hidden="true" className="my-2 border-t-2 list-none" />
               <ListItem icon={LucideHandshake}>
                 Rapid integration & premium onboarding
```
xxvis: minor: but its better to encode the ampersand `&amp;`

homepage/design-system/src/components/organisms/Footer.tsx
```diff
@@ -36,7 +36,11 @@ export function Footer({
       <div className="container grid gap-8 md:gap-12">
         <div className="grid grid-cols-12 gap-y-3 sm:items-center pb-8 border-b">
           <div className="col-span-full sm:col-span-6 md:col-span-8">
-            <Link href="https://garden.co" target="_blank">
+            <Link
+              href="https://garden.co"
+              target="_blank"
+              aria-label="Garden Computing Homepage"
+            >
               {logo}
```
xxvis, line 40: Something like this where you want to supply additional text would better be handled using tailwind `sr-only` class. 
So just below the {logo} you can add `<span className="sr-only">Garden Computing Homepage</span>`

homepage/design-system/src/components/atoms/Heading.tsx
mculotta120: I was having trouble overwriting the Tailwind responsive size classes. I think filtering out the default size classes when a custom className is provided should resolve the issue. Let me know if you think this isn’t the right approach.

homepage/homepage/app/(others)/cloud/page.tsx
mculotta120: I agree, I’m hoping my change to the Heading component works better for now. Admittedly, I’m still getting used to some of the nuances of Tailwind. I’d definitely love to improve the styling functionality while still maintaining accessibility best practices.

mculotta120:
@joeinnes @xxvis - I think it's ready for another look, see what you think of the change for the Heading component. I ran another audit on /docs and I had a different topic selected. As a result, the sidenav was flagged for insufficient clickable area. I need to implement a solution that won't cause problems with the nested links there. I'm out tomorrow/thursday. LMK if you have any suggestions. We could probably follow up with another run through (separate issue/ phase 2). I think I need to do an audit of just docs for accessibility anyway, I noticed some of the heading orders were off in one mdx file, I'm guessing there are probably others.

homepage/design-system/src/components/atoms/Heading.tsx
joeinnes: Ah yeah, the responsive styles would need to be overridden at every breakpoint, and it's not immediately clear to a consumer of the component which styles they'd need to override. Maybe we just throw away all of the styling if the user specifies a className? What do you reckon @xxvis?

homepage/homepage/app/(others)/cloud/page.tsx
xxvis: I agree @joeinnes in the past this flexibility of setting a specific heading level but styled as some other heading has come across in just about every project I've ever worked on. So makes sense to extend this functionality in the future

homepage/design-system/src/components/atoms/Heading.tsx
xxvis: @joeinnes I think thats a reasonable take until we create a <Heading/> specific component that allows you to set levels and styles independently

xxvis:
I'm happy to tackle more of these accessibility related things in the future as well.

joeinnes:
> the sidenav was flagged for insufficient clickable area

this is my fault, because I'm doing something silly here: https://github.com/garden-co/jazz/blob/192ebdeb8947ddd4531abaffb276be4c6daf7b7c/homepage/homepage/components/SideNavSection.tsx#L60

I think ideally we should not support these triggers being links, and instead they should be single-function elements which expand and collapse the disclosure. I'll put in another PR for the nav in the next few days which fixes the actual content structure (which means checking links too) and I'll fix this with an 'overview' link in the same way we currently have for Authentication.

Once that's done, the whole `<summary>` element can be the tap target, and that should be big enough I think.

PR #3020, opened 2025-10-08
Fix macOS trackpad scrollbar bug

joeinnes:
# Description
macOS has this weird behaviour where you cannot control overlay scrollbars, they always overlay the content.

This PR adds checks to see if the nav menu's scrollbar is 0 (indicating we have overlay scrollbars), and if so, adds additional right padding so macOS doesn't layer scrollbars over content.

Unplug your mouse, use a browser with overlay scrollbars (Firefox for example), and check if the nav content is obscured. (I actually couldn't quite get it to overlay the content exactly, but this is not comfortable padding with the summary's marker.

Before:
Screenshot 2025-10-08 at 17 21 14

After:
Screenshot 2025-10-08 at 17 22 40

## Tests

- [ ] Tests have not been updated, because: N/A

## Checklist

### DIFF
--- a/homepage/homepage/components/SideNav.tsx
+++ b/homepage/homepage/components/SideNav.tsx
@@ -1,8 +1,9 @@
+"use client";
 import { SideNavItem } from "@/components/SideNavItem";
 import { DoneStatus } from "@/content/docs/docNavigationItemsTypes";
 import { clsx } from "clsx";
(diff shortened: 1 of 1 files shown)


nrainhart (approved):
Looks amazing, thanks Joe!

PR #3030, opened 2025-10-09
fix: reveal child groups' writeOnly keys to parent groups

nrainhart:
# Description

When adding a `writeOnly` key to a group that had another group as a member, the `writeOnly` key was not being revealed to the parent group, which meant members of the parent group were not able to read writes performed by members with `writeOnly` permission.

Since https://github.com/garden-co/jazz/pull/2290 made it possible to extend a group without having permissions over it by reusing `writeOnly` keys, this bug also affected active accounts extending a group they don't have permissions over. In this case, members of _grandparent_ groups could not see the changes made in the child groups.

This PR:
- reveals new `writeOnly` keys to parent groups when adding a new member with `writeOnly` permission
- reveals ALL existing `writeOnly` keys in a child group when adding a new parent group as a member
- reveals ALL new `writeOnly` keys in a child group to parent groups when rotating keys

I tested the repro script provided by @tobiaslins and confirmed that now all accounts have access to the CoValues owned by the client group.

## Tests

## Checklist

### DIFF
--- a/.changeset/spicy-chefs-retire.md
+++ b/.changeset/spicy-chefs-retire.md
@@ -1,0 +1,5 @@
+---
+"cojson": patch
+---
+
+fix: reveal child groups' writeOnly keys to parent groups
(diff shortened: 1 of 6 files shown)


gdorsi (approved):
LGTM

PR #3038, opened 2025-10-10
Add "Performance tips" section

joeinnes:
# Description
Adds a "performance tips" to the docs.

N/A

## Tests

- [X] Tests have not been updated, because: Docs only

## Checklist

### DIFF
--- a/homepage/homepage/content/docs/performance.mdx
+++ b/homepage/homepage/content/docs/performance.mdx
@@ -1,0 +1,38 @@
+export const metadata = {
+  description: "Advanced performance tips for using Jazz."
+};
+
+# Tips for maximising Jazz performance
+
+## Use the best crypto implementation for your platform
+The fastest implementations are (in order):
+
+1. [Node-API crypto](/docs/server-side/setup#node-api) (only available in some Node/Deno environments) or [RNQuickCrypto on React Native](/docs/react-native/project-setup/providers#quick-crypto)
+2. [WASM crypto](/docs/server-side/setup#wasm-on-edge-runtimes) (default)
+3. JavaScript fallback (slowest, but most compatible)
+
+## Minimise group extensions
+Group extensions are exceptionally powerful for cascading permissions and are performant enough for the vast majority of use cases. However, if huge numbers of parent groups need to be loaded as part of the dependency chain, there can be a performance impact. 
+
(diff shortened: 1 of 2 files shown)


homepage/homepage/content/docs/performance.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+## Use the best crypto implementation for your platform
+
+The fastest implementations are (in order):
+
+1. [Node-API crypto](/docs/server-side/setup#node-api) (only available in some Node/Deno environments) or [RNQuickCrypto on React Native](/docs/react-native/project-setup/providers#quick-crypto)
+2. [WASM crypto](/docs/server-side/setup#wasm-on-edge-runtimes)
+3. JavaScript fallback (slowest, but most compatible)
+
+Check particularly if you are running Jazz in an environment which supports Node-API, or if you're using an edge runtime which may not enable WASM out-of-the-box.
```
mculotta120:
```suggestion
Check whether your environment supports Node-API.
Some edge runtimes may not enable WASM by default.
```

Nitpick: The language could be simplified a bit here to help with scannability.

homepage/homepage/content/docs/performance.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+## Minimise group extensions
+
+Group extensions are very powerful for cascading permissions and are performant enough for the vast majority of use cases. However, if huge numbers of parent groups need to be loaded as part of the dependency chain, there can be a performance impact. 
+
+You can optimise performance by manually creating groups, reusing groups where the permissions will be the same on both CoValues throughout the lifetime of the CoValue.
```
mculotta120:
```suggestion
Group extensions make it easy to cascade permissions and they’re fast enough for most cases.
However, performance can slow down when many parent groups need to load in the dependency chain.
To avoid this, create and reuse groups manually when their permissions stay the same for both CoValues over time.
```
\The original is clear, but I wondered if shorter sentences might make it easier for readers (especially non-native English speakers, or those using translation tools) to follow.
Totally optional, just sharing in case it feels like an improvement :)

mculotta120 (commented):
Looks great! just a couple of optional simplifications

joeinnes:
Thanks!

mculotta120 (approved):
🥇 looks great!

PR #3040, opened 2025-10-10
fix: validate write permissions against merge time

gdorsi:
Refactored the merge operation in order to make it possible to validate write permissions against merge time instead off referring to the original transaction time.

This fixes the case where account that does the merge has got write access after some of the changes happened in the branch.

This PR preserves the conflict management in order to keep the idempotency property of the merge action, which we rely on in CoList.

We could apply a different conflict management (and even make it customizable) on CoMap, as the strategy is applied in the `compareTransactions` to make more intuitive.

### DIFF
--- a/packages/cojson/src/coValueCore/branching.ts
+++ b/packages/cojson/src/coValueCore/branching.ts
@@ -27,7 +27,7 @@ export type BranchPointerCommit = {
  */
 export type MergedTransactionMetadata = {
   mi: number; // Transaction index and marker of a merge commit
-  t?: number;
+  t?: number; // The difference between the current time and the madeAt value of the original transaction. Used to calculate the original madeAt of the transaction, stored this way to reduce the size of the meta information.
   s?: SessionID;
   b?: RawCoID;
 };
--- a/packages/cojson/src/coValueCore/branching.ts
+++ b/packages/cojson/src/coValueCore/branching.ts
@@ -27,6 +27,7 @@ export type BranchPointerCommit = {
  */
 export type MergedTransactionMetadata = {
   mi: number; // Transaction index and marker of a merge commit
+  t?: number;
   s?: SessionID;
   b?: RawCoID;
 };
@@ -242,11 +243,20 @@ export function mergeBranch(branch: CoValueCore): CoValueCore {
   // To reduce the cost of the meta we skip the repeated information
   let lastSessionId: string | undefined = undefined;
   let lastBranchId: string | undefined = undefined;
+  let lastOriginalMadeAt: number = 0;
+
+  const now = Date.now();
 
   for (const tx of branchValidTransactions) {
     const mergeMeta: MergedTransactionMetadata = {
       mi: tx.txID.txIndex,
     };
+
+    // We compress the original made at, to reduce the size of the meta information
+    if (tx.madeAt !== lastOriginalMadeAt) {
+      // Storing the diff with madeAt to consume less bytes in the meta information
+      mergeMeta.t = now - tx.madeAt;
+    }
 
     if (lastSessionId !== tx.txID.sessionID) {
       mergeMeta.s = tx.txID.sessionID;
@@ -256,9 +266,10 @@ export function mergeBranch(branch: CoValueCore): CoValueCore {
       mergeMeta.b = tx.txID.branch;
     }
 
-    target.makeTransaction(tx.changes, tx.tx.privacy, mergeMeta, tx.madeAt);
+    target.makeTransaction(tx.changes, tx.tx.privacy, mergeMeta, now);
     lastSessionId = tx.txID.sessionID;
     lastBranchId = tx.txID.branch;
+    lastOriginalMadeAt = tx.madeAt;
   }
 
   // Track the merged transactions for the branch, so future merges will know which transactions have already been merged
--- a/packages/cojson/src/coValueCore/coValueCore.ts
+++ b/packages/cojson/src/coValueCore/coValueCore.ts
@@ -63,13 +63,13 @@ export class VerifiedTransaction {
   author: RawAccountID | AgentID;
   // An object containing the session ID and the transaction index
   originalTxID: TransactionID;
-  // If this is a merged transaction, the TxID of the source transaction
-  sourceTxID: TransactionID | undefined;
+  // If this is a merged transaction, the TxID of the merged transaction
+  mergedTxID: TransactionID | undefined;
   tx: Transaction;
   // The Unix time when the transaction was made
   originalMadeAt: number;
-  // If this is a merged transaction, the madeAt of the source transaction
-  sourceTransactionMadeAt: number | undefined;
+  // If this is a merged transaction, the madeAt of the merged transaction
+  mergedTransactionMadeAt: number | undefined;
   // Whether the transaction has been validated, used to track if determinedValidTransactions needs to check this
   isValidated: boolean;
   // The decoded changes of the transaction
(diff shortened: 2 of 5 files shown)


packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -920,14 +926,20 @@ export class CoValueCore {
...
-      const previousTransaction = transaction.previous?.txID;
-      const sessionID = meta.s ?? previousTransaction?.sessionID;
+      const previousTransaction = transaction.previous;
+      const sessionID = meta.s ?? previousTransaction?.txID.sessionID;
+
+      if (meta.t) {
+        transaction.madeAt = transaction.madeAt - meta.t;
+      } else if (previousTransaction) {
+        transaction.madeAt = previousTransaction.madeAt;
+      }
```
gdorsi, line 973: Overriding the `madeAt` property, this way it will show in the edits and used in the transaction sorting everywhere.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -998,7 +1010,8 @@ export class CoValueCore {
 
       options?.knownTransactions?.add(transaction.tx);
 
-      const { txID } = transaction;
+      // Using the orginalTxID to filter the transactions, because the TxID is modified by the merge meta
+      const txID = transaction.orginalTxID;
```
gdorsi: Noticed that the merged transactions were showing up in the branch after the merge.

The problem was that the txID is the one of the branch, so it was passing through the `to` filter.
Fixed by using the original txID instead of the one we use for the CoValue views.

packages/cojson/src/permissions.ts
```diff
@@ -220,13 +218,10 @@ function determineValidTransactionsForGroup(
   initialAdmin: RawAccountID | AgentID,
   extendChain?: Set<CoValueCore["id"]>,
 ): { memberState: MemberState } {
-  coValue.verifiedTransactions.sort((a, b) => {
-    return a.madeAt - b.madeAt;
-  });
+  coValue.verifiedTransactions.sort(coValue.compareTransactions);
```
gdorsi, line 223: The sorting should always use `compareTransactions`:
- it uses the sessionID to fix some edge cases
- we might add extra logic there in the future

packages/cojson/src/permissions.ts
```diff
@@ -220,13 +218,10 @@ function determineValidTransactionsForGroup(
...
 
   const memberState: MemberState = {};
   const writeOnlyKeys: Record<RawAccountID | AgentID, KeyID> = {};
-  const validTransactions: ValidTransactionsResult[] = [];
```
gdorsi, line 229: Unused variable

packages/cojson/src/coValueCore/branching.ts
```diff
@@ -27,6 +27,7 @@ export type BranchPointerCommit = {
  */
 export type MergedTransactionMetadata = {
   mi: number; // Transaction index and marker of a merge commit
+  t?: number;
```
nrainhart: Mind adding a comment of what `t` represents?

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -63,9 +63,12 @@ export type VerifiedTransaction = {
   author: RawAccountID | AgentID;
   // An object containing the session ID and the transaction index
   txID: TransactionID;
+  // The unmodified TxID that refers to the current position in the session map
+  orginalTxID: TransactionID;
   tx: Transaction;
   // The Unix time when the transaction was made
   madeAt: number;
+  originalMadeAt: number;
```
nrainhart: Same thing here

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -63,9 +63,12 @@ export type VerifiedTransaction = {
   author: RawAccountID | AgentID;
   // An object containing the session ID and the transaction index
   txID: TransactionID;
+  // The unmodified TxID that refers to the current position in the session map
+  orginalTxID: TransactionID;
```
```suggestion
  originalTxID: TransactionID;
```

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -63,9 +63,12 @@ export type VerifiedTransaction = {
   author: RawAccountID | AgentID;
   // An object containing the session ID and the transaction index
   txID: TransactionID;
+  // The unmodified TxID that refers to the current position in the session map
```
nrainhart: I'd add a bit more context to make it clear this is relevant for branching:
```suggestion
  // The unmodified TxID that refers to the current position in the session map.
  // When merging branches, the txID will differ from the originalTxId.
```

nrainhart (approved):
Even with the PR description's context, I found the "write permissions are validated against the time the branch was merged" idea hard to grasp by looking at the code.

I think this can be improved by having better comments around the pieces of state involved in this check

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -58,15 +58,19 @@ export function idforHeader(
...
+  originalTxID: TransactionID;
+  // If this is a merged transaction, the TxID of the merged transaction
+  mergedTxID: TransactionID | undefined;
   tx: Transaction;
   // The Unix time when the transaction was made
-  madeAt: number;
-  // Whether the transaction has been validated, used to track if determinedValidTransactions needs to be check this
+  originalMadeAt: number;
+  // If this is a merged transaction, the madeAt of the merged transaction
+  mergedTransactionMadeAt: number | undefined;
```
nrainhart: I find the "original" and "merged" names confusing. "original" to me suggest "the time at which the change was originally made in the branch", while "merged" is "the time at which the change was merged".

How do you feel about going with `madeAt` and `sourceTxMadeAt` instead?

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -86,13 +90,73 @@ export type VerifiedTransaction = {
...
+  }
+
+  // The TxID that refers to the current position in the session map
+  // If this is a merged transaction, the txID is the TxID of the merged transaction
+  get txID() {
+    return this.mergedTxID ?? this.originalTxID;
+  }
+
+  // The madeAt that refers to the time when the transaction was made
+  // If this is a merged transaction, the madeAt is the time when the transaction has been merged
```
nrainhart: Shouldn't this be
```suggestion
  // If this is a merged transaction, the madeAt is the time when the transaction has been made in the branch
```
?

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -86,13 +90,73 @@ export type VerifiedTransaction = {
...
+  get madeAt() {
+    if (this.mergedTransactionMadeAt) {
+      return this.mergedTransactionMadeAt;
+    }
+
+    return this.originalMadeAt;
```
nrainhart: Nit
```suggestion
    return this.mergedTransactionMadeAt ?? this.originalMadeAt;
```

nrainhart (approved):
Thanks for iterating through this! I like this final version way better. Left a few minor comments

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: Agree in the naming confusion.

`madeAt` is being used for the getter property so that won't do

How about `currentMadeAt`?

nrainhart: currentMadeAt is fine. Alternatively, maybe we can go with `txMadeAt`?

PR #3046, opened 2025-10-13
fix: enable branch creation for accounts with reader access

gdorsi:
Add a special permission rule to allow users with reader access to create branch pointers when creating a private branch, and block the branch creation to users without enough access rights.

This prevents situations when users would create bugged branches, switching to:
- a full featured branch for accounts that have reader access and are creating a private branch
- return the source value when the access rights are not enough to create a proper branch

I had to refactor a big the decoding pipeline, moving the decode of trusting transactions to the ValidTransaction class and specializing the decoding step into a decrypt step.


packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -70,24 +71,18 @@ export class VerifiedTransaction {
...
   // The decoded meta information of the transaction
   meta: JsonObject | undefined;
-
   // Whether the transaction is valid, as per membership rules
-  isValid: boolean;
-
-  // True if we encountered an error while decoding the changes
-  hasInvalidChanges: boolean;
-  // True if we encountered an error while parsing the meta
-  hasInvalidMeta: boolean;
```
gdorsi, line 86: Simplified these two into a single `isDecrypted`

packages/cojson/src/coValueCore/decryptTransactionChangesAndMeta.ts
```diff
@@ -0,0 +1,55 @@
...
+export function decryptTransactionChangesAndMeta(
+  coValue: AvailableCoValueCore,
+  transaction: VerifiedTransaction,
+  ignorePrivateTransactions: boolean,
+) {
+  if (
+    !transaction.isValid ||
+    transaction.isDecrypted ||
+    transaction.tx.privacy !== "private" ||
+    ignorePrivateTransactions
```
nrainhart: Shouldn't this be
```suggestion
    transaction.tx.privacy === "private" && ignorePrivateTransactions
```
? I'm surprised no tests are breaking

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -115,20 +110,21 @@ export class VerifiedTransaction {
...
-
-    this.isValid = false;
-    this.hasInvalidChanges = false;
-    this.hasInvalidMeta = false;
-    this.hasMetaBeenParsed = false;
     this.previous = previous;
+
+    // Decoding the trusting transactions here because they might be useful in the permissions checks
+    if (this.tx.privacy === "trusting") {
+      this.changes = safeParseJSON(this.tx.changes);
```
nrainhart: We were previously checking that changes are meta are successfully parsed. Are these checks no longer necessary?
```suggestion
      const changes = safeParseJSON(this.tx.changes);
      if (changes) {
        this.changes = changes;
      }
```

packages/cojson/src/coValueCore/decryptTransactionChangesAndMeta.ts
```diff
@@ -0,0 +1,55 @@
...
+      transaction.txID.txIndex,
+      readKey,
+    );
+
+    if (meta) {
+      transaction.meta = meta;
+    }
+  }
+
+  transaction.isDecrypted = true;
```
nrainhart, line 55: We mark the transaction as decrypted even if the result was undefined?

packages/cojson/src/coValueCore/branching.ts
```diff
@@ -143,6 +143,17 @@ export function createBranch(
     return coValue;
   }
 
+  const myRole = coValue.safeGetGroup()?.myRole();
+
+  // We allow branch creation to accounts with at least read access to the source group
+  if (!myRole || (myRole === "reader" && !ownerId)) {
+    return coValue;
```
nrainhart, line 155: I feel like it'd be better to throw an error rather than do nothing. A user may think that the covalue was successfully branched and make changes to the original covalue without noticing

packages/cojson/src/tests/branching.test.ts
```diff
@@ -528,6 +528,118 @@ describe("Branching Logic", () => {
...
+        originalTxID,
+      );
+      expect(
+        validSortedTransactions[mergedTransactionIdx + 1]?.txID.branch,
+      ).toBe(undefined);
+    });
+  });
+
+  describe("Branching permissions", () => {
+    test("should allow the creation of private branches to accounts with read access to the source group", async () => {
```
nrainhart, line 577: Should we add another test that checks that an account with read access than then gets write access (or another account with write access) can merge this branch?

nrainhart (commented):
The overall approach makes sense, but I've got a few questions

packages/cojson/src/coValueCore/branching.ts
gdorsi: My rationale about this choice is that, throwing in the middle of the checkout of multiple values might make the API harder to use.

Also, an account with reader access can't do changes anyway, so there would not be much more difference by returning the source value instead of a branch.

The only sensible difference is that branching stops the branch from reiceving updates from the source value, in this case it won't happen.

Maybe we could return a warning log, but I think throwing here would be to much.

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: That was done to avoid continuing parsing a broken JSON on each decode step.

Now that this is done in the transaction constructor it is no longer necessary.

packages/cojson/src/coValueCore/decryptTransactionChangesAndMeta.ts
gdorsi: It is correct as we don't decrypt trusting transactions as I moved the decoding in the VerifiedTransaction constructor.

Going to refactor the check to make it more clear.

packages/cojson/src/coValueCore/decryptTransactionChangesAndMeta.ts
gdorsi: Yes, if we've got the readKey it is meaningless to try to decrypt things more than once even if the decrypt fails.

packages/cojson/src/tests/branching.test.ts
gdorsi: Good point, added!

packages/cojson/src/coValueCore/decryptTransactionChangesAndMeta.ts
gdorsi: Added a comment to explain the design, thanks for the question!

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -121,14 +116,21 @@ export class VerifiedTransaction {
...
+    if (parsingCache) {
+      this.changes = parsingCache.changes;
+      this.meta = parsingCache.meta;
+    } else {
+      // Decoding the trusting transactions here because they might be useful in the permissions checks
+      if (this.tx.privacy === "trusting") {
+        this.changes = safeParseJSON(this.tx.changes);
+
+        if (this.tx.meta) {
+          this.meta = safeParseJSON(this.tx.meta);
```
nrainhart, line 130: Shouldn't we set `hasMetaBeenParsed` here?

nrainhart: Same thing if we get a parsing cache hit

packages/cojson/src/tests/branching.test.ts
```diff
@@ -528,6 +528,157 @@ describe("Branching Logic", () => {
...
+
+      expect(branch.get("key")).toBe("bob");
+
+      const branchOnAlice = await loadCoValueOrFail(alice.node, branch.id);
+      branchOnAlice.core.mergeBranch();
+
+      expect(map.get("key")).toBe("bob");
+    });
+
+    test("should not allow the creation of branches to accounts with read access", async () => {
```
nrainhart: Nit: my brain glitched for a second until I reminded we'd already tested private branch creation
```suggestion
    test("should not allow the creation of public branches to accounts with read access", async () => {
```

packages/cojson/src/coValueCore/branching.ts
nrainhart: +1 to at least show a warning

nrainhart (approved):
Thanks for addressing my comments! Have a few more comments, but this looks good to go once they're addressed

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: `hasMetaBeenParsed` is for the `parseMeta` step

We don't need a flag here because we try to decode meta once.

Should I rename the `hasMetaBeenParsed` flag into something else?

PR #3047, opened 2025-10-14
Add .md route to all docs pages

joeinnes:
# Description
When autogenerating the llms.txt file, we also now output MD files (compiled from MDX), which can be accessed by appending '.md' to the URL of any existing docs page.

These are output to the /public folder to avoid needing to compile them, which does result in some duplication at the moment to ensure we've got the right files for each framework

1. Navigate to a docs page
2. Add '.md' at the end of the URL
3. Check that the markdown accurately represents the content that was hosted at that URL
4. Check the framework specific llms-full.txt available at eg jazz.tools/svelte/llms-full.txt

## Tests

- [X] Tests have not been updated, because: Docs only

## Checklist

### DIFF
--- a/homepage/homepage/.gitignore
+++ b/homepage/homepage/.gitignore
@@ -42,3 +42,4 @@ codeSamples
 
 # LLM docs
 public/llms*.txt
+public/docs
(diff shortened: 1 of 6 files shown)


mculotta120 (approved):
Working great!

PR #3050, opened 2025-10-14
feat: introducing manager role

Takeno:
# Description
Replaces https://github.com/garden-co/jazz/pull/2995

This PR introduces the "manager" role. Like admins, it can invite readers, writers, writerOnly. They can also remove them. They can't remove or invite other managers. Managers can be invited and removed by admins.

---

This PR also introduces a refactoring on addMember/removeMember to validate permissions at a lower level (inside permission.ts) 

Previously, cojson/group.ts was trying to understand why an operation failed. Now, if the role setting fails, the error message is unified and enriched by useful information for debugging: `Failed to set role ${role} to ${memberKey} (role of current account is ${this.myRole()})`.

To be consistent and to give appropriate feedback, `group.removeMember` now throws an error if it fails (it was failing silently before). Due to this change, this might be considered a minor update instead of a patch.

## TODO:
- [x] generate test coverage to check uncovered branches
- [x] add some tests on group extension
- [ ] write docs when PR is confirmed /cc @joeinnes 

## Tests

## Checklist

### DIFF
--- a/.changeset/stupid-ducks-try.md
+++ b/.changeset/stupid-ducks-try.md
@@ -1,0 +1,6 @@
+---
+"jazz-tools": minor
+"cojson": minor
+---
+
+feat: introduced new "manager" role who can invite and remove members, and being removed by admins. Group.removeMember() now throws in case of invalid permissions.
--- a/.changeset/stupid-ducks-try.md
+++ b/.changeset/stupid-ducks-try.md
@@ -1,6 +1,6 @@
 ---
-"jazz-tools": minor
-"cojson": minor
+"jazz-tools": patch
+"cojson": patch
 ---
 
-feat: introduced new "manager" role who can invite and remove members, and being removed by admins. Group.removeMember() now throws in case of invalid permissions.
+feat: introduced new "manager" role who can invite and remove members, and being removed by admins.
--- a/packages/jazz-tools/src/tools/coValues/account.ts
+++ b/packages/jazz-tools/src/tools/coValues/account.ts
@@ -196,7 +196,7 @@ export class Account extends CoValueBase implements CoValue {
     );
   }
 
-  canAdmin(value: CoValue): boolean {
+  canManage(value: CoValue): boolean {
     const valueOwner = value.$jazz.owner;
     if (!valueOwner) {
       if (value[TypeSym] === "Group") {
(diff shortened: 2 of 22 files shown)


.changeset/stupid-ducks-try.md
```diff
@@ -0,0 +1,6 @@
+---
+"jazz-tools": minor
```
gdorsi: Why is this a minor?

gdorsi (approved):
Great job!

Do we need to update the docs?

.changeset/stupid-ducks-try.md
Takeno: > This PR also introduces a refactoring on addMember/removeMember to validate permissions at a lower level (inside permission.ts)
> 
> Previously, cojson/group.ts was trying to understand why an operation failed. Now, if the role setting fails, the error message is unified and enriched by useful information for debugging: `Failed to set role ${role} to ${memberKey} (role of current account is ${this.myRole()}).`
> 
> To be consistent and to give appropriate feedback, group.removeMember now throws an error if it fails (it was failing silently before). Due to this change, this might be considered a minor update instead of a patch.

packages/jazz-tools/src/tools/coValues/account.ts
```diff
@@ -188,22 +188,31 @@ export class Account extends CoValueBase implements CoValue {
...
         return value.$jazz.id === this.$jazz.id;
       }
       return false;
     }
-    return valueOwner.getRoleOf(this.$jazz.id) === "admin";
+
+    return (
+      valueOwner.getRoleOf(this.$jazz.id) === "admin" ||
+      valueOwner.getRoleOf(this.$jazz.id) === "manager"
+    );
```
gdorsi, line 215: hmmm, shall we add a specialized `canManage`?

Mangers can't add other managers, and this difference can be useful in some UIs

gdorsi (commented):
Would it be possible to test the new manager role in an example?

I think that the music player would be an interesting one to use as testing ground.

.changeset/stupid-ducks-try.md
gdorsi: We could make it an error log and upgrade it into a throw new Error in the next minor.

This way we can ship the manager role without breaking changes.

Takeno:
immagine

Added to the music-player

packages/jazz-tools/src/tools/coValues/account.ts
Takeno: `canManage` added https://github.com/garden-co/jazz/pull/3050/commits/94bbec440f3b6e14758f7054ad913a7841ab4eee#diff-d17ec25df0b8c9feddc7dcaa0cc3dd3b2ef8a0d73512811069d5fd92fc3b91aeL199

.changeset/stupid-ducks-try.md
Takeno: The throw has been removed and now it is a patch upgrade. I will reintroduce it later as minor.

PR #3052, opened 2025-10-14
fix(permissions): fix validation for self-revocation in groups

gdorsi:
Fixed the `isSelfRevocation` check in permissions.

Before the validation was done against the current account id, and so passing when validated by the account that removed themselves and and failing when validated by other accounts.


Takeno (approved):
Nice catch!

PR #3056, opened 2025-10-15
feat: Request authentication APIs

Elfo404:
# Description
Add `generatateAuthToken` and `authenticateRequest` functions to sign and authenticate requests.

`authenticateRequest` only performs basic authentication when the token is provided, and only fails when a provided token is invalid or expired. It's up to the user to then do authorization or restrict access to resources based on the result.

Fixes https://linear.app/garden-co/issue/GCO-888/feature-request-auth-token-api-for-user-identification

## Tests

- [X] Tests have been added and/or updated

## Checklist

- [X] I've updated the part of the docs that are affected the PR changes
- [X] I've generated a changeset, if a version bump is required
- [X] I've updated the jsDoc comments to the public APIs I've modified, or added them when missing


packages/jazz-tools/src/tools/tests/authenticate-request.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    const me = await createJazzTestAccount({
+      isCurrentActiveAccount: true,
+    });
+
+    const token = generateAuthToken();
+
+    const { account, error } = await authenticateRequest(
+      new Request("https://api.example.com/api/user", {
+        headers: {
+          Authorization: `Jazz ${token}`,
```
gdorsi, line 20: What does happen when the Jazz prefix is missing?

packages/jazz-tools/src/tools/coValues/request.ts
```diff
@@ -631,3 +634,173 @@ async function loadWorkerAccountOrGroup(id: string, loadAs: Account) {
...
+export async function authenticateRequest(
+  request: Request,
+  options?: {
+    expiration?: number;
+    loadAs?: Account;
+    getToken?: (request: Request) => string | undefined | null;
+  },
+): Promise<
+  | {
+      account?: Account;
```
gdorsi, line 688: Isn't account always set if error is missing?

gdorsi (approved):
Good job!

packages/jazz-tools/src/tools/coValues/request.ts
Elfo404: no, if i do a fetch without specifying a token, the authentication doesn't fail, there's simply no user to authenticate, thus no error in authenticating the request - but also no account.

packages/jazz-tools/src/tools/tests/authenticate-request.test.ts
Elfo404: good catch, thanks!

changed behavior in a7502f14a and added a test for it making the authentication scheme required (when using defaults)

PR #3057, opened 2025-10-15
fix: createImage return type

Takeno:
# Description
The `createImage` return type was not coherent with the implementation. Additionally, the `$each: true` blocks `original` property to be considered already loaded.

Now the return types are synced, and `original` (the only public filestream reference) is marked as loaded. 

## Tests

## Checklist


nrainhart (approved):
Thanks for the quick fix @Takeno! 🫶

PR #3060, opened 2025-10-15
feat: centralize knownState mutators and types

gdorsi:
# Description

Done a bit of cleanup, to put most of the knownState manipulation logic inside a `knownState` module and making more intentional and consistent where we do mutate the knownState and where we do create a new object on change.

Also simplified and optimized a little the straming state tracking.

Used this refactoring to cover the knownState modifiers with tests.

## Tests

## Checklist

- [ ] I've updated the part of the docs that are affected the PR changes (not applicable)
- [ ] I've generated a changeset, if a version bump is required (not applicable)

### DIFF
--- a/packages/cojson/src/PeerKnownStates.ts
+++ b/packages/cojson/src/PeerKnownStates.ts
@@ -1,41 +1,51 @@
 import { RawCoID, SessionID } from "./ids.js";
 import {
+  cloneKnownState,
+  combineKnownStates,
   CoValueKnownState,
-  combinedKnownStates,
   emptyKnownState,
-} from "./sync.js";
+  updateSessionCounter,
+} from "./knownState.js";
 
 export class PeerKnownStates {
   private coValues = new Map<RawCoID, CoValueKnownState>();
 
+  getKnownState(id: RawCoID) {
+    const knownState = this.coValues.get(id);
+
+    if (!knownState) {
+      const knownState = emptyKnownState(id);
+      this.coValues.set(id, knownState);
+      return knownState;
+    }
+
+    return knownState;
+  }
+
   updateHeader(id: RawCoID, header: boolean) {
-    const knownState = this.coValues.get(id) ?? emptyKnownState(id);
+    const knownState = this.getKnownState(id);
     knownState.header = header;
-    this.coValues.set(id, knownState);
-    this.triggerUpdate(id);
+    this.triggerUpdate(id, knownState);
   }
 
   combineWith(id: RawCoID, value: CoValueKnownState) {
-    const knownState = this.coValues.get(id) ?? emptyKnownState(id);
-    this.coValues.set(id, combinedKnownStates(knownState, value));
-    this.triggerUpdate(id);
+    const knownState = this.getKnownState(id);
+    combineKnownStates(knownState, value);
+    this.triggerUpdate(id, knownState);
   }
 
   updateSessionCounter(id: RawCoID, sessionId: SessionID, value: number) {
-    const knownState = this.coValues.get(id) ?? emptyKnownState(id);
-    const currentValue = knownState.sessions[sessionId] || 0;
-    knownState.sessions[sessionId] = Math.max(currentValue, value);
+    const knownState = this.getKnownState(id);
+    updateSessionCounter(knownState.sessions, sessionId, value);
 
-    this.coValues.set(id, knownState);
-    this.triggerUpdate(id);
+    this.triggerUpdate(id, knownState);
   }
 
-  set(id: RawCoID, knownState: CoValueKnownState | "empty") {
-    this.coValues.set(
-      id,
-      knownState === "empty" ? emptyKnownState(id) : knownState,
-    );
-    this.triggerUpdate(id);
+  set(id: RawCoID, payload: CoValueKnownState | "empty") {
+    const knownState =
+      payload === "empty" ? emptyKnownState(id) : cloneKnownState(payload);
+    this.coValues.set(id, knownState);
+    this.triggerUpdate(id, knownState);
   }
 
   get(id: RawCoID) {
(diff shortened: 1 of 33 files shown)


packages/cojson/src/PeerKnownStates.ts
```diff
@@ -1,41 +1,51 @@
...
+    const knownState = this.coValues.get(id);
+
+    if (!knownState) {
+      const knownState = emptyKnownState(id);
+      this.coValues.set(id, knownState);
+      return knownState;
+    }
+
+    return knownState;
+  }
```
gdorsi, line 23: Created a known state getter to centralize the empty fallback and remove the `coValue.set` when not necessary.

packages/cojson/src/PeerKnownStates.ts
```diff
@@ -1,41 +1,51 @@
...
-    this.triggerUpdate(id);
+    this.triggerUpdate(id, knownState);
   }
 
   combineWith(id: RawCoID, value: CoValueKnownState) {
-    const knownState = this.coValues.get(id) ?? emptyKnownState(id);
-    this.coValues.set(id, combinedKnownStates(knownState, value));
-    this.triggerUpdate(id);
+    const knownState = this.getKnownState(id);
+    combineKnownStates(knownState, value);
```
gdorsi, line 33: Changed this from an immutable update into a mutation.

It was the only immutable change on knownState here, so it should be unnecessary.

packages/cojson/src/PeerKnownStates.ts
```diff
@@ -1,41 +1,51 @@
...
 
-  set(id: RawCoID, knownState: CoValueKnownState | "empty") {
-    this.coValues.set(
-      id,
-      knownState === "empty" ? emptyKnownState(id) : knownState,
-    );
-    this.triggerUpdate(id);
+  set(id: RawCoID, payload: CoValueKnownState | "empty") {
+    const knownState =
+      payload === "empty" ? emptyKnownState(id) : cloneKnownState(payload);
```
gdorsi, line 46: This was leading us to share the same object between optimistic and knownState.

It wasn't a problem because the only optimistic update we do uses `combineWith` which was immutable, but now that I changed it to be mutable I found the bug here.

Fixed by cloning the input, which should be a good practice to do everywhere we do manipulate a knownState,.

packages/cojson/src/PeerKnownStates.ts
```diff
@@ -48,14 +58,18 @@ export class PeerKnownStates {
 
   clone() {
     const clone = new PeerKnownStates();
-    clone.coValues = new Map(this.coValues);
+
+    for (const [id, knownState] of this.coValues) {
+      clone.coValues.set(id, cloneKnownState(knownState));
+    }
```
gdorsi, line 64: This is a bit more expensive than before, but before we were depending on `combineWith` for doing "some" immutable updates.

It felt too weak as safety net.

packages/cojson/src/PeerKnownStates.ts
```diff
@@ -48,14 +58,18 @@ export class PeerKnownStates {
...
+
     return clone;
   }
 
   listeners = new Set<(id: RawCoID, knownState: CoValueKnownState) => void>();
 
-  triggerUpdate(id: RawCoID) {
-    this.trigger(id, this.coValues.get(id) ?? emptyKnownState(id));
+  triggerUpdate(id: RawCoID, knownState: CoValueKnownState) {
+    this.trigger(id, knownState);
```
gdorsi, line 72: Made the method private and optimized away the map access.

packages/cojson/src/coValueCore/SessionMap.ts
```diff
@@ -24,15 +28,31 @@ export type SessionLog = {
...
+  sessionID: SessionID;
 };
 
 export class SessionMap {
   sessions: Map<SessionID, SessionLog> = new Map();
 
+  // Known state related properies, mutated when adding transactions to the session map
+  knownState: CoValueKnownState = { id: this.id, header: true, sessions: {} };
+  knownStateWithStreaming: CoValueKnownState | undefined;
+  streamingKnownState?: KnownStateSessions;
```
gdorsi, line 40: Moved the knownState and streaming state here to co-locate the knownState update code with the transactions storage.

This way has been easy to keep track of the straming state changes and ensure that they are consistent with the transactions updates.

packages/cojson/src/coValueCore/SessionMap.ts
```diff
@@ -24,15 +28,31 @@ export type SessionLog = {
...
+    streamingKnownState?: KnownStateSessions,
+  ) {
+    if (streamingKnownState) {
+      this.streamingKnownState = { ...streamingKnownState };
+      this.knownStateWithStreaming = {
+        id: this.id,
+        header: true,
+        sessions: { ...streamingKnownState },
+      };
+    }
```
gdorsi, line 54: Always clone the input! (learned the hard way)

packages/cojson/src/coValueCore/verifiedState.ts
```diff
@@ -289,72 +287,42 @@ export class VerifiedState {
   }
 
   /**
-   * Returns the known state considering the known state of the streaming source
+   * Returns the known state of the CoValue
    *
-   * Used to correctly manage the content & subscriptions during the content streaming process
+   * The return value identity is going to be stable as long as the CoValue is not modified.
+   *
+   * On change the knownState is invalidated and a new object is returned.
```
gdorsi, line 294: Experimented a bit with returing always the same object as knownState.

`cojson` handles it correctly, but since this is a public API it was causing unexpected errors in the tests.

So I added a clone + cache to ensure that when the CoValue changes we return a new knownState object (same behaviour as before)

antoniomuso (approved):

PR #3062, opened 2025-10-16
docs: Update jazz syntax in better-auth example

Wizzel1:
# Description
The better-auth docs were using the pre 0.18 value assignment syntax

## Tests

- [x] Tests have not been updated, because: Only docs were changed

## Checklist

- [X] I've updated the part of the docs that are affected the PR changes


gdorsi (approved):
Thanks for the fix!

PR #3066, opened 2025-10-16
fix(core): fix concurrent loading from storage and refactor loading status management

gdorsi:
While cleaning up the code I found a possible bug on loadFromStorage and fixed it.


packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -1369,40 +1379,34 @@ export class CoValueCore {
       peer.trackLoadRequestSent(this.id);
     }
 
-    return new Promise<void>((resolve) => {
```
gdorsi, line 1372: The returned promise was unused

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -181,8 +181,9 @@ export class CoValueCore {
   get verified() {
     return this._verified;
   }
-  private readonly peers = new Map<
-    PeerID,
+
+  private readonly loadingStatuses = new Map<
+    string,
```
nrainhart: Why stop using PeerID?

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -265,53 +266,38 @@ export class CoValueCore {
   }
 
   isErroredInPeer(peerId: PeerID) {
-    return this.peers.get(peerId)?.type === "errored";
+    return this.getLoadingStateForPeer(peerId) === "errored";
   }
 
-  waitForAvailableOrUnavailable(): Promise<CoValueCore> {
+  waitFor(callback: (value: CoValueCore) => boolean) {
```
nrainhart, line 253: Sweet 💅

packages/cojson/src/tests/coValueCore.loadFromStorage.test.ts
```diff
@@ -0,0 +1,551 @@
...
+
+beforeEach(() => {
+  metricReader = createTestMetricReader();
+});
+
+afterEach(() => {
+  tearDownTestMetricReader();
+});
+
+function setup() {
```
nrainhart, line 21: This could live in `testUtils`. It could be something like `createCoValueCoreWithoutContent`, that returns the covalue and the header so that you can mark it as available later

nrainhart (approved):

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: Because this also accepts "storage" and PeerID is just string anyway.

But good point, I think being explicit helps on documenting what's in there so changing into `PeerID | "storage"`

packages/cojson/src/tests/coValueCore.loadFromStorage.test.ts
gdorsi: Created a `createUnloadedCoValue` utility, and killed the unused static methods on CoValueCore.

PR #3068, opened 2025-10-16
feat: split useAccount and merge selectors into useAccount & useState

nrainhart:
# Description

Splits the `useAccount` hook into three separate hooks _*_:
- `useAccount`: now only returns an Account CoValue
- `useLogOut`: returns a function for logging out of the current account
- `useAgent`: returns the current agent

It also adds a `select` option (and an optional `equalityFn`) to `useAccount` and `useCoState`, and removes `useAccountWithSelector` and `useCoStateWithSelector`_**_.

_*_ This change is applied both in the React and Vue bindings.
_**_ This change is applied only in React, as Vue bindings do not support selector functions.

## Discussion

I thought of splitting Svelte's [AccountCoState](https://github.com/garden-co/jazz/blob/feat/merge-useCoState-useCoStateWithSelector/packages/jazz-tools/src/svelte/jazz.class.svelte.ts#L142) class, but saw almost no benefit in doing so (apart from aligning the React & Svelte APIs more closely), so I ended up keeping it as is. If you think we should also make the split for Svelte, let me know.

## Tests

### DIFF
--- a/examples/betterauth/src/app/(app)/page.tsx
+++ b/examples/betterauth/src/app/(app)/page.tsx
@@ -12,7 +12,7 @@ import {
 import Image from "next/image";
 
 export default function Home() {
-  const { me } = useAccount(Account, { resolve: { profile: {} } });
+  const me = useAccount(Account, { resolve: { profile: {} } });
 
   if (!me.$isLoaded) {
     return null;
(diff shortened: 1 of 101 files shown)


gdorsi (approved):
LGTM

Agree on Svelte

PR #3078, opened 2025-10-20
fix(upsertUnique): fix concurrent calls on upsertUnique

gdorsi:
We found that the following snippet was filing with an error provideHeader:

```ts
    const Project = co.map({
      name: z.string(),
    });

    const owner = Group.create();

    const promises = Array.from({ length: 3 }, (_, i) =>
      Project.upsertUnique({
        owner,
        unique: "concurrent",
        value: { name: `Project ${i}` },
      }),
    );

    await Promise.all(promises);
```

This PR fixes the issue, and centralizes the loadUnique logic


packages/jazz-tools/src/tools/coValues/coMap.ts
```diff
@@ -434,19 +436,17 @@ export class CoMap extends CoValueBase implements CoValue {
...
-    return CoMap._findUnique(unique, ownerID, as);
+    const header = CoMap._getUniqueHeader(unique, ownerID);
+
+    return getIdFromHeader(header, as);
   }
 
   /** @internal */
-  static _findUnique<M extends CoMap>(
-    this: CoValueClass<M>,
+  static _getUniqueHeader(
```
nrainhart, line 445: We can make this method private, right?

packages/jazz-tools/src/tools/coValues/coMap.ts
```diff
@@ -495,34 +492,27 @@ export class CoMap extends CoValueBase implements CoValue {
       unique: CoValueUniqueness["uniqueness"];
       owner: Account | Group;
       resolve?: RefsToResolveStrict<M, R>;
+      upsert?: boolean;
```
nrainhart: Is this option being used?

packages/jazz-tools/src/tools/coValues/interfaces.ts
```diff
@@ -446,6 +448,92 @@ export function parseGroupCreateOptions(
...
+    // if load returns unavailable, we check the state in localNode
+    // to ward against race conditions that would happen when
+    // running the same upsert unique concurrently
+    if (!result && node.getCoValue(id).hasVerifiedContent()) {
+      result = await loadCoValueWithoutMe(cls, id, {
+        loadAs,
+      });
+    }
+
+    if (!result) {
```
nrainhart, line 502: What if the result is not available because the viewer is not authorized to read it? Should we create a new covalue?

packages/jazz-tools/src/tools/coValues/interfaces.ts
```diff
@@ -446,6 +448,92 @@ export function parseGroupCreateOptions(
...
+        resolve: options.resolve,
+      });
+    }
+  }
+
+  if (!result) return result;
+
+  if (options.onUpdateWhenFound) {
+    // we deeply load the value, retrying any failures
+    const loaded = await loadCoValueWithoutMe(cls, id, {
```
nrainhart, line 516: I think this is fine for now to not introduce breaking changes. But longer term, do we really need to handle updating as part of the load operation? Why not just first load the unique covalue and then update it if it's available?

packages/jazz-tools/src/tools/tests/coList.unique.test.ts
```diff
@@ -0,0 +1,244 @@
...
+    );
+
+    await Promise.all(promises);
+
+    const result = await ItemList.loadUnique("concurrent", owner.$jazz.id);
+    assert(result);
+
+    expect(result.length).toBe(2);
+    expect(result[0]).toBe(`Item 2`);
+    expect(result[1]).toBe(`Second 2`);
```
nrainhart, line 242: Should we also check that the other values are preserved as part of the history?

nrainhart (commented):
Looks good overall. I just have a few questions before approving

packages/jazz-tools/src/tools/coValues/interfaces.ts
gdorsi: Nothing bad should happen, because the created CoValue would be empty and the load would return null anyway.

Conflicts on unique values are managed at the sync level, so are ok on this side.

But, I can cover this with a test if it's not covered already.

packages/jazz-tools/src/tools/coValues/interfaces.ts
gdorsi: > But longer term, do we really need to handle updating as part of the load operation?

Agree that just having the create on missing would be better, but haven't introduced any breaking changed because I want to collect all the best practices around unique before writing an API that enforces them.

For example one issue we have now is the conflict management on nested references.

It would be good if we could either find a solution to that or push users to not nest references on unique values.

One idea I'd like to try is to propagate the uniqueness to childs, so conflicts would be merged n-levels deep.

packages/jazz-tools/src/tools/coValues/coMap.ts
gdorsi: Nope, thanks!

packages/jazz-tools/src/tools/coValues/coMap.ts
gdorsi: Typescript complains on the Record factory method:

Screenshot 2025-10-20 at 11 37 00

packages/jazz-tools/src/tools/tests/coList.unique.test.ts
gdorsi: Unsure, it would be pretty complex and not really part of this specific issue.

We have other tests that cover the create and update steps, which are more targeted and so easier to debug and understand.

nrainhart: I feel like we should document if the other changes are lost of if they are kept as part of the coValue's history

nrainhart (approved):
Thanks for addressing my comments!

PR #3080, opened 2025-10-21
Enable MDX Rust support

joeinnes:
# Description
On my laptop at least:
Before: pnpm run build  302.45s user 52.44s system 195% cpu 3:01.46 total
After: pnpm run build  66.28s user 21.27s system 218% cpu 40.121 total

±4.5&times; faster

Run `time pnpm build` with the flag set true vs false.

## Tests

## Checklist


joeinnes:
It's fast, but it breaks stuff with our CodeGroup stuff

PR #3082, opened 2025-10-21
fix inspector render in Svelte - GCO-930

Takeno:
# Description
`gobber` lib failed in non-react environments, like our Svelte examples.

The previous call of `setup` was invoked too late.

Run the Svelte example.

### DIFF
--- a/packages/jazz-tools/src/inspector/index.tsx
+++ b/packages/jazz-tools/src/inspector/index.tsx
@@ -1,3 +1,5 @@
+import React from "react";
+
 export { JazzInspectorInternal } from "./viewer/new-app.js";
 export { PageStack } from "./viewer/page-stack.js";
 export { Breadcrumbs } from "./viewer/breadcrumbs.js";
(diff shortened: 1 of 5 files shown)


packages/jazz-tools/src/inspector/index.tsx
```diff
@@ -30,8 +27,6 @@ export function JazzInspector({ position = "right" }: { position?: Position }) {
   const localNode = context.node;
   const me = "me" in context ? context.me : undefined;
 
-  if (process.env.NODE_ENV !== "development") return null;
```
Takeno, line 33: I've removed this line because the check is already done at the top level in [register-custom-element.ts](https://github.com/garden-co/jazz/blob/2e63a50d41b426827302e512236f707d7a0897d0/packages/jazz-tools/src/inspector/register-custom-element.ts)

gdorsi (approved):
LGTM, thanks!

PR #3084, opened 2025-10-21
feat: allow injecting a custom meter for usage metering

Elfo404:
Allows injection of a custom meter to `createWebsocketPeer` and `BatchedOutgoingMessages`. If no custom meter is provided a default one is retrieved.

[Fixes GCO-952
](https://linear.app/garden-co/issue/GCO-952/cojson-transport-ws-accept-a-custom-meter-when-dealing-with-usage)

### DIFF
--- a/packages/cojson/src/coValueCore/SessionMap.ts
+++ b/packages/cojson/src/coValueCore/SessionMap.ts
@@ -35,7 +35,7 @@ export class SessionMap {
   sessions: Map<SessionID, SessionLog> = new Map();
 
   // Known state related properies, mutated when adding transactions to the session map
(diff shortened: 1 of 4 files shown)


packages/cojson/src/coValueCore/SessionMap.ts
```diff
@@ -44,6 +44,7 @@ export class SessionMap {
     private readonly crypto: CryptoProvider,
     streamingKnownState?: KnownStateSessions,
   ) {
+    this.knownState = { id: this.id, header: true, sessions: {} };
```
Elfo404, line 47: this is unrelated to this PR, but i was getting TS errors when using it from an external project

gdorsi (approved):

PR #3086, opened 2025-10-21
Inspector's group view enhancements

Takeno:
# Description
The group's parents and children are now shown in the group view, and it is now possible to add/remove members directly from the inspector.

## Before
immagine

## After
immagine

### DIFF
--- a/packages/jazz-tools/src/inspector/viewer/group-view.tsx
+++ b/packages/jazz-tools/src/inspector/viewer/group-view.tsx
@@ -1,4 +1,10 @@
-import { JsonObject, LocalNode, RawAccount } from "cojson";
+import {
+  JsonObject,
+  LocalNode,
+  RawAccount,
+  RawCoValue,
+  RawGroup,
+} from "cojson";
 import { CoID } from "cojson";
 import {
   Table,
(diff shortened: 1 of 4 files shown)


packages/jazz-tools/src/inspector/viewer/group-view.tsx
```diff
@@ -8,60 +16,387 @@ import {
...
+      </div>
 
-          {Object.entries(data).map(([key, value]) =>
-            isCoId(key) ? (
-              <TableRow key={key}>
+      {childGroups.length > 0 && (
+        <Table>
+          <TableHead>
+            <TableRow>
+              <TableHeader>Children</TableHeader>
```
nrainhart: I find the term "children" hard to grasp. What do you think about renaming this to something more intention-revealing? E.g. "Included in" or "Member of"

packages/jazz-tools/src/inspector/viewer/group-view.tsx
```diff
@@ -8,60 +16,387 @@ import {
...
+              <Button type="submit" variant="primary">
+                Add
+              </Button>
+            </div>
+          </div>
+        </form>
+      </Modal>
+
+      <Modal
+        isOpen={isAddGroupModalOpen}
```
nrainhart: Nit: both modals are pretty similar, maybe we can extract them into a single component?

nrainhart (approved):
This is amazing!! We really need better debugging tools for permissions

packages/jazz-tools/src/inspector/viewer/group-view.tsx
Takeno: Changed into "Member of", much clearer.

packages/jazz-tools/src/inspector/viewer/group-view.tsx
Takeno: Initially, I kept both roles separated to clarify their different responsibilities in the reading. I merged them, as once merged, I could read them as easily as before.

Issue #3076, opened 2025-10-20
SQL storage for Cloudflare Durable Objects


Nipsuli:
In my app I rely heavily on durable objects, and many LLM powered functionality pulls big portions of the data so was thinking that having data stored on the the Durable Objects SQLite db could speed up operations. Discussed also in [discord](https://discord.com/channels/1139617727565271160/1139621689882321009/1428824411019935804).

I'll take a stab at this, and will open draft pr when I have something.

PR #3087, opened 2025-10-21
Add durable object SQL storage

Nipsuli:
# Description

Add `cloudflare-do-storage` package to allow utilizing the SQLite store in Cloudflare Durable Objects for Jazz. In my app I utilize Cloudflare Durable Objects heavily, and being able to load data mainly from local store instead over internet should make things faster, especially when there are repeated operations over larger collections of data, example when it comes to search and retrieval type of workloads in AI backends.

To make this work the `startWorker` in the `jazz-tools/worker` now accepts optional `storage: StorageAPI` option.

All the implementation is isolated in the new `cojson-storage-do-sqlite` package. Do to the limitations of SQLite on Durable Objects needed to modify the migration queries on the fly to store the migration version in separate table instead of using `PRAGMA user_version = <version>;`. If you have better ideas how to mange this, or if that logic should be moved to somewhere else, let me know and I'll fix.

Due to some bug in the `@cloudflare/vitest-pool-workers` related to loading wasm modules could not use that for implementing tests so the tests are in `tests/cloudflare-do-storage` and logic of running the operations are implemented within durable object and the test starts a dev server of wrangler. The wasm loading works perfectly fine in dev and in actual Cloudflare environment, the issue is in the test package. The tests are exact replica of the tests from the `cojson-storage-sqlite`. 

In addition there is simple example app in `examples/durable-object` showcasing the usage with simple KV store type of api. (didn't have any better ideas to do simple example 😅).

Related issue https://github.com/garden-co/jazz/issues/3076

No AI coding agent's were used to create this code.

For manual testing one can run the example app and see how it behaves.

## Tests

Tests have been added to `tests/cloudflare-do-storage`, but let me know if there are something you'd want to improve there.

## Checklist


packages/cojson-storage-do-sqlite/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+        ]);
+      } else {
+        throw new Error(`Unsupported pragma in migration: ${migration}`);
+      }
+    } else {
+      driver.run(migration, []);
+    }
+  }
+  return new StorageApiSync(new SQLiteClient(driver));
+}
```
Nipsuli, line 60: This is the migration bit that I'm bit unsure if this is the best possible approach. I'm open for better solutions.

examples/durable-object/src/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+    this.jazzConn = await startWorker({
+      accountID: this.env.JAZZ_WORKER_ACCOUNT,
+      accountSecret: this.env.JAZZ_WORKER_SECRET,
+      syncServer: `wss://cloud.jazz.tools/?key=${this.env.JAZZ_API_KEY}`,
+      asActiveAccount: false,
+      storage: getDurableObjectSqlStorage(this.ctx.storage),
+    });
+    return this.jazzConn;
+  }
```
Nipsuli, line 24: Example usage here

packages/cojson-storage-do-sqlite/src/index.ts
gdorsi: Maybe we could move the version tracking up to the driver?

Something like:
```ts
getCurrentVersion() {
  return this.get<{ version: number }>(
    "SELECT max(version) as version FROM _migration_version;",
    [],
  )?.version;
},
setCurrentVersion(version: number) {
  return this.run("INSERT INTO _migration_version (version) VALUES (?);", [
          parsedVersion,
        ]);
}
```

We could fallback to the current strategy if those methos are missing in the driver for backward compat.

packages/cojson-storage-do-sqlite/src/tests/worker-configuration.d.ts
```diff
@@ -0,0 +1,9681 @@
+/* eslint-disable */
```
gdorsi, line 1: Does this file need to be versioned?

gdorsi (commented):
Thanks for the contribution!

Code looks mostly good, the only detail to finalize are the migrations.

packages/cojson-storage-do-sqlite/src/index.ts
Nipsuli: That is actually good idea to move the methods to the driver! Will do that.

packages/cojson-storage-do-sqlite/src/tests/worker-configuration.d.ts
Nipsuli: To my best of understanding in worker projects (in general) it's best practice to generate the worker configuration.d.ts from the wrangler.jsonc and local .dev.vars to have the all the types correctly based on the actual configuration, including env variables. But in this case where there aren't really much going on and it's for tests, it's not that important. I'm fine either way.

gdorsi: SGTM

Thanks for the explanation!

packages/cojson-storage-do-sqlite/src/index.ts
Nipsuli: @gdorsi moved the migration management to the driver with the optional methods falling back to the behavior there was in case those methods are not implemented. As such removed the hard coded pragma queries from migrations and changed the migration shape a bit. All tests run on my local machine.

gdorsi (approved):
Looks good to me.

Good job, thanks!

PR #3092, opened 2025-10-22
Add "Testing" section

joeinnes:
# Description
Add a section on testing to the docs

N/A

## Tests

## Checklist

### DIFF
--- a/homepage/homepage/content/docs/reference/testing.mdx
+++ b/homepage/homepage/content/docs/reference/testing.mdx
@@ -97,20 +97,47 @@ During your tests, you may need to manage the currently active account after acc
 ### `setActiveAccount`
 Use `setActiveAccount()` to switch between active accounts during a test run.
 
+You can use this to test your app with multiple accounts.
+
 <CodeGroup preferWrap>
 ```ts twoslash
-import { co } from 'jazz-tools';
+import { co, z } from 'jazz-tools';
 import { createJazzTestAccount, setActiveAccount } from 'jazz-tools/testing';
+const MyMap = co.map({
+  text: z.string()
+})
+function expect (actual?: any) {
+    return { 
+      toBe (expected?: any) {
+        return null;
+      },
+      toThrow() {
+        return null;
+      }
+    }
+};
 // ---cut---
 const account1 = await createJazzTestAccount({ 
   isCurrentActiveAccount: true,
 });
-const group1 = co.group().create(); // Group is owned by account1;
-
 const account2 = await createJazzTestAccount();
-const group2 = co.group().create(); // Group is still owned by account1;
+
+const group1 = co.group().create(); // Group is owned by account1;
+group1.addMember(account2, 'reader');
+
+const myMap = MyMap.create({
+  text: "Created by account1"
+}, { owner: group1 });
+const myMapId = myMap.$jazz.id;
+
 setActiveAccount(account2);
-const group3 = co.group().create(); // Group is owned by account2;
+// myMap is still loaded as account1, so we need to load again as account2
+const myMapFromAccount2 = await MyMap.load(myMapId);
+
+expect(myMapFromAccount2?.text)
+  .toBe('Created by account1');
+expect(() => myMapFromAccount2?.$jazz.set('text', 'Updated by account2'))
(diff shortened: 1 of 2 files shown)


homepage/homepage/content/docs/reference/testing.mdx
```diff
@@ -0,0 +1,214 @@
...
+    });
+    expect(account1).not.toBeUndefined();
+    // ...
+  });
+});
+```
+</CodeGroup>
+
+<Alert variant="info" title="Important" className="mt-4">
+  The account returned by `setupJazzTestSync` is not created based on any custom schemas you may have defined, and no migrations will be run on this account.
```
gdorsi: The account value returned by `setupJazzTestSync` is only meant to be used to check the status of CoValues on the sync node.

It's a pretty advanced use-case, we could remove it if you think it might create confusion.

homepage/homepage/content/docs/reference/testing.mdx
```diff
@@ -0,0 +1,214 @@
...
+# Testing Jazz Apps
+
+As you develop your Jazz app, you might find yourself needing to test functionality relating to sync, identities, and offline behaviour. The `jazz-tools/testing` utilities provide helpers to enable you to do so.
+
+## Core test helpers
+Jazz provides some key helpers that you can use to simplify writing complex tests for your app's functionality.
+
+### `setupJazzTestSync`
+
+This should normally be the first thing you call in your test setup, for example in a `beforeEach` or `beforeAll` block. This function creates and returns a plain `Account`, which will also act as an in-memory sync node for the test session. You'll often pair this with `createJazzTestAccount()` for more complex testing. Test data is not persisted, and no clean-up is needed between test runs.
```
gdorsi: This is actually useful only for testing multi-account interactions, but it's good to have this as a rule to avoid wasting time trying to understand why two accounts don't sync

homepage/homepage/content/docs/reference/testing.mdx
```diff
@@ -0,0 +1,214 @@
...
+const account1 = await createJazzTestAccount({
+  isCurrentActiveAccount: true,
+});
+runWithoutActiveAccount(() => {
+  expect(() => co.group().create()).toThrow(); // can't create new group
+});
+```
+</CodeGroup>
+
+### `linkAccounts`
```
gdorsi: We can avoid documenting this one, the test sync now it is stable enough to replace linkAccounts

homepage/homepage/content/docs/reference/testing.mdx
```diff
@@ -0,0 +1,214 @@
...
+</ContentByFramework>
+
+<ContentByFramework framework="svelte">
+ You can render your components for testing by passing a mocked Jazz context to the `@testing-library/svelte` `render` helper.
+
+ [You can see an example of how we do that here](https://github.com/garden-co/jazz/blob/main/packages/jazz-tools/src/svelte/tests/testUtils.ts).
+</ContentByFramework>
+
+
+The `TestJazzContextManager` mocks the `JazzContextManager` to allow you to instantiate a Jazz context as a user or a guest, allowing you to run tests which depend on an authenticated or a guest session.
```
gdorsi, line 185: This one is not really meant to be a public API, but just a building block for creating test providers.

gdorsi (commented):
LGTM

I would add an example on how to write multi-account tests.

It would also be useful sharing [this](https://github.com/garden-co/jazz/blob/main/packages/jazz-tools/src/react-core/tests/testUtils.tsx) as example on how to integrate the JazzProvider with [testing library](https://testing-library.com/)

homepage/homepage/content/docs/reference/testing.mdx
joeinnes: I will change some of the wording around this to make it clear that it's not really intended to use the account.

homepage/homepage/content/docs/reference/testing.mdx
joeinnes: I think we might still need this for vanilla apps, but maybe I can put this in a ContentByFramework?

gdorsi (approved):

PR #3108, opened 2025-10-25
perf: remove child links on parent groups

gdorsi:
We found that child attiributes don't scale and that they cause severe performance issues when using extend inside of a loop:

```ts
  for (let i = 1; i <= 25000; i++) {
    const group = Group.create();
    group.addMember(recordGroup);

    const item = Item.create(
      {
        name: `Item ${i}`,
      },
      group,
    );

    record.$jazz.set(i, item);
  }
```

This PR replaces the child attributes for the key rotations into a dual rotation strategy:
- when rotating the parent key, all the child groups loaded in memory rotate their key
- on load, every group check if their key has been revealed to the latest parent readKey. If not, it means that the parent changed their key.

This change removes the cost for the parent of being extended, and brings down the test above from hours to 53s


packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -1147,12 +1147,14 @@ export class CoValueCore {
...
-    if (
-      !dependencyCoValue.isAvailable() &&
-      !this.isCircularMissingDependency(dependencyCoValue)
-    ) {
-      this.missingDependencies.add(dependency);
+    if (this.isCircularMissingDependency(dependencyCoValue)) {
+      return true;
+    }
+
+    dependencyCoValue.addDependant(this.id);
```
gdorsi, line 1154: Added dependant tracking, it's going to be useful also for GC and the check on "completely loaded"

packages/cojson/src/coValues/group.ts
```diff
@@ -260,27 +294,18 @@ export class RawGroup<
   }
 
   forEachChildGroup(callback: (child: RawGroup) => void) {
-    for (const key of this.keys()) {
-      if (isChildGroupReference(key)) {
-        // Check if the child group reference is revoked
-        if (this.get(key) === "revoked") {
-          continue;
-        }
+    for (const id of this.core.dependant) {
```
nrainhart, line 299: It'd be nice to keep this as a comment somewhere:
> when rotating the parent key, all the child groups loaded in memory rotate their key
on load, every group check if their key has been revealed to the latest parent readKey. If not, it means that the parent changed their key

nrainhart (approved):
🚀

antoniomuso (approved):

packages/cojson/src/coValues/group.ts
```diff
@@ -161,7 +191,11 @@ export class RawGroup<
     super(core, options);
     this.crypto = core.node.crypto;
 
-    healMissingKeyForEveryone(this);
+    // Checks if this is not an account
+    if (core.isGroup()) {
+      rotateReadKeyIfNeeded(this);
```
Takeno, line 196: Does this mean the key rotation has been postponed to the first Group load? If multiple peers are connected, will this result in multiple concurrent rotations, leading to a significant increase in transactions?

gdorsi: The key rotation is triggered only on the Group load, so the case is limited to when multiple peers are loading concurrently the Group.

This kind-of weaken the key rotation, but we are going to improve the security soon with authentication on sync.

PR #3116, opened 2025-10-28
fix: wait for full download of dependencies before running migrations

gdorsi:
# Description

With this PR we re-enable the progressive child-key rotation and add checks to validate the full download of all the dependency graph of a CoValue.

This fixes some edge cases, where we were starting migrations before some parent group was fully downloaded, missing some vital permissions information on how to validate and decrypt the transactions.

## Tests


packages/cojson/src/coValues/group.ts
```diff
@@ -123,6 +123,13 @@ function healMissingKeyForEveryone(group: RawGroup) {
 }
 
 function needsKeyRotation(group: RawGroup) {
+  const me = group.core.node.getCurrentAccountOrAgentID();
+
+  // Checking only direct membership because inside the migrations we can't navigate the parent groups
```
nrainhart, line 128: I don't get why we have this limitation. All parent groups are completely loaded by this point

nrainhart: Also: what's the impact of delaying the key rotation until an account with admin/manager loads the group? Could this mean an account with reader/writer permission won't be able to read/write, even when they technically have permissions to do so?

Edit: this [test case](https://github.com/garden-co/jazz/pull/3116/files#diff-6f24efd5bb3f0d39cdc0ebb56cb2da3023f3dd278af0130abc2df1f024e3e7cfR363) answers my question

packages/cojson/src/coValues/group.ts
```diff
@@ -190,12 +197,30 @@ export class RawGroup<
...
+  migrate() {
+    if (!this.core.isGroup()) {
+      return;
     }
+
+    // We need the group and their parents to be completely downloaded to correctly handle the migrations
+    if (!this.core.isCompletelyDownloaded()) {
+      this.core.subscribe((core, unsubscribe) => {
+        if (core.isCompletelyDownloaded()) {
+          unsubscribe();
```
nrainhart: This pattern is coming up a lot, what about extracting it into something like:
```typescript
this.core.waitFor(core => core.isCompletelyDownloaded()).then(()=> {
	rotateReadKeyIfNeeded(this);
	healMissingKeyForEveryone(this);
});
```
So that we don't need to handle the subscription lifecycle everywhere

nrainhart (commented):
Thanks for adding the tests! They are super detailed and clear.

I have only [one important question](https://github.com/garden-co/jazz/pull/3116#discussion_r2472707989)

packages/cojson/src/tests/coValueCore.isCompletelyDownloaded.test.ts
```diff
@@ -0,0 +1,491 @@
...
+    // Disconnect the admin node to sync the content manually
+    admin.disconnect();
+
+    // Make the original map large enough to require multiple messages to be synced
+    for (let i = 0; i < 100; i++) {
+      originalMap.set(`key${i}`, "1".repeat(1024));
+    }
+
+    // Create a branch from the original map
+    const branch = expectMap(
```
nrainhart: Is this situation the same if we first branch the map and then add new content to the original map? I.e. do we always need to wait for streaming of the original covalue to complete.

It shouldn't be necessary in this second case, right? If so, don't think it's something we need to optimize right now, but maybe we should have a separate test case for that scenario.

packages/cojson/src/coValues/group.ts
gdorsi: That was my first solution, but refactored into a `subscribe` to not pay for the Promise.

I guess we can write an abstraction that does the same without using a Promise

antoniomuso (approved):

packages/cojson/src/coValues/group.ts
gdorsi: Abstracted! https://github.com/garden-co/jazz/pull/3116/commits/b866e5977614452222d69e4c42f14dbb15b92d58

packages/cojson/src/tests/coValueCore.isCompletelyDownloaded.test.ts
gdorsi: I suppose that waiting for the full update is correct in most situations, the branch is actually the only exception because the branch would probably just discard the update.

I don't feel like it's relevant enough for us to have an exception in the code, for now.

Going to refactor these tests to handle both the "cold start" case and the "streaming in the middle" case

gdorsi: Ok, added the tests and the check actually returns `true` if the map is loaded and new content comes with streaming.

Need to figure out if this could be an issue for groups, but I think it isn't.

packages/cojson/src/coValues/group.ts
gdorsi: You are right, switched to use `myRole`

Had some issues before related to the way it tries to acces the current account ID, but I fixed that so it is ok now.

PR #3121, opened 2025-10-29
fix: track streaming state on content updates and fix empty initial content on streaming

gdorsi:
# Description

Noticed that there are some gaps in the way we manage content streaming.
This PR consolidates the way we handle it, and adds more tests to cover more streaming scenarios.

This PR brings a bundle of fixes on content streaming:
- added `expectContentUntil` on the local transactions sync (before all the local changes were never marked as streaming content!)
- removed the first empty content piece when streaming from `newContentSince`
- added streaming state tracking on updates (missing before)

It also brings some sanity to the implementation of `newContentSince` documenting the logic with tests.

## Tests


packages/cojson/src/tests/sync.upload.test.ts
```diff
@@ -407,41 +474,40 @@ describe("client to server upload", () => {
       }),
     ).toMatchInlineSnapshot(`
       [
-        "client -> server | LOAD Map sessions: empty",
-        "server -> client | CONTENT Group header: true new: After: 0 New: 5",
-        "server -> client | CONTENT Map header: true new:  expectContentUntil: header/1024",
-        "server -> client | CONTENT Map header: false new: After: 0 New: 73",
```
gdorsi, line 413: This test highlights that we removed the first empty chunk

packages/cojson/src/coValueCore/verifiedState.ts
```diff
@@ -160,119 +161,147 @@ export class VerifiedState {
     );
   }
 
+  setStreamingKnownState(streamingKnownState: KnownStateSessions) {
+    this.sessions.setStreamingKnownState(streamingKnownState);
+  }
+
   newContentSince(
```
nrainhart, line 159: It'd be nice to add some TS docs to the most important methods in CoValueCore (such as this one). It'd be enough to explain what this method does (at a high level), and what its primary use-case is

packages/cojson/src/coValueCore/verifiedState.ts
```diff
@@ -160,119 +161,147 @@ export class VerifiedState {
...
     knownState: CoValueKnownState | undefined,
   ): NewContentMessage[] | undefined {
-    const isKnownStateEmpty = !knownState?.header && !knownState?.sessions;
-
-    if (isKnownStateEmpty && this._cachedNewContentSinceEmpty) {
-      return this._cachedNewContentSinceEmpty;
```
nrainhart, line 169: Is `this._cachedNewContentSinceEmpty` still being used somewhere else? Can we remove it?

packages/cojson/src/tests/sync.load.test.ts
```diff
@@ -1311,8 +1386,8 @@ describe("loading coValues from server", () => {
       }),
     ).toMatchInlineSnapshot(`
       [
-        "client -> server | CONTENT Group header: true new: After: 0 New: 3",
-        "client -> server | CONTENT ParentGroup header: true new: After: 0 New: 5",
+        "client -> server | CONTENT Group header: true new: After: 0 New: 3 expectContentUntil: header/5",
+        "client -> server | CONTENT ParentGroup header: true new: After: 0 New: 5 expectContentUntil: header/7",
```
nrainhart, line 1390: I understand why `expectContentUntil` appears here now, but I don't get why it didn't appear before. The followup content messages didn't change, so we were already streaming content, right?

packages/cojson/src/tests/coValueCore.isStreaming.test.ts
```diff
@@ -0,0 +1,271 @@
...
+    const dataSize = 1 * 1024 * 100;
+    const chunkSize = 1024; // 1KB chunks
+    const chunks = dataSize / chunkSize;
+
+    const value = Buffer.alloc(chunkSize, `value$`).toString("base64");
+
+    for (let i = 0; i < chunks; i++) {
+      const key = `key${i}`;
+      map.set(key, value, "trusting");
+    }
```
nrainhart, line 161: Nit: we're doing this in a ton of places, might be worth extracting it into a test factory fn

packages/cojson/src/tests/coValueCore.isStreaming.test.ts
```diff
@@ -0,0 +1,271 @@
...
+    for (const chunk of content) {
+      newSession.node.syncManager.handleNewContent(chunk, "import");
+    }
+
+    const mapInNewSession = await loadCoValueOrFail(newSession.node, map.id);
+
+    expect(mapInNewSession.core.isStreaming()).toBe(false);
+  });
+
+  test("should be false when getting streaming content that's not really streaming", async () => {
```
nrainhart, line 241: Is this a real-world scenario?

packages/cojson/src/tests/coValueCore.newContentSince.test.ts
nrainhart, line 1: ✨

nrainhart (approved):
The `newContentSince` refactor is amazing! It's so easy to follow now. Left a few comments, nothing major.

PS: don't forget to add a changeset :)

packages/cojson/src/tests/sync.load.test.ts
gdorsi: Yes, it is before we weren't sending the `expectContentUntil` when the streaming was starting from the local transactions.

Now I've fixed that.

packages/cojson/src/tests/coValueCore.isStreaming.test.ts
gdorsi: Good point, going to do that in a follow-up PR

packages/cojson/src/tests/coValueCore.isStreaming.test.ts
gdorsi: No, but it gives us an expectation of what would happen in case of bugs on the definition of the `expectContentUntil` value

PR #3127, opened 2025-10-30
feat: Schema-level CoValue resolution rules

nrainhart:
# Description

When working with deeply-nested CoValue schemas, users need to handle complex resolve queries, both in CoValue-loading APIs and at the type level. This is notoriously cumbersome, which leads many users to just ignore resolve queries and resort to autoloading.

Schema-level CoValue resolution rules aim to provide an alternative solution: you specify the CoValue load strategy at the schema level, and then the resolve query will be inferred automatically:
```typescript
const Pet = co.map({
  name: z.string(),
}).resolved();
const Pets = co.list(Pet).resolved();
const Person = co.map({
  name: z.string(),
  pets: Pets(),
}).resolved();

// The following resolve query will be automatically inferred:
// { pets: { $each: true }
const person = await Person.load(personId);
assertLoaded(person);
console.log(person.pets[0]?.name); // "Rex"
```

TODO:
- update bindings for React & Svelte
- user-provided resolve queries currently override default resolve queries. We probably want to merge both queries

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/AccountSchema.ts
+++ b/packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/AccountSchema.ts
@@ -2,11 +2,16 @@ import {
   Account,
   AccountCreationProps,
   BranchDefinition,
+  CoMapSchemaDefinition,
   coOptionalDefiner,
   Group,
   MaybeLoaded,
   RefsToResolveStrict,
+  RefsToResolve,
+  Resolved,
   Simplify,
+  SubscribeListenerOptions,
+  unstable_mergeBranchWithResolve,
 } from "../../../internal.js";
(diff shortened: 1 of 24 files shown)


packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/AccountSchema.ts
```diff
@@ -35,60 +46,163 @@ export type DefaultAccountShape = {
   root: CoMapSchema<{}>;
 };
 
-export interface AccountSchema<
+export class AccountSchema<
```
nrainhart, line 49: I've been wanting to convert the CoMapSchema and the AccountSchema into classes for ages, and this gave me the perfect excuse. This aligns them with all other CoValue schemas, which were already defined using classes

nrainhart:
Superseded by https://github.com/garden-co/jazz/pull/3147

PR #3131, opened 2025-10-31
fix: fix streaming updates tracking and simplify syncLocalTransaction

gdorsi:
This PR started as a refactoring for `syncLocalTransaction` to reuse the logic from `newContentSince` and got into a bug introduced in the last "streaming on updates" improvements.

Surprisingly the refactoring also higlighted a difference in handling the chunking between `syncLocalTransaction` and `newContentSince` which is now fixed.


packages/cojson/src/coValueCore/verifiedState.ts
```diff
@@ -285,10 +288,12 @@ export class VerifiedState {
...
     if (piecesWithContent.length > 1 || this.isStreaming()) {
-      // Flag that more content is coming
-      firstPiece.expectContentUntil = {
-        ...this.knownStateWithStreaming().sessions,
-      };
+      if (!opts?.skipExpectContentUntil) {
+        // Flag that more content is coming
+        firstPiece.expectContentUntil = {
+          ...this.knownStateWithStreaming().sessions,
+        };
```
gdorsi, line 305: TODO: apply getSessionsDifference with the input knownState

PR #3133, opened 2025-10-31
feat: declare resolve fragments in music player schema

gdorsi:


examples/music-player/src/1_schema.ts
```diff
@@ -45,6 +45,15 @@ export const Playlist = co.map({
   tracks: co.list(MusicTrack), // CoList is the collaborative version of Array
 });
 export type Playlist = co.loaded<typeof Playlist>;
+
+export const PlaylistResolveWithTracks = {
+  tracks: { $each: { $onError: null } },
+} satisfies ResolveQuery<typeof Playlist>;
+export type PlaylistResolveWithTracks = co.loaded<
```
nrainhart, line 52: I find it confusing to use the same name for the resolve query and the loaded CoValue type. Other than that, I love the refactor!

PR #3147, opened 2025-11-03
feat: Schema-level resolve queries

nrainhart:
# Description
When working with deeply-nested CoValue schemas, users need to handle complex resolve queries, both in CoValue-loading APIs and at the type level. This is notoriously cumbersome, which leads many users to just ignore resolve queries and resort to autoloading.

Schema-level resolve queries rules aim to provide an alternative solution: you specify the resolve query once at the schema level, and that query will be used both when loading CoValues from that schema (when no resolve query is provided by the user) and in `co.loaded` types:

```typescript
const Pet = co.map({
  name: z.string(),
});
const Person = co.map({
  name: z.string(),
  pets: co.list(Pet),
});
const PersonWithPets = Person.resolved({
  pets: { $each: true }
});

const person = await PersonWithPets.load(personId);
assertLoaded(person);
console.log(person.pets[0]?.name); // "Rex"
```

Pending:
- add docs (will do it as a FUP once the API is set)

## Discussion

As part of your review, I'd appreciate your input on the following items:

- While discussing this feature we talked about _queries_ a lot, but that concept does not exist in this implementation. We just have schemas that may or may not have a resolve query. Do we want to have a clearer distinction between _schemas_ and _queries_?
- I wonder if it will be confusing for our users to decide which schema to use (e.g. when building a new schema or when creating a CoValue) if they have multiple "versions" of the same schema (e.g. Task & TaskWithText). Separating schemas from queries would help with that.

- I migrated several example apps to use schema resolve queries as part of this PR. 
- You can try this PR's pre-release version in any app you want!

## Tests

## Checklist


packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
```diff
@@ -207,7 +273,54 @@ export interface CoMapSchema<
...
+      // @ts-expect-error the partial shape contains all required keys
+      return partialCoMapSchema.catchall(
+        this.catchAll as unknown as AnyZodOrCoValueSchema,
+      );
+    }
+    // @ts-expect-error the partial shape contains all required keys
+    return partialCoMapSchema;
+  }
+
+  resolved<
```
nrainhart, line 333: Wonder if we should name it `unstable_resolved` until we figure out the merge semantics (both with user-provided resolve queries and with nested schemas' resolve queries)

examples/form/src/schema.ts
```diff
@@ -18,19 +18,27 @@ export const BubbleTeaBaseTeaTypes = [
...
+    addOns: true,
+    instructions: true,
+  });
 export type BubbleTeaOrder = co.loaded<typeof BubbleTeaOrder>;
 
 export const PartialBubbleTeaOrder = BubbleTeaOrder.partial({
   baseTea: true,
   deliveryDate: true,
   withMilk: true,
+}).resolved({
```
nrainhart, line 39: Right now `partial` and `pick` don't preserve the schema's resolve query. I don't think we have enough usage examples to decide if it's best to preserve them or not, so for now I'd prefer we don't

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
aeplay: nah, it's good enough not to be prefixed

aeplay (approved):

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/AccountSchema.ts
```diff
@@ -35,60 +43,151 @@ export type DefaultAccountShape = {
...
-  ) => Promise<AccountInstance<Shape>>;
-
-  unstable_merge: <R extends ResolveQuery<AccountSchema<Shape>>>(
+  ): Promise<AccountInstance<Shape>> {
+    // @ts-expect-error
+    return this.coValueClass.createAs(as, options);
+  }
+
+  unstable_merge<
+    // @ts-expect-error
```
gdorsi: Could you add an explaination for these `@ts-expect-error`?

packages/jazz-tools/src/tools/implementation/schemaUtils.ts
```diff
@@ -16,3 +18,16 @@ export function removeGetters<T extends object>(obj: T): Partial<T> {
...
+/**
+ * Adds a CoValue schema's resolve query to a load options object
+ * if no resolve query is provided.
+ */
+export function withSchemaResolveQuery<
+  const T extends { resolve?: CoreResolveQuery },
+>(loadOptions: T | undefined, schemaResolveQuery: CoreResolveQuery): T {
+  const newOptions: CoreResolveQuery = loadOptions ? { ...loadOptions } : {};
+  // TODO merge the schema resolve query with the user-provided resolve query
+  newOptions.resolve ||= schemaResolveQuery;
```
gdorsi, line 31: Do we really want to merge/override the schema queries?

It creates some degree of unpredictability, and Typescript is not going to help much on understanding the output of a merged schema query.

I would go far to not allowing a resolve option if the Schema already have one.

What do you think?

packages/jazz-tools/src/tools/tests/schema.resolved.test.ts
```diff
@@ -0,0 +1,696 @@
...
+        expect(loadedList[0]?.toUpperCase()).toBe("TEST");
+      });
+
+      // TODO fix - this is not working when providing an explicit resolve query either:
+      // TestFeed.load(feed.$jazz.id, {
+      //   loadAs: clientAccount,
+      //   resolve: {
+      //     $each: true,
+      //   },
+      // })
```
gdorsi, line 193: Could you explain this better?

packages/jazz-tools/src/tools/tests/schema.resolved.test.ts
```diff
@@ -0,0 +1,696 @@
...
+      });
+
+      const TestMapWithName = TestMap.resolved({
+        name: true,
+      });
+
+      expect(TestMapWithName.resolve).toEqual({
+        name: true,
+      });
+    });
```
gdorsi, line 58: Could you add a test for:

```
const Parent = co.map({ child: TestMapWithName }).resolved({
  child: true,
})
```

If there is already a test for this, feel free to resolve this comment.

packages/jazz-tools/src/react-core/tests/useCoState.test.ts
```diff
@@ -602,6 +602,43 @@ describe("useCoState", () => {
...
+      },
+    );
+
+    await waitFor(() => {
+      expect(result.current).not.toBeNull();
+    });
+
+    assertLoaded(result.current);
+    expect(result.current.name.toUpperCase()).toBe("JOHN DOE");
+  });
```
gdorsi, line 640: Could you add a test for the inline resolve:

```ts
useCoState(Person.resolved({ name: true }), person.$jazz.id)
```

packages/jazz-tools/src/svelte/jazz.class.svelte.ts
```diff
@@ -306,3 +307,18 @@ export class SyncConnectionStatus {
...
+): ResolveQuery<any> {
+  if (resolveQuery) {
+    return resolveQuery;
+  }
+  // Check the schema is a CoValue schema (and not a CoValue class)
+  if ("resolve" in Schema) {
+    return Schema.resolve;
+  }
+  return true;
+}
```
gdorsi, line 327: Could you add a test for this?

I can help with the setup.

packages/jazz-tools/src/tools/implementation/schemaUtils.ts
nrainhart: > I would go far to not allowing a resolve option if the Schema already have one.

This works to avoid conflicts between resolve queries provided in load APIs and resolve queries defined at the schema level, and I agree we should go this route.

The other case where we may want to merge resolve queries is with nested schemas, when both contain resolve queries. I feel like that's an expectation people could have when nesting schemas that contain resolve queries inside other schemas (otherwise, there's no difference between nesting the original schema or the schema with the resolve query). And it'll also be necessary if we want to have derived schema properties that are present even when you load that schema through a parent schema.

packages/jazz-tools/src/svelte/jazz.class.svelte.ts
nrainhart: Had to split the tests into a Svelte component test (otherwise it's impossible to test the classes' behavior) and a type test: [`8f6b7e4` (#3147)](https://github.com/garden-co/jazz/pull/3147/commits/8f6b7e4a047e3f145a2762374b538ec262152b75)

packages/jazz-tools/src/react-core/tests/useCoState.test.ts
nrainhart: [`2b24c5b` (#3147)](https://github.com/garden-co/jazz/pull/3147/commits/2b24c5b11f5366ec207a4215c4893b380ed3ba46)

packages/jazz-tools/src/tools/tests/schema.resolved.test.ts
nrainhart: [`b61c772` (#3147)](https://github.com/garden-co/jazz/pull/3147/commits/b61c772d3279388332dcaa80f1cbd2baffd3ac91)

gdorsi (approved):
LGTM, good job!

PR #3151, opened 2025-11-04
fix: handle correctly permission verification in groups when members are revoked

gdorsi:
Found a group membership that wasn't validated correctly.

With some divide & conquer I've isolated the issue on the parent group membership calculation.

Replacing the current membership algorithm with an actual group fixes the issue.

Noticed that there are a bunch of cases we don't currently handle:
- the membership isn't calculated based on the time of the transaction to verify
- if an account membership is revoked in the child group, their transactions on the child group are considered invalid even if they are admin in the parent group
- when the parent groups are partially loaded, we validate against only a partial of the transactions (moved in another PR)

### DIFF
--- a/packages/cojson/src/permissions.ts
+++ b/packages/cojson/src/permissions.ts
@@ -8,6 +8,7 @@ import {
   ParentGroupReferenceRole,
   RawGroup,
   isInheritableRole,
+  isSelfExtension,
 } from "./coValues/group.js";
 import { KeyID } from "./crypto/crypto.js";
 import {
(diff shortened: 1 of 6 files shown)


Takeno (approved):
LGTM

packages/cojson/src/permissions.ts
```diff
@@ -193,79 +192,81 @@ function isHigherRole(a: Role, b: Role | undefined) {
...
-      } else if (isHigherRole(parentRole, currentRole)) {
-        memberState[agent] = parentRole;
+      if (!parentRole || !isInheritableRole(parentRole)) {
+        continue;
+      }
+
+      const resolvedParentRole =
+        roleMapping === "extend" ? parentRole : roleMapping;
+
+      if (isHigherRole(resolvedParentRole, role)) {
```
nrainhart, line 232: We could bail out earlier if the role is admin, right?

PR #3153, opened 2025-11-04
Add integration with existing auth docs page

joeinnes:
# Description
Adds a new docs page describing how one could integrate Jazz with an existing authentication system by programmatically creating Jazz accounts.

Also fixes a regression on framework selector, and adds some framework specific components for minimising repetition in TabbedCodeGroups.

Check the new docs page under Authentication

## Tests

- [X] Tests have not been updated, because: N/A

## Checklist

### DIFF
--- a/homepage/homepage/content/docs/code-snippets/key-features/authentication/integrating-with-existing-auth/index.ts
+++ b/homepage/homepage/content/docs/code-snippets/key-features/authentication/integrating-with-existing-auth/index.ts
@@ -1,0 +1,29 @@
+import { MyAppAccount } from "./schema";
+import { startWorker } from "jazz-tools/worker";
+
+const { worker } = await startWorker({
+  accountID: process.env.JAZZ_WORKER_ACCOUNT,
+  accountSecret: process.env.JAZZ_WORKER_SECRET,
+});
+
+// Get all users from your existing database
+// @ts-expect-error This is a virtual implementation
+const allUsers = await db.query.users.findMany();
+
+for (const user of allUsers) {
+  const { credentials } = await MyAppAccount.createAs(worker, {
+    creationProps: { name: user.name },
+  });
+  const { accountID, accountSecret } = credentials;
+  // @ts-expect-error This is a virtual implementation
+  const encryptedSecret = await encrypt(accountSecret); // use your implementation here
+  // Persist the Jazz account ID and secret in your existing database
+  // @ts-expect-error This is a virtual implementation
+  await db.query.users.update({
+    where: { id: user.id },
+    data: {
+      jazzAccountID: accountID,
+      jazzAccountSecret: encryptedSecret,
+    },
+  });
+}
--- a/homepage/homepage/content/docs/code-snippets/key-features/authentication/integrating-with-existing-auth/schema.ts
+++ b/homepage/homepage/content/docs/code-snippets/key-features/authentication/integrating-with-existing-auth/schema.ts
@@ -1,0 +1,30 @@
+import { co, Group, z } from "jazz-tools";
+
+const MyAppAccountRoot = co.map({
+  someMap: co.map({}),
+});
+export const MyAppAccount = co
+  .account({
+    root: MyAppAccountRoot,
+    profile: co.profile(),
+  })
+  .withMigration(async (account, creationProps?: { name: string }) => {
+    if (!account.$jazz.has("root")) {
+      if (!process.env.JAZZ_WORKER_ACCOUNT)
+        throw new Error("JAZZ_WORKER_ACCOUNT is not set");
+      // Load the worker
+      const worker = await co.account().load(process.env.JAZZ_WORKER_ACCOUNT);
+      const group = Group.create();
+      // Add it as a member of a group
+      worker.$isLoaded && group.addMember(worker, "admin");
+      // Create the root using the group to grant your server worker admin access on the user's account root.
+      const myRoot = MyAppAccountRoot.create(
+        {
+          someMap: {},
+        },
+        group,
+      );
+
+      account.$jazz.set("root", myRoot);
+    }
+  });
--- a/homepage/homepage/content/docs/code-snippets/key-features/authentication/integrating-with-existing-auth/schema.ts
+++ b/homepage/homepage/content/docs/code-snippets/key-features/authentication/integrating-with-existing-auth/schema.ts
@@ -16,6 +16,7 @@ export const MyAppAccount = co
       const worker = await co.account().load(process.env.JAZZ_WORKER_ACCOUNT);
       const group = Group.create();
       // Add it as a member of a group
+      // Note that by doing this, we grant the server worker full access to the user's account root. Consider whether this is appropriate for your use case.
       worker.$isLoaded && group.addMember(worker, "admin");
       // Create the root using the group to grant your server worker admin access on the user's account root.
       const myRoot = MyAppAccountRoot.create(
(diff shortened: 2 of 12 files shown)


homepage/homepage/content/docs/key-features/authentication/integrating-with-existing-auth.mdx
```diff
@@ -0,0 +1,130 @@
...
+const { worker } = await startWorker({
+  accountID: PUBLIC_JAZZ_WORKER_ACCOUNT,
+  accountSecret: JAZZ_WORKER_SECRET,
+});
+
+// Get all users from your existing database
+const allUsers = await db.query.users.findMany();
+
+for (const user of allUsers) {
+  const { accountID, accountSecret } = await Account.createAs(worker, { name: user.name });
```
Takeno: I discussed with @gdorsi about `createAs`, which isn't resilient across multiple calls. Is it really, Guido?

nit: I would add the import or the definition of `Account`, and refer to the account's migration system as well.

homepage/homepage/content/docs/key-features/authentication/integrating-with-existing-auth.mdx
```diff
@@ -0,0 +1,130 @@
...
+import { useJazzContext } from 'jazz-tools/react';
+
+function logIn(username, password) {
+  // Whatever your existing auth system is
+  const myOldAppUser = myApp.logIn(username, password);
+  const accountID = myOldAppUser.jazzAccountID;
+  // If you've stored this in an encrypted form, make sure to decrypt it first
+  const accountSecret = myOldAppUser.jazzAccountSecret;
+  const context = useJazzContext();
+  context.authenticate({ accountID, accountSecret });
```
Takeno: I need to double-check, but if you don't pass the "provider" property (or use "anonymous"), the user will not be considered authenticated 🤔

homepage/homepage/content/docs/key-features/authentication/integrating-with-existing-auth.mdx
```diff
@@ -0,0 +1,130 @@
...
+```tsx
+import { useJazzContext } from 'jazz-tools/react';
+
+function logIn(username, password) {
+  // Whatever your existing auth system is
+  const myOldAppUser = myApp.logIn(username, password);
+  const accountID = myOldAppUser.jazzAccountID;
+  // If you've stored this in an encrypted form, make sure to decrypt it first
+  const accountSecret = myOldAppUser.jazzAccountSecret;
+  const context = useJazzContext();
```
Takeno: nit: I think the concepts are clear, but people like to copy-paste snippets

```diff
+ const context = useJazzContext();
- function logIn(username, password) {
+ async function logIn(username, password) {
  // Whatever your existing auth system is
-  const myOldAppUser = myApp.logIn(username, password);
+  const myOldAppUser = await myApp.logIn(username, password);
  const accountID = myOldAppUser.jazzAccountID;
  // If you've stored this in an encrypted form, make sure to decrypt it first
  const accountSecret = myOldAppUser.jazzAccountSecret;
-   const context = useJazzContext();
```

homepage/homepage/content/docs/key-features/authentication/integrating-with-existing-auth.mdx
joeinnes: Yup! Thanks!

https://github.com/garden-co/jazz/blob/feecdaeb55fc8a360f4a117fee978c875dc50c9e/packages/jazz-tools/src/tools/auth/AuthSecretStorage.ts#L135

homepage/homepage/content/docs/key-features/authentication/integrating-with-existing-auth.mdx
gdorsi: Correct!

I could quickly fix that

gdorsi: And also agree on the migration bits, it is the only way to setup the initial root and profile and give the worker write access to those.

joeinnes: Added some verbiage there around setting up the migration with a short illustration. It would be useful to have `createAs` stable to enable patterns like this. If it's a quick fix, I think that would be better, but I can also rewrite to show 'on-demand' account creation when users log in to the legacy system.

joeinnes: Does https://github.com/garden-co/jazz/pull/3216 fix this? If so I'll fix the merge conflicts ready for review

gdorsi: Yes, it does

homepage/homepage/content/docs/code-snippets/key-features/authentication/integrating-with-existing-auth/schema.ts
```diff
@@ -0,0 +1,30 @@
...
+      worker.$isLoaded && group.addMember(worker, "admin");
+      // Create the root using the group to grant your server worker admin access on the user's account root.
+      const myRoot = MyAppAccountRoot.create(
+        {
+          someMap: {},
+        },
+        group,
+      );
+
+      account.$jazz.set("root", myRoot);
```
Takeno, line 29: Adding the worker as Admin might not be the desired behaviour. It is the case when the server wants to manage data in the future in a hybrid approach.
But there are also cases where Jazz fully manages an application slice and only wants to create the credentials programmatically. 
I wonder if this part needs to be explained more clearly, given its security implications (a server account that can read everything).

joeinnes: Added a note :)

homepage/homepage/content/docs/code-snippets/key-features/authentication/integrating-with-existing-auth/index.ts
```diff
@@ -0,0 +1,29 @@
...
+});
+
+// Get all users from your existing database
+// @ts-expect-error This is a virtual implementation
+const allUsers = await db.query.users.findMany();
+
+for (const user of allUsers) {
+  const { credentials } = await MyAppAccount.createAs(worker, {
+    creationProps: { name: user.name },
+  });
```
gdorsi, line 16: You can give to the worker admin access here:

```ts
const { credentials } = await MyAppAccount.createAs(worker, {
  creationProps: { name: user.name },
  onCreate(account, worker) {
    account.root.$jazz.owner.addMember(worker, "admin")
  }
});
```

Or even create the root on the worker side:
```ts
const root = MyAppAccountRoot.create({
  userData: {}
}, Group.create(worker));

const { credentials, account } = await MyAppAccount.createAs(worker, {
  creationProps: { name: user.name },
  onCreate(account) {
    account.$jazz.set("root", root) // Set the initial root
  }
});

root.$jazz.owner.addMember(account, "writer"); // This way the user doesn't have admin rights on the root
```

gdorsi (approved):
LGTM

homepage/homepage/content/docs/code-snippets/key-features/authentication/integrating-with-existing-auth/index.ts
Takeno: Agree. The `onCreate` usage is something really important introduced with `createAs` refactoring. That's the only moment you can set root in the account.

joeinnes: @gdorsi @Takeno will update to document this, and I think `onCreate` probably should be the canonical way we recommend doing it because it's easier and more explicit.

However, if we create a root during a migration during a `createAs` creation, the migration will run as the current active user (i.e. the worker), right? So in this example:

https://github.com/garden-co/jazz/pull/3153/changes#diff-6b353d816357d12353bb571b7159e7210dc4446bf9d7c6b927c1af0da5a9943fR17

the `Group` is actually owned by the worker, not the account itself?

So this would be the most straight forward pattern:

```ts
const { credentials } = await MyAppAccount.createAs(worker, {
  creationProps: { name: user.name },
  onCreate(account, worker) {
    // If this root was created during a migration with `createAs`, then this line I *think* is unnecessary, and we only need to add the account itself to its own root, right? The only reason we'd need to do this would be if we do something unusual during a migration and assign the root ownership to something other than the current account at migration time?
    account.root.$jazz.owner.addMember(worker, "admin");
    // This I think we still need
    account.root.$jazz.owner.addMember(account, "writer");
  }
});
```

Is that all correct?

PR #3156, opened 2025-11-04
Added an utility to generate large comaps and a test to cover streaming from storage with multiple sessions

gdorsi:


nrainhart (approved):
🚀

PR #3162, opened 2025-11-06
fix: keep the same instance of content when revalidating

gdorsi:
Since now we had this weird thing that, when the group updates a new "currentContent" is created.

This has always been a pain while testing, because something like this would fail in an hard-to-understand way:
```ts
test("the map should rebuild when the group permissions change", async () => {
  const alice = setupTestNode({
    connected: true,
  });
  const bob = setupTestNode({
    connected: true,
  });
  const group = alice.node.createGroup();
  const map = group.createMap();
  
  group.addMember("everyone", "writer");  // new content here!

  const mapOnBob = await loadCoValueOrFail(bob.node, map.id);
  map.set("fromBob", true);

  await waitFor(() => {
    expect(map.get("fromBob")).toBeTrue() // This is no more the current content, so it won't be never updated
  });
});
```

The same thing happens on `jazz-tools`.

In this PR I'm changing the approach to reset the state of the current view, keeping the same reference.
Had to introduce a `version` field to let SubscriptionScope know when we've got a "revalidation" because only checking the `totalValidTransactions` is not enough for some edge cases (when after the revalidation we have the same amount of valid transactions but different content, yes we already have a test covering this case).

We've got no one complaining about this, besides me, but since it can make things fail in an unpredictable way I think it's time to improve the flow.

### DIFF
--- a/packages/cojson/src/coValues/coMap.ts
+++ b/packages/cojson/src/coValues/coMap.ts
@@ -40,10 +40,6 @@ export class RawCoMap<
   type = "comap" as const;
   /** @category 6. Meta */
   core: AvailableCoValueCore;
-  /** @internal */
-  latest: {
-    [Key in keyof Shape & string]?: MapOp<Key, Shape[Key]>;
-  };
 
   /** @internal */
   get latestTxMadeAt(): number {
@@ -55,12 +51,25 @@ export class RawCoMap<
     return this.core.earliestTxMadeAt;
   }
 
-  /** @internal */
+  /** The internal state of the RawCoMap */
   ops: {
     [Key in keyof Shape & string]?: MapOp<Key, Shape[Key]>[];
-  };
-  /** @internal */
-  knownTransactions: Record<RawCoID, number>;
+  } = {};
+  latest: {
+    [Key in keyof Shape & string]?: MapOp<Key, Shape[Key]>;
+  } = {};
+  // We track the knownTransactions in multiple CoValues because branches
+  // need to retrieve the transactions from both the source and the branch
+  knownTransactions: Record<RawCoID, number> = {};
+  totalValidTransactions: number = 0;
+  version: number = 0;
+
+  private resetInternalState() {
+    this.ops = {};
+    this.latest = {};
+    this.knownTransactions = { [this.core.id]: 0 };
+    this.totalValidTransactions = 0;
+  }
 
   /** @internal */
   ignorePrivateTransactions: boolean;
@@ -68,9 +77,6 @@ export class RawCoMap<
   atTimeFilter?: number = undefined;
   /** @category 6. Meta */
   readonly _shape!: Shape;
-
-  totalValidTransactions: number = 0;
-  version: number = 0;
 
   /** @internal */
   constructor(
@@ -84,12 +90,6 @@ export class RawCoMap<
 
     this.ignorePrivateTransactions =
       options?.ignorePrivateTransactions ?? false;
-    this.ops = {};
-    this.latest = {};
-
-    // We track the knownTransacions in multiple CoValues because branches
-    // need to retrieve the transactions from both the source and the branch
-    this.knownTransactions = { [core.id]: 0 };
 
     this.processNewTransactions();
   }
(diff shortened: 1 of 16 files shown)


packages/cojson/src/coValues/coStream.ts
```diff
@@ -69,6 +69,14 @@ export class RawCoStreamView<
     this.processNewTransactions();
   }
 
+  rebuildFromCore() {
+    this.version++;
+    this.items = {};
+    this.knownTransactions = { [this.core.id]: 0 };
+    this.totalValidTransactions = 0;
+    this.processNewTransactions();
```
nrainhart: This is the same thing we do in the constructor, right? Since we can't extract it into a function, we should add a comment to avoid forgetting to keep both in sync

packages/cojson/src/coValues/coMap.ts
```diff
@@ -353,18 +354,7 @@ export class RawCoMapView<
       listener(core.getCurrentContent() as this);
     });
   }
-}
 
-/** A collaborative map with precise shape `Shape` and optional static metadata `Meta` */
-export class RawCoMap<
```
nrainhart, line 359: No more split between read & write methods? I don't have a strong opinion either way, but curious if there was any reason to do this now

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
```diff
@@ -172,6 +173,7 @@ export class SubscriptionScope<D extends CoValue> {
     } else {
       const hasChanged =
         update.totalValidTransactions !== this.totalValidTransactions ||
+        update.version !== this.version ||
         // Checking the identity of the raw value makes us cover the cases where the group
         // has been updated and the coValues that don't update the totalValidTransactions value (e.g. FileStream)
         this.value.value.$jazz.raw !== update;
```
nrainhart: Is this still necessary?

packages/cojson/src/tests/sync.load.test.ts
```diff
@@ -481,10 +481,7 @@ describe("loading coValues from server", () => {
     map.set("fromServer", "updated", "trusting");
 
     await waitFor(() => {
-      const coValue = expectMap(
-        client.node.expectCoValueLoaded(map.id).getCurrentContent(),
-      );
-      expect(coValue.get("fromServer")).toEqual("updated");
+      expect(map.get("fromServer")).toEqual("updated");
```
nrainhart, line 484: ✨

nrainhart (approved):
Nice!!

packages/cojson/src/coValues/coMap.ts
gdorsi: Because this was making the refactoring to make `processNewTransactions` and `rebuildFromCore` more complicated.

I don't have a strong opinion either, but I've never got value from this separation and it's only been an hindrance so far.

Good observation thou, I should have added context on the PR for this change.

packages/cojson/src/coValues/coStream.ts
gdorsi: Good point, going to try to extract this into a function.

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
gdorsi: Probably no, let's try to remove it!

packages/cojson/src/coValues/coStream.ts
nrainhart: I don't think you'll be able to. Typescript is very strict with field assignments in constructors 👎

gdorsi: This should make it more clear that internal state needs to be part of the reset: https://github.com/garden-co/jazz/pull/3162/commits/94afaebf462eee9cbbe23012e5a5ba503c67cf05

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
gdorsi: Removed: https://github.com/garden-co/jazz/pull/3162/commits/c3cd9127ebfa8fdc8b8956a3623996c27b742647

PR #3166, opened 2025-11-06
[Inspector] CoMaps editor

Takeno:
Untitled

Two new features for CoMaps in the inspector:
- in app add/remove/edit properties
- Time traveling (multiple CoMap set/delete in a single transaction)


packages/cojson/src/coValues/coList.ts
```diff
@@ -188,6 +192,13 @@ export class RawCoList<
   }
 
   processNewTransactions() {
+    if (this.isTimeTravelEntity()) {
+      throw new Error("Cannot process transactions on a time travel entity");
+    }
```
gdorsi: This check is not necessary, because the new changes would be filtered out anyway.

packages/cojson/src/coValues/coList.ts
```diff
@@ -298,8 +323,13 @@ export class RawCoList<
    *
    * @category 4. Time travel
    */
-  atTime(_time: number): this {
-    throw new Error("Not yet implemented");
+  atTime(time: number): this {
+    if (time >= (this.lastValidTransaction ?? 0)) {
+      return this;
```
gdorsi: This could potentially lead to bugs, because this doesn't have a time filter and new transactions might "overflow".

We needed this check in CoMap because time traviling there is used to rollback Group in time and use it to validate permissions rights on transactions.

This case is not that performance-critical, so it's ok to go always for the slow path (for now)

gdorsi (approved):
Fantastic!

Going to merge right away!

PR #3167, opened 2025-11-06
Add image upload capability to the RN and Expo example apps

joeinnes:
# Description
While working through the docs and trying to make them more friendly towards other frameworks, I realised that the RN and Expo chat examples don't have the image upload we have for React and Svelte. This PR adds it.

If you are testing this in the monorepo, you must delete @bam.tech/react-native-image-resizer from the root node_modules because otherwise this https://github.com/garden-co/jazz/blob/df7ce42a4071b5cbb58a1dac948d1eb730a2266d/packages/jazz-tools/src/media/create-image/react-native.ts#L52 will succeed and then it won't work because that package doesn't work with Expo any more.

## Tests

- [x] Tests have not been updated, because: no tests to update

## Checklist


package.json
```diff
@@ -92,5 +92,8 @@
       "wasm-pack",
       "workerd"
     ]
+  },
+  "dependencies": {
+    "expo-image-manipulator": "^14.0.7"
```
nrainhart: @joeinnes was this change intended? I don't think we should be adding `expo-image-manipulator` as a dependency for the whole project

nrainhart (changes requested):

examples/chat-rn-expo/src/chat.tsx
```diff
@@ -59,6 +102,44 @@ export default function ChatScreen() {
...
+
+      const result = await launchImageLibrary({
+        mediaType: "photo",
+        quality: 0.8,
+      });
+
+      if (!result.didCancel && result.assets?.[0].uri) {
+        const image = await createImage(result.assets[0].uri, {
+          owner: loadedChat.$jazz.owner,
+          // Work out why I can't do this in Expo but in RN it works fine.
```
nrainhart: Should we remove this comment?

nrainhart (approved):
Changes look good!

PR #3173, opened 2025-11-12
feat: add CoValue serialization for Server Component -> Client Component

gdorsi:
This PR makes it possible to pass CoValues from Server Components to Client Components.

This way it is possible to do SSR of CoValue data that's also reactive.

Example:
```ts
export default async function ServerSidePage(props: {
  params: Promise<{ id: string }>;
}) {
  const { id } = await props.params;
  const account = await JazzAccount.load(id, {
    loadAs: jazzSSR,
  });

  if (!account.$isLoaded) {
    return <div>{account.$jazz.loadingState}</div>;
  }

  return (
    <TodoList id={account.$jazz.id} preloaded={account.$jazz.export()} />
  );
}
```

On the client component:
```ts
export function TodoList({
  id,
  preloaded,
}: {
  id: string;
  preloaded?: ExportedCoValue;
}) {
  const me = useAgent();
  const list = useCoState(JazzAccount, id, { // The list is available during the SSR
    preloaded,
    select: (account) => {
      if (!account.$isLoaded) return account;

      return account.profile.todos;
    },
  });
```

Live example: https://jazz-nextjs-git-feat-ssr-data-passing-gcmp.vercel.app/account/co_zFwYcUJSgEC5F179Hamwx66c2SB

Only works with public data, for now...


examples/jazz-nextjs/src/app/layout.tsx
```diff
@@ -2,6 +2,7 @@ import type { Metadata } from "next";
 import { Geist, Geist_Mono } from "next/font/google";
 import "./globals.css";
 import { Jazz } from "../jazz";
+import { ErrorBoundary } from "next/dist/client/components/error-boundary";
```
nrainhart: Don't think this is necessary

packages/jazz-tools/src/inspector/index.tsx
```diff
@@ -30,6 +30,15 @@ export function JazzInspector({ position = "right" }: { position?: Position }) {
   const localNode = context.node;
   const me = "me" in context ? context.me : undefined;
 
+  const [isCSR, setIsCSR] = useState(false);
+  useEffect(() => {
+    setIsCSR(true);
+  }, []);
```
nrainhart, line 36: ⚠️ Won't this always hide the inspector?

packages/jazz-tools/src/tools/coValues/interfaces.ts
```diff
@@ -710,6 +709,29 @@ export async function exportCoValue<
...
+  return {
+    id: subscription.id,
+    contentPieces,
+  };
+}
+
+export type ExportedCoValue = {
+  id: string;
+  contentPieces: CojsonInternalTypes.NewContentMessage[];
+};
```
nrainhart: Right now exported covalues are opaque, so there's no way to know which schema to load it with, unless you know how it was generated. We should at least keep the CoValue type as part of the `ExportedCoValue` type, and maybe also add some metadata about the CoValue schema?

packages/jazz-tools/src/tools/tests/exportImport.test.ts
```diff
@@ -262,6 +263,83 @@ describe("exportCoValue", () => {
...
+    const person = Person.create({ name: "Jane", age: 28 }, group);
+
+    const exported = person.$jazz.export();
+
+    expect(exported.id).toBe(person.$jazz.id);
+
+    const bob = await createJazzTestAccount();
+    importContentPieces(exported.contentPieces, bob);
+
+    const loadedPerson = await Person.load(person.$jazz.id, { loadAs: bob });
```
nrainhart, line 287: Should load methods support the preloaded option as well, to be consistent?

nrainhart (commented):
I really like the idea of supporting exports, not just for SSR->CSR but also for general use. It could be an interesting way to share CoValues without relying on a sync server. I do have some concerns with the current implementation

packages/jazz-tools/src/inspector/index.tsx
gdorsi: Only on SSR, the effects run when the rendering is client side

packages/jazz-tools/src/tools/coValues/interfaces.ts
gdorsi: My issue was that I didn't want to pollute the type with virtual fields, but managed to get it working with a branded type for the id

packages/jazz-tools/src/tools/tests/exportImport.test.ts
gdorsi: This can come later, it's out of scope for this change.

nrainhart (approved):
The branded type's great, thanks for adding it!

packages/jazz-tools/src/tools/coValues/interfaces.ts
gdorsi: https://github.com/garden-co/jazz/pull/3173/commits/b01fdd8dabf01f3fc336bc6ccda84297712a8cdf

PR #3179, opened 2025-11-13
Fix homepage build issues

joeinnes:
# Description
Homepage builds are failing since merging 3178 but 3178 built successfully in preview: https://vercel.com/gcmp/jazz-homepage/4GFURBFsp8apUe3zuKqcnKDyCyKW

I still don't know why the `// @ts-expect-error` is causing issues on Vercel but not locally (in fact, the opposite), but this will unblock homepage builds while I work it out.

Check the homepage builds in production.

## Tests

## Checklist

### DIFF
--- a/homepage/homepage/content/docs/code-snippets/core-concepts/covalues/imagedef/rn.tsx
+++ b/homepage/homepage/content/docs/code-snippets/core-concepts/covalues/imagedef/rn.tsx
@@ -29,10 +29,15 @@ async function handleImagePicker() {
     quality: 1,
   });
 
-  if (!result.didCancel && result.assets && result.assets.length > 0 && me.profile.$isLoaded) {
+  if (
+    !result.didCancel &&
+    result.assets &&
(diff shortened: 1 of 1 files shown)


joeinnes:
OK, figured it out.
#3167 introduced `react-native-image-picker` as a dependency in the example chat apps. This causes it to now be installed when you run `pnpm install` in the monorepo.

As a result, it's now available for TypeScript to properly type the import, so the `// @ts-expect-error` line is no longer appropriate (as there's no longer an error).

This had the side effect of uncovering another issue which was that I wasn't checking that the URI was a string, and it can technically be undefined, causing the second TS failure.

https://vercel.com/gcmp/jazz-homepage/4GFURBFsp8apUe3zuKqcnKDyCyKW built fine because it was build purely based on branch 3178, and the issue is actual an interplay between the changes introduced in 3167 and 3178.

mculotta120 (approved):

PR #3189, opened 2025-11-18
fix: avoid 'dianamically' loaded demo section

Elfo404:
Currently the demo section is rendered only when the chat ids are generated, the qrcode available, etc... this causes a bif of a shift when opening the page (especially in slower connections). this PR should fix it by rendering empty iframes and a white "qr code" while waiting for data, instead of rendering nothing

Before:
https://github.com/user-attachments/assets/9f19cb06-3fa9-4d5a-86e6-365a9874daaf

After:
https://github.com/user-attachments/assets/7465225d-9c1c-4727-a131-a0ea8dd9e7fd


joeinnes (approved):
LGTM

PR #3192, opened 2025-11-19
[Inspector] avoid crashes when covalue has not group

Takeno:
# Description
When the inspector doesn't recognize a CoValue as an account, it is shown as CoMap. Accounts have no `.group`, and the inspector crashed.
This PR addresses this case, improving stability during debugging unstable covalues.

### DIFF
--- a/.changeset/eighty-forks-shake.md
+++ b/.changeset/eighty-forks-shake.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+In the inspector, accounts are now identified by header's meta type
--- a/packages/jazz-tools/src/inspector/viewer/use-resolve-covalue.ts
+++ b/packages/jazz-tools/src/inspector/viewer/use-resolve-covalue.ts
@@ -48,10 +48,6 @@ export type ResolvedAccount = {
   [key: string]: JSON;
 };
 
-export const isAccount = (coValue: JSONObject): coValue is ResolvedAccount => {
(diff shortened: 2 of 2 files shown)


packages/jazz-tools/src/inspector/viewer/page.tsx
```diff
@@ -238,22 +238,24 @@ export function Page(props: PageProps) {
...
-                onClick={() => {
-                  onNavigate([{ coId: value.group.id, name: "owner" }]);
-                }}
-              />
-            </Text>
-          </>
-        )}
+        {extendedType !== "account" &&
+          extendedType !== "group" &&
+          value.core.safeGetGroup()?.id && (
```
nrainhart: Do you know in which case the type is not an account, but the CoValue is actually an account? May be worth leaving a comment about it here

nrainhart (approved):

packages/jazz-tools/src/inspector/viewer/page.tsx
gdorsi: The value type detection was based in the profile property, which wasn't very stable.

He switched to checking the header, which is the right way to detect an account.

gdorsi (approved):

packages/jazz-tools/src/inspector/viewer/page.tsx
gdorsi: Maybe we should revert these safeGroup access?

Having it throw helped us on finding the bug in the CoValue detection

PR #3196, opened 2025-11-20
refactor: move unloaded management inside the SubscriptionScope

gdorsi:
In order to make it easier to implement `value.$jazz.promise` I'm moving the unloaded management inside of SubscriptionScope.

Unloaded values coming from there will always have a stable reference, which makes it easier to build a promise getter that also returns a stable promise.

### DIFF
--- a/packages/jazz-tools/src/react-core/hooks.ts
+++ b/packages/jazz-tools/src/react-core/hooks.ts
@@ -26,7 +26,7 @@ import {
   SchemaResolveQuery,
   SubscriptionScope,
   coValueClassFromCoValueClassOrSchema,
-  createUnloadedCoValue,
+  getUnloadedCoValueWithoutId,
   type BranchDefinition,
 } from "jazz-tools";
 import { JazzContext, JazzContextManagerContext } from "./provider.js";
@@ -172,39 +172,15 @@ export function useCoValueSubscription<
   return subscription.subscription as CoValueSubscription<S, R>;
 }
 
-function getSubscriptionValue<C extends CoValue>(
-  subscription: SubscriptionScope<C> | null,
-): MaybeLoaded<C> {
-  if (!subscription) {
-    return createUnloadedCoValue("", CoValueLoadingState.UNAVAILABLE);
-  }
-  const value = subscription.getCurrentValue();
-  if (typeof value === "string") {
-    return createUnloadedCoValue(subscription.id, value);
-  }
-  return value;
-}
-
 function useGetCurrentValue<C extends CoValue>(
   subscription: SubscriptionScope<C> | null,
 ) {
-  const previousValue = useRef<MaybeLoaded<CoValue> | undefined>(undefined);
-
   return useCallback(() => {
-    const currentValue = getSubscriptionValue(subscription);
-    // Avoid re-renders if the value is not loaded and didn't change
-    if (
-      previousValue.current !== undefined &&
-      previousValue.current.$jazz.id === currentValue.$jazz.id &&
-      !previousValue.current.$isLoaded &&
-      !currentValue.$isLoaded &&
-      previousValue.current.$jazz.loadingState ===
-        currentValue.$jazz.loadingState
-    ) {
-      return previousValue.current as MaybeLoaded<C>;
+    if (!subscription) {
+      return getUnloadedCoValueWithoutId(CoValueLoadingState.UNAVAILABLE);
     }
-    previousValue.current = currentValue;
-    return currentValue;
+
+    return subscription.getCurrentValue();
   }, [subscription]);
 }
 
--- a/packages/jazz-tools/src/tools/coValues/interfaces.ts
+++ b/packages/jazz-tools/src/tools/coValues/interfaces.ts
@@ -107,6 +107,23 @@ export function isCoValueClass<V extends CoValue>(
  */
 export type ID<T> = string;
 
+const unloadedCoValueStates = new Map<
+  NotLoadedCoValueState,
+  NotLoaded<CoValue>
+>();
+
+export function getUnloadedCoValueWithoutId<T extends CoValue>(
+  loadingState: NotLoadedCoValueState,
+): NotLoaded<T> {
+  const value = unloadedCoValueStates.get(loadingState);
+  if (value) {
+    return value;
+  }
+  const newValue = createUnloadedCoValue("", loadingState);
+  unloadedCoValueStates.set(loadingState, newValue);
+  return newValue;
+}
+
 export function createUnloadedCoValue<T extends CoValue>(
   id: ID<T>,
   loadingState: NotLoadedCoValueState,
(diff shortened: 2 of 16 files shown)


packages/jazz-tools/src/react-core/hooks.ts
```diff
@@ -172,39 +172,15 @@ export function useCoValueSubscription<
...
-      previousValue.current !== undefined &&
-      previousValue.current.$jazz.id === currentValue.$jazz.id &&
-      !previousValue.current.$isLoaded &&
-      !currentValue.$isLoaded &&
-      previousValue.current.$jazz.loadingState ===
-        currentValue.$jazz.loadingState
-    ) {
-      return previousValue.current as MaybeLoaded<C>;
+    if (!subscription) {
+      return getUnloadedCoValueWithoutId(CoValueLoadingState.UNAVAILABLE);
```
gdorsi, line 180: This value is cached, so should always return a stable reference now

packages/jazz-tools/src/react-core/hooks.ts
```diff
@@ -172,39 +172,15 @@ export function useCoValueSubscription<
...
     }
-    previousValue.current = currentValue;
-    return currentValue;
+
+    return subscription.getCurrentValue();
```
gdorsi, line 183: This now returns a stable reference of a MaybeLoaded value

packages/jazz-tools/src/tools/coValues/interfaces.ts
```diff
@@ -107,6 +107,23 @@ export function isCoValueClass<V extends CoValue>(
...
+  loadingState: NotLoadedCoValueState,
+): NotLoaded<T> {
+  const value = unloadedCoValueStates.get(loadingState);
+  if (value) {
+    return value;
+  }
+  const newValue = createUnloadedCoValue("", loadingState);
+  unloadedCoValueStates.set(loadingState, newValue);
+  return newValue;
+}
```
gdorsi, line 125: Made this helper to have stable references for unloaded values when the subscription scope is not available

packages/jazz-tools/src/tools/coValues/promise.ts
```diff
@@ -0,0 +1,34 @@
+export class CoValuePromise<T> extends Promise<T> {
```
gdorsi, line 1: This is unused now, but going to use this API in a stacked PR

packages/jazz-tools/src/tools/coValues/request.ts
```diff
@@ -621,20 +616,29 @@ async function loadWorkerAccountOrGroup(id: string, loadAs: Account) {
   const coValue = await node.loadCoValueCore(id as `co_z${string}`);
 
   if (!coValue.isAvailable()) {
-    return createUnloadedCoValue(id, CoValueLoadingState.UNAVAILABLE);
+    throw new JazzRequestError("Worker account not found", 400);
```
gdorsi, line 619: Refactored to get rid of `createUnloadedCoValue` here

packages/jazz-tools/src/tools/implementation/refs.ts
```diff
@@ -53,29 +53,21 @@ export class Ref<out V extends CoValue> {
...
 
-    const value = node.value;
+    const value = node.getCurrentValue();
 
-    if (value?.type === CoValueLoadingState.LOADED) {
-      return value.value as V;
+    if (value.$isLoaded) {
+      return value as V;
     } else {
       return new Promise((resolve) => {
```
gdorsi, line 64: We will be able to replace this with `value.$jazz.promise` 🚀

packages/jazz-tools/src/tools/coValues/promise.ts
```diff
@@ -0,0 +1,34 @@
...
+  value: T | undefined;
+  reason: unknown | undefined;
+
+  static getRejected<T = never>(reason?: unknown): CoValuePromise<T> {
+    return new CoValuePromise<T>((resolve, reject) => {
+      reject(reason);
+    });
+  }
+
+  static getFulfilled<T>(value: T): CoValuePromise<T> {
```
nrainhart, line 12: Do we need these static methods? According to [the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject#description)
> Promise.reject() is generic and supports subclassing, which means it can be called on subclasses of Promise, and the result will be a promise of the subclass type

packages/jazz-tools/src/tools/coValues/promise.ts
```diff
@@ -0,0 +1,34 @@
...
+    return new CoValuePromise<T>((resolve) => {
+      resolve(value);
+    });
+  }
+
+  constructor(
```
nrainhart, line 18: Same thing with the constructor. Is it really necessary?

packages/jazz-tools/src/tools/implementation/refs.ts
```diff
@@ -86,7 +78,7 @@ export class Ref<out V extends CoValue> {
     }
   }
 
-  get value(): V | null | undefined {
+  get value(): MaybeLoaded<V> {
```
nrainhart, line 81: This is a breaking change, just want to double check that we are ok with releasing it in a patch release

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
```diff
@@ -283,7 +284,41 @@ export class SubscriptionScope<D extends CoValue> {
...
+      return this.unloadedValue;
+    }
+
+    const unloadedValue: NotLoaded<D> = {
+      $jazz: {
+        id: this.id,
+        loadingState: reason,
+      },
+      $isLoaded: false,
+    };
```
nrainhart, line 302: Nit: you can use `createUnloadedCoValue`

packages/jazz-tools/src/tools/tests/SubscriptionScope.test.ts
```diff
@@ -0,0 +1,397 @@
...
+    });
+
+    it("returns LOADING state when shouldSendUpdates returns false", () => {
+      const person = Person.create({ name: "John" });
+      const node = person.$jazz.raw.core.node;
+      const id = person.$jazz.id;
+      const schema = {
+        ref: coValueClassFromCoValueClassOrSchema(Person),
+        optional: false,
+      };
```
nrainhart, line 310: Nit: you could use `getSubscriptionScope` instead of creating the subscription scope by hand

nrainhart (commented):
Thanks for doing the refactor, and also for fixing the `by` field for edits and cofeed entries. Left only a few minor comments

nrainhart (approved):
Heh, meant to approve the PR

packages/jazz-tools/src/tools/coValues/promise.ts
gdorsi: Typescript doesn't agree with this.

I think it has to do with [this issue](https://github.com/tc39/proposal-rm-builtin-subclassing?tab=readme-ov-file#type-ii-subclass-instance-creation-in-built-in-methods)

TS can't know how the subclass is going to be resolved, so it plays safe.

packages/jazz-tools/src/tools/implementation/refs.ts
gdorsi: Before the type wasn't aligned with the actual return type of `accessChildById`

It is already returning `MaybeLoaded<V>`, I just have fixed the return type

packages/jazz-tools/src/tools/coValues/promise.ts
nrainhart: I think this is an error in Typescript's types, but it doesn't look like they care too much about it: https://github.com/microsoft/TypeScript/issues/45427

> TS can't know how the subclass is going to be resolved, so it plays safe.

.resolve and .reject could just return the `this` type

PR #3202, opened 2025-11-24
fix(cojson-storage-indexeddb): resolve error when adding key to uniqueSessions in multi-tab Jazz apps

gdorsi:
Switching `getSingleCoValueSession` in the indexeddb client to use the `readwrite` transaction we use for the updates.

This way it should be executed on the same transaction we use for updates, ensuring consistency [during the update transaction.](https://github.com/garden-co/jazz/blob/main/packages/cojson/src/storage/storageAsync.ts#L283)

### DIFF
--- a/packages/cojson-storage-indexeddb/src/idbClient.ts
+++ b/packages/cojson-storage-indexeddb/src/idbClient.ts
@@ -1,4 +1,9 @@
-import type { CojsonInternalTypes, RawCoID, SessionID } from "cojson";
+import type {
+  CojsonInternalTypes,
+  RawCoID,
+  SessionID,
+  DBTransactionInterfaceAsync,
+} from "cojson";
 import type {
   CoValueRow,
   DBClientInterfaceAsync,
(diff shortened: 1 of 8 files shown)


packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -25,22 +24,14 @@ export class IDBClient implements DBClientInterfaceAsync {
...
-    if (this.autoBatchingTransaction?.isReusable()) {
-      return this.autoBatchingTransaction.handleRequest<T>(handler);
-    }
-
-    const tx = new CoJsonIDBTransaction(this.db);
-
-    this.autoBatchingTransaction = tx;
-
-    return tx.handleRequest<T>(handler);
+    throw new Error("No active transaction");
```
gdorsi: Removed the autobatching logic, which wasn't used before.

Here we want to enforce all the changes to run in the same activeTransaction, to ensure consistency when the app is used on multiple tabs.

Takeno (approved):

PR #3205, opened 2025-11-25
perf: improve the permission checks performance by optimizing the parent groups retrival

gdorsi:
Introduced a new caching mechanism for parent groups to make the access to `roleOf` faster.

### DIFF
--- a/packages/cojson/src/coValues/coMap.ts
+++ b/packages/cojson/src/coValues/coMap.ts
@@ -1,5 +1,8 @@
 import { CoID, RawCoValue } from "../coValue.js";
-import { AvailableCoValueCore } from "../coValueCore/coValueCore.js";
+import {
+  AvailableCoValueCore,
+  DecryptedTransaction,
+} from "../coValueCore/coValueCore.js";
 import { AgentID, RawCoID, TransactionID } from "../ids.js";
 import { JsonObject, JsonValue } from "../jsonValue.js";
 import { accountOrAgentIDfromSessionID } from "../typeUtils/accountOrAgentIDfromSessionID.js";
@@ -64,7 +67,7 @@ export class RawCoMap<
   totalValidTransactions: number = 0;
   version: number = 0;
 
-  private resetInternalState() {
+  protected resetInternalState() {
     this.ops = {};
     this.latest = {};
     this.knownTransactions = { [this.core.id]: 0 };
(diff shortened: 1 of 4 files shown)


packages/cojson/src/coValues/coMap.ts
```diff
@@ -175,16 +183,12 @@ export class RawCoMap<
...
-      const clone = Object.create(this) as RawCoMap<Shape, Meta>;
+    const clone = Object.create(this) as RawCoMap<Shape, Meta>;
 
-      clone.atTimeFilter = time;
-      clone.latest = {};
+    clone.atTimeFilter = time;
+    clone.latest = {};
 
-      return clone as this;
-    }
```
gdorsi, line 187: The check wasn't considering the dependencies, so I had to switch on always setting the filter

packages/cojson/src/coValues/group.ts
```diff
@@ -161,6 +162,40 @@ function rotateReadKeyIfNeeded(group: RawGroup) {
   }
 }
 
+class TimeBasedEntry<T> {
```
gdorsi, line 165: Planning to reuse this somewhere else, e.g. to cache roles resolutions.

packages/cojson/src/coValues/group.ts
```diff
@@ -200,6 +248,44 @@ export class RawGroup<
     this.migrate();
   }
 
+  handleNewTransaction(transaction: DecryptedTransaction): void {
```
nrainhart: Nit
```suggestion
  override handleNewTransaction(transaction: DecryptedTransaction): void {
```

packages/cojson/src/coValues/group.ts
```diff
@@ -189,6 +224,19 @@ export class RawGroup<
 
   _lastReadableKeyId?: KeyID;
 
+  // Not using class field initializers because they run after that the CoMap constructor
+  // calls processNewTransactions, which would reset the parentGroups map
+  private declare parentGroups: Map<
```
nrainhart: How about naming this `parentGroupRoles`?

.kiro/specs/group-role-cache/design.md
nrainhart, line 1: Are we keeping these files in the repo?

nrainhart (approved):

packages/cojson/src/coValues/group.ts
gdorsi: Renamed into `parentGroupsChanges` 

It's not just about roles, it also store `extends` and the whole history of changes

PR #3206, opened 2025-11-25
Add feedback buttons to every article

joeinnes:
# Description
Adds a 'Was this helpful?' prompt at the bottom of each article with buttons 'yes' and 'no' which trigger analytics events

Look at a docs page and check for the links at the bottom of the content.

## Tests

## Checklist


mculotta120 (approved):
Looks good! I think in the future, especially if we use another analytics tool we should consider adding a comment box so we can get some input on why a page wasn't helpful.

PR #3220, opened 2025-11-28
fix: avoid migrating unauthorized CoValues

nrainhart:
# Description

I found we were attempting to migrate CoValues even if we don't have permission to access them. Usually this just led to doing unnecessary work, but for `co.discriminatedUnion` in particular it was causing the following error:
```
co.discriminatedUnion() of collaborative types with no matching discriminator value found
```

I added a check to prevent migrating CoValues the user has no access to.

## Tests

## Checklist

---

> [!NOTE]
> Prevents migrating/processing unauthorized CoValues in `SubscriptionScope` and adds tests for discriminated unions and partial access handling.
> 
> - **Core subscription logic (`packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts`)**:
>   - Skip migrations unless the value is accessible via new `hasAccessToCoValue()` check.
>   - Early-return unauthorized updates by replacing inline ruleset checks with `hasAccessToCoValue()`.
>   - Add `hasAccessToCoValue(rawCoValue)` helper centralizing access logic.
> - **Tests (`packages/jazz-tools/src/tools/tests/coDiscriminatedUnion.test.ts`)**:
>   - Add test asserting error when creating union member without matching discriminator field.
>   - Add test confirming `co.list` of discriminated unions loads with mixed accessibility, marking inaccessible items as `UNAUTHORIZED`.
> - **Release**:
>   - Changeset: `jazz-tools` patch bump.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e113a79d401ecd8058970e7d9b8ae93668cd1ece. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
+++ b/packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
@@ -91,7 +91,11 @@ export class SubscriptionScope<D extends CoValue> {
         // - Run the migration only once
         // - Skip all the updates until the migration is done
         // - Trigger handleUpdate only with the final value
-        if (!this.migrated && value !== CoValueLoadingState.UNAVAILABLE) {
+        if (
+          !this.migrated &&
+          value !== CoValueLoadingState.UNAVAILABLE &&
+          hasAccessToCoValue(value)
+        ) {
           if (this.migrating) {
             return;
           }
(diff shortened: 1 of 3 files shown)


Takeno (approved):
LGTM

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
```diff
@@ -759,3 +756,13 @@ export class SubscriptionScope<D extends CoValue> {
...
   }
 }
+
+function hasAccessToCoValue(rawCoValue: RawCoValue): boolean {
+  const ruleset = rawCoValue.core.verified.header.ruleset;
+
+  // Groups and accounts are accessible by everyone, for the other coValues we use the role to check access
+  return (
+    ruleset.type !== "ownedByGroup" ||
+    myRoleForRawValue(rawCoValue) !== undefined
```
joeinnes, line 766: Question here, do we also want to avoid running migrations if the user only has read access?

See: https://discord.com/channels/1139617727565271160/1139621689882321009/1445690193213653116

PR #3224, opened 2025-12-01
feat: add an error reporter facility to subscriptions

boorad:
# Description

This PR adds enhanced debugging capabilities for Jazz subscription errors by capturing stack traces at subscription creation points. When a CoValue becomes unavailable or unauthorized, developers can now see exactly which component or function created the subscription, making it much easier to debug common issues.

**Key improvements:**
- Stack traces captured at subscription creation (React hooks, `loadCoValue`, `loadUnique`, etc.)
- Enhanced error messages showing both subscription creation location and error occurrence location
- New `jazzErrorReporter` API for programmatically subscribing to Jazz errors
- All debugging overhead gated behind `NODE_ENV === "development"` checks for zero production impact

**Example output in development:**
```
🔴 JAZZ: Value became UNAVAILABLE
  CoValue ID: co_xyz...
  Schema: TaskList
  📍 Subscription created at: TaskListView.tsx:45
  Full subscription creation stack: ...
```

1. Set `NODE_ENV=development` in your Jazz app
2. Create a subscription to a CoValue that will become unavailable (e.g., access a CoValue with insufficient permissions or network issues)
3. Observe the enhanced error messages in the console showing:
   - Where the subscription was created (component/file location)
   - Full stack trace from subscription creation
4. Test the `jazzErrorReporter` API:
   ```typescript
   import { jazzErrorReporter } from 'jazz-tools';
   
   jazzErrorReporter.onError((error) => {
     console.log('Caught error:', error.type, error.message);
   });
   ```
5. Verify in production build (`NODE_ENV=production`) that no debug logging appears and stack traces are not captured

## Tests

- [x] Tests have not been updated, because: This adds debugging/observability features that enhance existing error reporting without changing functional behavior. The changes are primarily logging and developer experience improvements.

## Checklist

---

> [!NOTE]
> Adds configurable subscription error reporting with stack traces captured at React hook creation and standardized error messages, exposed via jazzConfig APIs.
> 
> - **Subscriptions/Error handling**:
>   - Capture caller stack at subscription creation (React hooks, account/coValue subscriptions) and attach to errors for prettier traces.
>   - Centralize logging in `SubscriptionScope.logError()` with de-duplication; respect `skipRetry` for unavailable; optionally report via custom reporter.
>   - Standardize messages: `Jazz Authorization Error`, `Jazz Unavailable Error`, and `Jazz Validation Error`, including subscription origin and path context.
> - **Config & APIs**:
>   - New `jazzConfig.setCustomErrorReporter` and `jazzConfig.enableCaptureErrorCause` to intercept/send errors and toggle stack capture.
>   - Export `captureStack`; add `tools/config.ts` and re-export as `jazzConfig`.
> - **React hooks**:
>   - `useCoValueSubscription` and `useAccountSubscription` now capture and pass caller stacks to `SubscriptionScope`.
> - **Tests**:
>   - Add `errorReporting.test.ts` and update existing tests to assert new messages and custom reporter behavior.
> - **Changeset**:
>   - Patch bump for `jazz-tools` with improved subscription error logging and stack traces.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bfb5b57b7955d472538b84bbd6d5772dfebd4442. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/jazz-tools/src/tools/subscribe/errorReporting.ts
```diff
@@ -0,0 +1,135 @@
+import type { CoValue, ID } from "../internal.js";
+import { CoValueLoadingState } from "./types.js";
+
+/**
+ * Check if we're in development mode.
+ * Stack traces are only captured in development to avoid overhead in production.
+ */
+export function isDev(): boolean {
+  return process.env.NODE_ENV === "development";
```
Gabrola: Not sure if it's a good idea to use `process.env.NODE_ENV` in "framework/platform/tooling"-agnostic code. For example `process` is not available at all in browser environments, so this will not work in browsers without being transpiled.

It would be nice to provide a user-configurable setting `captureCreationStack` to allow users to choose the creation stacks to be attached even in production, and fallback to process.env.NODE_ENV === "development" for the latter. The creation stack would be really useful for us in production environments for example, so it would be nice to have this user-configurable.

packages/jazz-tools/src/tools/subscribe/errorReporting.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/**
+ * Capture a stack trace only in development mode.
+ * Returns empty string in production to avoid overhead.
+ */
+export function captureStack(): string {
+  return isDev() ? new Error().stack || "" : "";
+}
```
Gabrola: This would be a perfect use-case for `Error.cause`. Instead of passing around the stack trace and printing it manually, we can just pass the error itself, attach it as a `cause` for other errors, and let the JS engine handle everything including printing the stack trace correctly. And since this would follow an existing standard, this would allow libraries or tools like Sentry to process or display the error chain as expected.

```suggestion
export function captureCause() {
  return new Error(`The Jazz error originated here`);
}
```

and then something like this
```ts
console.error(new Error(`Jazz Unavailable Error: ${value.toString()}`, { cause: callerCause }));
```

would look like this on a browser

Image

packages/jazz-tools/src/tools/subscribe/JazzError.ts
```diff
@@ -11,7 +11,12 @@ export class JazzError {
   ) {}
```
Gabrola, line 11:
```suggestion
export class JazzError extends Error {
  constructor(
    public id: ID<CoValue> | undefined,
    public type:
      | typeof CoValueLoadingState.UNAVAILABLE
      | typeof CoValueLoadingState.UNAUTHORIZED,
    public issues: JazzErrorIssue[],
    cause?: unknown,
  ) {
    super(/* everything in toString() */, { cause });
  }
```

Extending `Error` will cause the construction of `JazzError` to automatically capture an "access" stack trace, without needing to manually capture or process the stack trace, as well as add the ability to attach a `cause` Error.

packages/jazz-tools/src/tools/subscribe/errorReporting.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export type JazzErrorEvent = {
+  type: "unavailable" | "unauthorized" | "validation";
+  coValueId: ID<CoValue> | undefined;
+  message: string;
+  stack: string;
+  timestamp: number;
+  errorObject: any;
+};
```
Gabrola: This could be redundant at this point and passing a `JazzError` could suffice

Gabrola (commented):
This would be super useful for us, thank you!

gdorsi:
Pushed changes to centralize the error logging in SubscriptionScope and refine the logs for React and Expo.

Removed the captureStack on the load/subscribe methods because wasn't giving good info and the stacktrace from console.error was already better than the custom one.

Error example from a web app:
Screenshot 2025-12-02 at 17 39 51


Error example from the expo chat app:
Screenshot 2025-12-02 at 17 35 36

Pleasant surprise, capturing the error stack pushes Expo into some pretty-print automation.

packages/jazz-tools/src/tools/subscribe/errorReporting.ts
gdorsi: Good point, going for it

packages/jazz-tools/src/tools/subscribe/errorReporting.ts
gdorsi: Looks nice!

Going to give it a try!

packages/jazz-tools/src/tools/subscribe/JazzError.ts
gdorsi: True, but this will add some performance cost in production.

We use this class also when errors are catched, so I'd like to keep it lightweight.

Gabrola:
@gdorsi Hiya 👋 I can see the error subscription API was removed. This would have been very useful for us in addition to having the ability to capture creation stack traces in production as that would allow us to capture these errors with something like Sentry, especially if they were passed as an Error object with the creation origin passed as an Error in `cause`.

packages/jazz-tools/src/tools/subscribe/errorReporting.ts
gdorsi: The stacktrace coming from SubscriptionScope doesn't help much:
Screenshot 2025-12-02 at 17 51 52

Expo ignores the cause:
Screenshot 2025-12-02 at 17 53 53

gdorsi:
> @gdorsi Hiya 👋 I can see the error subscription API was removed. This would have been very useful for us in addition to having the ability to capture creation stack traces in production as that would allow us to capture these errors with something like Sentry, especially if they were passed as an Error object with the creation origin passed as an Error in `cause`.

Good suggestion.

Makes sense for us to have a good integration with error trackers, going to look into how to provide a first class API for that.

gdorsi:
@Gabrola @boorad 

Added `setCustomErrorReporter` and `enableCaptureErrorCause` to both intercept errors and enable the stacktraces.

`setCustomErrorReporter` can be used to override the default error reporting (console.error) and to integrate error trackers.
 `enableCaptureErrorCause` to turn on the additonal debug info coming from React hooks (provided via Error.cause in the reporter)

I've also tweaked the error messages to make them easier to understand.

Authorization error:
Screenshot 2025-12-02 at 19 10 46

Unavailable error:
Screenshot 2025-12-02 at 19 10 30

packages/jazz-tools/src/tools/subscribe/JazzError.ts
Gabrola: Yeah, that makes sense. Also definitely not needed now that an error is constructed inside `logError`, so all good!

packages/jazz-tools/src/tools/subscribe/errorReporting.ts
Gabrola: Yeah, Metro doesn't log the cause in the console, but it gets logged in the DevTools console

Gabrola:
@gdorsi looks really good, thank you!

Semi-relevant to this PR: To enhance discoverability and use, would it make sense to expose all global settings through a single function, especially if any other global settings are added in the future. So something like `configureJazz({ captureErrorCause: true, ...})` or maybe even just export all the settings getters and setters from a single object/module similar to `cojsonInternals`?

aeplay (approved):

gdorsi:
@Gabrola 

Makes sense. Added a `jazzConfig` export and exported the reporting API through it.

PR #3246, opened 2025-12-08
Add createdBy getter to CoValue and tests

joeinnes:
# Description
Per suggestion from [Discord](https://discord.com/channels/1139617727565271160/1139621689882321009/1447411642806698004), implements a `createdBy` getter.

N/A, but you can check by using `myCoValue.$jazz.createdBy`

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/CoValueBase.ts
+++ b/packages/jazz-tools/src/tools/coValues/CoValueBase.ts
@@ -112,6 +112,18 @@ export abstract class CoValueJazzApi<V extends CoValue> {
     return this.raw.core.earliestTxMadeAt;
   }
 
+  get createdBy(): string | undefined {
+    const createdBy = this.raw.core.getValidSortedTransactions({
+      ignorePrivateTransactions: false,
+    })[0]?.author;
+
+    // Only return accounts, not sealer/signer strings
+    if (typeof createdBy === "string" && createdBy.startsWith("co_z"))
+      return createdBy;
+
+    return undefined;
+  }
+
   /**
    * The timestamp of the last updated time of the CoValue
    *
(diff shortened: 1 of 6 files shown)


gdorsi (approved):

packages/jazz-tools/src/tools/coValues/CoValueBase.ts
```diff
@@ -112,6 +112,30 @@ export abstract class CoValueJazzApi<V extends CoValue> {
...
+   * Creation is determined by inspecting the earliest valid transaction,
+   * Note: Where the author is a sealer/signer identifiers (e.g. accounts)
+   * nothing is returned intentionally
+   *
+   * @returns {string | undefined} The creating user's account ID, or
+   * `undefined` if no author can be determined
+   *
+   * @category Content
+   */
+  get createdBy(): string | undefined {
```
gdorsi, line 127: We could use this API in the Chat example now

joeinnes: Not sure it matters particularly, but I think in the chat example we get the *last* updater of a CoValue, rather than the creator: https://github.com/garden-co/jazz/blob/ed2c91bd99b3eeb54b007c55e57e8a1e7096e350/examples/chat/src/chatScreen.tsx#L124

Come to think of it, there's no reason not to have a `lastUpdatedBy` to go with `lastUpdatedAt`, I might as well add that in too.

Should I continue with that here, or would it make more sense to grab these details as we loop over the transactions here: https://github.com/garden-co/jazz/blob/87254a24ea9b0c6b67d6ee324d4e2b759b748fb0/packages/cojson/src/coValueCore/coValueCore.ts#L957, adding `earliest` and `latest` `TxMadeBy` here: https://github.com/garden-co/jazz/blob/87254a24ea9b0c6b67d6ee324d4e2b759b748fb0/packages/cojson/src/coValueCore/coValueCore.ts#L848?

PR #3261, opened 2025-12-10
Remove RNQuickCrypto and make RNCrypto the default

aeplay:
## Full native crypto (0.20.0)

With this release we complete the migration to a pure Rust toolchain and remove the JavaScript crypto compatibility layer. The native Rust core now runs everywhere: React Native, Edge runtimes, all server-side environments, and the web.

### Crypto providers / fallback behavior

- **Removed `PureJSCrypto`** from `cojson` (including the `cojson/crypto/PureJSCrypto` export).
- **Removed `RNQuickCrypto`** from `jazz-tools`.
- **No more fallback to JavaScript crypto**: if crypto fails to initialize, Jazz now throws an error instead of falling back silently.
- **React Native + Expo**: **`RNCrypto` (via `cojson-core-rn`) is now the default**.

Migration guide: https://jazz-homepage-git-remove-rn-quick-crypto-garden-co.vercel.app/docs/upgrade/0-20-0

---

> [!NOTE]
> **Full native crypto rollout**
> 
> - Removes JavaScript crypto providers and fallback: deletes `PureJSCrypto` (and its exports) and `RNQuickCrypto`; Jazz now errors instead of silently falling back
> - React Native default now `RNCrypto` via `cojson-core-rn`
> 
> **WASM initialization**
> 
> - Adds `initializeSync()` to `cojson-core-wasm` (`index.js`, `index.d.ts`, and `edge-lite.js`) and supports data-URL/base64 decoding for sync init
> 
> **Bench updates**
> 
> - Benchmarks now use `WasmCrypto` (and NAPI) only; removes PureJS paths in `bench/cojson/*`
> 
> **Build/packaging**
> 
> - Fixes `CojsonCoreRnFramework.xcframework` `Info.plist` library identifiers/variant ordering
> - Adds changeset documenting minor releases and migration guide
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4b5e1056c094c3d134cc4677c31681c3ae689adf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


.changeset/wild-melons-study.md
```diff
@@ -0,0 +1,29 @@
+---
+"cojson-core-rn": patch
+"jazz-tools": patch
+---
+
+## 💥 Breaking Changes
```
antoniomuso: I added a migration guid in the changeset, @joeinnes do you think it is better to have it even in the homepage docs? (Under `/upgrade`)

.changeset/wild-melons-study.md
```diff
@@ -0,0 +1,29 @@
...
+
+`RNQuickCrypto` has been removed to streamline the library. `RNCrypto` is now the default implementation.
+
+**Migration Guide:**
+
+If you are currently using `RNCrypto`, you need to update your import paths:
+
+**For React Native:**
+```diff
+- import { RNCrypto } from "jazz-tools/react-native-core/crypto/RNCrypto";
```
joeinnes: This looks wrong, should we not be removing the import for RNQuickCrypto?

.changeset/wild-melons-study.md
```diff
@@ -0,0 +1,29 @@
...
++ import { RNCrypto } from "jazz-tools/react-native-core/crypto";
+}
+```
+**For Expo:**
+```diff
+-  import { RNCrypto } from "jazz-tools/react-native-core/crypto/RNCrypto";
```
joeinnes: Same as above

.changeset/wild-melons-study.md
```diff
@@ -0,0 +1,29 @@
...
++  import { RNCrypto } from "jazz-tools/expo/crypto";
+}
+```
+
+If you are not currently using `RNCrypto`, please refer to the [setup guide](https://jazz.tools/docs/react-native-expo/project-setup/providers#rncrypto).
```
gdorsi: This part of the docs should go away, and we should make the installation of cojson-core-rn part of the installation guide

.changeset/wild-melons-study.md
```diff
@@ -0,0 +1,93 @@
...
+// BEFORE
+import { RNQuickCrypto } from "jazz-tools/react-native-core/crypto";
+
+// AFTER
+import { RNCrypto } from "jazz-tools/react-native-core/crypto";
+```
+
+## SSR / Workers (WASM)
+
+WASM crypto no longer falls back to JS crypto. If you want async initialization:
```
gdorsi: The change to a sync intialization is a minor, not even worth to be mentioned.

It's more imprtant that the lack of fallback to PureJSCrypto is going to break Cloudflare Workers.

It should be something like this:
> Before this release we were falling back to Javascript when the runtime doesn't support our default way of loading WASM (e.g. in Cloudflare Workers)
> This wasn't great because that fallback was coming with a 20x slowdown
> Now the load is going to fail with an error message that will guide you to the right solution
> More details in https://jazz.tools/docs/react/server-side/setup#wasm-on-edge-runtimes

homepage/homepage/content/docs/reference/performance.mdx
```diff
@@ -10,9 +10,8 @@ export const metadata = {
 
 The fastest implementations are (in order):
 
-1. [Node-API crypto](/docs/server-side/setup#node-api) (only available in some Node/Deno environments) or [RNCrypto](/docs/project-setup/providers#react-native-crypto) and [RNQuickCrypto](/docs/react-native/project-setup/providers#quick-crypto) on React Native
+1. [Node-API crypto](/docs/server-side/setup#node-api) (only available in some Node/Deno environments) and RNCrypto on React Native (Default)
 2. [WASM crypto](/docs/server-side/setup#wasm-on-edge-runtimes)
-3. JavaScript fallback (slowest, but most compatible)
 
 Check whether your environment supports Node-API.
```
gdorsi, line 16: We should add the suggetion about the async initialization and nextjs here

homepage/homepage/content/docs/code-snippets/server-side/ssr/jazzSSRInitWasm.ts
```diff
@@ -0,0 +1,9 @@
+import { createSSRJazzAgent } from "jazz-tools/ssr";
+import { initWasm } from "jazz-tools/wasm";
+
+// Init WASM asynchronously to avoid blocking the main thread.
+await initWasm();
```
gdorsi, line 5: This is not the only place where this should be done.

The `await initWasm();` should be done also when using the JazzReactProvider, to force the async initalization also in the browser (where it matters most)

We can avoid flagging it here, not important in the server side context I think.

homepage/homepage/content/docs/docNavigationItems.js
```diff
@@ -43,6 +43,11 @@ export const docNavigationItems = [
     collapse: true,
     prefix: "/docs/upgrade",
     items: [
+      {
+        name: "0.20.0 - RNCrypto default (Pure JS & RNQuickCrypto removed)",
```
gdorsi: The title should be "Full native crypto"

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,346 @@
...
+
+This release removes the PureJSCrypto and QuickCrypto implementations and makes **`RNCrypto` (Rust, via `cojson-core-rn`) the default crypto provider on React Native**
+and **`WasmCrypto` is the default crypto provider for Web/SSR/Workers (wasm only)**.
+
+So you can use only:
+- WASM for the web and edge runtimes
+- RNCrypto for React Native
+- Node-API or WASM for Node.js
+
+## What changed
```
gdorsi: We should document the missing fallback to PureJSCrypto: https://github.com/garden-co/jazz/pull/3261/files#r2620273782

This is the part that's going to create more issues, and we should check after the release how v0 behaves.

.changeset/wild-melons-study.md
gdorsi: I also think that the instruction on how to run Jazz on edge runtimes should be more visible

cc  @joeinnes

antoniomuso: Ok, I add it even in the homepage migration guide.

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,346 @@
...
+    <JazzReactNativeProvider
+      sync={{ peer: `wss://cloud.jazz.tools/?key=${apiKey}` }}
+      CryptoProvider={RNCrypto}
+    >
+      {children}
+    </JazzReactNativeProvider>
+  );
+}
+
+// AFTER (recommended: RNCrypto is the default)
```
gdorsi:
```suggestion
// AFTER (RNCrypto is the default)
```

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,346 @@
...
+export function MyJazzProvider({ children }: { children: React.ReactNode }) {
+  return (
+    <JazzReactNativeProvider
+      sync={{ peer: "wss://cloud.jazz.tools/?key=your-api-key" }}
+      CryptoProvider={RNCrypto}
+    >
+      {children}
+    </JazzReactNativeProvider>
+  );
+}
```
gdorsi: This middle step is not required.

We could also merge 4 with 3, as the breaking change is that we don't accept a CryptoProvider anymore.

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,346 @@
...
+  return (
+    <JazzReactNativeProvider sync={{ peer: "wss://cloud.jazz.tools/?key=your-api-key" }}>
+      {children}
+    </JazzReactNativeProvider>
+  );
+}
+```
+</CodeGroup>
+
+If you were using `RNCrypto` already (and still want to pass it explicitly), update the import path:
```
gdorsi: We can remove this part, why would someone go for this?

.changeset/wild-melons-study.md
joeinnes: > I also think that the instruction on how to run Jazz on edge runtimes should be more visible
> 
> cc [@joeinnes](https://github.com/joeinnes)

Yeah, we probably have enough content now to separate out the server-side setup page into two different pages, one which is just Jazz (creating a worker, spinning it up, etc.) and 'Finagling with your Runtime' (title TBD).

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -1,71 +1,97 @@
...
+## Motivation
 
-## What changed
+Before 0.20.0, Jazz relied on a JavaScript crypto implementation to work around compatibility issues in some runtimes—most notably React Native.
+As a result, our native Rust core wasn’t running in React Native and, on some Edge runtimes, 
+only worked when WASM was explicitly initialized (see: https://jazz.tools/docs/react/server-side/setup#wasm-on-edge-runtimes).
 
-### React Native / Expo
+This JavaScript path was also slower than native Rust crypto. Even RNQuickCrypto (a React Native provider) couldn’t match native performance because it only wrapped some native crypto libraries,
+ rather than running Jazz’s full Rust crypto implementation.
```
joeinnes:
```suggestion
The JavaScript crypto implementation is much slower than native Rust crypto. Although workarounds like RNQuickCrypto for React Native improved performance, they still only wrapped certain native libraries, rather than running Jazz's full Rust crypto.
```

Wording nit

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -1,71 +1,97 @@
...
+<CodeGroup>
+```
+Warning: WASM crypto is not available. Falling back to JavaScript crypto.
+```
+</CodeGroup>
 
 ### React Native (non-Expo)
 
+If you use React Native, you need to install `cojson-core-rn` and make sure the version matches `jazz-tools`. `cojson-core-rn` is our new high-performance crypto provider for React Native. 
+And it is **required** for React Native applications.
```
joeinnes:
```suggestion
If you use React Native, you need to install `cojson-core-rn` and make sure the version matches `jazz-tools`. `cojson-core-rn` is our new high-performance crypto provider for React Native, and it is **required** for React Native applications.
```

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -168,41 +177,36 @@ export function MyJazzProvider({ children }: { children: React.ReactNode }) {
...
-
-// AFTER
-import { RNCrypto } from "jazz-tools/react-native-core/crypto";
-```
-</CodeGroup>
+### Expo
 
-If you weren’t setting `CryptoProvider` at all, **no code changes are required** (it now defaults to `RNCrypto`) - but you **must** install `cojson-core-rn`.
+1) If you use Expo, you need to install `cojson-core-rn` and make sure the version matches `jazz-tools`. `cojson-core-rn` is our new high-performance crypto provider for Expo. 
+And it is **required** for Expo applications.
```
joeinnes:
```suggestion
1) If you use Expo, you need to install `cojson-core-rn` and make sure the version matches `jazz-tools`. `cojson-core-rn` is our new high-performance crypto provider for Expo, and it is **required** for Expo applications.
```

.changeset/wild-melons-study.md
antoniomuso: We can remove RNCrypto too, as it's now the default for RN and Expo.

antoniomuso:
@gdorsi @joeinnes I’m thinking of reducing the size of the changesets by removing the migration steps and linking directly to the documentation instead.

.changeset/wild-melons-study.md
gdorsi: > This looks wrong, should we not be removing the import for RNQuickCrypto?

Could you explain what looks wrong and why you think we should keep RNQuickCrypto?

We want to remove it to keep the dependencies clean and hopefully reduce the bundle size (it currently depends on PureJSCrypto), but we may have missed something.

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,293 @@
...
+## Motivation
+
+Before 0.20.0, Jazz relied on a JavaScript crypto implementation to work around compatibility issues in some runtimes-most notably React Native.
+As a result, our native Rust core wasn’t running in React Native and, on some Edge runtimes, 
+only worked when WASM was explicitly initialized (see [WASM on Edge runtimes](/docs/react/server-side/setup#wasm-on-edge-runtimes)).
+
+The JavaScript crypto implementation is much slower than native Rust crypto. Although workarounds like RNQuickCrypto for React Native improved performance, they still only wrapped certain native libraries, rather than running Jazz's full Rust crypto.
+
+With native Rust crypto now running everywhere, Jazz crypto is faster and more reliable. 
+This also helps us speed up the migration of Jazz Core to Rust which will improve the Jazz overall performance.
```
gdorsi:
```suggestion
This also helps us speed up the migration of Jazz Core to Rust which will help us improve the Jazz overall performance.
```

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,293 @@
...
+
+## Motivation
+
+Before 0.20.0, Jazz relied on a JavaScript crypto implementation to work around compatibility issues in some runtimes-most notably React Native.
+As a result, our native Rust core wasn’t running in React Native and, on some Edge runtimes, 
+only worked when WASM was explicitly initialized (see [WASM on Edge runtimes](/docs/react/server-side/setup#wasm-on-edge-runtimes)).
+
+The JavaScript crypto implementation is much slower than native Rust crypto. Although workarounds like RNQuickCrypto for React Native improved performance, they still only wrapped certain native libraries, rather than running Jazz's full Rust crypto.
+
+With native Rust crypto now running everywhere, Jazz crypto is faster and more reliable.
```
gdorsi:
```suggestion
With native Rust crypto now running everywhere, Jazz comes with good performance in every platform. 
```

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,293 @@
...
+  },
+};
+```
+</CodeGroup>
+
+if you don't do this, you will get an error like this:
+
+<CodeGroup>
+```
+Error: Failed to initialize WasmCrypto. WebAssembly is required for crypto operations on web platforms.
```
gdorsi: This error should be more informative, either by giving the instructions to fix or by giving the doc page.

The former is preferred so AIs are more likely to fix the issue automatically (thinking about v0)

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,293 @@
...
+    >
+      {children}
+    </JazzReactNativeProvider>
+  );
+}
+```
+</CodeGroup>
+
+If you previously had `CryptoProvider={...}` and/or `import { RNCrypto } ...`, you can remove both:
+
```
gdorsi: I don't think we need to explain this, people only need to know that the CryptoProvider prop is useless now

packages/cojson/src/crypto/WasmCrypto.ts
```diff
@@ -54,23 +72,40 @@ let wasmInit = initialize;
...
+  // @deprecated
+  static async create(): Promise<WasmCrypto> {
     try {
       await wasmInit();
     } catch (e) {
-      logger.warn(
-        "Failed to initialize WasmCrypto, falling back to PureJSCrypto",
-        { err: e },
+      throw new Error(
+        "Failed to initialize WasmCrypto. WebAssembly is required for crypto operations on web platforms.",
```
gdorsi: Already said this above, but these errors should guide the user to a solution and explain why this operation may fail.

To save bytes we can also use the same message or a template.

gdorsi (commented):
Have you checked the bundle size already?

Curious of the cost of the noble packages dropped or if we need to do something else.

This might be another selling point to explain in the motivations.

.changeset/wild-melons-study.md
gdorsi: Just realized that the Github Mobile was showing me outdated discussions on a different line!

Ok, from the web your question makes sense, closing the thread.

homepage/homepage/content/docs/upgrade/0-20-0.mdx
antoniomuso: done! Now `CryptoProvider` is not anymore supported in RN and Expo.

antoniomuso:
> Have you checked the bundle size already?
> 
> Curious of the cost of the noble packages dropped or if we need to do something else.
> 
> This might be another selling point to explain in the motivations.

The Noble packages are already quite small, so we only save about ~3 MB.

packages/cojson/src/crypto/WasmCrypto.ts
antoniomuso: I have done something better now.

PR #3267, opened 2025-12-12
jazz-tools cleanup 1 - hide CoValue classes

aeplay:
This is the first step in a big refactor that will enable us to slowly get rid of the CoValueClass / Zod Schema divide.

This PR
- moves the implementation of most static methods from CoValueClasses to their corresponding Zod Schemas
  - except `.create` because that's used internally in a myriad ways
- stops exporting CoValueClasses as values (so you can't call `Group.create` anymore, but have to do `co.group().create`)


examples/betterauth/src/app/(app)/page.tsx
```diff
@@ -12,7 +12,7 @@ import {
 import Image from "next/image";
 
 export default function Home() {
-  const me = useAccount(Account, { resolve: { profile: {} } });
```
gdorsi, line 15: I don't like this much, especially because `co.account()` will return a different schema on each render, resetting the subscrition.

We could:
1. Make `co.account()` return a cached schema (this should be easy)
2. Change the examples to export a JazzAccount from the schema, even if they don't use the root

examples/chat-rn/src/chat.tsx
```diff
@@ -105,7 +105,7 @@ export function ChatScreen({ navigation }: { navigation: any }) {
 
   const createChat = () => {
     if (!me.$isLoaded) return;
-    const group = Group.create({ owner: me });
```
gdorsi, line 108: We could do the same with `co.group()` and cache the output

gdorsi (commented):
Looks good!

I think we can delete the codemod packages, as we don't need to keep them compatible with the latest versions of Jazz.

Any fix on them could be done on the tag versions.

examples/betterauth/src/app/(app)/page.tsx
aeplay: I'd say 1.

PR #3268, opened 2025-12-12
fix: ensure that all the suspended promises are cached and port the chat example to Suspense

gdorsi:
Last migration to Suspense has broken the chat example.

Digging into the problem I found two issues:
1. React was nagging that promises weren't properly cached
2. The router.navigate in useHashRouter doesn't work when the target route is suspended

The first one was only a warning, because while we didn't cache the promises the data was.
But the checks on React are based on the promise identity, so I've changed the API to make them happy.

On the second one it's the comeback of the event propagation issue we had in the past with the same router.
Calling directly `onHashChange` fixes the bug, and I've added a `startTransition` to make it work better with Suspense.

### DIFF
--- a/examples/chat/src/app.tsx
+++ b/examples/chat/src/app.tsx
@@ -52,6 +52,8 @@ export function App() {
 
     // for https://jazz.tools marketing site demo only
     onChatLoad(chat);
+
+    return null;
   };
 
   const usernamePlaceholder = "Set username";
@@ -93,7 +95,7 @@ export function App() {
         )}
       </TopBar>
       {router.route({
-        "/": () => createChat() as never,
+        "/": () => createChat(),
         "/chat/:id": (id) => <ChatScreen chatID={id} />,
       })}
     </AppContainer>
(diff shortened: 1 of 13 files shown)


examples/chat/src/app.tsx
```diff
@@ -78,22 +67,19 @@ export function App() {
...
-            {localValue || usernamePlaceholder}
+            {usernamePlaceholder}
           </span>
           <input
             type="text"
             id={inputId}
-            value={localValue}
-            style={{ width: `${inputWidth}px` }}
+            value={me.profile.name}
+            style={{ width: `${me.profile.name.length}ch` }}
```
gdorsi, line 79: The user name was glitching while editing, so I've changed the solution to autosize the input

examples/chat/src/ui.tsx
```diff
@@ -95,7 +95,7 @@ export function BubbleImage(props: { image: ImageDefinition }) {
   return (
     <Image
       imageId={props.image.$jazz.id}
-      className="h-auto max-h-80 max-w-full rounded-t-xl mb-1"
+      className="h-auto max-h-80 max-w-full rounded-t-xl mb-1 object-contain"
```
gdorsi, line 98: The images were stretched, switched to `object-contain` to always keep the ratio

aeplay (approved):

PR #3270, opened 2025-12-12
Adding other frameworks to the forms page

joeinnes:
# Description
Adds other frameworks to the forms pattern page

Review reference/design-patterns/form

## Tests

- [X] Tests have not been updated, because: Docs

## Checklist

### DIFF
--- a/homepage/homepage/content/docs/code-snippets/reference/design-patterns/form/CreateForm.svelte
+++ b/homepage/homepage/content/docs/code-snippets/reference/design-patterns/form/CreateForm.svelte
@@ -1,0 +1,26 @@
+<script lang="ts">
+  import { co, ID } from "jazz-tools";
+  import OrderForm from "./OrderForm.svelte";
+  import { BubbleTeaOrder, JazzAccount, PartialBubbleTeaOrder } from "./schema";
+  import { CoState, AccountCoState } from "jazz-tools/svelte";
+  const { id }: { id: ID<typeof BubbleTeaOrder> } = $props();
+
+  const me = new AccountCoState(JazzAccount, {
+    resolve: {
+      root: {
+        orders: true
+      }
+    }
+  });
+
+  const newOrder = new CoState(PartialBubbleTeaOrder, id);
+
+  const handleSave = async (data: any) => {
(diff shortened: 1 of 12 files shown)


homepage/homepage/content/docs/code-snippets/reference/design-patterns/form/CreateForm.svelte
```diff
@@ -0,0 +1,26 @@
...
+    resolve: {
+      root: {
+        orders: true
+      }
+    }
+  });
+
+  const newOrder = new CoState(PartialBubbleTeaOrder, id);
+
+  const handleSave = async (data: any) => {
```
gdorsi:
```suggestion
  const handleSave = async () => {
```

homepage/homepage/content/docs/code-snippets/reference/design-patterns/form/CreateForm.svelte
```diff
@@ -0,0 +1,26 @@
...
+    if (me.current.$isLoaded && newOrder.current.$isLoaded) me.current.root.orders.$jazz.push(newOrder.current as co.loaded<typeof BubbleTeaOrder>);
```
gdorsi, line 19: Why do we need the `loaded` casting here?

gdorsi (approved):
Nice job!

homepage/homepage/content/docs/code-snippets/reference/design-patterns/form/CreateForm.svelte
joeinnes: Short answer: because `newOrder` is a `PartialBubbleTeaOrder`, not a `BubbleTeaOrder`.

Medium answer: because that's what we had in the React example: https://github.com/garden-co/jazz/blob/57274c8c3c2cbba1e3d4bc77be813fff57ff059a/homepage/homepage/content/docs/code-snippets/reference/design-patterns/form/OrderForm.tsx#L91, and I was focussing on ensuring we had parity between docs for React and other frameworks.

Long answer: I'm actually rethinking this whole pattern. We need to create the `PartialBubbleTeaOrder` CoValue in a parent component, pass its ID as a prop to this form, and then pass to the child `OrderForm`. Seems a bit of a messy way of doing it. I think I might try to come up with a better design for this across all the frameworks.

PR #3276, opened 2025-12-14
Runtime performance metrics

Takeno:
This PR introduces comprehensive observability support for Jazz using the OpenTelemetry protocol. It instruments the SubscriptionScope with metrics and tracing spans, allowing developers to integrate Jazz's internal performance data with their existing observability stack.

Metrics are based on `performance.marks` and `performance.measure`, and they are visible in the DevTools.

immagine

A new `performance` page has been added to the Inspector, displaying load times with detailed breakdown (storage, peer, parsing). It is available only for in-app inspector, as it uses the same context.

immagine

---

> [!NOTE]
> Adds end-to-end observability for subscription performance and exposes it in-app.
> 
> - **Instrumentation (OTel):** Instrumented `SubscriptionScope` with `jazz.subscription.active` (UpDownCounter) and `jazz.subscription.first_load` (Histogram), plus tracing spans (`jazz.subscription.*`). New `unstable_setOpenTelemetryInstrumentationEnabled` toggle. CoJSON adds `performance` marks for storage/peer loads and exports `perf.performanceMarks`.
> - **Inspector UI:** New Perf tab shows subscription load times (storage/peer/memory), active subscriptions, and transport metrics; includes `jazzMetricReader` and auto meter setup in `JazzInspector`.
> - **Docs & examples:** New Observability docs and code snippets; Inspector docs updated for Perf tab; homepage adds OpenTelemetry deps and nav entry; example app embeds `JazzInspector` within provider.
> - **Tests:** Added OTel metric reader test utilities and tests for active subscriptions and first-load duration.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a11bedb8b854e88dda3cec93d4f9fdfe14e1b619. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


homepage/homepage/content/docs/code-snippets/tooling-and-resources/inspector/index.tsx
```diff
@@ -0,0 +1,19 @@
+// #region RecordMetrics
+import { JazzInspector, recordMetrics } from "jazz-tools/inspector";
+import { JazzReactProvider } from "jazz-tools/react";
+
+// Initialize metrics collection before rendering
+recordMetrics();
```
gdorsi: Would it be possible to do this "automatically" when the JazzInspector component is mounted?

homepage/homepage/content/docs/tooling-and-resources/observability.mdx
```diff
@@ -0,0 +1,110 @@
+import { CodeGroup } from '@/components/forMdx'
+
+export const metadata = {
+  description: "Learn how to integrate Jazz with OpenTelemetry for metrics and tracing."
+};
```
gdorsi, line 5: It's probably for the best to either mark this as "unstable" and add an alert saying that breaking changes are to be expected in patch upgrades.

In my experience it takes time to get the observability metrics right, and I don't expect any of these to be correct on the first try.

We will probably change them alot before coming with a stable API.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -1085,6 +1086,8 @@ export class CoValueCore {
     if (!this.isAvailable()) {
       return;
     }
+
+    trackPerformanceMark("transactionParsingStart", this.id);
```
gdorsi: `performance.mark` is too slow to be used in this kind of performance-critical code-paths.

Better to not track performance on this level of detail, not even in dev.

packages/jazz-tools/src/inspector/utils/instrumentation.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    return dp;
+  }
+
+  async collectMetrics() {
+    if (!this.initialized) {
+      // TODO: link to documentation
+      throw new Error(
+        "JazzOTelMetricReader not initialized. Have you called recordMetrics?",
+      );
```
gdorsi: We have docs now to link here, right?

packages/jazz-tools/src/inspector/utils/performances.ts
```diff
@@ -0,0 +1,212 @@
...
+    activeSubscriptions: ActiveSubscription[];
+    loadTimes: {
+      id: string;
+      source_id?: string;
+      parent_id?: string;
+      resolve?: string;
+      loadTime: number;
+      loadFromStorage?: number;
+      loadFromPeer?: number;
+      transactionParsing?: number;
```
gdorsi: I like the idea of measuring storage vs network load time, but instead of exposing it this way I think it would be easier to understand if we give people a `type` flag that's storage if the value was found on storage and network otherwise.

A single `loadTime` number is more than enough to catch slow queries, the rest is for advanced investigations, but I think that even in that case it would be hard to gain good insights given their very small granularity.

When things are going to be slow (I expect on thousands of CoValues) we need to show only the meaningful data, and avoid polluting the devtools with too many marks.

gdorsi (changes requested):
Love the feature!

I think we should start way more slowly, and minimize the information exposed as much as possible.

Other than the performance impact concerns I think we should carefully expose only the extremely valuable informations, otherwise we would end up like the React's performance marks that I myself never got to understand.

Advanced users can use the performance profiler that gives detailed info anyway.

Also, React Native is failing with `failure: com.facebook.react.common.JavascriptException: TypeError: undefined is not a function`

packages/jazz-tools/src/inspector/utils/instrumentation.ts
Takeno: Added the link in the inspector, if this throws.

packages/cojson/src/coValueCore/coValueCore.ts
Takeno: Removed the transaction parsing

Takeno:
PR updated with latest feedback:
- New OTel metrics and spans are disabled by default, and must be activated with `unstable_setOpenTelemetryInstrumentationEnabled`
- Removed marks from decrypt transactions
- Added a simple approach to clean up performance marks once read, to avoid useless pollution
- Inspector now reports only the peer from the CoValue has been loaded (storage or network)
immagine

- Implemented a clear button to filter out covalue load times at runtime
- Updated the documentation
- Auto-register the instrumentation when importing the inspector, to avoid manual configurations

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -1477,12 +1479,15 @@ export class CoValueCore {
...
     this.markPending("storage");
+
+    trackPerformanceMark("loadFromStorageStart", this.id);
     node.storage.load(
       this.id,
       (data) => {
         node.syncManager.handleNewContent(data, "storage");
       },
       (found) => {
+        trackPerformanceMark("loadFromStorageEnd", this.id);
```
gdorsi: We should track also if the value was found, to be able to exclude storage when presenting it to the user

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
```diff
@@ -525,6 +589,32 @@ export class SubscriptionScope<D extends CoValue> {
       this.subscribers.forEach((listener) => listener(value));
     }
 
+    if (!this.firstLoadRecorded) {
+      trackPerformanceMark("subscriptionLoadEnd", this.id);
```
gdorsi, line 601: This might not catch some subscriptions load time correctly.

We arrive here only if there are listeners, so e.g. when the values are already loaded we would end up recording the first update.

I think it should be placed above, [maybe here](https://github.com/garden-co/jazz/blob/8c59660e0383270159194195f62f624be6f43c24/packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts#L579) but not too sure.

We probably need a different approach or a refactoring of SubscriptionScope to have some more explicit lifecycles.

Takeno: Good catch! The implementation started there with a different aim, but actually the load time must be tracked even without a listener attached. This is the case with Jazz Context, which loads the account without any listeners.

Takeno:
PR Updated:
- Simplified inspector's UI: now only top-level subscriptions are reported
- Reduced the noise in the devTools' profiling, removing used marks and showing only top-level subscription's times
- Tracked also coValues without subscribers
-  Introduced a console.error when Inspector is rendered twice.

immagine

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
```diff
@@ -96,6 +138,27 @@ export class SubscriptionScope<D extends CoValue> {
...
+            resolve: JSON.stringify(this.resolve),
+          },
+        });
+
+    this.activeSubCounter.add(1, {
+      // It increments/decrements counters comparing the attributes
+      id: this.id,
+      source_id: this.sourceId,
+      resolve: JSON.stringify(this.resolve),
+    });
```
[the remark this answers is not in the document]
Takeno: This OTel counter is also used to track the number of active subscriptions in the inspector. We need an additional check if the inspector is installed or not.

Takeno:
Closed in favor of #3389

PR #3278, opened 2025-12-15
fix: prevent race conditions and reuse also the new accounts sessions in browser session lock system

gdorsi:
We've found that sometimes the provideLockSession in the browser was failing with `Couldn't get lock on session after 100x2 tries`.

Also while testing I've noticed that we weren't storing the first session generated for new accounts.

Added a e2e test that spawns 9 iframes and checks if each one acquires the sessions correctly, that runs on chrome, firefox and webkit to ensure that there are no deadlocks across the browsers implementations.


.github/workflows/playwright.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         uses: ./.github/actions/source-code/
 
       - name: Install Playwright Browsers
-        run: pnpm exec playwright install
+        run: pnpm exec playwright install --with-deps
```
gdorsi, line 33: Slows down the CI, but required for cross-browser runs

aeplay (approved):

PR #3287, opened 2025-12-16
docs: spec for coValue delete

gdorsi:
## Overview

Spec for a data deletion flow for coValues with immediate sync blocking and scheduled physical deletion. Deleted coValues leave tombstones sessions that prevent data resurrection.

## Key Design Choices

**Important Constraint**: Account and Group coValues cannot be deleted, otherwise we might not be able to verify the delete transactions.

- **Admin-only deletion**: Only admins can delete coValues, validated in `determineValidTransactions()`
- **Special delete session**: Uses `{accountId}_deleted_{uniqueId}` session naming to sync delete markers without carrying extra data or history
- **Unencrypted transactions**: Delete transactions are trusting (unencrypted) with meta `{ deleted: true }`
- **Immediate sync blocking**: Deleted coValues block all sync immediately, only delete session/transaction can sync
- **Storage API batch deletion**: `ereaseAllDeletedCoValues()` performs physical deletion while preserving tombstones
- **Tombstone preservation**: Delete session + coValue header preserved even after physical deletion
- **Storage shard support**: Shards with `skipVerify: true` mark as deleted and block sync without verification

Documents preview:
- Requirements https://github.com/garden-co/jazz/blob/spec/delete-coValues/.specs/data-delete-flow/requirements.md
- Design https://github.com/garden-co/jazz/blob/spec/delete-coValues/.specs/data-delete-flow/design.md

### DIFF
--- a/.specs/data-delete-flow/design.md
+++ b/.specs/data-delete-flow/design.md
@@ -32,7 +32,7 @@ Key points:
   - Checks that the current account has admin permissions on the coValue
   - Throws error if not admin
   - Creates delete transaction with meta `{ deleted: true }`
-  - Uses session naming pattern: `{accountId}_deleted_{uniqueId}`
+  - Uses session naming pattern: `{accountId}_session_z{uniqueId}_deleted`
 
 **Key Logic**:
 ```typescript
@@ -66,8 +66,7 @@ deleteCoValue(): void {
   }
 
   // Generate unique session ID for delete transaction
-  const uniqueId = this.crypto.newRandomSessionID(currentAccount.id).split('_session_z')[1];
-  const deleteSessionID = `${currentAccount.id}_deleted_${uniqueId}` as SessionID;
+  const deleteSessionID = this.crypto.newRandomSessionID(currentAccount.id) + `_deleted`;
 
   // Create unencrypted (trusting) transaction with delete meta
   // Pass deleteSessionID to makeTransaction to use the delete session
@@ -91,7 +90,7 @@ deleteCoValue(): void {
 - Ensure delete transactions are always unencrypted (trusting)
 
 **Rationale for Special Delete Session**:
-The special session naming pattern (`{accountId}_deleted_{uniqueId}`) is used to provide a simple way to sync the delete marker without carrying extra data. By using a dedicated session for delete transactions, the system can:
+The special session naming pattern (`{accountId}_session_z{uniqueId}_deleted`) is used to provide a simple way to sync the delete marker without carrying extra data. By using a dedicated session for delete transactions, the system can:
 - Easily identify delete operations by session ID pattern
 - Sync only the delete marker without carrying any history
 - Maintain a clean separation between regular transactions and delete operations
@@ -126,7 +125,7 @@ makeTransaction(
 
 **Changes**:
 - After signature verification succeeds, check for delete transactions
-- Only parse transaction meta if sessionID contains `_deleted_` (optimization)
+- Only parse transaction meta if sessionID ends with `_deleted` (optimization)
 - Validate that delete transaction author has admin permissions
 - Mark coValue as deleted if valid delete transaction found
 - Store delete state for sync blocking
@@ -140,8 +139,8 @@ tryAddTransactions(
   skipVerify: boolean = false,
 ) {
   let isDeleteOperation = false;
-  // Only check for delete transactions if sessionID contains '_deleted_'
-  if (sessionID.includes('_deleted_') && !this.isGroup() && !this.isAccount()) {
+  // Only check for delete transactions if sessionID contains '_deleted'
+  if (sessionID.endsWith('_deleted') && !this.isGroup() && !this.isAccount()) {
     // Check for delete transactions after successful addition
     for (const tx of newTransactions) {
       if (tx.privacy !== "trusting" || !tx.meta) continue;
@@ -260,7 +259,7 @@ This section defines exactly how existing sync message handlers behave once a co
 
 #### Terminology / constants
 
-- **Delete session**: `deleteSessionID` with pattern `{RawAccountID}_deleted_{string}`.
+- **Delete session**: `deleteSessionID` with pattern `{accountId}_session_z{uniqueId}_deleted`.
 - **Delete transaction count**: `deleteTxCount = knownState.sessions[deleteSessionID]` (typically `1`).
 - **Poison counter**: We reply with the peer optimistic known state plus the deleted session.
 
@@ -520,7 +519,7 @@ This section defines the **storage-level primitive** used by batch deletion (sec
 **Goal**: remove all persisted history for a deleted coValue while preserving:
 
 - the `coValues` row (header), and
-- all sessions matching the delete-session pattern (`*_deleted_*`) and their transactions/signatures.
+- all sessions matching the delete-session pattern (`*_deleted$`) and their transactions/signatures.
 
 Everything else for that coValue is physically deleted.
 
@@ -563,7 +562,7 @@ The SQLite storage schema is:
 SELECT rowID
 FROM sessions
 WHERE coValue = ?
-  AND sessionID NOT LIKE '%_deleted_%';
+  AND sessionID NOT LIKE '%_deleted';
 ```
 
 2. Delete all rows for those sessions:
@@ -573,19 +572,19 @@ DELETE FROM transactions
 WHERE ses IN (
   SELECT rowID FROM sessions
   WHERE coValue = ?
-    AND sessionID NOT LIKE '%_deleted_%'
+    AND sessionID NOT LIKE '%_deleted'
 );
 
 DELETE FROM signatureAfter
 WHERE ses IN (
   SELECT rowID FROM sessions
   WHERE coValue = ?
-    AND sessionID NOT LIKE '%_deleted_%'
+    AND sessionID NOT LIKE '%_deleted'
 );
 
 DELETE FROM sessions
 WHERE coValue = ?
-  AND sessionID NOT LIKE '%_deleted_%';
+  AND sessionID NOT LIKE '%_deleted';
 ```
 
 `deletedCoValues` is treated as a **work queue**, remove the row so the batch job won’t repeatedly re-process already-erased coValues:
@@ -612,7 +611,7 @@ The IndexedDB storage uses these stores:
 
 1. Load all sessions for the coValue via the `sessionsByCoValue` index.
 2. Partition them into:
-   - keep: `sessionID.includes("_deleted_")`
+   - keep: `sessionID.endsWith("_deleted")`
    - delete: everything else
 3. For each session in the delete set:
    - delete all `transactions` keys in range `IDBKeyRange.bound([ses, 0], [ses, +∞])` (cursor delete)
@@ -627,7 +626,7 @@ The IndexedDB storage uses these stores:
 ```typescript
 // Conceptual pseudocode inside one IDB transaction:
 const allSessions = await sessionsByCoValue.getAll(coValueRowID);
-const sessionsToDelete = allSessions.filter((s) => !s.sessionID.includes("_deleted_"));
+const sessionsToDelete = allSessions.filter((s) => !s.sessionID.endsWith("_deleted"));
 
 for (const s of sessionsToDelete) {
   // delete txs for ses = s.rowID
@@ -657,7 +656,7 @@ type DeleteTransactionMeta = {
   deleted: true;
 };
 
-type DeleteSessionID = `${RawAccountID}_deleted_${string}`;
+type DeleteSessionID = `${SessionID}_deleted`;
 ```
 
 ### CoValue Deleted State
--- a/.specs/data-delete-flow/design.md
+++ b/.specs/data-delete-flow/design.md
@@ -8,6 +8,18 @@ This design implements a secure, efficient data deletion flow for coValues that 
 
 The design leverages existing transaction validation mechanisms, adding delete-specific checks in `tryAddTransactions`, and implements sync restrictions to prevent deleted content from propagating while allowing delete operations to sync.
 
+## Backward compatibility (wire protocol + mixed-version peers)
+
+This design is intended to be rolled out without introducing new message types or changing the shape of existing sync messages (`load`, `known`, `content`, `done`). Instead, we change *semantics* for deleted coValues while remaining compatible with older peers.
+
+Key points:
+
+- **No protocol versioning required**: delete is represented as an ordinary unencrypted (`trusting`) transaction with meta `{ deleted: true }`, and a dedicated session ID pattern (`{accountId}_deleted_{uniqueId}`).
+- **Mixed-version safety**:
+  - New peers must **not accept** or **re-propagate** historical content for a deleted coValue (they only accept/sync the tombstone).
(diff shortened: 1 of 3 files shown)


aeplay (approved):

Gabrola:
This looks great, really looking forward to this!

Couple of questions:

1. Since physical deletion is purely for storage space reclamation. Would it be possible to have control over whether it happens at all or not? You can imagine there might be cases where we would not want to delete anything at all and keep a trail of everything à la blockchain.
2. Regarding the previous point, there might be a need for granular control over what gets physically deleted and what doesn't. A use-case for example would be not deleting anything to keep the history as an audit trail, but an explicit physical delete for sensitive data and compliance reasons. Would it make sense to have some sort of "soft-delete" primitive that might be as simple as flag in the `deletedCoValues` table? Technically achievable through userland soft-deletes, but less ergonomic.
3. Also regarding the previous points, if a coValue is not physically deleted and we want to see the history of a coValue and potentially restore it even. It seems like this would currently not be possible due to the tombstone blocking the sync of the non-delete transactions. Would it be possible (not now probably, but in the future) to explicitly bypass this somehow, say for those with `admin` permissions?
4. How would deleted coValues be handled when referenced in coLists or coMaps? Currently there are two approaches when doing userland soft-deletes.
    1. Make sure at delete-time that the referencing coValues are also updated to de-reference the deleted coValue
    2. Filter out deleted items at read-time.

    Both approaches are easily prone to errors due to forgetting to handle the deleted coValues, but the former less so because you'd need to remember only when a new or edited schemas starts newly referencing the deletable schema. The former has the advantage of keeping the `.length` property on coLists consistent, and also having the ability to handle any cascading logic. However, I'm not sure if the fist approach is even possible since jazz itself internally is not schema-aware. In this case:
    - Would is be possible at all for deleted coValues in a coList to not be included in `CoList.length`?
    - How would a deleted coValue be handled when referenced through a non-optional field on a coMap and when attempting to be deeply resolved?

Edit: I also realize that the last point is relatively non-trivial since its a referential integrity problem in a de facto graph database.

gdorsi:
Thanks @Gabrola for chiming in!

> Since physical deletion is purely for storage space reclamation

We are actually doing this for GDPR compliance (EU privacy laws) and making it possible for apps to implement full account delete.

> Would it be possible to have control over whether it happens at all or not?

Still TBD, but having two steps, tracking the delete and actually delete data could make it possible to enable hard-delete at sync server layer.

But, honestly, the focus is on hard-delete data and all the design revolves around making it distributed, so supporting soft-delete is a non-goal for now.

I'm also concerned that having a two phase delete on day one might be too much, I really want it for the performance implications, but need to validate benefit/cost balance.

> Would it make sense to have some sort of "soft-delete" primitive that might be as simple as flag in the deletedCoValues table? Technically achievable through userland soft-deletes, but less ergonomic.
> Also regarding the previous points, if a coValue is not physically deleted and we want to see the history of a coValue and potentially restore it even. It seems like this would currently not be possible due to the tombstone blocking the sync of the non-delete transactions. Would it be possible (not now probably, but in the future) to explicitly bypass this somehow, say for those with admin permissions?
> Would it be possible (not now probably, but in the future) to explicitly bypass this somehow, say for those with admin permissions?

This looks connected to me, and I'd say that's something we should make it possible via a different API.
My understanding is that you want a way to block the sync of certain coValues to non-admin accounts.

So the requirements are:
- given a coValue, you want to mark it as "soft-deleted"
- the soft-deleted state should be undoable by admins
- admins should be able to access to soft-deleted values
- non-admins should not be able to update or load soft-deleted values

I'm working on the implementation of the hard-delete at on the sync layer, and I have to say that's already complicated without any extra requirements.

But I appreciate your suggestions and can design the API in order to have a sort of "mode" toggle that would help us on supporting soft-deletes in the future. 

It should be possible to add this on top of the hard-delete implementation, it's only that I prefer to keep the scope of the current task at minimum to be able to have a better focus on the sync implications and the general design. 

> Technically achievable through userland soft-deletes, but less ergonomic.

Could you explain more about what you mean with "less ergonomic"?

gdorsi:
> How would deleted coValues be handled when referenced in coLists or coMaps?

The spec for this is still missing, because I can handle this problem separately and focus on the cojson layer for now.

Probably I'm going to add a `deleted` loadingState that would be considered the same as `unavailable`.

We could apply a `$onDeleted` behavoir that accounts can apply on read, similar to `$onError`.
Unsure about this, maybe $onError is enough for now.

I also think we should have both declarable in the schema, because what to do in these case is almost always the same and don't depends on where you are fetching the values.

Something like this:
```ts
export const MusicTrack = co.map({ ... }).onError("catch");
export const Playlist = co.map({ ... }).onError("catch");

// Where it would reflect
export const MusicaAccountRoot = co
  .map({
    rootPlaylist: Playlist,
    playlists: co.list(Playlist), // onError catch 
    activeTrack: MusicTrack.optional(), // onError catch
    activePlaylist: Playlist, // onError catch

    exampleDataLoaded: z.optional(z.boolean()),
    accountSetupCompleted: z.optional(z.boolean()),
  })
```

This would let me handle at schema level the situations where:
- access to a given Playlist is revoked, now I need to be careful on how to write the resolve to avoid making the app load fail when this happens
- MusicTrack is deleted, this could be a problem when loading a playlist on when I was listening that track

It's clear we need something better than catch, because I would prefer for the music track to be null when not valid rather than be both nullable and MaybeLoaded (which requires me to do an extra check!)

Same for when inside a collection, I want it to be skipped, not to make everything MaybeLoaded.

So, to give you an straight answer, I think we should leave the task of deleting the right values to the developer.

e.g. when deleting a MusicTrack you should also delete the file and the waveform. Then you just skip the deleted tracks and do periodically a purge if keeping them there becomes a performance issue.

> I also realize that the last point is relatively non-trivial since its a referential integrity problem in a de facto graph database.

In a _distributed graph database with offline writes_ which is even worse 😆 

One might add the track to a playlist while offine and another user deletes the track, and we need to support that case.

Gabrola:
@gdorsi Thank you for the in-depth reply! 

> This looks connected to me, and I'd say that's something we should make it possible via a different API. My understanding is that you want a way to block the sync of certain coValues to non-admin accounts.
> 
> So the requirements are:
> 
> * given a coValue, you want to mark it as "soft-deleted"
> * the soft-deleted state should be undoable by admins
> * admins should be able to access to soft-deleted values
> * non-admins should not be able to update or load soft-deleted values

Yes, exactly.

> I'm working on the implementation of the hard-delete at on the sync layer, and I have to say that's already complicated without any extra requirements.
> 
> But I appreciate your suggestions and can design the API in order to have a sort of "mode" toggle that would help us on supporting soft-deletes in the future.
> 
> It should be possible to add this on top of the hard-delete implementation, it's only that I prefer to keep the scope of the current task at minimum to be able to have a better focus on the sync implications and the general design.

Oh yes, definitely for sure! My suggestions/questions are more about the future, but absolutely not requesting anything with an expectation of changing the current spec or in-progress work.

> Could you explain more about what you mean with "less ergonomic"?

I actually failed to mention that userland soft-deletes wouldn't capable of the sync block functionality entirely. But what I meant was, when Jazz ever handles the references of deleted coValues, that would much less work than doing it manually.

> I also think we should have both declarable in the schema, because what to do in these case is almost always the same and don't depends on where you are fetching the values.

Fully agree. I think an `.onDelete()` handler similar to `.withMigration()` that is placed on the referencing coValue's schema would be really nice and it would map nicely to the familiar concept of `ON DELETE` constraints on foreign keys in relational DBs.

```ts
export const MusicaAccountRoot = co
  .map({
    rootPlaylist: Playlist,
    playlists: co.list(Playlist).onDelete((list, deletedValue) => {
      // where deletedValue.$jazz.loadingState === 'deleted'
      list.$jazz.remove((p) => p.$jazz.id === deletedValue.$jazz.id);
    }),
    activeTrack: MusicTrack.optional(),
    activePlaylist: Playlist,

    exampleDataLoaded: z.optional(z.boolean()),
    accountSetupCompleted: z.optional(z.boolean()),
  })
  .onDelete((musicaAccountRoot, deletedFieldName, deletedFieldValue) => {
    if (deletedFieldName === 'activeTrack') {
      musicaAccountRoot.$jazz.delete('activeTrack');
    }
  });
```

Only problem with this approach is that it would not work/run on accounts with no write permissions. So we'd still need the `.onError` handler additionally to deal with that case.

gdorsi:
> Fully agree. I think an .onDelete() handler similar to .withMigration() that is placed on the referencing coValue's schema would be really nice and it would map nicely to the familiar concept of ON DELETE constraints on foreign keys in relational DBs.

It would probably become a "hey, if you try to load this Playlist and it is deleted, delete also the entry from the list" without any callback.

This way we can handle the case where the user lacks the rights to update the list by simply skipping the entry when loading the list.

.specs/data-delete-flow/design.md
```diff
@@ -0,0 +1,770 @@
...
+**Changes**:
+- Modify `makeTransaction()` to accept optional `sessionID` parameter
+- When creating delete transaction, use the delete session ID pattern
+- Ensure delete transactions are always unencrypted (trusting)
+
+**Rationale for Special Delete Session**:
+The special session naming pattern (`{accountId}_session_z{uniqueId}_deleted`) is used to provide a simple way to sync the delete marker without carrying extra data. By using a dedicated session for delete transactions, the system can:
+- Easily identify delete operations by session ID pattern
+- Sync only the delete marker without carrying any history
+- Maintain a clean separation between regular transactions and delete operations
```
nrainhart, line 96: It'd be great if we can enforce this at the type level, so that delete sessions can only contain delete transactions, and regular sessions cannot contain delete transactions

.specs/data-delete-flow/design.md
```diff
@@ -0,0 +1,770 @@
...
+
+**Key Logic**:
+```typescript
+makeTransaction(
+  changes: JsonValue[],
+  privacy: "private" | "trusting",
+  meta?: JsonObject,
+  madeAt?: number,
+  sessionID?: SessionID, // Optional override for delete sessions
```
nrainhart, line 105: I worry that this option becomes a footgun, because it should only ever be used with delete transactions

.specs/data-delete-flow/design.md
```diff
@@ -0,0 +1,770 @@
...
+```
+
+### 3. Delete Transaction Validation in tryAddTransactions
+
+**Location**: `packages/cojson/src/coValueCore/coValueCore.ts` - `tryAddTransactions()`
+
+**Changes**:
+- After signature verification succeeds, check for delete transactions
+- Only parse transaction meta if sessionID ends with `_deleted` (optimization)
+- Validate that delete transaction author has admin permissions
```
nrainhart, line 129: We should rather validate that the author _had admin permissions_ at the time the transaction was created, right?

.specs/data-delete-flow/design.md
```diff
@@ -0,0 +1,770 @@
...
+
+**Throttling / batching**
+
+To prevent long pauses/locks:
+- process **at most** `maxCoValuesPerRun` (e.g. 50–500) per invocation.
+- optionally stop after a `maxDurationMs` budget (e.g. 100–300ms).
+- for each coValue, perform deletion in a **per-coValue transaction** (idempotent) so work can resume after interruptions.
+
+**Concurrency / locking**
+- The operation should be guarded so only one eraser runs per database at a time.
```
nrainhart, line 452: Is the following case possible? Do we need to do any special handling for it?
- sync server is streaming covalue to a client
- another client deletes the covalue
- streaming hasn't completed by the time `eraseAllDeletedCoValues` runs

.specs/data-delete-flow/design.md
```diff
@@ -0,0 +1,770 @@
...
+};
+
+type DeleteSessionID = `${SessionID}_deleted`;
+```
+
+### CoValue Deleted State
+
+```typescript
+type CoValueDeletedState = {
+  isDeleted: boolean;
```
nrainhart, line 666: Nit: `isDeleted` could just be a getter that's derived from `deleteSessionID`

nrainhart (commented):
The current approach makes sense to me, but I wonder how we can extend it to support partial deletes in the future. The delete transaction should contain information about which transactions need to be deleted.

A few other thoughts:
- The interaction between deletion and uniqueness really makes me think we should support some form of "CoValue resurrection" with new state.
- Even though this spec is scoped to cojson, I feel like we should have an idea on how deletion will be surfaced at the jazz-tools level before we commit to this approach

.specs/data-delete-flow/design.md
gdorsi: Unsure if this is possible given the current architecture, but going to give it a thought.

.specs/data-delete-flow/design.md
gdorsi: Yep, it's been removed from the final spec: https://github.com/garden-co/jazz/pull/3294/files#diff-391a59f58348bc03d6061bffd0d538ea75faeb0cf53659b50852a0f0b1b4e0ebR91

Should have closed this one, sorry!

.specs/data-delete-flow/design.md
gdorsi: Yep, that's what I've used in the final implemetation: https://github.com/garden-co/jazz/pull/3294/files#diff-391a59f58348bc03d6061bffd0d538ea75faeb0cf53659b50852a0f0b1b4e0ebR275

The spec is not aligned on the wording as you are pointing out, so going to fix it!

.specs/data-delete-flow/design.md
gdorsi: The final spec on this point is more in-depth and specifies what to do in each storage impl: https://github.com/garden-co/jazz/pull/3294/files#diff-391a59f58348bc03d6061bffd0d538ea75faeb0cf53659b50852a0f0b1b4e0ebR554

gdorsi: All these cases you mentioned, are not currently managed at the storage level of the spec, but on the sync layer.

Double deletes and streaming are supported by the design, and covered by the test suite.

There is no cancelation for the current in-flight streaming operations coming from the same storage we are going to delete.

While in theory this should work, I need to add test to be sure that things don't explode.
And maybe this could be a good opportunity to implement some explicit state management around streaming from storage.

.specs/data-delete-flow/design.md
gdorsi: Ended up removing `deleteSessionID` to simplify the management of multiple delete operations from different accounts.

gdorsi:
Closing in favor of https://github.com/garden-co/jazz/pull/3294/

gdorsi:
> The interaction between deletion and uniqueness really makes me think we should support some form of "CoValue resurrection" with new state.

Agree, and that needs to be spec'd before we ship the v1 of the delete because otherwise might be very complicated to add it in the future.

> Even though this spec is scoped to cojson, I feel like we should have an idea on how deletion will be surfaced at the jazz-tools level before we commit to this approach

State-wise there is something here: https://github.com/garden-co/jazz/pull/3294/files#diff-391a59f58348bc03d6061bffd0d538ea75faeb0cf53659b50852a0f0b1b4e0ebR809

It is bare-minimum, as I wanted to experiment with some apps before coming for a delete API with resolve. (done [here](https://github.com/garden-co/jazz/pull/3294/files#diff-d3594031d67c838b40000b56ff9286f71623a25a807a640b2efe7860027c6654R224), will try to add a spec for that tomorrow)

On error management I think for now `$onError` should be enough, as the issues I've faced are the same when I deal with shared values.

We need to improve `$onError` to be more flexible, but that's another topic.

Issue #2019, opened 2025-04-29
Implement an API to delete CoValues


gdorsi:
The idea is to give users the ability of requesting the delete of a given CoValue, from the sync servers and other users dbs.

PR #3294, opened 2025-12-17
feat: hard delete on coValues

gdorsi:
## Overview

Fixes https://github.com/garden-co/jazz/issues/2019

Implements the data deletion flow for coValues with immediate sync blocking and scheduled physical deletion. 
Deleted coValues leave tombstones sessions that prevent data resurrection.

## Example

```ts
export async function deleteMyMusicPlayerAccount() {
  const me = await MusicaAccount.getMe().$jazz.ensureLoaded({
    resolve: {
      root: {
        playlists: {
          $each: { $onError: "catch" },
        },
      },
    },
  });

  // Delete all playlists referenced in the user's root. (This may include invited playlists.)
  for (const playlist of me.root.playlists.values()) {
    if (!playlist.$isLoaded) continue;
    if (me.canAdmin(playlist)) {
      await deleteCoValues(Playlist, playlist.$jazz.id);
    } else {
      playlist.$jazz.owner.removeMember(me);
    }
  }

  // Delete all the user owned data
  await deleteCoValues(MusicaAccount, me.$jazz.id, {
    resolve: {
      profile: {
        avatar: {
          $each: true,
        },
      },
      root: {
        rootPlaylist: {
          tracks: {
            $each: {
              file: true,
              waveform: true,
            },
          },
        },
        playlists: true,
      },
    },
  });
}
```

## Key Design Choices

**Important Constraint**: Account and Group coValues cannot be deleted, otherwise we might not be able to verify the delete transactions.

- **Admin-only deletion**: Only admins can delete coValues, validated in `determineValidTransactions()`
- **Special delete session**: Uses `{SessionID}_deleted` session naming to sync delete markers, to quickly identify the delete transactions with zero cost for non-delete paths
- **Unencrypted transactions**: Delete transactions are trusting (unencrypted) with meta `{ deleted: true }`, to be inspected from the sync servers
- **Immediate sync blocking**: Deleted coValues block all sync immediately, only delete session/transaction can sync
- **Storage API batch deletion**: `ereaseAllDeletedCoValues()` performs physical deletion while preserving tombstones
- **Tombstone preservation**: Delete session + coValue header preserved even after physical deletion
- **Storage shard support**: Shards with `skipVerify: true` mark as deleted and block sync without verification

---

> [!NOTE]
> Introduces permanent CoValue deletion across core, storage, examples, and docs.
> 
> - **API/UX**: New `deleteCoValues` to hard-delete CoValues (supports nested via `resolve`, skips Groups/Accounts); introduces `deleted` loading state; examples (music-player) add delete track/playlist/account flows and UI dialog
> - **Core**: Enforces admin-only deletes, special delete sessions (`*_session_d…$`), prevents further writes to deleted values, and adjusts known-state/peer sync to block non-delete content
> - **Storage**: IndexedDB/SQLite add `deletedCoValues` work queue, APIs to mark and erase, and a scheduler for background erasure that preserves tombstones (header + delete session) while removing history; migrations and interrupt-safe behavior during load/store
> - **Docs**: New “Deleting CoValues” guide, code snippets, nav entry, and upgrade notes mentioning the `deleted` state
> - **Tests**: Extensive unit/e2e coverage for deletion, storage erasure, streaming edge cases, and app-level account deletion
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 29997b1dd19709663cac5a84dfc6df98f1595715. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


.specs/data-delete-flow/design.md
```diff
@@ -0,0 +1,920 @@
...
+3. Tombstone preserved after physical delete
+4. CoValue marked as deleted
+
+**Scenario 5: Sync Restriction**
+1. CoValue deleted
+2. Try to sync non-delete content
+3. Verify sync blocked
+4. Try to sync delete session
+5. Verify delete session syncs successfully
+
```
gdorsi:
```suggestion
```

.specs/data-delete-flow/design.md
```diff
@@ -0,0 +1,882 @@
...
+  }
+}
+```
+
+### 4. Delete Permission Validation
+
+**Location**: `packages/cojson/src/coValueCore/coValueCore.ts` - `#canAuthorDeleteCoValueAtTime(...)`
+
+**Changes**:
+- Check that transaction author has admin permissions when meta contains `{ deleted: true }`
```
gdorsi:
```suggestion
- Check that transaction author had admin permissions on `tx.madeAt` when meta contains `{ deleted: true }`
```

.specs/data-delete-flow/design.md
```diff
@@ -0,0 +1,882 @@
...
+    throw new Error("Cannot delete coValue without verified state");
+  }
+
+  if (this.isGroupOrAccount()) {
+    throw new Error("Cannot delete Group or Account coValues");
+  }
+
+  if (this.isDeleted) {
+    throw new Error("CoValue is already deleted");
+  }
```
nrainhart: What made you decide to throw an error here instead of making deletion idempotent? I've got no strong preference, but wondering what your rationale was

gdorsi: Agree, I got into this error in my music player experiment and it's quite annoying.

Going to replace with an early return.

packages/jazz-tools/src/tools/subscribe/types.ts
```diff
@@ -31,6 +35,7 @@ export type CoValueLoadingState =
 
 export type CoValueErrorState =
   | typeof CoValueLoadingState.UNAVAILABLE
+  | typeof CoValueLoadingState.DELETED
```
nrainhart, line 38: ⚠️ This is technically a breaking change. It will break switch statements with no default clause that were previously exhaustive. Not sure how strict we want to be with it.

packages/jazz-tools/src/tools/coValues/interfaces.ts
```diff
@@ -178,6 +179,7 @@ export function loadCoValue<
         syncResolution: true,
         skipRetry: options.skipRetry,
         onUnavailable: resolve,
+        onDeleted: resolve,
         onUnauthorized: resolve,
```
nrainhart, line 183: This will scale pretty badly if we keep adding loading states. Should we collapse them into a single `onError` callback?

Not sure if we consider these functions part of our public API or not

packages/jazz-tools/src/tools/coValues/interfaces.ts
```diff
@@ -688,23 +700,11 @@ export async function exportCoValue<
...
-      }
-    });
-  });
-
-  if (!value) {
+  try {
+    await rootNode.getPromise();
+    rootNode.destroy();
+  } catch (error) {
+    rootNode.destroy();
```
nrainhart, line 707: ✨

packages/jazz-tools/src/tools/coValues/interfaces.ts
```diff
@@ -857,19 +857,129 @@ export async function unstable_mergeBranchWithResolve<
...
+      id,
+      CoValueLoadingState.DELETED,
+      errors.flatMap((e) => e.issues),
+    );
+    throw new Error(combined.toString());
+  }
+
+  const deletedValues = deleteCoValueFromSubscription(rootNode);
+
+  await Promise.all(Array.from(deletedValues, (value) => value.waitForSync()));
```
nrainhart: Should we use `allSettled` here? This function shouldn't bail out early if one deletion fails to sync

packages/jazz-tools/src/tools/tests/deleteCoValues.test.ts
```diff
@@ -0,0 +1,148 @@
...
+      deleteCoValues(Note, note.$jazz.id, {
+        loadAs: writer,
+        resolve: { meta: true },
+      }),
+    ).rejects.toThrow(/admin permissions/i);
+
+    expect(note.$jazz.raw.core.isDeleted).toBe(false);
+  });
+
+  test("rejects deletion when a resolved child is not deletable (error includes path)", async () => {
```
nrainhart, line 94: Should we include tests for error when loading the CoValues to delete?

packages/jazz-tools/src/tools/tests/deletedState.test.ts
```diff
@@ -0,0 +1,110 @@
...
+  coValueClassFromCoValueClassOrSchema,
+  subscribeToCoValue,
+} from "../internal.js";
+import { loadCoValue } from "../exports.js";
+import { createJazzTestAccount, setupJazzTestSync } from "../testing.js";
+import { setupAccount, waitFor } from "./utils.js";
+
+cojsonInternals.setCoValueLoadingRetryDelay(50);
+
+describe("deleted loading state", () => {
```
nrainhart, line 17: It's kind of a pain to test because of the API surface, but I wonder if deletion will cause issues with any jazz-tools-level CoValue operation that's synchronously performed after the deletion. Ignoring changes made after the deletion makes sense at the cojson level, but maybe breaks some jazz-tool CoValue operations

packages/jazz-tools/src/react-core/tests/useSuspenseCoState.test.tsx
```diff
@@ -214,6 +214,50 @@ describe("useSuspenseCoState", () => {
...
+        </ErrorBoundary>,
+        {
+          account: owner,
+        },
+      );
+    });
+
+    await waitFor(
+      () => {
+        expect(container.textContent).toContain("Error!");
```
nrainhart: Could we assert the error we throw is indeed a "deleted" error?

nrainhart (commented):
Reviewed the jazz-tools changes, will do a second pass on the cojson changes later

packages/cojson/src/storage/DeletedCoValuesEraserScheduler.ts
```diff
@@ -0,0 +1,101 @@
...
+    }, delayMs);
+  }
+
+  private async run() {
+    if (this.disposed) return;
+
+    // Clear any pre-run scheduled state and enter running state.
+    this.state = "running";
+
+    const result = await this.runOnce();
```
antoniomuso: We should handle the case where `runOnce` throws an error, set the state to a safe state (retry?), and forward the error message (onError?).

packages/cojson/src/storage/DeletedCoValuesEraserScheduler.ts
```diff
@@ -0,0 +1,101 @@
...
+type SchedulerState =
+  | "idle"
+  | "startup_scheduled"
+  | "throttle_scheduled"
+  | "followup_scheduled"
+  | "running"
+  | "disposed";
+
+export class DeletedCoValuesEraserScheduler {
+  private readonly runOnce: () => Promise<DeletedCoValuesEraserSchedulerRunResult>;
```
antoniomuso: `runOnce` name makes me think it will run only one time. Problably it is better something like `runJob` WDYT?

packages/cojson/src/sync.ts
```diff
@@ -617,13 +624,19 @@ export class SyncManager {
...
+    let wasAlreadyDeleted = coValue.isDeleted;
+
     /**
      * The coValue is in memory, load the transactions from the content message
      */
     for (const [
       sessionID,
       newContentForSession,
     ] of getSessionEntriesFromContentMessage(msg)) {
+      if (coValue.isDeleted && !isDeletedSessionID(sessionID)) {
```
nrainhart: Nit: pretty sure both are equivalent at this point, but to me wasAlreadyDeleted is semantically clearer
```suggestion
      if (wasAlreadyDeleted && !isDeletedSessionID(sessionID)) {
```

nrainhart: Actually, shouldn't we ignore all new content in this case? Even the one coming from delete transactions

packages/cojson/src/sync.ts
```diff
@@ -818,6 +852,10 @@ export class SyncManager {
 
     const value = this.local.getCoValue(content.id);
 
+    if (value.isDeleted) {
+      storage.markCoValueAsDeleted(value.id);
+    }
```
nrainhart, line 1065: I wonder if this should happen in the same transaction that updates the content. Otherwise we risk ending up with inconsistent data in storage if `storage.markCoValueAsDeleted` completes and `storage.store` does not

packages/cojson/src/tests/sync.deleted.test.ts
```diff
@@ -0,0 +1,296 @@
...
+
+    const groupContent = group.core.newContentSince(undefined)?.[0];
+    assert(groupContent);
+
+    client.node.syncManager.handleNewContent(content, "import");
+    client.node.syncManager.handleNewContent(groupContent, "import");
+
+    await waitFor(() => {
+      expect(client.node.expectCoValueLoaded(map.id).isDeleted).toBe(true);
+    });
```
nrainhart, line 110: Should we check `isDeleted` is false until the group content is processed?

packages/jazz-tools/src/tools/subscribe/types.ts
gdorsi: This is a good point, we may need a minor to ship deletes

packages/cojson/src/tests/sync.deleted.test.ts
```diff
@@ -0,0 +1,296 @@
...
+    map.core.deleteCoValue();
+    await waitFor(() => {
+      expect(mapOnClient.core.isDeleted).toBe(true);
+    });
+
+    SyncMessagesLog.clear();
+
+    expect(() => {
+      mapOnClient.set("x", "y", "trusting");
+    }).toThrow(/deleted coValue/i);
```
nrainhart: I thought CoValue updates after a delete would be ignored instead of throwing (as it says in [this comment](https://github.com/garden-co/jazz/pull/3294/changes#diff-f4de88cd1dd209c7fe4f28e41a46d285f10a88f3c5106b944ba52f11c6f62601R631-R633)). Don't think throwing is a good idea in the user-facing api

packages/jazz-tools/src/tools/coValues/interfaces.ts
gdorsi: Yeah, my idea was to add the callback and then break the subscribe API to have a better way to handle errors.

This is a legacy of what we had to do before we had explicit loading states, maybe now we can directly put errors in the subscribe callback?

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -611,6 +663,58 @@ export class CoValueCore {
       };
     }
 
+    // Detect + validate delete transactions during ingestion (US-3)
```
```suggestion
    // Detect + validate delete transactions during ingestion
```

nrainhart: Nit: we could also extract all of this code for obtaining the delete transaction into a function

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -620,6 +724,13 @@ export class CoValueCore {
         skipVerify,
       );
 
+      // Mark deleted state when a delete marker transaction is accepted.
+      // - In skipVerify mode (storage shards), we accept + mark without permission checks.
+      // - In verify mode, we only reach here if the delete permission check passed.
+      if (deleteTransaction && !this.isGroupOrAccount()) {
```
nrainhart: Shouldn't we make the group/account check earlier and throw an error asap?

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -628,6 +739,75 @@ export class CoValueCore {
...
+        ok: false,
+        reason: "CoValueNotDeletable",
+        message: "Cannot delete Group or Account coValues",
+      };
+    }
+
+    const group = this.safeGetGroup();
+
+    // Today, delete permission is defined in terms of group-admin on the owning group.
+    // If we cannot derive that (non-owned coValues), we reject the delete when verification is required.
```
nrainhart, line 893: We should never hit this check because we already validated the covalue is not a group or an account, right? I'd add this to the comment

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -725,6 +905,70 @@ export class CoValueCore {
     };
   }
 
+  validateDeletePermissions() {
```
nrainhart, line 1014: This method could be private, right?

packages/cojson/src/coValueCore/SessionMap.ts
```diff
@@ -60,7 +61,32 @@ export class SessionMap {
     }
   }
 
+  markAsDeleted() {
+    this.isDeleted = true;
+
+    // We reset the known state to report only the deleted session/transaction
+    this.knownState = { id: this.id, header: true, sessions: {} };
+
+    // We remove the streaming stauses, because once deleted we don't need
```
nrainhart:
```suggestion
    // We remove the streaming statuses, because once deleted we don't need
```

packages/cojson/src/tests/deleteCoValue.test.ts
```diff
@@ -0,0 +1,423 @@
...
+    /Cannot delete Group or Account coValues/,
+  );
+
+  const group = client.node.createGroup();
+  expect(() => group.core.deleteCoValue()).toThrow(
+    /Cannot delete Group or Account coValues/,
+  );
+});
+
+test("deleteCoValue throws when called by a non-admin on a group-owned CoValue", async () => {
```
nrainhart, line 51: We could add a test for time travel when validating permissions as well (if the delete transaction was created before a user got admin rights, deleteCoValue should fail as well)

packages/cojson/src/storage/storageAsync.ts
```diff
@@ -388,11 +403,44 @@ export class StorageApiAsync implements StorageAPI {
...
+      runOnce: async () => {
+        // Async storage: no max-time budgeting; drain to completion when scheduled.
+        await this.eraseAllDeletedCoValues();
+        const remaining = await this.dbClient.getAllCoValuesWaitingForDelete();
+        return { hasMore: remaining.length > 0 };
+      },
+      opts: {
+        throttleMs: 60_000,
+        startupDelayMs: 1_000,
+        followUpDelayMs: 1_000,
```
nrainhart: Nit: reuse the constants in storageSync

packages/cojson/src/storage/DeletedCoValuesEraserScheduler.ts
```diff
@@ -0,0 +1,101 @@
...
+  private readonly opts: DeletedCoValuesEraserSchedulerOpts;
+
+  private state: SchedulerState = "idle";
+  private disposed = false;
```
nrainhart:
```suggestion
  private get disposed() {
  	return this.state === "disposed";
  }
```

packages/cojson/src/storage/sqlite/client.ts
```diff
@@ -141,6 +145,77 @@ export class SQLiteClient
     return result.rowID;
   }
 
+  markCoValueAsDeleted(id: RawCoID) {
```
nrainhart, line 150: We've got no tests for sqlite clients?

packages/cojson/src/storage/DeletedCoValuesEraserScheduler.ts
gdorsi: I don't think we need a retry, because the delete is performed on a local DB, but agree that we need to cover errors here.

packages/cojson/src/storage/sqlite/client.ts
```diff
@@ -141,6 +145,77 @@ export class SQLiteClient
...
+    this.db.run(
+      `INSERT INTO deletedCoValues (coValueID, status) VALUES (?, 'done')
+       ON CONFLICT(coValueID) DO UPDATE SET status='done'`,
+      [id],
+    );
+  }
+
+  eraseCoValueButKeepTombstone(coValueId: RawCoID) {
+    const coValueRow = this.db.get<RawCoValueRow & { rowID: number }>(
+      "SELECT * FROM coValues WHERE id = ?",
```
nrainhart:
```suggestion
      "SELECT rowID FROM coValues WHERE id = ?",
```

packages/jazz-tools/src/tools/tests/deletedState.test.ts
gdorsi: Could you do some examples?

packages/cojson/src/storage/sqlite/client.ts
```diff
@@ -141,6 +145,77 @@ export class SQLiteClient
...
+    // Single-transaction primitive: delete non-delete sessions and their data,
+    // preserve header + delete sessions (tombstone).
+    //
+    // Note: `sessions.coValue` references `coValues.rowID`.
+    this.db.run(
+      `DELETE FROM transactions
+       WHERE ses IN (
+         SELECT rowID FROM sessions
+         WHERE coValue = ?
+           AND sessionID NOT LIKE '%_deleted'
```
nrainhart: Do we need an index by coValue id to make this more performant? Same thing in the other tables

packages/cojson/src/storage/sqlite/client.ts
```diff
@@ -141,6 +145,77 @@ export class SQLiteClient
...
+      [coValueRow.rowID],
+    );
+  }
+
+  getAllCoValuesWaitingForDelete(): RawCoID[] {
+    return this.db
+      .query<DeletedCoValueQueueRow>(
+        `SELECT coValueID as id
+         FROM deletedCoValues
+         WHERE status = 'pending'`,
```
nrainhart: Status index?

packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -40,6 +45,75 @@ export class IDBTransaction implements DBTransactionInterfaceAsync {
...
+  async markCoValueDeletionDone(id: RawCoID): Promise<void> {
+    await this.run((tx) =>
+      tx.getObjectStore("deletedCoValues").put({
+        coValueID: id,
+        status: "done",
+      } satisfies DeletedCoValueQueueEntry),
+    );
+  }
+
+  async eraseCoValueButKeepTombstone(coValueID: RawCoID): Promise<void> {
```
nrainhart: Did you do any sort of performance testing? It looks like this could be quite slow. Mostly worried about deletes locking object stores and affecting other write operations

nrainhart (commented):
Second review pass, focusing on cojson

packages/jazz-tools/src/tools/coValues/interfaces.ts
nrainhart: > maybe now we can directly put errors in the subscribe callback?

I think I prefer having a separate onError callback, but we can try and see which option we like the most

packages/jazz-tools/src/tools/tests/deletedState.test.ts
nrainhart: Actually, this is related to this comment: https://github.com/garden-co/jazz/pull/3294#discussion_r2650821785.

I thought cojson updates were ignored instead of throwing. If we're throwing, then disregard my original comment.

packages/jazz-tools/src/tools/coValues/interfaces.ts
gdorsi: NVM, added `onError` and marked `onUnavailable` and `onUnauthorized` as deprecated

packages/cojson-storage-indexeddb/src/idbClient.ts
nrainhart: For browsers, I'm even wondering if we should use something like [requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback) to avoid stalling more important work

packages/cojson/src/sync.ts
gdorsi: That's for syncing delete txs coming from different accounts.

packages/cojson/src/sync.ts
gdorsi: Nice catch, going to fix

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: Good point

packages/cojson/src/storage/sqlite/client.ts
gdorsi: All the storage tests run by default in sqlite clients using libsql, so they are already on cojson

packages/cojson/src/sync.ts
gdorsi: Necessary to keep the sync status consistent across all clients

packages/cojson/src/storage/sqlite/client.ts
gdorsi: Ahh, good catch

We alredy have the index by coValue in the sessions table, but we don't have an index by session in both transactions and signatureAfter, that's a problem!

nrainhart: Not sure a session index is going to help with the LIKE query, though. That's why I was asking about the covalue idx. I see we have an index by ses+idx in the transactions table, but I don't think it'll be used if you query only by covalue id

gdorsi: Ah no, session is part of the primary key for both tables, so they are already indexed

packages/cojson/src/storage/sqlite/client.ts
gdorsi: Good point

Changed the field into INTEGER and added the index.

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: It is used in jazz-tools to validate the permission before the actual delete

packages/cojson-storage-indexeddb/src/idbClient.ts
gdorsi: `requestIdleCallback` is more suited for scheduling low priority work on the main thread.

The operation is not render-blocking, the actual delete is performed not on the main thread.

We could optimize this operation by running it in a single transaction, let me do some tests.

packages/cojson/src/storage/sqlite/client.ts
gdorsi: > Not sure a session index is going to help with the LIKE query, though

That one is a nested query that runs on session so should use the coValue index and then filter the results with LIKE.

Sub-optimal, we could add a `type` flag on session to filter out sessions without resorting on `LIKE` but I'd like to avoid to do alter tables.

>  I see we have an index by ses+idx in the transactions table, but I don't think it'll be used if you query only by covalue id

The query on the transactions is filtered by session.
Actually I'm not sure if indexes would work on compound primary keys when filtering by one of the fields, needs some testing.

packages/cojson-storage-indexeddb/src/idbClient.ts
gdorsi: As you have predicted, it's really slow.

Tried some experiments, but with IndexedDB you can bulk delete only using ranges on primary keys.

So I made load and store interrupt any in-flight delete process using AbortController.

This should avoid that delete batches would block any other storage operation.

https://github.com/garden-co/jazz/pull/3294/commits/8dff941ff301910a6d50bed473eeaca5d66b966c

WDYT?

packages/cojson/src/storage/sqlite/client.ts
nrainhart: > Actually I'm not sure if indexes would work on compound primary keys when filtering by one of the fields, needs some testing

I _think_ they won't, but yeah let's test it

packages/cojson-storage-indexeddb/src/idbClient.ts
nrainhart: Like the idea! Shouldn't the remainder of the delete process be rescheduled, though? Or is that already happening?

packages/cojson/src/storage/sqlite/client.ts
gdorsi: Yes, it uses an index. The ses column is the first column of the composite primary key (ses, idx). Since the table is WITHOUT ROWID, it's organized as a clustered index on the primary key. SQLite can use the leftmost prefix of a composite index, so filtering by ses will use the primary key index.

To be sure I've used EXPLAIN QUERY PLAN and it confirms my supposition:
```
[
  {
    id: 3,
    parent: 0,
    notused: 0,
    detail: 'SEARCH transactions USING PRIMARY KEY (ses=?)'
  },
  { id: 7, parent: 0, notused: 0, detail: 'LIST SUBQUERY 1' },
  {
    id: 10,
    parent: 7,
    notused: 0,
    detail: 'SEARCH sessions USING INDEX sessionsByCoValue (coValue=?)'
  }
]
```

packages/cojson-storage-indexeddb/src/idbClient.ts
nrainhart: I think we need better guidelines on when to add a method to the DBTransactionInterface vs when to add it to the DBClientInterface. Until now it looks like DBTransactionInterface contained individual operations that were meant to be run as part of a larger transaction, but now eraseCoValueButKeepTombstone is doing a bunch of things. What do you intend the scope of DBTransactionInterface to be?

gdorsi: It's already happening, the `hasMore` flag handles that.

gdorsi: The scope is everything that should run as part of a readwrite transaction.

It gives use the property of knowning that everything in this class is meant to be executed sequentially and never concurrently, across all the stores and tabs.

Sync state for example, probably needs a different abstraction because it runs on a single store.

nrainhart: Mind adding that as a TSdoc on top of `IDBTransaction`? Will be useful for future reference

gdorsi: Done: https://github.com/garden-co/jazz/pull/3294/commits/d2bbd00ca20da1265331223a78d00027a5e68101

homepage/homepage/content/docs/core-concepts/deleting.mdx
```diff
@@ -0,0 +1,66 @@
+import { Alert, CodeGroup } from "@/components/forMdx";
+
+export const metadata = {
+  description: "Learn how to permanently delete CoValues and their nested references using the deleteCoValues function"
+};
+
+# Deleting CoValues
+
+The `deleteCoValues` function allows you to permanently delete CoValues from Jazz. When a CoValue is deleted, it is marked with a tombstone, making it inaccessible to all users.
```
nrainhart:
```suggestion
The `deleteCoValues` function allows you to permanently delete CoValues from Jazz. When a CoValue is deleted, it becomes inaccessible to all users.
```
Not sure how relevant the tombstone part is for a high-level explanation of how deletion works. You also repeat it in the next paragraph

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -22,6 +22,7 @@ This also helps us speed up the migration of Jazz Core to Rust which will help u
 - We dropped support for `RNQuickCrypto`. RNCrypto is now the default crypto provider on React Native.
 - There is no more fallback to Javascript crypto. If the crypto fails to initialize, it will throw an error. 
 **NB**: if you use WASM crypto on Edge runtimes, you need to ensure that WASM is available and initialized. See more details at [WASM on Edge runtimes](/docs/react/server-side/setup#wasm-on-edge-runtimes).
+- We have intruduced a way to permanently [delete CoValues](/docs/core-concepts/covalues/deleting). This brings a new `deleted` loading state.
```
nrainhart:
```suggestion
- We have introduced a way to permanently [delete CoValues](/docs/core-concepts/covalues/deleting). This brings a new `deleted` loading state.
```

packages/jazz-tools/src/tools/coValues/interfaces.ts
```diff
@@ -223,12 +225,15 @@ type SubscribeListener<V extends CoValue, R extends RefsToResolve<V>> = (
...
+
 export type SubscribeListenerOptions<
   V extends CoValue,
   R extends RefsToResolve<V>,
 > = {
   resolve?: RefsToResolveStrict<V, R>;
   loadAs?: Account | AnonymousJazzAgent;
+  onError?: (value: NotLoaded<V>) => void;
   onUnauthorized?: (value: NotLoaded<V>) => void;
   onUnavailable?: (value: NotLoaded<V>) => void;
```
nrainhart, line 238: Should these two options be marked as deprecated?

packages/jazz-tools/src/tools/tests/deleteCoValues.test.ts
```diff
@@ -0,0 +1,231 @@
...
+      new RegExp(
+        `Subscription starts from ${root.$jazz.id}.*path ${child.$jazz.id}`,
+      ),
+    );
+
+    expect(root.$jazz.raw.core.isDeleted).toBe(false);
+    expect(child.$jazz.raw.core.isDeleted).toBe(false);
+  });
+
+  test("rejects deletion when a the value cannot be loaded", async () => {
```
nrainhart:
```suggestion
  test("rejects deletion when the value cannot be loaded", async () => {
```

packages/cojson/src/tests/sync.deleted.test.ts
```diff
@@ -0,0 +1,294 @@
...
+
+    client.node.syncManager.handleNewContent(content, "import");
+    client.node.syncManager.handleNewContent(groupContent, "import");
+
+    await waitFor(() => {
+      expect(client.node.expectCoValueLoaded(map.id).isDeleted).toBe(true);
+    });
+  });
+
+  test("outbound blocking: post-delete normal writes throw and do not produce content uploads", async () => {
```
nrainhart:
```suggestion
  test("outbound blocking: post-delete normal writes are ignored and do not produce content uploads", async () => {
```

packages/cojson-storage-indexeddb/src/tests/storage.indexeddb.test.ts
```diff
@@ -4,30 +4,31 @@ import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
...
   syncMessages = trackMessages();
   cojsonInternals.setSyncStateTrackingBatchDelay(0);
   cojsonInternals.setCoValueLoadingRetryDelay(10);
 });
 
 afterEach(async () => {
   syncMessages.restore();
-  cojsonInternals.setSyncStateTrackingBatchDelay(1000);
-
-  await clearObjectStore(DATABASE_NAME, "unsyncedCoValues");
```
nrainhart, line 30: We can also delete this utility function

packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -41,6 +52,75 @@ export class IDBTransaction implements DBTransactionInterfaceAsync {
...
+  async markCoValueAsDeleted(id: RawCoID): Promise<void> {
+    await this.run((tx) =>
+      tx.getObjectStore("deletedCoValues").put({
+        coValueID: id,
+        status: "pending",
+      } satisfies DeletedCoValueQueueEntry),
+    );
+  }
+
+  async markCoValueDeletionDone(id: RawCoID): Promise<void> {
```
nrainhart: Is this method used? I see there's a similar one in the `IDBClient` down below

packages/cojson/src/storage/sqlite/client.ts
```diff
@@ -142,6 +147,81 @@ export class SQLiteClient
...
+    if (!coValueRow) {
+      logger.warn(`CoValue ${coValueId} not found, skipping deletion`);
+      return;
+    }
+
+    // Single-transaction primitive: delete non-delete sessions and their data,
+    // preserve header + delete sessions (tombstone).
+    //
+    // Note: `sessions.coValue` references `coValues.rowID`.
+    this.db.run(
```
nrainhart: We're not doing all of these deletes inside a transaction?

nrainhart (approved):
Approved! Have a few comments, but nothing major

packages/cojson/src/storage/sqlite/client.ts
gdorsi: The transaction wrapper is applied by the StorageAPI [here](https://github.com/garden-co/jazz/blob/58fe8566e500f467ac70e0ff06e6b4f403d202cc/packages/cojson/src/storage/storageAsync.ts#L247).

Going make this a tx only API so we can run it only inside a transaction.

packages/jazz-tools/src/tools/tests/deletedState.test.ts
gdorsi: It doesn't throw anymore.

At jazz-tools level we make the value NotLoaded, so I guess it's ok on both cases.

packages/cojson-storage-indexeddb/src/idbClient.ts
gdorsi: Moved it inside eraseCoValueButKeepTombstone, so it's easier to run it inside a transaction

packages/cojson/src/storage/sqlite/client.ts
gdorsi: https://github.com/garden-co/jazz/pull/3294/commits/428b4a812737d3a10d1820e700ef717a110051d8

packages/cojson-storage-indexeddb/src/tests/storage.indexeddb.test.ts
gdorsi: Done https://github.com/garden-co/jazz/pull/3294/commits/62d1dec890a6035784cc40a9f4c00d4f19c55e93

homepage/homepage/content/docs/core-concepts/deleting.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+The `deleteCoValues` function allows you to permanently delete CoValues from Jazz. When a CoValue is deleted, it becomes inaccessible to all users.
+
+<Alert variant="warning" className="my-4" title="Irreversible Operation">
+  Deletion is permanent and cannot be undone. Once a CoValue is deleted, it cannot be recovered. Make sure you have proper confirmation flows in your application before calling `deleteCoValues`.
+</Alert>
+
+<Alert variant="warning" className="my-4" title="Admin Permissions Required">
+  Only users with admin permissions on a CoValue's group can delete it. Attempting to delete a CoValue without admin access will throw an error.
+</Alert>
```
joeinnes, line 17: Normally I'd say that we should try to separate these somehow so it doesn't look quite so dramatic, but I think actually here, they work.

homepage/homepage/content/docs/core-concepts/deleting.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+Deleted values are not deleted from storage immediately, but are marked with a tombstone.
+
+The physical deletion of the data from storage is done asynchronously in the background, and is dependent on the sync server's configuration.
+
+Jazz Cloud and `jazz-run sync` have delete enabled by default. If you're running a custom self-hosted sync server, you need to enable this feature.
```
joeinnes:
```suggestion
Jazz Cloud and `jazz-run sync` have delete enabled by default on a one minute schedule. If you're running a custom self-hosted sync server, you need to enable this feature.

Deleted CoValues stored in Jazz Cloud may persist in back-ups until they are overwritten. These back-ups undergo no additional processing, and will only be restored for disaster recovery purposes.
```
I'm not sure what our retention period is for back-ups at the moment, but adding that in here might also be re-assuring. Also assuming the 1 minute default in `DEFAULT_DELETE_SCHEDULE_OPTS` applies for Jazz Cloud

homepage/homepage/content/docs/core-concepts/deleting.mdx
```diff
@@ -0,0 +1,66 @@
...
+  Deletion is permanent and cannot be undone. Once a CoValue is deleted, it cannot be recovered. Make sure you have proper confirmation flows in your application before calling `deleteCoValues`.
+</Alert>
+
+<Alert variant="warning" className="my-4" title="Admin Permissions Required">
+  Only users with admin permissions on a CoValue's group can delete it. Attempting to delete a CoValue without admin access will throw an error.
+</Alert>
+
+Deleted values are not deleted from storage immediately, but are marked with a tombstone.
+
+The physical deletion of the data from storage is done asynchronously in the background, and is dependent on the sync server's configuration.
```
joeinnes:
```suggestion
To balance performance considerations, the actual physical deletion of the data from storage is done asynchronously in the background, and is dependent on the sync server's configuration.
```

As far as I understand, although the delete is async from a technical perspective, from a practical/legal perspective, the CoValue will be deleted from the sync server in *almost* real time (1 minute or so). 

It might be worth clarifying that the reason for async is to allow the sync server to delete CoValues when it's otherwise idle, rather than 'it's going to delete your data in 90 business days'.

Folks building serious apps may need to provide details on this when executing data processing addenda/completing security reviews with their customers.

joeinnes (commented):
Add review

PR #3296, opened 2025-12-19
feat: removed stableStringify in some code paths

antoniomuso:
# Description

Moved stable JSON serialization from JavaScript to Rust in SessionLog operations

### Changes

- **`tryAdd`**: Stable serialization now happens in Rust. The Rust layer parses each transaction and re-serializes it to ensure a stable JSON representation for signature verification. JavaScript side now uses `JSON.stringify` instead of `stableStringify`.

- **`addNewPrivateTransaction`** and **`addNewTrustingTransaction`**: Removed `stableStringify` usage since the data is either encrypted (private) or already in string format (trusting), making stable serialization unnecessary on the JS side.

## Tests

## Checklist

---

> [!NOTE]
> Simplifies `SessionLog.try_add` by forwarding raw JSON strings to core instead of pre-parsing them.
> 
> - Core N-API and RN: remove `serde_json::RawValue` parsing and call `internal.try_add(transactions_json, ...)` directly
> - WASM: regenerated `cojson_core_wasm.d.ts`/`.js` with reordered exports and `InitOutput` entries; no net API change
> - iOS: adjust `CojsonCoreRnFramework.xcframework/Info.plist` to correctly label `ios-arm64` vs `ios-arm64-simulator` and simulator variant
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8dda2473c02cc2b9ed99f6fbaa1a81b5ef0385b6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/cojson/src/crypto/NapiCrypto.ts
```diff
@@ -190,7 +190,7 @@ class SessionLogAdapter {
     skipVerify: boolean,
   ): void {
     this.sessionLog.tryAdd(
-      transactions.map((tx) => stableStringify(tx)),
+      transactions.map((tx) => JSON.stringify(tx)),
```
antoniomuso, line 193: Here, we are replacing `stableStringify` with `JSON.stringify`. The only overhead is parsing the transaction inside Rust.

packages/cojson/src/crypto/NapiCrypto.ts
```diff
@@ -205,12 +205,14 @@ class SessionLogAdapter {
     meta: JsonObject | undefined,
   ): { signature: Signature; transaction: PrivateTransaction } {
     const output = this.sessionLog.addNewPrivateTransaction(
-      stableStringify(changes),
+      // We can avoid stableStringify because it will be encrypted.
+      JSON.stringify(changes),
```
antoniomuso, line 209: `stableStringify` is not necessary here, deterministic serialization is only required for the transaction attributes. We can use non-stable serialization for the `changes`/`meta` since they are exchanged as strings and treated as immutable data. This guarantees the hash remains consistent.

packages/cojson/src/crypto/NapiCrypto.ts
```diff
@@ -229,8 +231,10 @@ class SessionLogAdapter {
     madeAt: number,
     meta: JsonObject | undefined,
   ): { signature: Signature; transaction: TrustingTransaction } {
-    const stringifiedChanges = stableStringify(changes);
-    const stringifiedMeta = meta ? stableStringify(meta) : undefined;
+    // We can avoid stableStringify because the changes will be in a string format already.
```
antoniomuso, line 234: `stableStringify` is not necessary here, deterministic serialization is only required for the transaction attributes. We can use non-stable serialization for the `changes`/`meta` since they are exchanged as strings and treated as immutable data. This guarantees the hash remains consistent.

packages/cojson/src/tests/coreWasm.test.ts
```diff
@@ -133,4 +134,58 @@ describe("SessionLog WASM", () => {
 
     expect(decrypted).toEqual(fixtures.decrypted);
   });
+
+  it("Signature is still valid after shuffling the transaction keys", () => {
+    const { agent, session, node } = createTestNode();
+
+    const group = node.createGroup();
+    const sessionContent =
+      group.core.newContentSince(undefined)?.[0]?.new[session];
```
gdorsi: This only covers trusting transactions, and depends on the init of Group.

How about being more explicit on the test input using a CoMap filled with test data> (e.g. 100 k/v entries)
To fully cover the changes should write tests for trusting and private transactions, same goes with metas.

antoniomuso: done!

gdorsi (approved):

PR #3306, opened 2025-12-24
docs: spec for CoValue sync status tracking & sync resumption

nrainhart:
Preview links:

- Requirements: [.specs/unsynced-covalues-tracking/requirements.md](https://github.com/garden-co/jazz/blob/docs/spec-covalue-sync-tracking/.specs/unsynced-covalues-tracking/requirements.md)
- Design: [.specs/unsynced-covalues-tracking/design.md](https://github.com/garden-co/jazz/blob/docs/spec-covalue-sync-tracking/.specs/unsynced-covalues-tracking/design.md)
- Tasks: [.specs/unsynced-covalues-tracking/tasks.md](https://github.com/garden-co/jazz/blob/docs/spec-covalue-sync-tracking/.specs/unsynced-covalues-tracking/tasks.md)

---

> [!NOTE]
> Introduces specifications for tracking unsynced `CoValue`s and resuming their sync after restarts, with reactive sync status APIs and storage-backed persistence.
> 
> - Extend `StorageAPI` with `trackCoValueSyncStatus`, `getUnsyncedCoValueIDs`, and `stopTrackingSyncStatus`
> - New `UnsyncedCoValuesTracker` to maintain per-peer unsynced sets, persist state, and notify listeners
> - Integrate into `SyncManager`: track on `syncContent`/`handleNewContent`, add `trackSyncState`, and async `resumeUnsyncedCoValues`; call during peer reconciliation
> - Add subscriptions: `CoValueCore.subscribeToSyncStatus` and `SyncManager.subscribeToSyncStatus`
> - Refactor `CoValueCore.waitForSync` to use tracker-based status
> - Define storage schemas for IndexedDB (`unsyncedCoValues`) and SQLite (`unsynced_covalues`) and associated operations
> - Outline unit/integration/E2E/performance tests and edge cases
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 718bd1e457c7cd6e0754d13473f64978baacd6c2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.specs/unsynced-covalues-tracking/design.md
+++ b/.specs/unsynced-covalues-tracking/design.md
@@ -1,0 +1,239 @@
+# Design: Track Unsynced CoValues & Resume Sync
+
+## Overview
+
+This design implements automatic tracking of CoValues with unsynced changes and provides mechanisms to resume syncing them on app restart. The solution integrates with the existing sync infrastructure and provides reactive APIs for monitoring sync status.
+
+The core idea is to maintain a set of CoValue IDs that have pending changes not yet fully synced to all peers. This set is persisted across app restarts and used to:
+1. Automatically resume syncing on startup
+2. Provide efficient sync status queries
+3. Enable reactive sync status subscriptions
+
+## Architecture / Components
+
+### 1. Extended StorageAPI Interface
+
+Extend the existing `StorageAPI` interface to include methods for tracking unsynced CoValues.
+
+**Location:** `packages/cojson/src/storage/types.ts`
+
+**New Methods:**
+```typescript
+export interface StorageAPI {
+  // ... existing methods ...
+  
+  trackCoValueSyncStatus(id: RawCoID, peerId: PeerID, synced: boolean): void;
+  getUnsyncedCoValueIDs(callback: (data: RawCoID[]) => void);
+}
+```
+
+**Implementation Strategy:**
+- Storage implementations will add these methods to persist the set of unsynced CoValue IDs
+- For IndexedDB: Add a new object store `"unsyncedCoValues"`
+- For SQLite: Add a table `unsynced_covalues`
+
+### 2. UnsyncedCoValuesTracker
+
+A new class that manages the set of unsynced CoValue IDs.
+
+**Location:** `packages/cojson/src/sync/UnsyncedCoValuesTracker.ts`
+
+**Responsibilities:**
+- Maintain an in-memory Set of unsynced CoValue IDs (including peers pending sync)
+- Persist the set to storage using `StorageAPI.trackCoValueSyncStatus` if available. Persistence can be batched and performed asynchronously, to avoid the cost of N extra storage writes per update (where N is the number of peers). 
+- Load persisted unsynced CoValues on initialization using `StorageAPI.getUnsyncedCoValueIDs` + `LocalNode.loadCoValueCore` if storage is available
+- Notify listeners when the set changes
+
+**Key Methods:**
+- `constructor(storage?: StorageAPI)`: Initialize with optional storage for persistence
+- `add(id: RawCoID, peerId: PeerID)`: Add a CoValue to the unsynced set (triggers persistence if storage available)
+- `remove(id: RawCoID, peerId: PeerID)`: Remove a CoValue from the unsynced set (triggers persistence if storage available)
+- `getAll()`: Returns all unsynced CoValue IDs 
+- `isAllSynced()`: Check if all CoValues are synced (O(1), returns `size() === 0`)
+- `subscribe(id: RawCoID, listener: (synced: boolean) => void)`: Subscribe to changes in whether a CoValue is synced
+- `subscribe(listener: (synced: boolean) => void)`: Subscribe to changes in whether all CoValues are synced
+
+### 3. Integration with SyncManager
+
+**Location:** `packages/cojson/src/sync.ts`
+
+**Changes:**
+- Add `unsyncedTracker: UnsyncedCoValuesTracker` property to `SyncManager`
+- Initialize tracker in `SyncManager` constructor: `new UnsyncedCoValuesTracker(local.storage, this)`
+- Update `syncContent()` method to keep track of unsynced CoValues:
+ ```ts
+ syncContent(content: NewContentMessage) {
+  const coValue = this.local.getCoValue(content.id);
+
+  this.storeContent(content);
+
+  const contentKnownState = knownStateFromContent(content);
+
+  for (const peer of this.getPeers(coValue.id)) {
+    this.unsyncedTracker.add(coValue.id, peer.id);
+
+    // ...
+
+    this.trySendToPeer(peer, content);
+    const unsubscribe = this.syncState.subscribeToPeerUpdates(
+      peer.id,
+      (knownState, syncState) => {
+        if (syncState.uploaded && knownState.id === coValue.id) {
+          this.unsyncedTracker.remove(coValue.id, peer.id);
+          unsubscribe();
+        }
+      },
+    );
+
+    peer.combineOptimisticWith(coValue.id, contentKnownState);
+    peer.trackToldKnownState(coValue.id);
+  }
+}
+ ```
+- Add method `async resumeUnsyncedCoValues()` to load and resume syncing unsynced CoValues. This happens asynchronously and doesn't block initialization
+- Call `resumeUnsyncedCoValues` as part of `SyncManager.startPeerReconciliation`.
+
+### 4. Sync Status Subscriptions
+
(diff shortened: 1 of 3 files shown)


.specs/unsynced-covalues-tracking/requirements.md
```diff
@@ -0,0 +1,66 @@
...
+**As a** Jazz user  
+**I want** Jazz to automatically track CoValues that have unsynced changes  
+**So that** these CoValues can be synced even if they're not currently loaded in memory
+
+**Acceptance Criteria:**
+- [ ] When a CoValue has local changes that haven't been fully uploaded to at least one peer, it is tracked as unsynced
+- [ ] When a CoValue becomes fully synced to all peers, it is removed from the unsynced tracking
+- [ ] The tracking persists across app restarts using platform-appropriate storage
+
+### US2: Resume Sync on App Start
```
gdorsi, line 23: Also on sign up when the dev configured `sync.when: "signUp"` - [docs](https://jazz.tools/docs/react/key-features/authentication/authentication-states#controlling-sync-for-different-authentication-states)

We probably need a different strategy for that, because there would be no server peers in that case.

We either track every coValue as unsynced when there are no server peers, or we try to load all the values on syncUp.

.specs/unsynced-covalues-tracking/design.md
```diff
@@ -0,0 +1,239 @@
...
+    this.trySendToPeer(peer, content);
+    const unsubscribe = this.syncState.subscribeToPeerUpdates(
+      peer.id,
+      (knownState, syncState) => {
+        if (syncState.uploaded && knownState.id === coValue.id) {
+          this.unsyncedTracker.remove(coValue.id, peer.id);
+          unsubscribe();
+        }
+      },
+    );
```
gdorsi, line 89: We maybe want to change this API to work better with many unsynced CoValues.

With this implementation, the `syncState` is going to trigger 1 callback x unsynced CoValue.
It would be probably more scalable to pass the `coValueID` as param and use a 2-level map to reduce the complexity here.

.specs/unsynced-covalues-tracking/design.md
```diff
@@ -0,0 +1,239 @@
...
+### 3. Integration with SyncManager
+
+**Location:** `packages/cojson/src/sync.ts`
+
+**Changes:**
+- Add `unsyncedTracker: UnsyncedCoValuesTracker` property to `SyncManager`
+- Initialize tracker in `SyncManager` constructor: `new UnsyncedCoValuesTracker(local.storage, this)`
+- Update `syncContent()` method to keep track of unsynced CoValues:
+ ```ts
+ syncContent(content: NewContentMessage) {
```
gdorsi, line 66: This won't be enough if we want to track unsynced coValues also on the sync server side.

We probably need to intercept also calls to handleNewContent for that, or just rely on the optimisticKnownState updates.

.specs/unsynced-covalues-tracking/design.md
```diff
@@ -0,0 +1,239 @@
...
+
+Storage persists unsynced CoValue-to-peer relationships as individual rows, with one row per (CoValue ID, Peer ID) pair.
+
+**IndexedDB Schema:**
+- Object store: `"unsyncedCoValues"`
+- Key: `[coValueId, peerId]` (composite key)
+- Value: `{ coValueId: RawCoID, peerId: PeerID }`
+- Indexes:
+  - Index on `coValueId` for efficient queries by CoValue
+  - Index on `peerId` for efficient queries by peer (optional, for cleanup)
```
gdorsi, line 238: How are we going to use this `PeerID` information?

What if we do change it, let's say the dev changes the API Key?

.specs/unsynced-covalues-tracking/requirements.md
```diff
@@ -0,0 +1,66 @@
...
+## User Stories
+
+### US1: Track Unsynced CoValues
+**As a** Jazz user  
+**I want** Jazz to automatically track CoValues that have unsynced changes  
+**So that** these CoValues can be synced even if they're not currently loaded in memory
+
+**Acceptance Criteria:**
+- [ ] When a CoValue has local changes that haven't been fully uploaded to at least one peer, it is tracked as unsynced
+- [ ] When a CoValue becomes fully synced to all peers, it is removed from the unsynced tracking
```
gdorsi:
```suggestion
- [ ] When a CoValue becomes fully synced to all server peers, it is removed from the unsynced tracking
```

I would also add that a peer must be flagged in some way, because even by filtering via "server peers" could catch us into tracking the sync state for temporary connections, [like this one](https://github.com/garden-co/jazz/blob/docs/spec-covalue-sync-tracking/packages/cojson/src/localNode.ts#L792).

We could either rely on [this persistent flag](https://github.com/garden-co/jazz/blob/docs/spec-covalue-sync-tracking/packages/cojson-transport-ws/src/createWebSocketPeer.ts#L193) which gives us more stability, or exclude all the temporary peers.

.specs/unsynced-covalues-tracking/design.md
nrainhart: What do you mean by "rely on the optimisticKnownState updates"?

gdorsi: Subscribe directly to the peer and capture all the known state updates.

.specs/unsynced-covalues-tracking/requirements.md
nrainhart: I think scoping this to "persistent server peers" makes sense

.specs/unsynced-covalues-tracking/requirements.md
nrainhart: How do you feel about forwarding `when` to the local node? To me the correct behaviour if there are no server peers would be:
- if `when="signup"` (and the user is not signed up), track all coValues as unsynced
- if `when="never"`, not track coValues as unsynced

gdorsi: Yes, makes sense.

.specs/unsynced-covalues-tracking/design.md
gdorsi: It would probably also be helpful to keep track of the input order, to start the sync resume on the oldest update before.

.specs/unsynced-covalues-tracking/design.md
```diff
@@ -0,0 +1,239 @@
...
+- Uses `syncManager.unsyncedTracker.subscribe()` to get notified when the unsynced set changes
+- Calls listener immediately with current state on subscription (check `unsyncedTracker.isAllSynced()`)
+
+### 5. Refactored waitForSync
+
+**Location:** `packages/cojson/src/coValueCore/coValueCore.ts` and `packages/cojson/src/sync.ts`
+
+**Changes:**
+- Replace the current implementation of `CoValueCore.waitForSync()` based on CoValue subscription with a simple call to `syncManager.unsyncedTracker.subscribe(coValueId)`.
+
```
gdorsi, line 211: We should also specify how the sync resume process will work on reload.

I think it would be nice to make it a some sort of background task, with values sync resumed in chunks.

.specs/unsynced-covalues-tracking/design.md
nrainhart: > How are we going to use this PeerID information?

My idea with storing the peer id is avoid needing to check the known state is synced with ALL peers each time the known state of any of the peers is updated.

> What if we do change it, let's say the dev changes the API Key?

Is it possible for persistent server peers to change at runtime? If reloading the app with new peers, we could retrieve the list of unique peers and delete any that we're no longer connected to. I guess we could do the same on `removePeer` if necessary.

Regarding new peers that are added: do we want to sync ALL coValues with them on startup? Or is it ok to only sync covalues as they're are loaded, as we do now?

> It would probably also be helpful to keep track of the input order, to start the sync resume on the oldest update before.

Agree. We can switch to using autoincremental ids instead.

.specs/unsynced-covalues-tracking/design.md
nrainhart: Updated the spec to also track unsynced coValues in handleNewContent. I'm thinking of only tracking changes coming from client peers. Do you think we need to include other types of peers as well?

.specs/unsynced-covalues-tracking/design.md
nrainhart: Sketched a proposal in the spec, we can keep iterating on it

nrainhart:
Closing in favor of this draft PR: https://github.com/garden-co/jazz/pull/3320

PR #3307, opened 2025-12-24
refactor: cleanup ahead of CoValue sync status tracking

nrainhart:
# Description

I did some minor codebase cleanup while working on the spec for CoValue sync status tracking: https://github.com/garden-co/jazz/pull/3306.

That includes removing unused methods, attributes and imports, make internal methods private and adding some TS docs.

All commits are self-contained.

---

> [!NOTE]
> General cleanup and encapsulation hardening ahead of sync status work.
> 
> - Mark internal fields/methods as `private` across `PeerState`, `CoValueCore`, `SubscriptionScope`, and related modules; convert some arrow properties to methods
> - Rename storage `knwonStates` to `knownStates` and update all usages in async/sync storage APIs
> - Remove unused code/imports (e.g., `hasLoadingSources`, `firstChunks`, stray `NewContentMessage` imports)
> - Add short TS doc comments to `tryAddTransactions` and `makeTransaction`; make `internalLoadFromPeer` private
> - Minor call-site/flow tweaks (drop unnecessary `void` on reconciliation call, small batching/queue cleanups)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f9a5cb9b2a1c3c33c27042a3d3032329a02dc4e6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/cojson/src/SyncStateManager.ts
```diff
@@ -24,20 +18,11 @@ export type PeerSyncStateListenerCallback = (
...
     Set<PeerSyncStateListenerCallback>
   >();
 
-  subscribeToUpdates(listener: GlobalSyncStateListenerCallback) {
-    this.listeners.add(listener);
-
-    return () => {
-      this.listeners.delete(listener);
-    };
-  }
```
gdorsi, line 39: Users are relying on this internal method, I'd say to keep it until we can offer a viable alternative to:
- resume unsynced values
- subscribe to a sync status in the UI

packages/cojson/src/storage/storageAsync.ts
```diff
@@ -40,10 +40,10 @@ export class StorageApiAsync implements StorageAPI {
     this.dbClient = dbClient;
   }
 
-  knwonStates = new StorageKnownState();
+  knownStates = new StorageKnownState();
```
gdorsi, line 43: Ahhh, good catch!

gdorsi (approved):
Thanks for this cleanup

packages/cojson/src/SyncStateManager.ts
nrainhart: Reverted [`f9a5cb9` (#3307)](https://github.com/garden-co/jazz/pull/3307/commits/f9a5cb9b2a1c3c33c27042a3d3032329a02dc4e6)

wrangelvid:
@nrainhart I noticed you made some attributes of the subscription scope private.
We have an internal method `useCoSignal` that iterates through the subscription snapshot and accesses the subscription scopes `resolve` attribute to decide if deeper nested refs should be included in the signal or not.

We are stuck on jazz 0.19.14 now, but maybe we can think of a public api on the jazz covalue object?

nrainhart:
Hey @wrangelvid, sorry to hear that. As a quick solution, I wonder if you can check `SubscriptionScope.childNodes` instead to decide if more refs need to be included. If you're fine with sharing the code of `useCoSignal` with me I may be able to suggest other options, or can think of a public API we can provide (we can follow-up via DM)

wrangelvid:
I see. So instead of checking if the raw key is in the resolve we can check he child nodes:

```ts
...
for (const key of coMap.$jazz.raw.keys() {
    const rawValue = coMap.$jazz.raw.get(key)
    if (
     typeof rawValue === "string" &&
     rawValue.startsWith("co_") &&
     coMap.$jazz._subscriptionScope !== undefined &&
     (coMap.$jazz._subscriptionScope.childNodes.has(rawValue) || (coMap.$jazz._subscriptionScope.childValues.has(rawValue)
     ) {
        // Ref is being resolved, access via property.
        const sourceValue = coMap[key]
        syncPropertyToSignal(target, key, sourceValue);
     } else {
          target[key] = undefined;
     }
     ...
 }
 ...
```

PR #3317, opened 2025-12-30
feat: remove `JazzContextManagerContext`, enforce error on nested `JazzProvider` components, and rename `useJazzContext` to `useJazzContextValue` for clarity

gdorsi:
This PR simplifies the internals of JazzProvider and reduces the providers we use to one.

This way it's going to be easiers for users to build their workarounds in case some bad library doesn't use the React builtin methods to build portal components.

It's also better for us because it makes the JazzProvider more stable and reduce the re-renders during auth changes because we don't subscribe to the contextValue anymore in there.

This is a minor because it changes how we expose the provider primitives, and should go with the 0.20

---

> [!NOTE]
> Simplifies React context architecture and tightens provider usage.
> 
> - Removes `JazzContextManagerContext`; `JazzContext` now holds the context manager only
> - API change: `useJazzContext` now returns the manager; new `useJazzContextValue` returns the context value; repo-wide migration to the new hook
> - Enforces error when nesting `JazzProvider`/`JazzReactProvider`/`JazzExpoProvider` and adds `fallback`/“ready” gating based on manager readiness
> - Refactors web/native providers to stop subscribing to context value directly; use `useSyncExternalStore` for readiness and cleanup via `done()`
> - Updates auth hooks/components (BetterAuth, Clerk, Passkey, Demo), inspector, RN/web hooks to use `useJazzContextValue`
> - Tests and exports adjusted to new APIs; adds changeset with migration instructions
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f6d800c4605200bf8ecf7ea3e19664b8f96ff516. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/jazz-tools/src/react-core/hooks.ts
```diff
@@ -33,40 +31,58 @@ import {
...
+  const context = useSyncExternalStore(
+    useCallback(
+      (callback) => {
+        return contextManager.subscribe(callback);
+      },
+      [contextManager],
+    ),
+    () => contextManager.getCurrentValue(),
+    () => contextManager.getCurrentValue(),
+  );
```
gdorsi, line 63: Re-rendering wise it is better to subscribe here instead of having JazzProvider do the subscription

packages/jazz-tools/src/react-native-core/provider.tsx
```diff
@@ -53,46 +60,41 @@ export function JazzProviderCore<
...
         return contextManager.subscribe(callback);
       },
-      [sync, guestMode].concat(storage as any),
+      [contextManager],
     ),
-    () => contextManager.getCurrentValue(),
-    () => contextManager.getCurrentValue(),
+    () => Boolean(contextManager.getCurrentValue()),
+    () => Boolean(contextManager.getCurrentValue()),
   );
```
gdorsi, line 104: I have simplified the provider code a bit by:
- switched to `useMemo` to make the flow more simple
- using the `sync.peer, sync.when` values as deps, which should bring more stability
- subscribe to `Boolean(contextManager.getCurrentValue())` so this provider now re-render only when the state changes into isReady

packages/jazz-tools/src/react-core/hooks.ts
```diff
@@ -33,40 +31,58 @@ import {
...
     );
   }
 
-  return value;
+  return context;
 }
 
 export function useAuthSecretStorage() {
-  const value = useContext(JazzContextManagerContext);
+  const value = useContext(JazzContext);
```
Takeno: nit
```suggestion
  const value = useJazzContext<Account>();
```

packages/jazz-tools/src/react-core/tests/useAccount.selector.test.ts
```diff
@@ -2,7 +2,7 @@
 
 import { Account, RefsToResolve, co, z, Group } from "jazz-tools";
 import { assert, beforeEach, describe, expect, it } from "vitest";
-import { useAccount, useJazzContextManager } from "../hooks.js";
+import { useAccount, useJazzContext } from "../hooks.js";
```
Takeno: nit
```suggestion
import { useAccount } from "../hooks.js";
```

packages/jazz-tools/src/react-core/hooks.ts
```diff
@@ -33,40 +31,58 @@ import {
...
+export function useJazzProviderCheck() {
+  const contextManager = useContext(JazzContext);
+
+  if (contextManager) {
+    throw new Error(
+      "You can't nest a JazzProvider inside another JazzProvider.",
+    );
+  }
+}
+
```
Takeno: Do we really need to expose this hook from react-core? Isn't enough something like this in context provider?
```ts
  if (useContext(JazzContext)) {
    throw new Error(
      "You can't nest a JazzProvider inside another JazzProvider.",
    );
  }
  ```

Takeno (approved):

nrainhart (approved):

gdorsi:
Merged manually in https://github.com/garden-co/jazz/pull/3318

PR #3318, opened 2025-12-30
Jazz 0.20

gdorsi:
With this release we complete the migration to a pure native Rust toolchain and remove the JavaScript crypto compatibility layer. The native Rust core now runs everywhere: React Native, Edge runtimes, all server-side environments, and the web.

The JavaScript crypto implementation is much slower than native Rust crypto. Although workarounds like RNQuickCrypto for React Native improved performance, they still only wrapped certain native libraries, rather than running Jazz's full Rust crypto.

With native Rust crypto now running everywhere, Jazz delivers good performance on every platform. This also helps us speed up the migration of Jazz Core to Rust which will improve Jazz overall performance.

Changes:
- **Removed `PureJSCrypto`** from `cojson` (including the `cojson/crypto/PureJSCrypto` export).
- **Removed `RNQuickCrypto`** from `jazz-tools`.
- **No more fallback to JavaScript crypto**: if crypto fails to initialize, Jazz now throws an error instead of falling back silently.
- **React Native + Expo**: **`RNCrypto` (via `cojson-core-rn`) is now the default**.
- Optimized the JS <--> Rust communication, by removing the compat logic we had with JS crypto
- Added permanent CoValue deletion with a new `deleted` loading state.
- Restricted `unique` parameters to strings or string records for deterministic serialization.
- `removeMember` now throws when the caller is unauthorized.
- React context changes: `useJazzContextValue` replaces value access, `useJazzContext` returns the manager, and nested `JazzProvider` now throws.


homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,200 @@
...
+
+The JavaScript crypto implementation is much slower than native Rust crypto. Although workarounds like RNQuickCrypto for React Native improved performance, they still only wrapped certain native libraries, rather than running Jazz's full Rust crypto.
+
+With native Rust crypto now running everywhere, Jazz comes with good performance in every platform. 
+This also helps us speed up the migration of Jazz Core to Rust which will help us improve the Jazz overall performance.
+
+## What changed
+
+- We dropped support for `PureJSCrypto`. Jazz crypto now runs natively in each environment.
+- We dropped support for `RNQuickCrypto`. RNCrypto is now the default crypto provider on React Native.
```
joeinnes:
```suggestion
- - As Jazz crypto now runs natively in all environments, we no longer support `PureJSCrypto`. 
- RNCrypto is now the default crypto provider on React Native. As a result, we're also discontinuing support for `RNQuickCrypto`. 
```

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,200 @@
...
+```ts
+// Before (integers cause TypeScript errors but still work at runtime)
+Task.create({ text: "X" }, { unique: 123 }); // TYPE ERROR (deprecated)
+
+// After - Recommended: convert to string or object with string values
+Task.create({ text: "X" }, { unique: "123" }); // SUPPORTED (RECOMMENDED)
+
+// Alternative: bypass type check (integers still work at runtime, but deprecated)
+// @ts-expect-error - We are bypassing the type check
+Task.create({ text: "X" }, { unique: 123 }); // SUPPORTED (but use string instead)
```
joeinnes:
```suggestion
```

Although this is technically possible, I don't think there's a good reason for us to illustrate this. I think we should just drop this third example, and let the second stand on its own.

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,200 @@
...
+export default {
+  async fetch(request: Request, env: Env, ctx: ExecutionContext) {
+    // Jazz application logic
+    return new Response("Hello from Jazz on Cloudflare!");
+  },
+};
+```
+</CodeGroup>
+
+if you don't do this, you will get an error like this:
```
joeinnes:
```suggestion
If you don't do this, you will get an error like this:
```

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,200 @@
...
+Before 0.20.0, you would get a warning like this:
+<CodeGroup>
+```
+Warning: WASM crypto is not available. Falling back to JavaScript crypto.
+```
+</CodeGroup>
+
+### Expo
+
+If you had previously installed QuickCrypto, then you should follow the steps below. Otherwise, no action is needed to take advantage of RNCrypto!
```
joeinnes:
```suggestion
If you had previously installed QuickCrypto, then you should follow the steps below. Otherwise, no action is needed to take advantage of RNCrypto.
```

antoniomuso (approved):
🚀

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,256 @@
...
+This also helps us speed up the migration of Jazz Core to Rust which will help us improve the Jazz overall performance.
+
+## What changed
+
+- There is no more fallback to Javascript crypto. If the crypto fails to initialize, it will throw an error. See [Edge runtimes](#edge-runtimes).
+**NB**: if you use WASM crypto on Edge runtimes, you need to ensure that WASM is available and initialized. See more details at [WASM on Edge runtimes](/docs/react/server-side/setup#wasm-on-edge-runtimes).
+- We dropped support for `PureJSCrypto`. Jazz crypto now runs natively in each environment. See [Expo](#expo) and [React Native](#react-native-non-expo).
+- We dropped support for `RNQuickCrypto`. RNCrypto is now the default crypto provider on React Native. See [Expo](#expo) and [React Native](#react-native-non-expo).
+- Optimized the JS-to-Rust communication by implementing native data type exchange, eliminating serialization overhead.
+- You can now only use strings or records of strings as [`unique` parameters](/docs/react/core-concepts/covalues/comaps#uniqueness). This ensures that serialisation is deterministic. See [A more strict `unique` param](#a-more-strict-unique-param).
```
antoniomuso:
```suggestion
- You can now only use strings or records of strings as [`unique` parameters](/docs/react/core-concepts/covalues/comaps#uniqueness). This ensures that serialisation is deterministic. See [a more strict `unique` param](#a-more-strict-unique-param).
```

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -0,0 +1,256 @@
...
+- The `removeMember` method now throws an error if the caller is not authorized.
+- We have introduced a way to permanently [delete CoValues](/docs/core-concepts/covalues/deleting). This brings a new `deleted` loading state. See [New `deleted` loading state](#new-deleted-loading-state).
```
antoniomuso:
```suggestion
- We have introduced a way to permanently [delete CoValues](/docs/core-concepts/deleting). This brings a new `deleted` loading state. See [new `deleted` loading state](#new-deleted-loading-state).
```

joeinnes (approved):

CHANGELOG.md
```diff
@@ -1,4 +1,25 @@
 
+Released Jazz 0.20.0:
+
+With this release we complete the migration to a pure native Rust toolchain and remove the JavaScript crypto compatibility layer. The native Rust core now runs everywhere: React Native, Edge runtimes, all server-side environments, and the web.
```
joeinnes:
```suggestion
With this release, we introduce a new, simple to use API for [permanently deleting CoValues](https://jazz.tools/docs/react/core-concepts/deleting). 

For auditing and data recovery purposes, we still recommend using soft-deletes wherever possible. However, we appreciate that for various reasons (data privacy, storage space), it may be preferable to delete data permanently. From Jazz 0.20.0 onwards, you'll be able to do this easily, and build experiences where your users can manage their own data.

Additionally, with this release we complete the migration to a pure native Rust toolchain and remove the JavaScript crypto compatibility layer. The native Rust core now runs everywhere: React Native, Edge runtimes, all server-side environments, and the web.
```

homepage/homepage/content/docs/upgrade/0-20-0.mdx
gdorsi: Removed the entire snippet.

It's not really a fix to change the unique value.

PR #3320, opened 2025-12-30
feat: resume interrupted CoValues sync on restart

nrainhart:
# Description

Currently, if CoValue sync is interrupted and the app is restarted, the sync process won't resume until the unsynced CoValue is loaded again.

This PR implements sync state tracking for CoValues in persistent storage, so that the sync process can be automatically resumed when the app restarts. Sync resumption is performed in batches, to avoid slowing down the performance of the app on startup.

## Manual testing

Ran some performance testing in the browser to analyze the additional cost of tracking CoValue sync state. After applying some performance optimizations (persist pending sync updates in batches, minimize the number of times the sync state tracking callbacks are executed), managed to reduce it to what I consider a reasonable cost:

Screenshot_2025-12-30_at_2 56 34_PM

This measures the time it takes to create 10k tasks in the stress test app (which means creating 30k CoValues, including the tasks themselves, their group owners & their text content).

## Tests

## Checklist

- [ ] I've updated the part of the docs that are affected the PR changes => don't think it's necessary. If you do, let me know

---

> [!NOTE]
> Introduces persisted tracking of unsynced `CoValue`s and automatic sync resumption after restart, focused on persistent server peers and batched writes for performance.
> 
> - New `UnsyncedCoValuesTracker` with batched storage persistence, per-`CoValue`/global subscriptions, and flush on `gracefulShutdown`
> - Extend `StorageAPI` with `trackCoValuesSyncState`, `getUnsyncedCoValueIDs`, `stopTrackingSyncState`
> - IndexedDB: new `unsyncedCoValues` store, schema v5, CRUD in `idbClient`; SQLite: new `unsynced_covalues` table + migrations and client methods
> - `SyncManager` integration: `trackSyncState`, `resumeUnsyncedCoValues()` (batched), only track persistent server peers, resume during `startPeerReconciliation`
> - `SyncStateManager.subscribeToPeerUpdates` now keyed by `(peerId, coValueId)`; related call sites updated
> - `LocalNode`/contexts accept `syncWhen` and propagate; logout/shutdown now awaits tracker flush
> - Tests added for tracking, persistence, resumption; IDB transaction supports custom store sets; minor test/config updates
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 633321e69f7c0dbab3a58c3f2ece766019e97daa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/twelve-bats-repeat.md
+++ b/.changeset/twelve-bats-repeat.md
@@ -1,0 +1,28 @@
+---
+"cojson-core-napi-linux-arm-gnueabihf": patch
+"cojson-core-napi-linux-arm64-musl": patch
+"cojson-core-napi-linux-arm64-gnu": patch
+"cojson-core-napi-linux-x64-musl": patch
+"cojson-core-napi-linux-x64-gnu": patch
+"cojson-core-napi-darwin-arm64": patch
+"cojson-core-napi-darwin-x64": patch
+"cojson-storage-do-sqlite": patch
+"cojson-storage-indexeddb": patch
+"cojson-storage-sqlite": patch
+"cojson-transport-ws": patch
+"svelte-passkey-auth": patch
+"community-jazz-vue": patch
+"jazz-react-passkey-auth-starter": patch
+"cojson-core-napi": patch
+"cojson-core-wasm": patch
+"durable-object": patch
+"jazz-sveltekit": patch
+"cojson-core-rn": patch
+"jazz-webhook": patch
+"chat-svelte": patch
+"jazz-tools": patch
+"jazz-run": patch
+"cojson": patch
+---
+
+Resume interrupted CoValue sync on app restart (without requiring CoValues to be manually reloaded)
--- a/.changeset/twelve-bats-repeat.md
+++ b/.changeset/twelve-bats-repeat.md
@@ -1,27 +1,5 @@
 ---
-"cojson-core-napi-linux-arm-gnueabihf": patch
-"cojson-core-napi-linux-arm64-musl": patch
-"cojson-core-napi-linux-arm64-gnu": patch
-"cojson-core-napi-linux-x64-musl": patch
-"cojson-core-napi-linux-x64-gnu": patch
-"cojson-core-napi-darwin-arm64": patch
-"cojson-core-napi-darwin-x64": patch
-"cojson-storage-do-sqlite": patch
-"cojson-storage-indexeddb": patch
-"cojson-storage-sqlite": patch
-"cojson-transport-ws": patch
-"svelte-passkey-auth": patch
-"community-jazz-vue": patch
-"jazz-react-passkey-auth-starter": patch
-"cojson-core-napi": patch
-"cojson-core-wasm": patch
-"durable-object": patch
-"jazz-sveltekit": patch
-"cojson-core-rn": patch
-"jazz-webhook": patch
-"chat-svelte": patch
 "jazz-tools": patch
-"jazz-run": patch
 "cojson": patch
 ---
 
--- a/packages/cojson-storage-indexeddb/src/CoJsonIDBTransaction.ts
+++ b/packages/cojson-storage-indexeddb/src/CoJsonIDBTransaction.ts
@@ -2,7 +2,8 @@ export type StoreName =
   | "coValues"
   | "sessions"
   | "transactions"
-  | "signatureAfter";
+  | "signatureAfter"
+  | "unsyncedCoValues";
 
 // A access unit for the IndexedDB Jazz database
 // It's a wrapper around the IDBTransaction object that helps on batching multiple operations
@@ -28,7 +29,13 @@ export class CoJsonIDBTransaction {
 
   refresh() {
     this.tx = this.db.transaction(
-      ["coValues", "sessions", "transactions", "signatureAfter"],
+      [
+        "coValues",
+        "sessions",
+        "transactions",
+        "signatureAfter",
+        "unsyncedCoValues",
+      ],
       "readwrite",
     );
 
--- a/packages/cojson-storage-indexeddb/src/CoJsonIDBTransaction.ts
+++ b/packages/cojson-storage-indexeddb/src/CoJsonIDBTransaction.ts
@@ -5,13 +5,19 @@ export type StoreName =
   | "signatureAfter"
   | "unsyncedCoValues";
 
+const DEFAULT_TX_STORES: StoreName[] = [
+  "coValues",
+  "sessions",
+  "transactions",
+  "signatureAfter",
+];
+
 /**
  * An access unit for the IndexedDB Jazz database.
  * It's a wrapper around the IDBTransaction object that helps on batching multiple operations
  * in a single transaction.
  */
 export class CoJsonIDBTransaction {
-  db: IDBDatabase;
   declare tx: IDBTransaction;
 
   pendingRequests: ((txEntry: this) => void)[] = [];
@@ -23,23 +29,16 @@ export class CoJsonIDBTransaction {
   failed = false;
   done = false;
 
-  constructor(db: IDBDatabase) {
-    this.db = db;
-
+  constructor(
+    public db: IDBDatabase,
+    // The object stores this transaction will operate on
+    private storeNames: StoreName[] = DEFAULT_TX_STORES,
+  ) {
     this.refresh();
   }
 
   refresh() {
-    this.tx = this.db.transaction(
-      [
-        "coValues",
-        "sessions",
-        "transactions",
-        "signatureAfter",
-        "unsyncedCoValues",
-      ],
-      "readwrite",
-    );
+    this.tx = this.db.transaction(this.storeNames, "readwrite");
 
     this.tx.oncomplete = () => {
       this.done = true;
--- a/packages/cojson-storage-indexeddb/src/CoJsonIDBTransaction.ts
+++ b/packages/cojson-storage-indexeddb/src/CoJsonIDBTransaction.ts
@@ -5,9 +5,11 @@ export type StoreName =
   | "signatureAfter"
   | "unsyncedCoValues";
 
-// A access unit for the IndexedDB Jazz database
-// It's a wrapper around the IDBTransaction object that helps on batching multiple operations
-// in a single transaction.
+/**
+ * An access unit for the IndexedDB Jazz database.
+ * It's a wrapper around the IDBTransaction object that helps on batching multiple operations
+ * in a single transaction.
+ */
 export class CoJsonIDBTransaction {
   db: IDBDatabase;
   declare tx: IDBTransaction;
@@ -49,12 +51,6 @@ export class CoJsonIDBTransaction {
 
   rollback() {
     this.tx.abort();
-  }
-
-  startedAt = performance.now();
-  isReusable() {
-    const delta = performance.now() - this.startedAt;
-    return !this.done && !this.failed && delta <= 100;
   }
 
   getObjectStore(name: StoreName) {
--- a/packages/cojson-storage-indexeddb/src/idbClient.ts
+++ b/packages/cojson-storage-indexeddb/src/idbClient.ts
@@ -22,7 +22,7 @@ import {
 export class IDBTransaction implements DBTransactionInterfaceAsync {
   constructor(private tx: CoJsonIDBTransaction) {}
 
-  run<T>(
+  private async run<T>(
     handler: (txEntry: CoJsonIDBTransaction) => IDBRequest<T>,
   ): Promise<T> {
     return this.tx.handleRequest<T>(handler);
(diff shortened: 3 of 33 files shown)


packages/cojson/src/localNode.ts
```diff
@@ -178,12 +179,14 @@ export class LocalNode {
     crypto: CryptoProvider;
     initialAgentSecret?: AgentSecret;
     peers?: Peer[];
+    syncWhen?: SyncWhen;
```
nrainhart, line 184: I'd prefer both params to be part of a single [SyncConfig](https://github.com/garden-co/jazz/blob/feat/resume-sync-on-restart/packages/jazz-tools/src/tools/types.ts#L56) object, but didn't want to introduce a breaking change in the local node API

packages/cojson/src/storage/sqlite/client.ts
```diff
@@ -193,4 +194,40 @@ export class SQLiteClient
...
+  }
+
+  trackCoValuesSyncState(
+    operations: Array<{ id: RawCoID; peerId: PeerID; synced: boolean }>,
+  ): void {
+    if (operations.length === 0) {
+      return;
+    }
+
+    this.db.transaction(() => {
```
gdorsi: Do we need to run these inside a transaction?

packages/cojson/src/sync.ts
```diff
@@ -259,7 +275,76 @@ export class SyncManager {
...
+        }
+
+        const BATCH_SIZE = 10;
+        let processed = 0;
+
+        const processBatch = async () => {
+          const batch = unsyncedCoValueIDs.slice(
+            processed,
+            processed + BATCH_SIZE,
+          );
```
gdorsi: We could filter these IDs by already available values, to focus on what's not loaded yet.

packages/cojson/src/storage/sqlite/client.ts
nrainhart: Good point, we don't. I did this as an optimization for IndexedDB and mindlessly carried it over to sqlite

packages/cojson/src/UnsyncedCoValuesTracker.ts
```diff
@@ -0,0 +1,248 @@
...
+import type { StorageAPI } from "./storage/types.js";
+
+/**
+ * Used to track a CoValue that hasn't been synced to any peer,
+ * because none is currently connected.
+ */
+const ANY_PEER_ID: PeerID = "any";
+
+// Flush pending updates to storage after 1 second
+let BATCH_DELAY_MS = 1000;
```
gdorsi: 1s looks a bit too much, how the performance would look if we lower this to 200ms?

packages/cojson/src/sync.ts
```diff
@@ -121,6 +131,7 @@ export class SyncManager {
   constructor(local: LocalNode) {
     this.local = local;
     this.syncState = new SyncStateManager(this);
+    this.unsyncedTracker = new UnsyncedCoValuesTracker(() => local.storage);
```
gdorsi: How about instead of providing storage here we do set it directly when calling `setStorage`?

I think we also should make the persistence of the `unsyncedTracker` optional, so we can enable it only on client nodes.

packages/cojson/src/sync.ts
```diff
@@ -259,7 +275,74 @@ export class SyncManager {
...
+        let processed = 0;
+
+        const processBatch = async () => {
+          const batch = coValuesToLoad.slice(processed, processed + BATCH_SIZE);
+
+          await Promise.all(
+            batch.map(async (coValueId) => {
+              try {
+                // Load the CoValue from storage (this will trigger sync if peers are connected)
+                const coValue = await this.local.loadCoValueCore(coValueId);
```
gdorsi: It would be nice to load this specifically from storage.

This way we don't fallback to server peers if the value is not found on storage.

packages/cojson/src/sync.ts
```diff
@@ -259,7 +275,74 @@ export class SyncManager {
...
+        processBatch().catch(reject);
+      });
+    });
+  }
+
   startPeerReconciliation(peer: PeerState) {
+    // Resume syncing unsynced CoValues asynchronously
+    this.resumeUnsyncedCoValues().catch((error) => {
+      logger.warn("Failed to resume unsynced CoValues:", error);
+    });
```
gdorsi: Shall we call this only for persistent peers?

packages/cojson/src/sync.ts
```diff
@@ -811,6 +900,44 @@ export class SyncManager {
...
+
+      const unsubscribe = this.syncState.subscribeToCoValueUpdates(
+        coValueId,
+        (_peerId, _knownState, syncState) => {
+          if (syncState.uploaded) {
+            this.unsyncedTracker.remove(coValueId);
+            unsubscribe();
+          }
+        },
+      );
```
gdorsi: Why do we need this subscription here?

When a Peer is added and we go through `startPeerReconciliation` aren't we going in `trackSyncState` again?

packages/cojson/src/sync.ts
```diff
@@ -259,7 +275,77 @@ export class SyncManager {
...
+
+          await Promise.all(
+            batch.map(async (coValueId) => {
+              try {
+                // Clear previous tracking (as it may include outdated peers)
+                this.local.storage?.stopTrackingSyncState(coValueId);
+
+                // Resume tracking sync state for this CoValue
+                // This will add it back to the tracker and set up subscriptions
+                this.trackSyncState(coValueId);
```
gdorsi: Inside here we are calling `this.unsyncedTracker.add(coValueId, peer.id)` which persists the unsynced state again, after we removed it above with `stopTrackingSyncState`.

If I'm understanding correctly, this is done to make the whole process "reset" the peer ids stored with the ones coming from the new session.

I think this flow should be documented in the code, to make it more explicit

packages/cojson-storage-indexeddb/src/CoJsonIDBTransaction.ts
```diff
@@ -28,7 +31,13 @@ export class CoJsonIDBTransaction {
 
   refresh() {
     this.tx = this.db.transaction(
-      ["coValues", "sessions", "transactions", "signatureAfter"],
+      [
+        "coValues",
+        "sessions",
+        "transactions",
+        "signatureAfter",
+        "unsyncedCoValues",
```
gdorsi: If we scope the readwrite transactions to `unsyncedCoValues` the updates we do there won't be blocking for the rest of the stores.

For that we would need a dedicated tx abstraction, but I think it's worth it.

packages/cojson-storage-indexeddb/src/idbNode.ts
```diff
@@ -47,6 +47,21 @@ export async function getIndexedDBStorage(name = DATABASE_NAME) {
           keyPath: ["ses", "idx"],
         });
       }
+      if (ev.oldVersion <= 4) {
+        const unsyncedCoValues = db.createObjectStore("unsyncedCoValues", {
+          autoIncrement: true,
+          keyPath: "rowID",
+        });
+        unsyncedCoValues.createIndex("byCoValueId", "coValueId");
+        unsyncedCoValues.createIndex("byPeerId", "peerId");
```
gdorsi: Is this index in use?

packages/cojson/src/sync.ts
```diff
@@ -811,6 +899,44 @@ export class SyncManager {
...
+      }
+
+      const unsubscribe = this.syncState.subscribeToPeerUpdates(
+        peer.id,
+        coValueId,
+        (_knownState, syncState) => {
+          if (syncState.uploaded) {
+            this.unsyncedTracker.remove(coValueId, peer.id);
+            unsubscribe();
+          }
```
gdorsi, line 942: Is this callback called only on the first peer state change?

If yes, do we need to add a sync state check before subscribing?

packages/cojson/src/sync.ts
```diff
@@ -787,6 +873,8 @@ export class SyncManager {
 
     this.storeContent(content);
 
+    this.trackSyncState(coValue.id);
```
gdorsi, line 891: We should also call `trackSyncState` when accepting an invite: https://github.com/garden-co/jazz/blob/ea38b24e6d3af9ba3177d7021bd65036f261120a/packages/cojson/src/localNode.ts#L674-L680

It doesn't go through `syncContent` so it requires a special handling.

Would make sense to do it in general when importing content.

packages/cojson-storage-indexeddb/src/idbNode.ts
nrainhart: It isn't. Removed!

packages/cojson/src/sync.ts
nrainhart: Exactly. Already added comments, if you think it's still unclear let me know

packages/cojson/src/sync.ts
nrainhart: Good point. Removed it

packages/cojson/src/sync.ts
nrainhart: Yeah! Modified it to only be triggered when adding persistent server peers

packages/cojson/src/sync.ts
nrainhart: Added the setter. Don't think we need to make persistence optional, because we do not create covalues in server nodes

packages/cojson/src/sync.ts
nrainhart: You mean in case the covalue was already synced to all peers before we reached this point?

packages/cojson-storage-indexeddb/src/CoJsonIDBTransaction.ts
nrainhart: I refactored `CoJsonIDBTransaction` to support defining which stores will be locked when creating the transaction: [`fd53d65` (#3320)](https://github.com/garden-co/jazz/pull/3320/commits/fd53d65b371d16da425141cb93961d780c2f6e4c)

packages/cojson/src/UnsyncedCoValuesTracker.ts
gdorsi: Just got into this issue: https://github.com/garden-co/jazz/pull/3328

I think it would be better to avoid batching, we could consider using a different storage persistence for this data.

Thinking about either localStorage or OPFS in the browser, and keeping checking SQLite on mobile (should be ok)

packages/cojson/src/sync.ts
nrainhart: Added a synchronous sync state check before tracking sync state (and checked it doesn't introduce any noticeable performance impact): https://github.com/garden-co/jazz/pull/3320/commits/7501d69971b90afb523ec9b40c47766df5e6ec10

packages/cojson/src/sync.ts
nrainhart: Now we track sync state for all imported content: [`6d51a03` (#3320)](https://github.com/garden-co/jazz/pull/3320/commits/6d51a031f7908327b74664293de9ab5f1c06d17f)

packages/cojson/src/UnsyncedCoValuesTracker.ts
nrainhart: After evaluating both localStorage and OPFS we decided to stay with indexedDB both for performance and code complexity reasons. Decreased the persistence delay to 200ms and confirmed in does not affect performance noticeably: [`4aa45bb` (#3320)](https://github.com/garden-co/jazz/pull/3320/commits/4aa45bb26edec5c52f1abdb82dc403e4df038b0e)

.changeset/twelve-bats-repeat.md
```diff
@@ -0,0 +1,28 @@
...
+"cojson-core-napi": patch
+"cojson-core-wasm": patch
+"durable-object": patch
+"jazz-sveltekit": patch
+"cojson-core-rn": patch
+"jazz-webhook": patch
+"chat-svelte": patch
+"jazz-tools": patch
+"jazz-run": patch
+"cojson": patch
```
gdorsi:
```suggestion
"jazz-tools": patch
"cojson": patch
```

packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -171,4 +222,73 @@ export class IDBClient implements DBClientInterfaceAsync {
...
+                  : {
+                      coValueId: update.id,
+                      peerId: update.peerId,
+                    },
+              );
+            }
+          }),
+        );
+      },
+      ["unsyncedCoValues"],
```
gdorsi, line 265: I like this API!

gdorsi (approved):
Let's go!!!

PR #3322, opened 2025-12-31
feat: narrow accepted types for the `uniqueness` parameter

antoniomuso:
# Description

This PR restricts the types accepted by the `uniqueness` parameter, moving from a generic `JsonValue` to a stricter union type.

**New Type Definition:**
```ts
type Uniqueness = 
  | string
  | boolean
  | null
  | undefined
  | { [key: string]: string };
 ```

**Motivation**

- **Performance & Simplification**: This change allows us to remove the dependency on stableStringify. Serializing static data is significantly easier and more efficient especially, regarding our Rust integration, compared to handling dynamic JsonValue structures.
- **Use Case Optimization**: Deeply nested objects were found to be unnecessary for this specific purpose.
- **Precision**: We have explicitly excluded the number type to prevent potential issues with floating-point precision.

**Backward Compatibility**

To ensure this is not a breaking change, we performed an analysis of the production database. The investigation confirmed that all existing uniqueness values comply with this new strict typing, ensuring no regressions for current users.
  
## Tests

## Checklist

---

> [!NOTE]
> - **Type narrowing**: Replace `JsonValue` with strict `Uniqueness` (`string | boolean | null | undefined | { [key: string]: string }`) in `verifiedState.ts`.
> - **Runtime validation**: Add `validateUniqueness` and enforce it in `CoValueCore.provideHeader`; invalid values are logged and headers rejected (arrays, non-integer numbers, objects with non-string values).
> - **Tests**: New tests cover rejection of invalid `uniqueness` values.
> - **Docs/changeset**: Upgrade guide adds a "unique validation" note; changeset marks a minor release and deprecates numeric uniqueness.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6c5ebbb35e87f9e5c472688a8ff096a85cb635c4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/strict-uniqueness-type.md
+++ b/.changeset/strict-uniqueness-type.md
@@ -1,0 +1,11 @@
+---
+"cojson": minor
+---
+
+Restricted the `Uniqueness` type to only accept specific values instead of any `JsonValue`. 
+The allowed types are now: `string`, `number` (as integers), `boolean`, `null`, `undefined`, 
+or an object with string values. Arrays, non-integer numbers, and objects with non-string values are no longer 
+accepted and will throw an error.
+
+Deprecated `number`, even if it is an integer, as it is not a valid uniqueness type in TS, if you want to continue using numbers, 
+use a string instead or ignore the type check.
--- a/packages/cojson/src/coValueCore/coValueCore.ts
+++ b/packages/cojson/src/coValueCore/coValueCore.ts
@@ -27,7 +27,12 @@ import {
   getDependenciesFromGroupRawTransactions,
   getDependenciesFromHeader,
 } from "./utils.js";
-import { CoValueHeader, Transaction, VerifiedState } from "./verifiedState.js";
+import {
+  CoValueHeader,
+  Transaction,
+  Uniqueness,
+  VerifiedState,
+} from "./verifiedState.js";
 import { SessionMap } from "./SessionMap.js";
 import {
   MergeCommit,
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: packages/cojson/src/coValueCore/coValueCore.ts [hunk 1]
```diff
@@ -51,10 +56,35 @@ import {
...
+      }
+    }
+  }
+}
+
 export function idforHeader(
   header: CoValueHeader,
   crypto: CryptoProvider,
 ): RawCoID {
+  assertValidUniqueness(header.uniqueness);
```
gdorsi: We can't throw like this, it could cause issues in the sync flow.

We alreay have a validation in place for headers, I'd suggest you to do this there.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -51,10 +56,35 @@ import {
 } from "../knownState.js";
 import { safeParseJSON } from "../jsonStringify.js";
 
+function assertValidUniqueness(uniqueness: Uniqueness) {
+  if (typeof uniqueness === "number") {
+    throw new Error("Uniqueness cannot be a number, got " + uniqueness);
+  }
```
gdorsi, line 72: We have people using `number` as uniqueness, and this check would break their apps without an easy migration path.

I'd suggest to keep supporting number at first level, and consider only floating point values as invalid.

On type level we can keep it as "invalid" but suggest people to turn-off the error.

packages/cojson/src/coValueCore/verifiedState.ts
```diff
@@ -35,10 +35,20 @@ export type CoValueHeader = {
...
 
 export type CoValueUniqueness = {
-  uniqueness: JsonValue;
+  uniqueness: Uniqueness;
   createdAt?: `2${string}` | null;
 };
 
+export type Uniqueness =
+  | string
+  | `${number}`
```
gdorsi: This is redundant, as it is a subset of string.

gdorsi (changes requested):

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: The validation is in provideHeader: https://github.com/garden-co/jazz/blob/main/packages/cojson/src/coValueCore/coValueCore.ts#L471

.changeset/strict-uniqueness-type.md
```diff
@@ -0,0 +1,9 @@
+---
+"cojson": patch
```
gdorsi:
```suggestion
"cojson": minor
```

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -474,6 +504,16 @@ export class CoValueCore {
...
   ) {
     if (!skipVerify) {
+      try {
+        assertValidUniqueness(header.uniqueness);
+      } catch (e) {
+        logger.error("Invalid uniqueness", {
+          header,
+          error: e instanceof Error ? e.message : String(e),
+        });
+        return false;
```
gdorsi, line 527: How about renaming `assertValidUniqueness` into `validateUniqueness` and making it return a validation value rather than throw an error?

It should be easier to provide an informative error.

gdorsi (approved):

packages/cojson/src/coValueCore/coValueCore.ts
antoniomuso: Done!

PR #3325, opened 2026-01-02
fix: resolve race condition in Clerk auth signup flow to prevent duplicate login attempts

gdorsi:
While debugging an issue I've found that when signing up with Clerk the user also goes through the logIn flow, creating a new context and interrupting the previous sync ops.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 Released Jazz 0.19.16:
 - Improved sync timeout error messages to include known state, peer state, and any error information when waiting for sync times out
+- Bugfix: fixed a race condition in Clerk auth where the signup flow could trigger a duplicate login attempt
 
 Released Jazz 0.19.15:
 - Added a locking system for session IDs in React Native to make mounting multiple JazzProviders safer (still not advised as duplicate the data loading effort)
--- a/packages/jazz-tools/src/tools/auth/clerk/index.ts
+++ b/packages/jazz-tools/src/tools/auth/clerk/index.ts
@@ -53,18 +53,21 @@ export class JazzClerkAuth {
   }
 
   private isFirstCall = true;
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 Released Jazz 0.19.16:
 - Improved sync timeout error messages to include known state, peer state, and any error information when waiting for sync times out
+- Bugfix: fixed a race condition in Clerk auth where the signup flow could trigger a duplicate login attempt
```
nrainhart, line 3: Is this needed? I thought our release process took care of updating this file

packages/jazz-tools/src/tools/auth/clerk/index.ts
```diff
@@ -53,18 +53,21 @@ export class JazzClerkAuth {
   }
 
   private isFirstCall = true;
+  private previousUser: Pick<
+    NonNullable<MinimalClerkClient["user"]>,
+    "unsafeMetadata"
+  > | null = null;
```
nrainhart, line 59: Nit: I'd extract the repeated type into a named type

nrainhart (approved):

CHANGELOG.md
gdorsi: No, it only updates the individual changelogs in each package.

The result is unreadable, and can't be used when publishing on discord, so I've started to manually keep track of the changes to release here.

packages/jazz-tools/src/tools/auth/clerk/index.ts
gdorsi: Trying to clean up the module, there is a bit of unnecessary code that's here for tech debt.

Going to do it in a follow-up PR, so I can release this fix ASAP.

PR #3326, opened 2026-01-02
refactor(JazzClerkAuth): remove unnecessary logic and add Zod validation

gdorsi:


packages/jazz-tools/src/tools/auth/clerk/types.ts
```diff
@@ -1,26 +1,38 @@
 import { AgentSecret } from "cojson";
 import { Account, ID } from "jazz-tools";
+import { z } from "zod/v4";
+
+const ClerkUserSchema = z.object({
+  fullName: z.string().nullable().optional(),
```
nrainhart:
```suggestion
  fullName: z.string().nullish(),
```

packages/jazz-tools/src/tools/auth/clerk/types.ts
```diff
@@ -1,26 +1,38 @@
...
+  firstName: z.string().nullable().optional(),
+  lastName: z.string().nullable().optional(),
+  id: z.string().optional(),
+  primaryEmailAddress: z
+    .object({
+      emailAddress: z.string().nullable(),
+    })
+    .nullable()
+    .optional(),
+  unsafeMetadata: z.record(z.string(), z.any()),
```
nrainhart: Could we use z.json instead of z.any?

packages/jazz-tools/src/tools/auth/clerk/index.ts
```diff
@@ -53,32 +55,30 @@ export class JazzClerkAuth {
...
   private isFirstCall = true;
-  private previousUser: Pick<
-    NonNullable<MinimalClerkClient["user"]>,
-    "unsafeMetadata"
-  > | null = null;
+  private previousUser: Pick<ClerkUser, "unsafeMetadata"> | null = null;
 
   registerListener(clerkClient: MinimalClerkClient) {
-    this.previousUser = clerkClient.user ?? null;
+    this.previousUser = ClerkEventSchema.parse(clerkClient).user ?? null;
```
nrainhart, line 61: ⚠️ If parsing fails we just let the zod error bubble up? Should we either handle the error and throw a more meaningful jazz error? Or just default to null in that case?

packages/jazz-tools/src/tools/auth/clerk/types.ts
gdorsi: z.json doesn't accept undefined values, but it was actually easy to provide a schema of what `unsafeMetadata` should contain, making the validation even better.

packages/jazz-tools/src/tools/auth/clerk/index.ts
gdorsi: Changed with safeParse

nrainhart (approved):
Thanks for addressing my comments!

PR #3327, opened 2026-01-02
chore: upgrade to vitest 4

gdorsi:
Upgrades vitest to the latest major.

The notable breaking changes for us are:
- vi.spyOn now returns the mock if the method is already mocked
- had to provide a config for each package because otherwise it was running the tests inside the dist
- removed the bench because we don't want to continue to use vitest for benchmarks


tests/cloudflare-workers/tests/integration.test.ts
```diff
@@ -1,6 +1,6 @@
 import { dirname, join } from "path";
 import { fileURLToPath } from "url";
-import { execa } from "execa";
+import { execa, $ } from "execa";
```
Takeno:
```suggestion
import { execa } from "execa";
```

It has been imported by mistake

Takeno (approved):

PR #3330, opened 2026-01-05
feat: add JazzMessageChannel for cross-context communication

gdorsi:
The JazzMessageChannel feature enables synchronization between different JavaScript execution contexts using the Web [MessageChannel API](https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel). This allows cojson nodes to communicate across:

- Main thread ↔ Web Worker
- Main thread ↔ Shared Worker
- Main thread ↔ Service Worker
- Window ↔ iframe
- Window ↔ Window (via `postMessage`)
- Electron main process ↔ renderer (via `MessageChannelMain`)

Unlike WebSocket peers which communicate over the network, MessageChannel peers communicate within the same browser/runtime using structured cloning, which is faster and doesn't require serialization to strings.

The feature exposes a `JazzMessageChannel` class (inspired by [Comlink](https://github.com/GoogleChromeLabs/comlink)) that manages:
- Creating and exposing MessageChannel ports
- Accepting ports from the other side
- Performing a handshake to establish the connection
- Creating the cojson `Peer` once connected

Usage:
```ts
// Main thread
import { experimental_JazzMessageChannel as JazzMessageChannel } from "jazz-tools";

const worker = new Worker("./worker.ts");

// Uses active account context by default
await JazzMessageChannel.expose(worker);

// Or explicitly provide an account
await JazzMessageChannel.expose(worker, { loadAs: me });
```

```ts
// Worker thread
import { startWorker, JazzMessageChannel } from "jazz-tools";

// Accept the peer from main thread
const peer = await JazzMessageChannel.waitForConnection();

// Start the worker with the peer
const { worker } = await startWorker({
  peer,
  AccountSchema: MyAccountSchema,
});
```

NOTE: no changeset here because the API is to be shipped for user testing before making it "public"


packages/jazz-tools/src/worker/JazzMessageChannel.ts
```diff
@@ -0,0 +1,73 @@
...
+    const peer = await CojsonMessageChannel.expose(target, cojsonOpts);
+
+    node.syncManager.addPeer(peer);
+  }
+
+  /**
+   * Accept an incoming Jazz connection.
+   * Same as cojson CojsonMessageChannel.accept().
+   */
+  static accept(opts?: AcceptOptions): Promise<Peer> {
```
nrainhart: I'd personally prefer something like `waitForConnection`, but feel free to ignore this comment 😛

packages/cojson/src/CojsonMessageChannel/CojsonMessageChannel.ts
```diff
@@ -0,0 +1,330 @@
...
+        }
+
+        // Validate origin if in Window context and allowedOrigins is specified
+        if (opts.allowedOrigins && opts.allowedOrigins.length > 0) {
+          const origin = messageEvent.origin;
+          const isAllowed = opts.allowedOrigins.some(
+            (allowed) => allowed === "*" || allowed === origin,
+          );
+          if (!isAllowed) {
+            return; // Ignore messages from non-allowed origins
```
nrainhart, line 172: Should we log a warning/error in this case?

packages/cojson/src/CojsonMessageChannel/types.ts
```diff
@@ -0,0 +1,220 @@
...
+ * Type guard to check if a message is a PortTransferMessage.
+ */
+export function isPortTransferMessage(
+  msg: unknown,
+): msg is PortTransferMessage {
+  return (
+    typeof msg === "object" &&
+    msg !== null &&
+    "type" in msg &&
+    (msg as { type: unknown }).type === "jazz:port" &&
```
nrainhart: Nit: you could reuse `isControlMessage` in the rest of the type guards

nrainhart (approved):
Nice! Let's try it out 🚀

packages/jazz-tools/src/worker/JazzMessageChannel.ts
gdorsi: Agree, renamed!

packages/cojson/src/CojsonMessageChannel/CojsonMessageChannel.ts
gdorsi: Yep, done!

packages/cojson/src/CojsonMessageChannel/types.ts
gdorsi: Done!

PR #3333, opened 2026-01-06
Runtime validation

Takeno:
# Description

Cursor's summary is almost complete. Although the scope is clear and documented on Linear, here are the known weird behaviours:

- Validations are applied only to writes: invalid data may still exist and be loaded from cojson. Once loaded, the schema validation is applied again to any mutation. 
- Inputs can be transformed by validators
- As validations are not applied on loading, it might happen to set a non-validated comap as a reference inside another CoValue.  E.g.
```ts
const Box = co.map({x: z.number()});
const Boxes = co.list(Box);

const invalidBox = Box.create({}, {validation: "loose"});

const list = Boxes.create([invalidBox])); // This will not throw
Boxes.create([co.map({}).create({})); // This doesn't either, only the type is checked (CoMap) and uses the reference.
```

Check [this](https://github.com/garden-co/jazz/pull/3333#discussion_r2706451837) conversation too.

- Until the moment we change the default validation mode to "strict", default validation is set to "warn" and an additional `console.warn` reminds users that the default will be "strict" in the next major release.
- ~~`z.default()` now works as expected on writes. They don't work on reads. This is a nice feature, but it can be misunderstood.~~

## Still missing
- [ ] Docs

---

> [!NOTE]
> **Medium Risk**
> Touches core CoValue creation/mutation paths and schema plumbing; while validation is opt-in/defaults to `warn`, it can change runtime behavior (errors/warnings and transformed inputs) across many write operations.
> 
> **Overview**
> Adds **runtime schema validation on writes** for schema-based CoValues: `create` and mutation APIs (`CoMap.$jazz.set/applyDiff`, `CoList.$jazz` mutators, `CoFeed.$jazz.push`, and `upsertUnique`) now accept a `validation` option, and a new global `setDefaultValidationMode()` controls default behavior (currently `warn`, with `strict` encouraged). Validation is implemented via new Zod-backed `getValidationSchema()` hooks on core schema types and enforced via `executeValidation`, while schema descriptors are now sourced from `coValueSchema` rather than legacy per-class `_schema` wiring.
> 
> Removes legacy schema runtime (`coField`/`Encoders` and related `SchemaInit`-based descriptor injection), introduces schema invariants (`assertCoValueSchema`) and built-in schema initialization, and updates tests/docs/examples (examples opt into `setDefaultValidationMode("strict")`, discriminated-union tests add validation coverage, and some adapter defaults/test expectations are adjusted).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bbe9596535dc1116b0147b169a547104221d17fe. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/RichTextSchema.ts
```diff
@@ -25,6 +26,7 @@ export function createCoreCoRichTextSchema(): CoreRichTextSchema {
     collaborative: true as const,
     builtin: "CoRichText" as const,
     resolveQuery: true as const,
+    getValidationSchema: () => z.string().or(z.instanceof(CoRichText)),
```
nrainhart: I don't think we need to add validation schemas in core schemas. Having them in the actual (hydrated) schemas should be enough

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoOptionalSchema.ts
```diff
@@ -28,6 +30,20 @@ export class CoOptionalSchema<
 
   constructor(public readonly innerType: Shape) {}
 
+  getValidationSchema = () => {
+    if (this.innerType.builtin === "CoMap") {
+      return z.optional(
+        z.instanceof(CoMap).or(this.innerType.getValidationSchema()),
```
nrainhart: Shouldn't `z.instanceof(CoMap)` be part of `CoMapSchema.getValidationSchema()`? Same thing with CoList

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoListSchema.ts
```diff
@@ -49,6 +50,18 @@ export class CoListSchema<
    */
   permissions: SchemaPermissions = DEFAULT_SCHEMA_PERMISSIONS;
 
+  getValidationSchema = () => {
+    return z
+      .instanceof(CoList)
```
nrainhart: I wonder if we should also validate the content of the list if it's not shallowly loaded.

And even if it's a shallowly-loaded colist, shouldn't we validate that it contains unloaded elements? I think the current implementation will accept any colist, which means trying to assign a `CoList<number>` into a `CoList<string>` field wouldn't throw any errors

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/schemaValidators.ts
```diff
@@ -0,0 +1,22 @@
+import { Account, Group } from "../../../internal.js";
+import { z } from "../zodReExport.js";
+import type { CoreCoValueSchema } from "./CoValueSchema.js";
+
+export const isCoValueSchema = (item: any): item is CoreCoValueSchema => {
```
nrainhart: Nit: we already have a `isCoValueSchema` function in `coValueSchemaTransformation.ts`

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
```diff
@@ -46,6 +48,57 @@ export class CoMapSchema<
...
   getDefinition: () => CoMapSchemaDefinition;
 
+  getValidationSchema = () => {
+    const plainShape: Record<string, AnyZodSchema> = {};
+
+    for (const key in this.shape) {
+      const item = this.shape[key];
+      // item is a Group
+      if (item?.prototype?.[TypeSym] === "Group") {
+        plainShape[key] = z.instanceof(Group);
```
nrainhart: Can this be simplified as `item.getValidationSchema()`?

nrainhart: Actually, that's what you're doing down below. Why is it necessary to treat Account and Group schemas as special cases?

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
```diff
@@ -46,6 +48,57 @@ export class CoMapSchema<
   catchAll?: CatchAll;
   getDefinition: () => CoMapSchemaDefinition;
 
+  getValidationSchema = () => {
+    const plainShape: Record<string, AnyZodSchema> = {};
+
+    for (const key in this.shape) {
+      const item = this.shape[key];
```
nrainhart, line 73: We should have a function that receives a zod or covalue schema and returns the validation schema. We're going to need it for CoMap, CoList and CoFeed

packages/jazz-tools/src/tools/implementation/zodSchema/runtimeConverters/coValueSchemaTransformation.ts
```diff
@@ -81,8 +82,12 @@ export function hydrateCoreCoValueSchema<S extends AnyCoreCoValueSchema>(
     const ClassToExtend = schema.builtin === "Account" ? Account : CoMap;
 
     const coValueClass = class ZCoMap extends ClassToExtend {
+      static coValueSchema: CoreCoValueSchema;
       constructor(options: { fromRaw: RawCoMap } | undefined) {
-        super(options);
+        super({
+          ...options,
+          schema: ZCoMap.coValueSchema,
```
nrainhart: Do we need `coValueSchema` as a static field? Can't we just capture `coValueSchema` from the enclosing scope?

packages/jazz-tools/src/tools/coValues/coMap.ts
```diff
@@ -562,17 +568,50 @@ export class CoMap extends CoValueBase implements CoValue {
...
+      throw new Error("Cached schema is not a union");
+    }
+
+    // @ts-expect-error as union, it has options fields and 2nd is the plain shape
+    const fieldSchema = this.cachedSchema.options[1]?.shape?.[key] as
+      | z.core.$ZodTypes
+      | undefined;
+
+    // ignore codecs/pipes
+    // even if they are optional and nullable
```
nrainhart: Are there any other zod constructs we should ignore? E.g. transforms & default values

nrainhart (commented):
Great work! I know this is still a draft, but I couldn't stop myself from leaving a bunch of comments 😅

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/RichTextSchema.ts
Takeno: That's totally true. This was in the first round, when I didn't understand the correct usage of `createCoreCoRichTextSchema` and similar. These `getValidationSchema` implementations are unused.

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoListSchema.ts
Takeno: You're right; you can potentially assign a `CoList<number>` to a `CoList<string>`.

The assumption was that there is no validation on reads (for now), so we can't ensure the loaded list has validated content.

MAYBE now that we have coupled CoValue instances and schemas together, it is possible to compare them at least. I'll give a test

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
Takeno: This was due to the old usage of `Group` in schema definitions instead of `co.group()`.
https://github.com/garden-co/jazz/blob/7c63a23994b41c013df082bf21fb980d9ca82238/packages/jazz-tools/src/tools/tests/coMap.test.ts#L2565-L2568

If we force the usage of `co.group()`, then we can remove that condition. I will add a comment anyway.

nrainhart: Yeah, or maybe this generic function I proposed [here](https://github.com/garden-co/jazz/pull/3333#discussion_r2706486441) could also handle CoValue classes

packages/jazz-tools/src/tools/implementation/zodSchema/runtimeConverters/coValueSchemaTransformation.ts
Takeno: I use the static field on static methods like `create`:

https://github.com/garden-co/jazz/pull/3333/files#diff-15d7a1ee65e2eaa1e050e043345fcab92bac6d5ceb844e7857e1cfa20aa4de94R171-R188

But, from this comment, I understood how to replace the constructor override and use `(this.constructor as any).coValueSchema` inside.

Thx!

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
Takeno: Yep, a good [refactoring](https://github.com/garden-co/jazz/pull/3333/files#diff-e845add98a58068db00aa279b30ca5b5848160680bb49da2920ea618b5dc947bR12) came out

Takeno:
Rebased with 0.19.22 for pre-release publish

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoListSchema.ts
Takeno: I tried to compare the schemas. It works fine for cases like:
```ts
const ListOfNumbers = co.list(z.number());
const ListOfStrings = co.list(z.string());

const Data = co.map({ list: ListOfNumbers});

Data.create({
  list: ListOfStrings.create(["test"])
})
```

It works because they use the same Schema instance. But comparing this is a way harder:

```ts
const Data = co.map({
  list: co.list(z.number()
});

Data.create({
  list: co.list(z.string()).create(["test"])
})
```

Comparing schema's items/shape full of references and getters is hard as hell, isn't it?

examples/inspector/tests/lib/data.ts
```diff
@@ -72,19 +80,25 @@ export const createFile = () => {
...
-    originalSize: [1920, 1080],
-    placeholderDataURL: "data:image/jpeg;base64,...",
-  });
+  return ImageDefinition.create(
+    {
+      originalSize: [1920, 1080],
+      placeholderDataURL: "data:image/jpeg;base64,...",
+    },
+    { validation: "loose" },
+  );
```
gdorsi: Why are you using loose validation here?

packages/jazz-tools/src/tools/coValues/coMap.ts
```diff
@@ -562,17 +576,49 @@ export class CoMap extends CoValueBase implements CoValue {
...
   }
 
   get owner(): Group {
     return getCoValueOwner(this.coMap);
   }
 
+  private getPropertySchema(key: string): z.ZodType {
+    if (this.cachedSchema === undefined) {
+      return z.any();
+    }
```
gdorsi, line 622: Could you add some comments to explain why are we turing off some of the schemas?

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -518,16 +531,56 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
...
+      fieldSchema?.def?.type === "pipe" ||
+      // @ts-expect-error
+      fieldSchema?.def?.innerType?.def?.type === "pipe" ||
+      // @ts-expect-error
+      fieldSchema?.def?.innerType?.def?.innerType?.def?.type === "pipe"
+    ) {
+      return z.any();
+    }
+
+    return fieldSchema ?? z.any();
```
gdorsi: How about writing an helper function for these mappings?

Looks like the checks are the same in all the coValue types, the only thing that changes is how to get the fieldSchema.

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoFeedSchema.ts
```diff
@@ -51,32 +58,60 @@ export class CoFeedSchema<
     return this.#permissions ?? DEFAULT_SCHEMA_PERMISSIONS;
   }
 
+  getValidationSchema = () => {
+    return z
+      .instanceof(CoFeed)
+      .or(z.array(generateValidationSchemaFromItem(this.element)));
+  };
```
gdorsi, line 76: These `getValidationSchema` look quite expensive to me, as I expect that zod prioritize validation performance over schema creation performance.

Would it be possible to compute this validation schema once on schema creation?

packages/jazz-tools/src/tools/tests/runtimeValidation.test.ts
```diff
@@ -0,0 +1,438 @@
...
+    );
+
+    profile.$jazz.set("nickname", "dj");
+    expect(profile.nickname).toEqual("dj");
+
+    profile.$jazz.set("nickname", undefined);
+    expect(profile.nickname).toBeUndefined();
+  });
+
+  test("applies defaults when values are omitted", () => {
```
gdorsi: Does this happen only when the validation is strict?

Could you add a test with `loose` validation?

gdorsi (commented):
Great work!

packages/jazz-tools/src/tools/tests/coMap.test.ts
```diff
@@ -1582,6 +1750,56 @@ describe("CoMap applyDiff", async () => {
...
+
+    expect(() =>
+      // @ts-expect-error - age should be a number
+      map.$jazz.applyDiff(newValues, { validation: "loose" }),
+    ).not.toThrow();
+
+    expect(map.age).toEqual("35");
+  });
+
   test("applyDiff with nested changes", () => {
```
gdorsi, line 1960: Could you also add a test for validation in upsertUnique?

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoFeedSchema.ts
Takeno: It was cached internally when used by `JazzApi`. But yes, it's better to move the creation to this level.

packages/jazz-tools/src/tools/coValues/coMap.ts
Takeno:
```
    /**
     * coMapSchema may be undefined if the CoMap is created directly with its constructor,
     * without using a co.map().create() to create it.
     * In that case, we can't validate the values.
     */
```

examples/inspector/tests/lib/data.ts
Takeno: Because, for testing purposes, we were skipping the initialization of the `original` and `progressive` properties. 
But you're right, they can be easily injected with fake data. Fixed 👍 

See: f3f01d3

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoFeedSchema.ts
Takeno: Refactored for all CoValues in f3f01d3

packages/jazz-tools/src/tools/tests/runtimeValidation.test.ts
Takeno: Yes, it happens only on strict validation because on `loose`, it skips the zod parsing at all.
I still think it is a good feature, and it will be documented with "working only in strict mode, only on writes".

packages/jazz-tools/src/tools/tests/coMap.test.ts
Takeno: Nice catch. For upsertUnique, only on creation, it was skipping the validation.
Moved down from CoValueSchema.create to CoValue.create, to cover all cases. And added tests.
See in 9c658cc

packages/jazz-tools/src/tools/coValues/account.ts
```diff
@@ -67,6 +67,9 @@ export type AccountCreationProps = {
 /** @category Identity & Permissions */
 export class Account extends CoValueBase implements CoValue {
   declare [TypeSym]: "Account";
+  static {
+    this.prototype[TypeSym] = "Account";
+  }
```
nrainhart: Is this necessary? `[TypeSym]` is already defined in the constructor

packages/jazz-tools/src/tools/exports.ts
```diff
@@ -67,6 +67,8 @@ export {
   unstable_loadUnique,
   getUnloadedCoValueWithoutId,
   setDefaultSchemaPermissions,
+  setDefaultValidationMode,
+  getDefaultValidationMode,
```
nrainhart: Do we need to export this? I'd expect people to just override the default once per project

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoValueSchema.ts
```diff
@@ -17,6 +19,8 @@ export interface CoreCoValueSchema {
   builtin: string;
 
   resolveQuery: CoreResolveQuery;
+
+  getValidationSchema: () => z.ZodType;
```
nrainhart, line 23: I thought you were going to remove `getValidationSchema` from CoreCoValueSchema. We could add a CoValueSchema interface that contains just this method, and make all actual/hydrated CoValue schemas implement it

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoDiscriminatedUnionSchema.ts
```diff
@@ -55,6 +55,11 @@ export class CoDiscriminatedUnionSchema<
   readonly builtin = "CoDiscriminatedUnion" as const;
   readonly getDefinition: () => CoDiscriminatedUnionSchemaDefinition<Options>;
 
+  getValidationSchema = () => {
+    // Discriminated union schema can apply only if data are plain objects.
+    return z.any();
```
nrainhart: Shouldn't this be something like
```suggestion
    const { discriminator, options } = this.getDefinition();
    return z.discriminatedUnion(discriminator, options.map(schema => schema.getValidationSchema()));
```
?

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoListSchema.ts
nrainhart: Yeah, I guess this is fine for now. Our main focus is on validating scalar values anyway, so I wouldn't go too crazy with collaborative schemas

packages/jazz-tools/src/tools/implementation/zodSchema/validationSettings.ts
```diff
@@ -0,0 +1,111 @@
+import { z } from "./zodReExport.js";
+
+/**
+ * Global validation mode that can include "warn" for logging-only validation.
+ * - "strict": validate and throw on error (default)
+ * - "loose": skip validation entirely
```
nrainhart, line 6: I feel like "loose" is kind of misleading, because it makes me think there's some amount of runtime validation going on. What do you think about renaming this to "disabled" or "skip"?

packages/jazz-tools/src/tools/implementation/zodSchema/validationSettings.ts
```diff
@@ -0,0 +1,111 @@
...
+ * - "warn": validate but only log errors to console (don't throw)
+ */
+export type GlobalValidationMode = "strict" | "loose" | "warn";
+
+/**
+ * Local validation mode for per-operation overrides.
+ * Only supports "strict" | "loose" - "warn" is only available as a global setting.
```
nrainhart, line 13: Why not support "warn" here? I don't see any issues with it, and maybe someone wants to be conservative about validations in a specific part of their codebase

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
```diff
@@ -52,6 +55,41 @@ export class CoMapSchema<
...
+      if (isCoValueSchema(item)) {
+        // Inject as getter to avoid circularity issues
+        Object.defineProperty(plainShape, key, {
+          get: () => generateValidationSchemaFromItem(item),
+          enumerable: true,
+          configurable: true,
+        });
+      } else {
+        plainShape[key] = generateValidationSchemaFromItem(this.shape[key]);
```
nrainhart:
```suggestion
        plainShape[key] = generateValidationSchemaFromItem(item);
```

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
```diff
@@ -52,6 +55,41 @@ export class CoMapSchema<
...
+      }
+    }
+
+    let validationSchema = z.strictObject(plainShape);
```
nrainhart, line 86: I wonder if we should throw an error when unknown keys are found, or just use `z.object` here. Being more lenient would allow doing things like `Dog.create(anotherDog.toJSON())`, which would otherwise fail because `toJSON` includes the `$jazz.id` as part of its output.

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/schemaValidators.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export function normalizeZodSchema(schema: z.ZodType): z.ZodType {
+  // ignore codecs/pipes
+  // even if they are nested into optional and nullable
+  if (
+    schema.def?.type === "pipe" ||
+    // @ts-expect-error
+    schema.def?.innerType?.def?.type === "pipe" ||
+    // @ts-expect-error
+    schema.def?.innerType?.def?.innerType?.def?.type === "pipe"
```
nrainhart, line 131: ⚠️ I think this validation should be made as early as possible (instead of doing it at the CoValue level), and we should be more thorough and traverse the whole zod schema to ensure it contains no pipes. E.g. this doesn't catch pipes inside a `z.object` field schema

packages/jazz-tools/src/tools/coValues/coFeed.ts
```diff
@@ -362,6 +420,22 @@ export class CoFeedJazzApi<F extends CoFeed> extends CoValueJazzApi<F> {
...
+    }
+    this.pushLoose(...items);
+  }
+
+  /**
+   * Push items to this `CoFeed` without applying schema validation.
+   *
+   * @category Content
+   */
+  pushLoose(...items: CoFieldInit<CoFeedItem<F>>[]): void {
```
nrainhart, line 528: Do we want this method to be part of the public API? I'd personally vote against it.

Same thing with all other "loose" methods

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -518,16 +548,48 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
...
     private coList: L,
     private getRaw: () => RawCoList,
+    private coListSchema?: CoreCoValueSchema,
   ) {
     super(coList);
   }
 
+  private getItemSchema(): z.ZodType {
+    /**
+     * coMapSchema may be undefined if the CoMap is created directly with its constructor,
```
nrainhart:
```suggestion
     * coListSchema may be undefined if the CoList is created directly with its constructor,
```

packages/jazz-tools/src/tools/tests/coList.test.ts
```diff
@@ -406,6 +607,26 @@ describe("Simple CoList operations", async () => {
         list.$jazz.splice(1, 0, "lettuce");
         expect(list[1]?.toString()).toBe("lettuce");
       });
+
+      test("splice with validation errors", () => {
```
nrainhart, line 625: We should also validate that the colist wasn't modified (same thing with the other colist methods)

packages/jazz-tools/src/tools/tests/runtimeValidation.test.ts
```diff
@@ -0,0 +1,438 @@
...
+import {
+  co,
+  z,
+  setDefaultValidationMode,
+  getDefaultValidationMode,
+} from "../exports.js";
+import { createJazzTestAccount, setupJazzTestSync } from "../testing.js";
+import { expectValidationError } from "./utils.js";
+
+describe("runtime validation", () => {
```
nrainhart, line 10: Do we have tests for filtering out codecs/pipes/transforms?

packages/jazz-tools/src/tools/tests/runtimeValidation.test.ts
```diff
@@ -0,0 +1,438 @@
...
+    expect(profile.nickname).toBeUndefined();
+  });
+
+  test("applies defaults when values are omitted", () => {
+    const Document = co.map({
+      title: z.string().min(1).default("Untitled"),
+      pageCount: z.number().int().min(1).default(1),
+    });
+
+    // @ts-expect-error - missing required fields
```
nrainhart, line 154: It's kind of weird that this would technically work, but the type system still fails. I'm inclined to disallow default until we have proper support for it at the type level (and maybe also on reads?).

nrainhart (commented):
This is looking great! Besides the comments in my review, the bugs found by Cursor also seem important to address

packages/jazz-tools/src/tools/coValues/account.ts
Takeno: It was to use the same check we do for Groups, when people use the Account class in schema definition. E.g.
```ts
const RequestToJoin = co.map({
  account: Account,
  status: z.literal(["pending", "approved", "rejected"]),
});
```

But at the very end, the checks for both can be easier for those cases:
```diff
- if ("prototype" in item && item.prototype?.[TypeSym] === "Account") {
+ if (item === Account) {
    return z.instanceof(Account);
  }
```

Thx!

nrainhart: Hmmmm not sure I follow. What I meant is that `[TypeSym]` is already being set [here](https://github.com/garden-co/jazz/blob/main/packages/jazz-tools/src/tools/coValues/account.ts#L106), so it shouldn't be necessary to have that static block for the check you want to make.

I don't quite understand the change you made: in `if (item === Account)` you're checking that an Account _object_ equals the Account _class_?

Takeno: Sorry, I wasn't clear at all.
That `TypeSym` check was made [here](https://github.com/garden-co/jazz/pull/3333/files#diff-e845add98a58068db00aa279b30ca5b5848160680bb49da2920ea618b5dc947bR17-R23) because of the usage of Account and Group classes in schema validations
`co.map({ account: Account })` vs `co.map({ account: co.account() })`

I initially used the `TypeSym` static property in the class, as done in other parts of the codebase (as almost every coValue has instance's and class' TypeSym property [e.g.](https://github.com/garden-co/jazz/blob/038fa14204cc82f96541696c02b5fb3a57ad94e6/packages/jazz-tools/src/tools/coValues/coList.ts#L120-L123))

For our use cases, if we receive a schema definition like `co.map({ account: Account })`, we can verify it just using the strict equality (class vs class). If the schema definition contains `co.map({ account: co.account() })`, it will be managed like any other CoValueSchema.

packages/jazz-tools/src/tools/exports.ts
Takeno: Make sense. We will add it later if we find any use case.

packages/jazz-tools/src/tools/coValues/account.ts
nrainhart: Oh, i see! Yeah, your current solution makes total sense, then. Thanks for the explanation!

packages/jazz-tools/src/tools/tests/runtimeValidation.test.ts
Takeno: Disabled, as I found some more weird corner cases to be managed.

Takeno:
PR updated, addressing feedback.

**Notice:** `.default()` and other transformations are currently disabled. Although default values are a useful feature, using the `z.parse` method breaks some internal references when applied to existing coValues, leading to the creation of new CoValues. We are reverting to the original plan to keep this outside of the initial iteration.

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoDiscriminatedUnionSchema.ts
```diff
@@ -55,6 +55,32 @@ export class CoDiscriminatedUnionSchema<
   readonly builtin = "CoDiscriminatedUnion" as const;
   readonly getDefinition: () => CoDiscriminatedUnionSchemaDefinition<Options>;
 
+  getValidationSchema = () => {
+    const { discriminator, options } = this.getDefinition();
+    return z.discriminatedUnion(
```
gdorsi: Why the caching is missing in the discriminatedUnion validation schema?

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -543,8 +613,26 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
...
    */
   push(...items: CoFieldInit<CoListItem<L>>[]): number {
+    const validationMode = resolveValidationMode();
+    if (validationMode !== "loose" && this.coListSchema) {
+      const schema = z.array(this.getItemSchema());
+      executeValidation(schema, items, validationMode) as CoFieldInit<
+        CoListItem<L>
+      >[];
+    }
+    return this.unsafePush(...items);
```
gdorsi:
```suggestion
    return this.pushLoose(...items);
```

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
```diff
@@ -52,6 +55,41 @@ export class CoMapSchema<
...
+    if (this.catchAll) {
+      validationSchema = validationSchema.catchall(
+        generateValidationSchemaFromItem(
+          this.catchAll as unknown as AnyZodOrCoValueSchema,
+        ),
+      );
+    }
+
+    this.#validationSchema = z.instanceof(CoMap).or(validationSchema);
```
gdorsi: Could you add a comment to describe this design choice?

gdorsi (approved):

PR #3336, opened 2026-01-07
feat: add getJazzErrorType to simplify error handing in React error boundaries

gdorsi:
Added `getJazzErrorType` helper function to identify the type of Jazz error from an Error object thrown by suspense hooks. 

This enables error boundaries to display appropriate UI based on whether the error is "unauthorized", "unavailable", or "unknown".


joeinnes (approved):
Tested this by removing myself from the example track's ownership group. Did what it was supposed to do!

PR #3338, opened 2026-01-07
Account.createAs timeout + onCreate credentials

Takeno:
# Description
This PR adds a timeout option to `Account.createAs` method and use it in `waitForAllCoValuesSync`. This is needed because in the `onCreate` callback, we can't know what will happen. 

Additionally, the created account's credentials are passed to the `onCreate` callback, as they might be useful inside.

---

> [!NOTE]
> Enhances account creation APIs and tests to support initialization hooks with credentials and configurable sync timing.
> 
> - Expands `Account.createAs` to pass `credentials` to `onCreate` and adds `waitForSyncTimeout`, forwarding it to `account.$jazz.waitForAllCoValuesSync({ timeout })`
> - Mirrors these options/signature in `AccountSchema.createAs` with detailed docs and example
> - Updates tests to validate credential propagation, migration-before-`onCreate` ordering, and root modifications from the worker
> - Adds a changeset for a patch release
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 69e23bb3cda62072cc0dfec9c118bca5ac7ebeaa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/jazz-tools/src/tools/coValues/account.ts
+++ b/packages/jazz-tools/src/tools/coValues/account.ts
@@ -271,7 +271,12 @@ export class Account extends CoValueBase implements CoValue {
     worker: Account,
     options: {
       creationProps: { name: string };
-      onCreate?: (account: A, worker: Account) => Promise<void>;
+      onCreate?: (
+        account: A,
+        worker: Account,
+        credentials: { accountID: string; accountSecret: AgentSecret },
+      ) => Promise<void>;
+      waitForSyncTimeout?: number;
     },
   ): Promise<{
     credentials: {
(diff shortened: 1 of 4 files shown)


packages/jazz-tools/src/tools/coValues/account.ts
```diff
@@ -271,7 +271,12 @@ export class Account extends CoValueBase implements CoValue {
     worker: Account,
     options: {
       creationProps: { name: string };
-      onCreate?: (account: A, worker: Account) => Promise<void>;
+      onCreate?: (
+        account: A,
+        worker: Account,
+        credentials: { accountID: string; accountSecret: AgentSecret },
+      ) => Promise<void>;
+      waitForSyncTimeout?: number;
```
nrainhart, line 279: Mind adding a comment here about when this param can be useful? I think it's not super obvious

nrainhart: It'd be good to also add a test case

Takeno: Added JSDoc in https://github.com/garden-co/jazz/pull/3338/files#diff-5aa81758f83bade9ddf1ed6dd735a98f34a9f2c48f2edf5be31af8369d9a708aR97-R121

The method is for more advanced usage, and we should explain it in the documentation later.

nrainhart (approved):

PR #3339, opened 2026-01-07
feat(music-player): music streaming

gdorsi:
This PR implements streaming audio playback for the music player example, enabling music to start playing as soon as enough data has been loaded rather than waiting for the entire file to download.

I've also added a passphrase login so I can backup my precious access key in another place alongside the passkey.

Wanted to do this for the Advent of Jazz, but been a little late! 🎄 

### Known limitations

To be able to do proper streaming with a decent codec support I'm using **mediabunny**, which relies on the Web Audio API’s [AudioDecoder](https://developer.mozilla.org/en-US/docs/Web/API/AudioDecoder).

Unfortunately `AudioDecoder` is currently not supported on:
  * Firefox for Android
  * Safari versions earlier than 26


nrainhart:
Not sure what's happening, but I managed to break playback by switching very quickly between different tracks.

Once it happens those tracks get permanently "blocked" until I reload the page

gdorsi:
@nrainhart Good catch, it was a race condition in loadAudio.

Should be fixed now

PR #3345, opened 2026-01-08
feat(storage-streaming): implement priority-based streaming queue for content management

gdorsi:
## Summary
This PR introduces a priority-based streaming queue for storage content that integrates with the existing sync scheduling infrastructure. It replaces the previous setTimeout-based yielding approach with a unified scheduler that coordinates both incoming peer messages and storage content streaming.

## Problem
The previous storage content streaming implementation had several limitations:
- No coordination with incoming messages - Storage streaming ran independently from IncomingMessagesQueue, leading to suboptimal scheduling when both were active
- No priority awareness - All CoValues were streamed with equal priority, regardless of importance (accounts/groups vs binary streams)
- Unpredictable yielding - The setTimeout approach yielded after each chunk regardless of actual time spent
Solution

## Solution

Introduce a priority-based StorageStreamingQueue and unify scheduling in SyncManager:
- HIGH priority content (accounts, groups) bypasses the queue entirely and streams directly
- MEDIUM/LOW priority content goes through the queue and is processed by SyncManager.processQueues()
- Both incoming messages and storage streaming share the same time budget (50ms default)
- Lazy evaluation: callbacks are stored in the queue, content is only fetched when processing

## Load balancing test results

Command: `pnpm run run --workers 8 --durationMs 120000 --inflight 8 --mix 3f:1m`

Main:
```
{
  "elapsedMs": 127737,
  "ops": {
    "total": 5232,
    "fileOps": 3924,
    "mapOps": 1308,
    "errors": 0,
    "opsPerSec": 40.95915827050894
  }
}
```

This branch, without the backlog optimization:
```
{
  "elapsedMs": 127410,
  "ops": {
    "total": 5922,
    "fileOps": 4440,
    "mapOps": 1482,
    "errors": 0,
    "opsPerSec": 46.47986814221804
  }
}
```

This branch, with the backlog optimization:
```
{
  "elapsedMs": 128406,
  "ops": {
    "total": 6649,
    "fileOps": 4990,
    "mapOps": 1659,
    "errors": 0,
    "opsPerSec": 51.78106942043206
  }
}
```


packages/cojson/src/storage/storageSync.ts
```diff
@@ -106,78 +118,116 @@ export class StorageApiSync implements StorageAPI {
...
+    // Send the first chunk
+    this.pushContentWithDependencies(coValueRow, contentMessage, callback);
+    this.knownStates.handleUpdate(coValueRow.id, knownState);
 
-    // If there is no new content but steaming is not active, it's the case for a coValue with the header but no transactions
-    // For streaming the push has already been done in the loop above
-    if (hasNewContent || !contentStreaming) {
-      this.pushContentWithDependencies(coValueRow, contentMessage, callback);
+    // Groups and accounts are streamed directly, other CoValues are queued
+    const useQueue = priority !== CO_VALUE_PRIORITY.HIGH;
```
gdorsi: Unsure about streaming groups directly, going to probably change this and put them in the queue.

packages/cojson/src/storage/storageSync.ts
```diff
@@ -106,78 +118,116 @@ export class StorageApiSync implements StorageAPI {
...
+
+    for (const pushStreamingContent of streamingQueue) {
+      if (useQueue) {
+        this.streamingQueue.push(pushStreamingContent, priority);
+      } else {
+        pushStreamingContent();
+      }
```
gdorsi: With this refactoring we are chunking content in a different way.

The first content message will contain the first chunks of all the sessions, before only the sessions until the first streaming session was encountered.

I think this is better because the chunking applied for the sync goes through `newContentSince` which will lead to a more stable chunking.

packages/cojson-transport-ws/src/BatchedOutgoingMessages.ts
```diff
@@ -116,25 +116,33 @@ export class BatchedOutgoingMessages
...
+      this.backlog += `\n${stringifiedMsg}`;
     } else {
-      this.backlog = payload;
+      this.backlog = stringifiedMsg;
+    }
+
+    // If message itself exceeds the chunk size, send it immediately
+    if (msgSize >= WEBSOCKET_CONFIG.MAX_OUTGOING_MESSAGES_CHUNK_BYTES) {
+      this.sendMessagesInBulk();
     }
```
gdorsi, line 146: Refactored the logic here to produce less strings.

Before we were joining strings to calculate their size, now we only join the strings when we know we are below the budget

packages/cojson/src/sync.ts
```diff
@@ -426,17 +428,89 @@ export class SyncManager {
...
       return;
     }
 
-    this.messagesQueue.processQueue((msg, peer) => {
-      this.handleSyncMessage(msg, peer);
-    });
+    this.processing = true;
+    let lastTimer = performance.now();
+
+    try {
```
nrainhart: Any reason for using this try-finally block? It seems like all things that could throw are inside their own try-catch block

packages/cojson/src/queue/StorageStreamingQueue.ts
```diff
@@ -0,0 +1,96 @@
...
+  [CO_VALUE_PRIORITY.HIGH]: 0,
+  [CO_VALUE_PRIORITY.MEDIUM]: 1,
+  [CO_VALUE_PRIORITY.LOW]: 2,
+} as const;
+
+type StreamingQueueTuple = [
+  LinkedList<ContentCallback>,
+  LinkedList<ContentCallback>,
+  LinkedList<ContentCallback>,
+];
```
nrainhart, line 21: Nit: use named tuples to remember how queues are sorted
```suggestion
type StreamingQueueTuple = [
  high: LinkedList<ContentCallback>,
  medium: LinkedList<ContentCallback>,
  low: LinkedList<ContentCallback>,
];
```

packages/cojson/src/storage/storageSync.ts
```diff
@@ -30,22 +29,25 @@ import type {
...
 
 export class StorageApiSync implements StorageAPI {
-  private streamingCounter: UpDownCounter;
-
   private readonly dbClient: DBClientInterfaceSync;
   private loadedCoValues = new Set<RawCoID>();
 
+  /**
+   * Queue for streaming content that will be pulled by SyncManager.
+   * Only used when content requires streaming (multiple chunks) and priority is not HIGH.
```
nrainhart: This part is outdated, right?
```suggestion
   * Only used when content requires streaming (multiple chunks)
```

packages/cojson/src/storage/storageSync.ts
```diff
@@ -30,22 +29,25 @@ import type {
...
+  readonly streamingQueue: StorageStreamingQueue;
+
   constructor(dbClient: DBClientInterfaceSync) {
     this.dbClient = dbClient;
-    this.streamingCounter = metrics
-      .getMeter("cojson")
-      .createUpDownCounter(`jazz.storage.streaming`, {
-        description: "Number of streaming coValues",
-        unit: "1",
-      });
```
nrainhart, line 47: We're doing away with this metric?

packages/cojson/src/tests/SyncManager.processQueues.test.ts
```diff
@@ -0,0 +1,320 @@
...
+  setupTestNode,
+  tearDownTestMetricReader,
+  waitFor,
+} from "./testUtils.js";
+
+describe("SyncManager.processQueues", () => {
+  let jazzCloud: ReturnType<typeof setupTestNode>;
+
+  beforeEach(async () => {
+    createTestMetricReader();
```
nrainhart: I don't get why this is necessary

packages/cojson/src/tests/SyncManager.processQueues.test.ts
```diff
@@ -0,0 +1,320 @@
...
+      // Push LOW first, then MEDIUM
+      storage.streamingQueue?.push(lowCallback, CO_VALUE_PRIORITY.LOW);
+      storage.streamingQueue?.push(mediumCallback, CO_VALUE_PRIORITY.MEDIUM);
+      storage.streamingQueue?.emit();
+
+      // Wait for both to be processed
+      await waitFor(() => order.length === 2);
+
+      // MEDIUM should be processed first
+      expect(order).toEqual(["medium", "low"]);
```
nrainhart: Nit: add HIGH as well

nrainhart (commented):
The new approach looks good, thanks for adding the benchmarks!

Not approving because this is still a draft and there are some test failures, but I did a full review and don't have any major comments

packages/cojson/src/sync.ts
gdorsi: Right!

packages/cojson/src/storage/storageSync.ts
gdorsi: I'm going to add metrics in the queue, which should be better.

packages/cojson/src/tests/SyncManager.processQueues.test.ts
gdorsi: Copy/pasta 😆

packages/cojson-transport-ws/src/createWebSocketPeer.ts
```diff
@@ -87,32 +87,42 @@ export function createWebSocketPeer({
...
+    // Only some WebSocket implementations emit the error event, so we need to do casting to any to avoid type errors
+    websocket.removeEventListener("error" as any, handleError);
+    pingTimeoutListener.clear();
+    outgoing.drain();
+  }
+
   function handleClose() {
     incoming.push("Disconnected");
     emitClosedEvent();
+    cleanup();
```
gdorsi, line 102: The metrics in production about the outgoing messages don't add up.

This cleanup should be done by SyncManager that calls `outgoing.close` when gets a disconnected message.
But the cleanup function is idempotent, and so I prefer to keep this kind of logic self-contained in this module and get it called twice.

nrainhart (approved):

PR #3348, opened 2026-01-09
feat: added lazy loading storage

antoniomuso:
# Description

This PR implements lazy loading of CoValues from storage to optimize memory usage and response times on sync servers.

### The Problem
When handling load requests from peers, the server was fully loading CoValues from storage (header + all transaction data) even when the peer already had the latest content. This was inefficient for sync servers handling many CoValues.

### The Solution
Instead of doing a full load, we now first query only the knownState (header presence + session transaction counts) and compare it with the peer's state:
- If the peer already has all content → Reply with known message, skip full load
- If the peer needs new content → Proceed with full load from storage

For more details see: [design.md](https://github.com/garden-co/jazz/blob/59a4fdc2d63c50f4dc60843f84e89b2117ba7e72/.specs/lazy-storage-load/design.md)

## Tests

## Checklist


.specs/lazy-storage-load/design.md
```diff
@@ -0,0 +1,549 @@
...
+Add a new method that loads only the knownState from storage, enabling callers to decide whether a full load is needed:
+
+```typescript
+/**
+ * Lazily load only the knownState from storage without loading full transaction data.
+ * This is useful for checking if a peer needs new content before committing to a full load.
+ * 
+ * @param done - Callback with the storage knownState, or undefined if not found in storage
+ */
+lazyLoadFromStorage(done: (knownState: CoValueKnownState | undefined) => void) {
```
gdorsi: How about naming this `getKnownStateFromStorage`?

It would be more explicit on what it does.

.specs/lazy-storage-load/design.md
```diff
@@ -0,0 +1,549 @@
...
+      if (updatedState === "available" || state.isAvailable()) {
+        unsubscribe();
+        done(state.knownState());
+      } else if (updatedState === "errored" || updatedState === "unavailable") {
+        unsubscribe();
+        done(undefined);
+      }
+    });
+    return;
+  }
```
gdorsi: How about directly calling `storage.loadKnownState` here?

The flow would be way more straightforward this way.

The caching/deduplication should be handled at storage level.

.specs/lazy-storage-load/design.md
```diff
@@ -0,0 +1,549 @@
...
+
+    // Peer needs content - do full load from storage
+    this.loadFromStorage((found) => {
+      if (found && this.isAvailable()) {
+        callbacks.onNeedsContent();
+      } else {
+        callbacks.onNotFound();
+      }
+    });
+  });
```
gdorsi: I think this logic might fit better inside SyncManager, since it’s closely related to what we already handle in handleLoad and doesn’t seem to be needed elsewhere. 

Centralizing it there could also make the overall flow easier to follow and reason about.

.specs/lazy-storage-load/design.md
```diff
@@ -0,0 +1,549 @@
...
+**Helper function** (can be in `knownState.ts` or as a method):
+
+```typescript
+/**
+ * Check if the peer already has all the content from storage.
+ */
+function peerHasAllContent(
+  storageKnownState: CoValueKnownState,
+  peerKnownState: CoValueKnownState | undefined,
+): boolean {
```
gdorsi, line 233: We already have this helper https://github.com/garden-co/jazz/blob/main/packages/cojson/src/knownState.ts#L129

The comment needs to be updated thou.

.specs/lazy-storage-load/design.md
```diff
@@ -0,0 +1,549 @@
...
+) {
+  const coValue = this.local.getCoValue(msg.id);
+  const peer = from === "storage" || from === "import" ? undefined : from;
+  
+  // ... existing code ...
+
+  if (!coValue.hasVerifiedContent()) {
+    if (!msg.header) {
+      // NEW: Use lazyLoadFromStorage to check if CoValue exists in storage
+      coValue.lazyLoadFromStorage((storageKnownState) => {
```
gdorsi: We need to run this logic only if the content comes from a Peer or "import"

packages/cojson/src/storage/storageAsync.ts
```diff
@@ -66,7 +66,8 @@ export class StorageApiAsync implements StorageAPI {
     // Check if there's already a pending load for this ID (deduplication)
     const pending = this.pendingKnownStateLoads.get(id);
     if (pending) {
-      pending.then(callback);
+      // Ensure callback is always called, even if pending fails unexpectedly
+      pending.then(callback, () => callback(undefined));
```
antoniomuso, line 70: is it ok to use promises here @gdorsi ? or do you want to avoid them usage?

.specs/lazy-storage-load/design.md
```diff
@@ -0,0 +1,523 @@
...
+  const pending = this.pendingKnownStateLoads.get(id);
+  if (pending) {
+    pending.then(callback);
+    return;
+  }
+
+  // Start new load and track it for deduplication
+  const loadPromise = this.dbClient
+    .getCoValueKnownState(id)
+    .then((knownState) => {
```
antoniomuso, line 167: is it ok to use promises here @gdorsi ? or do you want to avoid them usage?

nrainhart (approved):
~Approved! Have a few comments, but nothing major~

I meant to approve Guido's hard delete PR lol. Disregard this comment

packages/cojson/src/storage/storageAsync.ts
aeplay: lgtm

.specs/lazy-storage-load/design.md
gdorsi: It is ok to use promises in the async storage.

packages/cojson/src/sync.ts
```diff
@@ -622,46 +697,25 @@ export class SyncManager {
...
-          // Should never happen
-          logger.error(
-            "Received new content with no header on a missing CoValue",
-            {
-              id: msg.id,
-            },
-          );
-        }
+        // Content from storage without header - this shouldn't happen normally
+        // as storage loads should include headers
```
gdorsi: This could happen in this case:
- The storage is streaming a CoValue
- The server is under load, so a streaming chunk isn't processed in 10minutes
- The GC cleanup unmounts the CoValue
- The chunk is processed, but the value is unavailable

We should also add a test for this.

gdorsi: As how to handle the issue, I'd say we should not unmount a CoValue if there is a streaming operation active.

Or that the unmount should clean up the streaming queue.

Here we should just log the error, because we should handle the issue somewhere else.

packages/cojson/src/tests/StorageApiAsync.test.ts
```diff
@@ -823,4 +823,90 @@ describe("StorageApiAsync", () => {
...
+
+      // All results should be the same
+      expect(results[0]).toEqual(results[1]);
+      expect(results[1]).toEqual(results[2]);
+      expect(results[0]?.id).toBe(group.id);
+
+      // Database should only be called once due to deduplication
+      expect(dbClientSpy).toHaveBeenCalledTimes(1);
+    });
+  });
```
gdorsi, line 961: Shall we also test the cache here?

packages/cojson/src/tests/knownState.lazyLoading.test.ts
```diff
@@ -0,0 +1,217 @@
...
+      opts.loadKnownState || vi.fn((id, callback) => callback(undefined)),
+    waitForSync: opts.waitForSync || vi.fn().mockResolvedValue(undefined),
+    trackCoValuesSyncState: opts.trackCoValuesSyncState || vi.fn(),
+    getUnsyncedCoValueIDs: opts.getUnsyncedCoValueIDs || vi.fn(),
+    stopTrackingSyncState: opts.stopTrackingSyncState || vi.fn(),
+    close: opts.close || vi.fn().mockResolvedValue(undefined),
+  };
+}
+
+describe("peerHasAllContent", () => {
```
gdorsi, line 43: nit: these tests should be under the knownState name, because they come from the knownState module

packages/cojson/src/tests/messagesTestUtils.ts
```diff
@@ -53,6 +56,10 @@ export function toSimplifiedMessages(
         return `${from} -> ${to} | DONE ${getCoValue(msg.id)}`;
       case "content":
         return `${from} -> ${to} | CONTENT ${getCoValue(msg.id)} header: ${Boolean(msg.header)} new: ${simplifyNewContent(msg.new)}${msg.expectContentUntil ? ` expectContentUntil: ${simplifySessions({ sessions: msg.expectContentUntil, header: true })}` : ""}`;
+      case "lazyLoad":
+        return `${from} -> ${to} | LAZY_LOAD ${getCoValue(msg.id)}`;
+      case "lazyLoadResult":
+        return `${from} -> ${to} | LAZY_LOAD_RESULT ${getCoValue(msg.id)} sessions: ${simplifySessions(msg)}`;
```
gdorsi: How about being more explicit on what's happening?
I don't think we are going to load something different in the future

```suggestion
      case "lazyLoad":
        return `${from} -> ${to} | GET_KNOWN_STATE ${getCoValue(msg.id)}`;
      case "lazyLoadResult":
        return `${from} -> ${to} | GET_KNOWN_STATE_RESULT ${getCoValue(msg.id)} sessions: ${simplifySessions(msg)}`;
```

gdorsi (approved):
LGTM, good job!

gdorsi (approved):

PR #3349, opened 2026-01-09
test(e2e): add a e2e test to cover the sync of values after the app reload

gdorsi:


nrainhart (approved):
Thanks for adding this test!!

PR #3352, opened 2026-01-12
Serialize secretSeed as array in auth header

joeinnes:
# Description
Secret Seeds are stored as Uint8Arrays which leads to them being serialised as objects not arrays. This fix ensures they are correctly serialised so that they can be restored on the client (e.g. for passphrase/passkey generation for a BetterAuth user).

N/A

## Tests

## Checklist

### DIFF
--- a/.changeset/ba-fix-serialisation.md
+++ b/.changeset/ba-fix-serialisation.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Fix serialisation of secret seed in Better Auth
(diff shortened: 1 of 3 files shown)


joeinnes:
Looks like Clerk provider is already handling this: 
https://github.com/garden-co/jazz/blob/eced19a315290c41e088d38e6af5447c5273b75f/packages/jazz-tools/src/tools/auth/clerk/index.ts#L138-L140

gdorsi (approved):

PR #3321, opened 2025-12-31
feat: useMultiCoState and useSuspenseMultiCoState

gdorsi:
### Summary

This PR introduces two new React hooks for subscribing to multiple CoValues in a single hook call, with unified suspense handling and proper TypeScript tuple typing.

### Motivation

Currently, subscribing to multiple CoValues requires calling `useSuspenseCoState` (or `useCoState`) multiple times, which has several limitations:

1. **React's Rules of Hooks**: Hooks cannot be called conditionally or in loops, making it difficult to subscribe to a dynamic list of IDs
2. **Waterfall suspense**: Each `useSuspenseCoState` call can trigger its own suspense boundary, leading to cascading loading states

#### `useSuspenseMultiCoState`

Subscribe to multiple CoValues with unified Suspense handling. Suspends until all values are loaded.

```ts 
const [project, task] = useSuspenseMultiCoState([
  { schema: Project, id: projectId },
  { schema: Task, id: taskId, resolve: { assignee: true } },
] as const);
```

#### `useMultiCoState`

Subscribe to multiple CoValues without Suspense. Returns `MaybeLoaded` values that can be checked via `$isLoaded`.

```ts
const [project, task] = useMultiCoState([
  { schema: Project, id: projectId },
  { schema: Task, id: taskId, resolve: { assignee: true } },
] as const);

if (!project?.$isLoaded || !task?.$isLoaded) {
  return <Loading />;
}
```

PR #3353, opened 2026-01-12
feat: `useCoStates` & `useSuspenseCoStates`

nrainhart:
# Description

This PR introduces two new React hooks for subscribing to multiple CoValues in a single hook call.

## Motivation

Currently, subscribing to multiple CoValues requires calling `useSuspenseCoState` (or `useCoState`) multiple times, which has several limitations:

- **React's Rules of Hooks:** Hooks cannot be called conditionally or in loops, making it difficult to subscribe to a dynamic list of IDs. The new hooks make common patterns (e.g. pagination) easier to use
- **Waterfall suspense:** Each `useSuspenseCoState` call can trigger its own suspense boundary, leading to cascading loading states

## Design choices

- The new hooks work with a single single shape (as opposed to [the previous POC](https://github.com/garden-co/jazz/pull/3321), that supported a different schema and resolve query per element)
 	- we want to avoid having a single hook that's really complex to use, and instead have simpler and more composable APIs
	- this means we can still run into waterfall suspense queries if we're loading multiple types of CoValues in the same component
	- in the future, all types of waterfall suspense queries will be avoided by composing simpler APIs (e.g. `useCoState`/`useCoStates` + `$jazz.promise` + `React.use`)
- Unlike `useCoState`, the new hooks don't allow null/undefined ids as inputs: user can always filter their list of ids beforehand
- Decided not to overload existing hooks (`useCoState` & `useAccount`) to support multiple ids. I expect both types of hooks to support different options in the future (e.g. a `combine` option would only make sense when loading multiple CoValues)
- `select` and `equalityFn` work on individual values, similar to the function that's passed to an `Array.map` operation
- I considered also adding a `combine` callback that transforms the whole list, but I'm not sure it's really needed. All use cases I could think of are also solved by composing `useCoStates` & `useMemo` 
- Re: naming. I wanted to keep the name as concise as possible. `useSuspenseCoState` is already a mouthful, and something like `useMultiSuspenseCoState` felt like too much

## Tests

## Checklist

- [x] I've updated the part of the docs that are affected the PR changes: [Preview](https://jazz-homepage-git-feat-use-multi-co-state-garden-co.vercel.app/docs/react/core-concepts/subscription-and-loading#loading-multiple-covalues-at-the-same-time)

---

> [!NOTE]
> Introduces multi-CoValue subscription hooks with Suspense support and updates examples/docs accordingly.
> 
> - Adds `useCoStates` and `useSuspenseCoStates` in `packages/jazz-tools/src/react-core/hooks.ts` with shared internals (`useCoValueSubscriptions`, `useSubscriptionsSelector`, `useSuspendUntilLoaded`) and selector/equality support; exported via `react` and `react-native-core`
> - Updates docs: new section on loading multiple CoValues in `subscription-and-loading.mdx` and code snippet `ProjectViewWithPagination` demonstrating paginated `Task` loading
> - Example changes: chat app now derives `messageIds` and uses `useCoStates(Message, ids)` for batched loading; simplifies `ChatBubble` to use `createdBy/createdAt`; `Message` schema now `.resolved({ text: true, image: true })`; `BubbleInfo` resolves author via `useCoState(Account, id)`
> - Stress test: `ProjectScreen` paginates tasks using `useCoStates` with shallow `tasks` list
> - Tests: extensive coverage for new hooks (`useCoStates`, `useSuspenseCoStates`) including selectors, equality, branching, id changes, and subscription cleanup
> - Types: `BranchDefinition` no longer allows `owner: null`
> - Changeset: patch release for `jazz-tools`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 05bca6a98f54319fac85274eb65e7f93c16074df. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/jazz-tools/src/react-core/hooks.ts
```diff
@@ -882,3 +915,250 @@ function getResolveQuery(
...
+  const combinedPromise = useMemo(() => {
+    const promises = subscriptions.map((sub) => {
+      if (!sub) {
+        // For null subscriptions (undefined/null IDs), resolve immediately with null
+        return Promise.resolve(null);
+      }
+      return sub.getCachedPromise();
+    });
+
+    return Promise.all(promises);
```
nrainhart, line 921: This will fail if any of the CoValues cannot be loaded. Wondering if we should support an option to filter inaccessible CoValues

packages/jazz-tools/src/react-core/hooks.ts
```diff
@@ -882,3 +915,250 @@ function getResolveQuery(
...
+     * @param name - A unique name for the branch. This identifies the branch
+     *   and can be used to switch between different branches of the same CoValue.
+     * @param owner - The owner of the branch. Determines who can access and modify
+     *   the branch. If not provided, the branch is owned by the current user.
+     *
+     * For more info see the [branching](https://jazz.tools/docs/react/using-covalues/version-control) documentation.
+     */
+    unstable_branch?: BranchDefinition;
+  },
+): TSelectorReturn[] {
```
nrainhart, line 1051: We could make the return type a tuple, to avoid needing to handle potentially undefined elements when getting elements out of the returned list (if the ids are also a tuple). Doing this would make implementing https://github.com/garden-co/jazz/pull/3353/changes#r2682004577 more tricky, though

packages/jazz-tools/src/react-core/hooks.ts
```diff
@@ -103,94 +102,134 @@ export function useCoValueSubscription<
...
+  const state = stateRef.current;
+
+  const branchName = branch?.name;
+  const branchOwnerId = branch?.owner?.$jazz.id;
+
+  const contextChanged =
+    state.contextManager !== contextManager || state.agent !== agent;
+
+  // Quick checks to avoid going through the subscription cache
+  const paramsChanged =
```
nrainhart: Not sure if this is still useful now that we have the subscription cache. Will do some benchmarks to see if we can remove this check

packages/jazz-tools/src/react-core/hooks.ts
gdorsi: I'd say, let's get some experience around usages for `useSuspenseCoStates` and then land some improvements.

packages/jazz-tools/src/react-core/hooks.ts
gdorsi: Would be nice, but before let's wait to see how it goes in the wild.

gdorsi (approved):
Good work!

packages/jazz-tools/src/react-core/hooks.ts
nrainhart: I removed this check and confirmed that hitting the subscription cache every time does not degrade performance.

To test this, I forced a component that loaded 1000 covalues using useCoStates to re-render at 60fps for 5s. In both cases the cost of running this hook is similar (and negligible): 
Screenshot 2026-01-12 at 4 18 21 PM

PR #3354, opened 2026-01-12
feat: remove error callbacks from subscription API

gdorsi:
## Summary

  - Remove onError, onUnauthorized, and onUnavailable callback options from all subscribe functions
  - Change the subscription listener to receive Settled<Resolved<V, R>> instead of just Resolved<V, R>
  - Add (id, listener) overloads to Schema types that only had (id, options, listener)

 ## Breaking Changes

  The subscribe API has been simplified. Error callbacks have been removed, and the listener now receives a Settled value that includes both loaded and error states.

```ts
  // Before
  MyCoMap.subscribe(id, {
    onUnavailable: (value) => console.log("Unavailable"),
    onUnauthorized: (value) => console.log("Unauthorized"),
  }, (value) => {
    console.log(value.field);
  });

  // After
  MyCoMap.subscribe(id, (value) => {
    if (!value.$isLoaded) {
      console.log("Error:", value.$jazz.loadingState);
      return;
    }
    console.log(value.field);
  });
 ```

### DIFF
--- a/homepage/homepage/content/docs/upgrade/0-20-0.mdx
+++ b/homepage/homepage/content/docs/upgrade/0-20-0.mdx
@@ -24,6 +24,7 @@ This also helps us speed up the migration of Jazz Core to Rust which will help u
 **NB**: if you use WASM crypto on Edge runtimes, you need to ensure that WASM is available and initialized. See more details at [WASM on Edge runtimes](/docs/react/server-side/setup#wasm-on-edge-runtimes).
 - We have intruduced a way to permanently [delete CoValues](/docs/core-concepts/covalues/deleting). This brings a new `deleted` loading state.
 - You can now only use strings or records of strings as [`unique` parameters](/docs/react/core-concepts/covalues/comaps#uniqueness). This ensures that serialisation is deterministic.
+- The `subscribe` API has been simplified: error callbacks (`onError`, `onUnauthorized`, `onUnavailable`) have been removed, and the listener now receives a `Settled` value that includes both loaded and error states.
 
 ## Breaking changes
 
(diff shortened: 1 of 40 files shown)

### CODE REVIEW

File: homepage/homepage/content/docs/upgrade/0-20-0.mdx [hunk 1]
```diff
@@ -24,6 +24,7 @@ This also helps us speed up the migration of Jazz Core to Rust which will help u
 **NB**: if you use WASM crypto on Edge runtimes, you need to ensure that WASM is available and initialized. See more details at [WASM on Edge runtimes](/docs/react/server-side/setup#wasm-on-edge-runtimes).
 - We have intruduced a way to permanently [delete CoValues](/docs/core-concepts/covalues/deleting). This brings a new `deleted` loading state.
 - You can now only use strings or records of strings as [`unique` parameters](/docs/react/core-concepts/covalues/comaps#uniqueness). This ensures that serialisation is deterministic.
+- The errors ins `subscribe` API has been made more explicit: error callbacks (`onError`, `onUnauthorized`, `onUnavailable`) have been removed, and the listener now receives a `Settled` value that includes both loaded and error states.
```
joeinnes, line 27:
```suggestion
- Error handling in the `subscribe` API has been made more explicit: error callbacks (`onError`, `onUnauthorized`, `onUnavailable`) have been removed, and the listener now receives a `Settled` value that includes both loaded and error states.
```

homepage/homepage/content/docs/upgrade/0-20-0.mdx
```diff
@@ -199,3 +200,60 @@ export function MyJazzProvider({ children }: { children: React.ReactNode }) {
...
+});
+```
+</CodeGroup>
+
+For most use cases, you can simply return early when the value is not loaded:
+
+<CodeGroup>
+```ts
+MyCoMap.subscribe(id, (value) => {
+  if (!value.$isLoaded) return; // Handle loading/error states
```
joeinnes, line 253:
```suggestion
  if (!value.$isLoaded) return;
```

If we're advising folks to simply return early, I think we can remove this comment.

PR #3358, opened 2026-01-13
Adds an Example for Svelte with Clerk

joeinnes:
# Description
Building on the fantastic work by @oliverga

Adds an example app using Clerk with Svelte, similar to our Better Auth example.

Run the Svelte Clerk example app 

## Tests

## Checklist


gdorsi (approved):

gdorsi (commented):
How about adding the e2e tests to the CI?

joeinnes:
> How about adding the e2e tests to the CI?

I did this. Actually, I cleared out the Clerk dashboard earlier, and so I realised that the React Clerk E2E test was failing so also updated those tests.

PR #3359, opened 2026-01-13
Add improved docs for Suspense/Error Boundaries

joeinnes:
# Description
Adds documentation and guidance on using Error Boundaries

N/A

## Tests

- [x] Tests have not been updated, because: Docs

## Checklist

### DIFF
--- a/homepage/homepage/content/docs/core-concepts/subscription-and-loading.mdx
+++ b/homepage/homepage/content/docs/core-concepts/subscription-and-loading.mdx
@@ -92,7 +92,15 @@ To handle the various loading states, instead of checking `$isLoaded` or `$jazz.
 ```
 </CodeGroup>
 
-[Check out an example error boundary here](https://github.com/garden-co/jazz/blob/main/examples/music-player/src/components/ErrorBoundary.tsx).
+#### Use Suspense with Error Boundaries [!framework=react,react-native,react-native-expo]
+
+ An error boundary serves as a fallback for unrecoverable errors you cannot handle gracefully during rendering. If errors [can be handled locally](#loading-errors) in the component&hairsp;—&hairsp;for example, with `$onError: 'catch'` in a resolve query&hairsp;—&hairsp;it's generally better to do so. 
+ 
+ As a rule of thumb, it’s best to wrap only your routes or major components in error boundaries.
+ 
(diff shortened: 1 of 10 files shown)


homepage/homepage/content/docs/core-concepts/subscription-and-loading.mdx
```diff
@@ -92,7 +92,15 @@ To handle the various loading states, instead of checking `$isLoaded` or `$jazz.
 ```
 </CodeGroup>
 
-[Check out an example error boundary here](https://github.com/garden-co/jazz/blob/main/examples/music-player/src/components/ErrorBoundary.tsx).
+#### Use Suspense with Error Boundaries [!framework=react,react-native,react-native-expo]
+
+ An error boundary serves as a fallback for unrecoverable errors you cannot handle gracefully during rendering. If errors [can be handled locally](#loading-errors) in the component&hairsp;—&hairsp;for example, with `$onError: 'catch'` in a resolve query&hairsp;—&hairsp;it's generally better to do so.
```
nrainhart, line 97: https://jazz.tools/docs/react/core-concepts/subscription-and-loading#catching-loading-errors appears later in this doc, maybe we should add a link?

nrainhart: Unrelated, but it's the first time I notice this `&hairsp;` spacing. What are you using it for?

nrainhart (approved):
Looks like there's a build error with the homepage, but the new content looks good to me!

homepage/homepage/content/docs/core-concepts/subscription-and-loading.mdx
joeinnes: On your main question: `#loading-errors` links to the heading above `#catching-loading-errors`. I think it's better to give the higher level heading because otherwise folks might miss the explanation of why we catch errors the way we do.

On your unrelated question, I love using em-dashes and always have, but now everyone associates em-dashes with LLM slop. I use the hair space because a) I like the look of it, but mostly b) it's a typesetting convention I've never seen an LLM use (and likely never will, it's *slightly* incorrect according to most style guides for exactly opposing reasons: most suggest either a full space, no space, or a thin space, which is a bit bigger than a hair space).

PR #3362, opened 2026-01-14
feat: wait for CoValues to be synced before garbage-collecting them

nrainhart:
# Description

The current GC implementation does not check the CoValue's sync state before unmounting it, which means CoValues may be unmounted while still pending sync.

### Impact on GC performance

In order to get an idea of the performance impact of adding this check as part of GC, we measured the cost of doing known state comparisons between a local node and a peer (using data taken from edge/core sync servers):
```
Benchmark Results
============================================================
Total known state comparisons: 32,327
Total time: 5.62 ms

Session Statistics:
  Total origin sessions: 84,155
  Total current sessions: 141,002
  Average origin sessions per comparison: 2.60
  Average current sessions per comparison: 4.36
```

If we assume that the cost grows linearly, comparing the known state of 3M CoValues would take ~500ms. We don't expect to collect that many CoValues in a single GC operation in the near future, so we should be good.

The code for this benchmark is in https://github.com/garden-co/jazz/tree/poc/gc-known-state-comparison

## Tests

## Checklist


packages/cojson/src/sync.ts
```diff
@@ -1076,6 +1080,18 @@ export class SyncManager {
...
+      // the CoValue is synced
+      const unsubscribe = this.syncState.subscribeToCoValueUpdates(
+        coValueId,
+        (peer, _knownState, syncState) => {
+          if (isPersistentServerPeer(peer) && syncState.uploaded) {
+            this.unsyncedTracker.remove(coValueId);
+            unsubscribe();
+          }
+        },
+      );
```
gdorsi, line 1094: Why are you adding this here?

packages/cojson/src/sync.ts
```diff
@@ -1128,6 +1144,13 @@ export class SyncManager {
     });
   }
 
+  /**
+   * Returns true if the local CoValue changes have been synced to all persistent server peers
+   */
+  isSyncedToServerPeers(id: RawCoID): boolean {
+    return !this.unsyncedTracker.has(id);
+  }
```
gdorsi, line 1158: Does this check verify only if the CoValue there are "local" changes that are not synced?

For the GC we need to check that the entire current CoValue content is synced to the sever peer, because we need it to enure that the content from edges is synced to core before umounting.

packages/cojson/src/sync.ts
nrainhart: This is necessary to mark values as synced when a server peer is added after the covalue's created

packages/cojson/src/sync.ts
nrainhart: You're right, I'll update it

nrainhart: Done [`a0c4e5f` (#3362)](https://github.com/garden-co/jazz/pull/3362/commits/a0c4e5f9d38767ce59e00ad1cd4a761dac1bc70f)

packages/cojson/src/sync.ts
gdorsi: Is this still necessary after the latest changes?

gdorsi (approved):
Good job!

packages/cojson/src/sync.ts
nrainhart: Yeah, this is actually a bug we had with sync state tracking. When adding new peers, we fetch unsynced covalues from storage but skip the ones that are already in memory to avoid unnecessary work (https://github.com/garden-co/jazz/blob/feat/gc-wait-for-sync/packages/cojson/src/sync.ts#L296). This meant we never marked those covalues as synced, even after they were uploaded to the new peer

PR #3364, opened 2026-01-14
fix: flickering when sending messages in chat example

nrainhart:
# Description

When sending messages in the chat example, the whole chat flickers for an instant:

https://github.com/user-attachments/assets/4532299a-fa88-4632-a4aa-b0c67c94f2fb

This regression was introduced by https://github.com/garden-co/jazz/pull/3353, which refactored the chat example to use pagination and avoid fetching all messages at once. 

## Manual Testing

https://github.com/user-attachments/assets/e6e2400f-b747-4657-aa0a-003972fc6ce7

### DIFF
--- a/examples/chat/src/chatScreen.tsx
+++ b/examples/chat/src/chatScreen.tsx
@@ -1,7 +1,7 @@
 import { Account } from "jazz-tools";
 import { createImage } from "jazz-tools/media";
 import { useSuspenseAccount, useSuspenseCoState } from "jazz-tools/react";
-import { useState } from "react";
+import { useRef, useState } from "react";
 import { Chat, Message } from "./schema.ts";
 import {
   BubbleBody,
@@ -25,6 +25,7 @@ export function ChatScreen(props: { chatID: string }) {
   const [showNLastMessages, setShowNLastMessages] = useState(
     INITIAL_MESSAGES_TO_SHOW,
   );
+  const initialLoadCompletedRef = useRef(false);
 
   const messageIds = Array.from(chat.$jazz.refs)
     // We call slice before reverse to avoid mutating the original array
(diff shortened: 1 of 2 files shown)


examples/chat/src/chatScreen.tsx
```diff
@@ -36,10 +43,16 @@ export function ChatScreen(props: { chatID: string }) {
   const messages = useCoStates(Message, messageIds);
 
   // The initial messages should be loaded all at once, so we can avoid flickering
-  if (
-    messages.slice(0, INITIAL_MESSAGES_TO_SHOW).some((msg) => !msg.$isLoaded)
-  ) {
-    return null;
+  // Only gate the initial render, not subsequent renders when new messages arrive
+  if (!initialLoadCompletedRef.current) {
```
nrainhart: If this feels too contrived for an example app, we could just remove this check. I haven't been able to repro a case where it's noticeable that the initial messages appear at different times

gdorsi: Yeah, let's remove the check completely

nrainhart: hmmmm removing the check actually messes up loading the message's author

https://github.com/user-attachments/assets/0aece80d-104a-452a-9609-4a63f1b11ca7

gdorsi: I think this can be solved on `BubbleInfo` by not rendering any text while the profile is loading.

Also this entire problem makes me think that either `useCoStates` or `useSuspenseCoStates` should be tweaked for this use case.

nrainhart: > Also this entire problem makes me think that either useCoStates or useSuspenseCoStates should be tweaked for this use case.

Agree. One thing to note is that you don't necessarily want the "wait until all CoValues are loaded" behaviour on every update, but rather only on the first render.

------------------

Btw, i fixed this in https://github.com/garden-co/jazz/pull/3364/changes/ce50efcc19467965498237f703c54e29d155c15b :

https://github.com/user-attachments/assets/bb4fada1-a684-4084-8b4c-990f0c819e41

gdorsi (approved):

examples/chat/src/chatScreen.tsx
gdorsi: > Agree. One thing to note is that you don't necessarily want the "wait until all CoValues are loaded" behaviour on every update, but rather only on the first render.

If React was able to support concurrent rendering on external stores, Suspense would have been the perfect solution.

PR #3365, opened 2026-01-14
feat: removed json serialization in try add, now it calls direcly ffi…

antoniomuso:
# Description

This feature optimizes the `tryAdd` method by replacing JSON serialization with direct FFI calls. Instead of serializing transaction objects to JSON in JavaScript and parsing them in Rust, the `tryAdd` method now iterates over transactions and passes primitive parameters directly across the FFI boundary.

## Tests

## Checklist


packages/cojson/src/crypto/WasmCrypto.ts
```diff
@@ -234,13 +234,24 @@ class SessionLogAdapter {
...
-      newSignature,
-      skipVerify,
-    );
+    // Use direct calls instead of JSON.stringify for better performance
+    for (const tx of transactions) {
+      if (tx.privacy === "private") {
+        this.sessionLog.addExistingPrivateTransaction(
+          tx.encryptedChanges,
+          tx.keyUsed,
+          BigInt(tx.madeAt),
```
gdorsi: Why are we casting to BigInt here?

In addNewPrivateTransaction we don’t do this, so I’d like to better understand the difference and confirm which approach is intended.

gdorsi (approved):
LGTM, good job!

packages/cojson/src/crypto/WasmCrypto.ts
antoniomuso: I originally added the cast as a precautionary measure, but strictly speaking, it isn't needed. We can use the `f64` representation directly, as it matches the JavaScript number format exactly.

PR #3370, opened 2026-01-15
fix(load-balancing): unmount files after full loading

gdorsi:


tests/load-balancing/src/worker.ts
```diff
@@ -122,32 +122,31 @@ async function main() {
...
 
-    const v = await node.loadCoValueCore(id as RawCoID);
+    const v = await node.loadCoValueCore(id as RawCoID, undefined, true);
 
     opsDone++;
-    if (!v.isAvailable()) {
-      console.error("coValue unavailable", id);
-      await sleep(50 + Math.floor(rng() * 50));
-    } else {
+    if (v.isAvailable()) {
```
nrainhart, line 128: We don't even log anymore if the covalue is unavailable?

nrainhart (approved):

tests/load-balancing/src/worker.ts
gdorsi: Wanted to add it as a metric, forgot about it.

Doing it now

PR #3371, opened 2026-01-15
feat: wait for dependencies to be gc-ed before collecting a CoValue

nrainhart:
# Description

Wait for CoValues' dependencies to be garbage-collected before collecting them. This makes accounts and groups safe to be collected.

This PR also:
- cleans up storage metadata (`loadedCoValues`) when a CoValue is unmounted
- removes the `garbageCollectGroups` option from `enableGarbageCollector`, since it's now always safe to collect groups and accounts

## Tests

## Checklist

### DIFF
--- a/.changeset/clever-rice-turn.md
+++ b/.changeset/clever-rice-turn.md
@@ -1,0 +1,6 @@
+---
+"jazz-run": patch
+"cojson": patch
+---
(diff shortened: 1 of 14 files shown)


gdorsi (approved):
Good job!

PR #3375, opened 2026-01-16
Fix expo placeholders

joeinnes:
# Description
Expo Image Manipulator returns a bare base64 string, which means that the images uploaded from Expo clients had invalid placeholders.

This fixes. 

Build and install the chat app, upload some images. Check that the placeholder URL is correctly set with the `data` prefix.

## Tests

- [x] Tests have not been updated, because: RN testing is not fully ready yet.

## Checklist

### DIFF
--- a/packages/jazz-tools/src/media/create-image/react-native.ts
+++ b/packages/jazz-tools/src/media/create-image/react-native.ts
@@ -1,8 +1,7 @@
-import { NativeModules } from "react-native";
 import type ImageResizerType from "@bam.tech/react-native-image-resizer";
 import type ImageManipulatorType from "expo-image-manipulator";
 import type { Account, Group } from "jazz-tools";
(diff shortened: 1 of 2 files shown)


packages/jazz-tools/src/media/create-image/react-native.ts
```diff
@@ -1,8 +1,7 @@
-import { NativeModules } from "react-native";
 import type ImageResizerType from "@bam.tech/react-native-image-resizer";
 import type ImageManipulatorType from "expo-image-manipulator";
 import type { Account, Group } from "jazz-tools";
-import { FileStream } from "jazz-tools";
+import { co, FileStream } from "jazz-tools";
```
Takeno: nit
```suggestion
import { co, type FileStream } from "jazz-tools";
```

Takeno (approved):

PR #3376, opened 2026-01-16
Add a 512px progressive image variant

joeinnes:
# Description
Currently, we resize images to 256, 1024, and 2048 (plus the original).

This is generally reasonable for thumbnails and full width content on different devices, but there's a big jump between 256 and 1024 for content that does not fill the width of a screen. This PR introduces the additional variant for users who opt in to progressive images

Upload an image with progressive enabled, check that a 512 variant is created

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/media/create-image/react-native.ts
+++ b/packages/jazz-tools/src/media/create-image/react-native.ts
@@ -1,8 +1,7 @@
-import { NativeModules } from "react-native";
 import type ImageResizerType from "@bam.tech/react-native-image-resizer";
 import type ImageManipulatorType from "expo-image-manipulator";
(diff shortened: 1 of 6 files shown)


packages/jazz-tools/src/media/create-image/react-native.ts
```diff
@@ -108,7 +107,8 @@ async function getPlaceholderBase64(filePath: string): Promise<string> {
       );
     }
 
-    return base64;
+    // Convert base64 to data URL
+    return "data:image/png;base64," + base64;
```
gdorsi, line 111: Is this a breaking change?

gdorsi (approved):

PR #3377, opened 2026-01-16
feat(storage): mark CoValue as being in-memory after a store

nrainhart:
# Description

While working on https://github.com/garden-co/jazz/pull/3371, I noticed that when storing CoValue content into storage, we're not tracking those CoValues as being in memory. We only tracked them as being in memory after a load. This led to unnecessarily reloading CoValue dependencies from storage.

This change also uncovered an issue: we should clear the in-memory CoValues set when closing the storage, otherwise we may carry state over to a new node that doesn't have those CoValues in memory (this is just a concern for tests, because we are not closing and reattaching storages at runtime in prod).

I also made node restart async in tests, otherwise we were not waiting for the storage to be closed.

## Tests

## Checklist

### DIFF
--- a/packages/cojson/src/storage/storageAsync.ts
+++ b/packages/cojson/src/storage/storageAsync.ts
@@ -34,6 +34,10 @@ import type {
 export class StorageApiAsync implements StorageAPI {
   private readonly dbClient: DBClientInterfaceAsync;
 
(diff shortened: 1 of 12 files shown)


gdorsi (approved):
LGTM

PR #3379, opened 2026-01-19
fix: flaky garbage collection test

nrainhart:
# Description

The test was flaky because syncing the covalues with the sync server introduced additional (and not deterministic) CoValue accesses in the client that occasionally prevented map2 from being collected. Since sync is not required for this test, I removed it to make the test deterministic.

### DIFF
--- a/packages/cojson/src/tests/GarbageCollector.test.ts
+++ b/packages/cojson/src/tests/GarbageCollector.test.ts
@@ -214,31 +214,31 @@ describe("garbage collector", () => {
     client.addStorage({
       ourName: "client",
     });
-    client.connectToSyncServer();
     client.node.enableGarbageCollector();
 
     const garbageCollector = client.node.garbageCollector;
 
     assert(garbageCollector);
 
+    await vi.advanceTimersByTimeAsync(100);
+
     const group = client.node.createGroup();
     const map1 = group.createMap();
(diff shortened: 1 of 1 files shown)


packages/cojson/src/tests/GarbageCollector.test.ts
```diff
@@ -227,22 +226,22 @@ describe("garbage collector", () => {
     const map1 = group.createMap();
     const map2 = group.createMap();
 
-    await new Promise((resolve) => setTimeout(resolve, 10));
-
-    map1.set("hello", "world", "trusting");
+    await client.node.syncManager.waitForAllCoValuesSync();
 
+    // Advance time
     getCurrentTime.mockReturnValue(2000);
```
gdorsi: suggestion: vitest fake timers also affect `getCurrentTime` which doesn't need to be specifically mocked.

I've found easier to use that kind of mock rather than going directly for `getCurrentTime` when I had to fix other flaky tests here

gdorsi (approved):

packages/cojson/src/tests/GarbageCollector.test.ts
nrainhart: Totally. I wanted to keep the diff small, but now that we switched to timers it's even better

PR #3381, opened 2026-01-19
feat: automate browser performance benchmarks with Playwright

gdorsi:
Adds a browser automation CLI for running statistically rigorous performance benchmarks on the browser-perf-tests scenarios.
- Implements Playwright-based browser automation to run grid load benchmarks
- Provides Tachometer-inspired statistical analysis with 95% confidence intervals, percentiles, and mean/median
- Supports warm storage (default) and cold storage modes for different testing scenarios
- Includes warmup run before benchmarking to ensure consistent results
- Allows reusing existing CoValue IDs to skip fixture generation

Running two subsequent test runs has given me the following results:

Screenshot 2026-01-19 at 17 47 08
Screenshot 2026-01-19 at 17 48 45

The results are quite stable, used 100 runs on a 20x20 grid and took only 2s to run the suite.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3044,6 +3044,12 @@ importers:
         specifier: ^6.16.0
         version: 6.28.0(react-dom@19.1.0(react@19.1.0))(react@19.1.0)
     devDependencies:
+      '@playwright/test':
+        specifier: ^1.50.1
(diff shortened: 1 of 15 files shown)


nrainhart (approved):
Let's gooooo!!

PR #3385, opened 2026-01-20
fix: `setDefaultSchemaPermissions` now modifies existing CoValue schemas

nrainhart:
# Description

`setDefaultSchemaPermissions` does not currently affect existing CoValue schemas. This turned out to be a footgun when having schemas on multiple files, because it means a schema's default permission depends on where the import happens. E.g. in the `browser-perf-tests` app, `setDefaultSchemaPermissions` is used in the top-level schema file ([link](https://github.com/garden-co/jazz/blob/fix/default-schema-permissions/tests/browser-perf-tests/src/schema.ts#L9)), but since that file already imported other schema files, those schemas will have outdated default permissions. 

## Tests

## Checklist

### DIFF
--- a/packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoFeedSchema.ts
+++ b/packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoFeedSchema.ts
@@ -42,11 +42,14 @@ export class CoFeedSchema<
    */
   resolveQuery: DefaultResolveQuery = true as DefaultResolveQuery;
 
+  #permissions: SchemaPermissions | null = null;
   /**
    * Permissions to be used when creating or composing CoValues
    * @internal
    */
-  permissions: SchemaPermissions = DEFAULT_SCHEMA_PERMISSIONS;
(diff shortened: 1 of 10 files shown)


packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoFeedSchema.ts
```diff
@@ -209,7 +212,7 @@ export class CoFeedSchema<
       hydrateCoreCoValueSchema(coreSchema);
     // @ts-expect-error TS cannot infer that the resolveQuery type is valid
     copy.resolveQuery = resolveQuery ?? this.resolveQuery;
-    copy.permissions = permissions ?? this.permissions;
+    copy.#permissions = permissions ?? this.permissions;
```
gdorsi: Does this mean that on copy the default permissions are not accessed via getter anymore?

nrainhart: Good catch! Updated

gdorsi (approved):

PR #3386, opened 2026-01-20
fix: peer reconciliation load all covalues

antoniomuso:
# Description

This design introduces explicit CoValue states (`garbageCollected` and `onlyKnownState`) to distinguish between CoValues we actually care about versus ones we just heard about. This prevents unnecessary data transfer during peer reconciliation by only restoring subscriptions for CoValues that were actively used.

## Tests

## Checklist

### DIFF
--- a/.specs/peer-reconciliation-lazy-load/design.md
+++ b/.specs/peer-reconciliation-lazy-load/design.md
@@ -1,0 +1,862 @@
+# Design: Peer Reconciliation Lazy Load Optimization (v2)
+
+## Overview
+
+This design introduces explicit CoValue states (`garbageCollected` and `onlyKnownState`) to distinguish between CoValues we actually care about versus ones we just heard about. This prevents unnecessary data transfer during peer reconciliation by only restoring subscriptions for CoValues that were actively used.
+
+**Problem with current approach:**
+```
+Client reconnects → iterates ALL CoValues in memory → for unavailable ones, sends LOAD
+→ "unavailable" includes truly unknown CoValues we never cared about
+→ Server loads and sends ALL content for every ID we ever heard of
+```
+
+**Root cause:** The original design treated all "unavailable" CoValues the same, but they're actually different:
+
+| Type | Meaning | Has lastKnownState? | Optimization |
+|------|---------|---------------------|--------------|
+| **Unknown** | Just heard the ID, never had/wanted the data | ❌ No | **SKIPPED** - no LOAD sent during reconciliation |
+| **GarbageCollected** | Had data in memory, was actively used, then GC'd | ✅ Yes | Sends last known state → server sends diff only |
+| **OnlyKnownState** | Checked storage, have knownState saved, but not full content | ✅ Yes | Sends last known state → server sends diff only |
+
+**New behavior:**
+```
+Client reconnects → iterates all CoValues → checks loadingState:
+  - "available" → send LOAD with in-memory knownState
+  - "garbageCollected" → send LOAD with lastKnownState (prevents full retransfer!)
+  - "onlyKnownState" → send LOAD with lastKnownState (prevents full retransfer!)
+  - "unknown" → SKIP (don't send LOAD - we never cared about this CoValue)
+  - "unavailable"/"loading"/"errored" → send LOAD with empty knownState
+```
+
+**Key optimization:** GC'd and onlyKnownState CoValues now send their `lastKnownState`, enabling the server to send only the diff instead of all content.
+
+**Design simplification:** The `knownState()` method is updated to return the `lastKnownState` when available, so callers can simply use `knownState()` in all cases without needing separate handling for different states.
+
+## Integration with Lazy Storage Load
+
+This design integrates with the **lazy-storage-load** spec (`/.specs/lazy-storage-load/design.md`). The `onlyKnownState` state emerges naturally from the existing `getKnownStateFromStorage` API:
+
+- When `getKnownStateFromStorage` is called and **finds data in storage**:
+  - Cache the returned knownState in `lastKnownState`
+  - Mark the CoValue as `onlyKnownState`
+
+This creates two complementary paths to the same outcome:
+
+| Trigger | State Set | When it happens |
+|---------|-----------|-----------------|
+| `getKnownStateFromStorage` returns data | `onlyKnownState` | Server handles LOAD, checks storage for knownState |
+| `unmount()` during GC | `garbageCollected` | Client GC removes full content from memory |
+
+Both states have `lastKnownState` available, enabling synchronous access during peer reconciliation.
+
+## Architecture / Components
+
+### 1. loadingStatuses Unchanged
+
+**Location:** `packages/cojson/src/coValueCore/coValueCore.ts`
+
+The `loadingStatuses` map remains unchanged - it only tracks loading state from sources (storage, peers):
+
+```typescript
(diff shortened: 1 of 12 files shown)


.specs/peer-reconciliation-lazy-load/design.md
```diff
@@ -0,0 +1,640 @@
...
+  this.counter.add(-1, { state: this.loadingState });
+  
+  // Delete the old CoValueCore
+  this.node.internalDeleteCoValue(this.id);
+  
+  // Notify storage that content was unmounted (for cleanup)
+  this.node.storage?.onCoValueUnmounted(this.id);
+  
+  // Create a new "shell" CoValueCore in garbageCollected state
+  this.node.createGarbageCollectedCoValue(this.id, cachedKnownState);
```
gdorsi: How about creating a `garbageCollectCoVaue` method that replaces `internalDeleteCoValue` and does all the other "cleanup" stuff?

I thing `internalDeleteCoValue` is currently used only for testing and we can probably replace it with `garbageCollectCoVaue`.

.specs/peer-reconciliation-lazy-load/design.md
```diff
@@ -0,0 +1,640 @@
...
+  if (this.isAvailable()) {
+    done(this.knownState());
+    return;
+  }
+
+  // Delegate to storage
+  this.node.storage.loadKnownState(this.id, (knownState) => {
+    if (knownState) {
+      // NEW: Cache the knownState and mark as onlyKnownState
+      this.cachedKnownState = knownState;
```
gdorsi: Consider renaming cachedKnownState — the "cached" prefix suggests caching semantics, but this is more of a snapshot or retained value.

Also, I'm concerned about the lifecycle here.
This property should get cleared on the next loading state change, not simply ignored.

.specs/peer-reconciliation-lazy-load/design.md
antoniomuso: We should retain `internalDeleteCoValue` since it provides a way to validate the `onlyKnownState`.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -449,10 +476,17 @@ export class CoValueCore {
...
+    const lastKnownState = this.knownState();
+
     this.counter.add(-1, { state: this.loadingState });
 
+    // Delete the old CoValueCore (this also calls storage?.onCoValueUnmounted)
     this.node.internalDeleteCoValue(this.id);
 
+    // Create a new "shell" CoValueCore in garbageCollected state
+    this.node.createGarbageCollectedCoValue(this.id, lastKnownState);
+
```
gdorsi: How about moving the logic completely inside localNode?

```ts
internalUnmountCoValue(id: string) {
  const current = this.getCoValue(id)
  const lastKnownState = current.knownState();
  const loadingStateBefore = current.loadingState;

  const placeholder = new CoValueCore(id, this);
  placeholder.setGarbageCollectedState(loadingStateBefore, lastKnownState); // provide the previous loading state to do a single counter update
  this.coValues.set(id, coValue); // a single update on the covalues map
}
```
This approach should be easier to optimize, because we can do a single counter update and a single operation on the covalues map.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -467,6 +501,43 @@ export class CoValueCore {
     this.loadingStatuses.set(peerId, { type: "available" });
     this.updateCounter(previousState);
     this.scheduleNotifyUpdate();
+
+    // Clean up if transitioning from garbageCollected/onlyKnownState
+    if (this.isAvailable()) {
+      this.cleanupLastKnownState();
+    }
```
gdorsi: The cleanup should be done when the verified state is provided, which happens on `provideHeader` because it's where the CoValue really is marked as available.

Here we just do state tracking for server peers.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -1927,14 +2016,29 @@ export class CoValueCore {
...
-    // If already available in memory, return the current knownState
-    if (this.isAvailable()) {
+    // If already available in memory or have last known state, return knownState()
+    // (knownState() handles both cases - verified content or lastKnownState)
+    const state = this.loadingState;
+    if (
+      this.isAvailable() ||
+      state === "garbageCollected" ||
+      state === "onlyKnownState"
+    ) {
```
gdorsi: How about:

```suggestion
    const knownState = this.knownState();
   
    if (knownState.header) {
```

Having the header means that we have a non-empty known state, and we do less spreading of the loading state checks.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -1945,7 +2049,12 @@ export class CoValueCore {
     for (const peer of peers) {
       const currentState = this.getLoadingStateForPeer(peer.id);
 
-      if (currentState === "unknown" || currentState === "unavailable") {
+      if (
+        currentState === "unknown" ||
+        currentState === "unavailable" ||
+        currentState === "garbageCollected" ||
+        currentState === "onlyKnownState"
```
gdorsi: nit: peers can't have the `garbageCollected` and `onlyKnownState` statuses

packages/cojson/src/localNode.ts
```diff
@@ -123,7 +123,13 @@ export class LocalNode {
...
     }
 
-    return coValue.loadingState !== "unknown";
+    const state = coValue.loadingState;
+    // garbageCollected and onlyKnownState shells don't have actual content loaded
+    return (
+      state !== "unknown" &&
+      state !== "garbageCollected" &&
+      state !== "onlyKnownState"
+    );
```
gdorsi, line 132: How about covering this with a test?

The logic has become complex enough now to need one.

gdorsi (commented):
Good job!

Definitiely a tricky bug to deal with.

packages/cojson/src/coValueCore/coValueCore.ts
antoniomuso: yeah I removed it

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -1927,14 +2009,25 @@ export class CoValueCore {
...
+    const knownState = this.knownState();
+    if (knownState.header) {
+      done(knownState);
       return;
     }
 
     // Delegate to storage - caching is handled at storage level
-    this.node.storage.loadKnownState(this.id, done);
+    this.node.storage.loadKnownState(this.id, (knownState) => {
+      if (knownState) {
```
gdorsi:
```suggestion
      if (knownState && !this.isAvailable()) {
```

The value could become available in the meantime.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -428,32 +449,22 @@ export class CoValueCore {
...
-    }
+    return this.node.internalUnmountCoValue(this.id);
+  }
 
+  /**
+   * Decrements the counter for the current loading state.
+   * Used during unmount to properly track state transitions.
+   * @internal
+   */
+  decrementCounter() {
```
gdorsi: How about `decrementLoadingStateCounter`?

gdorsi (approved):
Good job!

Could you add a changeset?

PR #3389, opened 2026-01-20
feat(inspector): performance tab

gdorsi:
Adds a new performance tab on the inspector to make it easier to debug Jazz loading flows.

I've added a slider to filter the subscriptions, a bar to give a gist of the timeline of the load operations and manipulated the stacktrace to give an idea of where the load started (where possible).

Used the opportunity to simplify SubscriptionScope types and provide the error stack on Svelte components.

Screenshot 2026-01-21 at 14 36 55


packages/jazz-tools/src/inspector/pages/performance/PerformancePage.tsx
```diff
@@ -0,0 +1,206 @@
...
+            ))}
+          </Grid>
+        </ListPanel>
+        {selectedEntry ? (
+          <SubscriptionDetailPanel
+            entry={selectedEntry}
+            onNavigate={handleNavigateToCoValue}
+            onClose={() => setSelectedRow(null)}
+          />
+        ) : null}
```
Elfo404: nit
```suggestion
        {selectedEntry && (
          <SubscriptionDetailPanel
            entry={selectedEntry}
            onNavigate={handleNavigateToCoValue}
            onClose={() => setSelectedRow(null)}
          />
        )}
```

packages/jazz-tools/src/inspector/pages/performance/SubscriptionDetailPanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+          fill="none"
+          xmlns="http://www.w3.org/2000/svg"
+        >
+          <path
+            d="M1 1L13 13M1 13L13 1"
+            stroke="currentColor"
+            strokeWidth="2"
+            strokeLinecap="round"
+          />
+        </svg>
```
Elfo404, line 131: would love to see this also to close the whole inspector rather than "Close"

packages/jazz-tools/src/inspector/pages/performance/SubscriptionDetailPanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+        <DetailLabel>CoValue</DetailLabel>
+        <DetailValue>
+          <a
+            title="Click to navigate to CoValue"
+            onClick={() => onNavigate(entry.id)}
+            style={{ color: "var(--j-link-color)", cursor: "pointer" }}
+          >
+            {entry.id}
+          </a>
```
Elfo404: it bothers me slighlty that this is not focusable, what about
```suggestion
          <button
            title="Click to navigate to CoValue"
            onClick={() => onNavigate(entry.id)}
            style={{ color: "var(--j-link-color)", cursor: "pointer" }}
          >
            {entry.id}
          </button>
```
semantically is not an anchor anyway

packages/jazz-tools/src/inspector/pages/performance/SubscriptionRow.tsx
```diff
@@ -0,0 +1,233 @@
...
+  return (
+    <RowWrapper
+      className="row-wrapper"
+      data-expanded={isSelected}
+      onClick={onSelect}
+      onKeyDown={handleKeyDown}
+      tabIndex={0}
+      role="button"
+      aria-label={`View details for ${entry.source} ${entry.id}`}
+    >
```
Elfo404, line 194: if it looks like a button and behaves like a button, shall we make it a button instead? now that you removed the "link" from the covalue id on the second row there's no additional interactive element within this one, and we won't need dedicated tests for the behavior

Elfo404 (approved):

PR #3391, opened 2026-01-21
Add Data Modelling Guide

joeinnes:
# Description
Adds a data modelling guide with code snippets.
Also adds an agent skill for designing schemas

Review the doc at reference/data-modelling
## Tests

## Checklist


homepage/homepage/content/docs/reference/data-modelling.mdx
```diff
@@ -0,0 +1,193 @@
...
+};
+
+import { Alert, CodeGroup, FileName } from "@/components/forMdx";
+
+# Data Modelling and Schema Design in Jazz
+
+To understand how best to model data for your Jazz application, it's helpful to first think about how your data is related.
+
+## Jazz as a Collaborative Graph
+In a traditional database, you might model different data types as 'tables' or 'collections'. With Jazz, you model data types as schemas, and you model your data as an explicitly linked graph.
```
nrainhart:
```suggestion
In a traditional database, you might model different data types as 'tables' or 'collections'. With Jazz, you model data types as schemas, and data as an explicitly linked graph.
```

homepage/homepage/content/docs/reference/data-modelling.mdx
```diff
@@ -0,0 +1,193 @@
...
+  <p>To define collaborative data types, Jazz provides utilities under the `co` namespace. For example, here, we use `co.richText()`, to indicate a collaborative value.</p>
+</details>
+
+## Permissions are part of the data model
+
+In traditional databases, permissions are often left to the application layer, or are a more advanced feature of the database that requires additional configuration.
+
+With Jazz, permissions are an integral part of the data model&hairsp;—&hairsp;you need to consider permissions when structuring your data, not just when you're accessing it. Each CoValue has an ownership group, and permissions are defined based on the ownership hierarchy. This is very powerful, as it allows you to easily build cascading permissions hierarchies, but it does mean that you need to ensure each piece of data which needs different permissions to be applied lives in a different container (e.g. a separate `CoMap` or `CoList`). It is not possible to change the group which owns a CoValue. In order to change the permissions applying to a particular CoValue, and admin (or manager) can [make changes to the group membership](/docs/permissions-and-sharing/overview).
```
nrainhart:
```suggestion
With Jazz, permissions are an integral part of the data model&hairsp;—&hairsp;you need to consider permissions when structuring your data, not just when you're accessing it. Each CoValue has an ownership group, and permissions are defined based on the ownership hierarchy. This is very powerful, as it allows you to easily build cascading permissions hierarchies, but it does mean that you need to ensure each piece of data which needs different permissions to be applied lives in a different container (e.g. a separate `CoMap` or `CoList`). It is not possible to change the group which owns a CoValue. In order to change the permissions applying to a particular CoValue, an admin (or manager) can [make changes to the group membership](/docs/permissions-and-sharing/overview).
```

.cursor/skills/jazz-schema-design/SKILL.md
```diff
@@ -0,0 +1,402 @@
...
+Jazz stores referenced ID. Use resolve queries to control reference traversal depth.
+
+### Inverse Relationships (Two-Way)
+
+**One-to-One:**
+
+```ts
+const Author = co.map({
+  name: z.string(),
+  post: Post
```
nrainhart: This should be a getter. Not sure if it's worth keeping this section, considering the next one focuses on recursive refs

.cursor/skills/jazz-schema-design/SKILL.md
```diff
@@ -0,0 +1,402 @@
...
+  author: Author
+});
+```
+
+**One-to-Many:**
+
+```ts
+const Author = co.map({
+  name: z.string(),
+  posts: co.list(Post)
```
nrainhart: Same thing here

.cursor/skills/jazz-schema-design/SKILL.md
```diff
@@ -0,0 +1,402 @@
...
+
+// Usage
+author.posts.$jazz.set(newPost.$jazz.id, newPost);
+```
+
+Note: CoRecords always use string keys. Validate IDs at application level.
+
+## Data Discovery Pattern
+
+CoValues are only addressable by unique ID. Discovery without ID requires reference traversal.
```
nrainhart, line 255: Is it worth including unique CoValues in this section?

homepage/homepage/content/docs/reference/data-modelling.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+You should consider the default permissions you would like a CoValue to have when you are designing your data model, and [consider specifying these in your schema](/docs/permissions-and-sharing/overview#defining-permissions-at-the-schema-level).
```
nrainhart:
```suggestion
You should consider the default permissions you would like a CoValue to have when you are designing your data model. You can [specify these in your schema](/docs/permissions-and-sharing/overview#defining-permissions-at-the-schema-level).
```

homepage/homepage/content/docs/reference/data-modelling.mdx
```diff
@@ -0,0 +1,193 @@
...
+* **admin**: can grant or revoke permissions for others, as well as writing.
+
+These permissions can be granted to individual users or to groups.
+
+<Alert variant="warning" title="Admins">
+  By default, **only** the user creating a CoValue is the admin. In contrast to traditional databases, Jazz does not have a concept of an 'application admin', or a 'root' or 'superuser'. 
+  
+  Unless explicitly defined otherwise in your data model, only the creator of the CoValue will be added as an admin. Once a CoValue is created, its permissions can **only** be changed by an admin (or a manager).
+
+  If creating a nested CoValue inline, then by default, permissions are inherited from the containing CoValue.
```
nrainhart:
```suggestion
  If [creating a nested CoValue inline](/docs/permissions-and-sharing/cascading-permissions#ownership-on-inline-covalue-creation), then by default, permissions are inherited from the containing CoValue.
```

homepage/homepage/content/docs/reference/data-modelling.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+* Add a version field to your schema
+* Only add fields, never remove them
+* Do not change the data type for existing fields
+* When adding fields, make them optional (so that you can load older data without these fields set).
+
+You can also use the [`withMigration()` method](/docs/core-concepts/covalues/comaps#running-migrations-on-comaps) which runs every time a CoValue is loaded (available on `CoMaps` and `Accounts`).
+
+<Alert variant="warning" title="Use migrations carefully">
+  Migrations run *every* time a CoValue is loaded. A poorly-written migration could cause a lot of unnecessary work being done, and potentially slow your app down. Exit as early as possible from the migration, and check before updating the update is necessary.
```
nrainhart:
```suggestion
  Migrations run *every* time a CoValue is loaded. A poorly-written migration could cause a lot of unnecessary work being done, and potentially slow your app down. Exit as early as possible from the migration, and check the update is necessary before updating.
```

nrainhart (approved):
This is great! Left a few comments, mostly details

.cursor/skills/jazz-schema-design/SKILL.md
joeinnes: I would suggest no. This skill is primarily about designing a schema, and using uniqueness is a separate concept. Technically you're right, CoValues are discoverable by uniqueness, but I don't think it's particularly useful to consider that during schema design. Happy to be convinced otherwise though if you disagree?

nrainhart: No, I agree. Actually, this question may be better suited for the data modelling section in our docs. What do you think about that?

gdorsi (approved):
Wondefrul job!

.cursor/skills/jazz-schema-design/SKILL.md
gdorsi: Yeah, and we need to revive the `getOrCreateUnique` API.

`loadUnique` is quite nasty to understand and use, and `upsertUnique` isn't what people look for when working with unique values.

.cursor/skills/jazz-schema-design/SKILL.md
joeinnes: I want to keep the idea of modelling different types of relationships (one-to-one, one-to-many, many-to-many) because I think LLMs will be familiar with those patterns, and scaffolding around those will help them pick the right Jazz tools. 

I've switched the sections around though to introduce the concept of getters first, and then the different relationship types.

.cursor/skills/jazz-schema-design/SKILL.md
joeinnes: And same comment :)

.cursor/skills/jazz-schema-design/SKILL.md
joeinnes: For now, I think I'll proceed with merging this as is, and open a new PR for documenting `unique` in its own right once this is ready: https://github.com/garden-co/jazz/pull/3396. Any objections @nrainhart, @gdorsi?

PR #3392, opened 2026-01-21
fix: correct CoList length descriptor implementation

gdorsi:
- Updated the `CoList` class to properly return the correct length when accessing the `length` property via `getOwnPropertyDescriptor`.
- Refactored related proxy handler methods to make them more readable.
- Added tests to verify the correct behavior of `CoList` proxy traps and ensure proper functionality of length-related operations.

### DIFF
--- a/packages/cojson/src/coValues/coList.ts
+++ b/packages/cojson/src/coValues/coList.ts
@@ -378,6 +378,10 @@ export class RawCoList<
     return arr;
   }
 
+  length() {
+    return this.entries().length;
+  }
+
   /** @internal */
   entriesUncached(): {
     value: Item;
--- a/packages/jazz-tools/src/tools/coValues/coList.ts
+++ b/packages/jazz-tools/src/tools/coValues/coList.ts
@@ -549,7 +549,7 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
       "private",
     );
 
-    return this.raw.entries().length;
+    return this.raw.length();
   }
 
   /**
@@ -567,7 +567,7 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
       this.raw.prepend(item);
     }
 
-    return this.raw.entries().length;
+    return this.raw.length();
   }
 
   /**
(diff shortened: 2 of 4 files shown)


packages/cojson/src/coValues/coList.ts
```diff
@@ -378,6 +378,10 @@ export class RawCoList<
     return arr;
   }
 
+  length() {
+    return this.entries().length;
+  }
```
gdorsi, line 383: Added this method for convenience, so we have the "internal" way of calculating length in one place

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -938,38 +938,57 @@ function toRawItems<Item>(
 
 const CoListProxyHandler: ProxyHandler<CoList> = {
   get(target, key, receiver) {
-    if (typeof key === "string" && !isNaN(+key)) {
-      const itemDescriptor = target.$jazz.schema[ItemsSym] as Schema;
+    if (typeof key === "symbol") {
+      return Reflect.get(target, key, receiver);
+    }
```
gdorsi, line 967: Early return, so we can know that key can only be a string

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -1009,23 +1038,24 @@ const CoListProxyHandler: ProxyHandler<CoList> = {
         writable: false,
         value: target[TypeSym],
       };
-    } else if (key in target) {
-      return Reflect.getOwnPropertyDescriptor(target, key);
-    } else if (typeof key === "string" && !isNaN(+key)) {
```
gdorsi, line 1014: This is where we had the bug.

"length" in target is always true, and using the internal descriptor leads to returning `value: 0`

This shouldn't be used normally by iterators, but can be the only place that was making the `coList.values()` return an empty iterator.

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -1009,23 +1038,24 @@ const CoListProxyHandler: ProxyHandler<CoList> = {
...
+    } else if (!isNaN(+key)) {
       const index = Number(key);
-      if (index >= 0 && index < target.$jazz.raw.entries().length) {
+      if (index >= 0 && index < target.$jazz.raw.length()) {
         return {
           enumerable: true,
           configurable: true,
-          writable: true,
+          writable: false,
+          value: target.$jazz.raw.get(index),
```
nrainhart: For covalue references, it it ok to return just the covalue id and not a reference to the covalue itself?

nrainhart (approved):
Thanks for the investigation! I only have one question. This is good to go once it's addressed

packages/jazz-tools/src/tools/coValues/coList.ts
gdorsi: Good point, going to fix

PR #3396, opened 2026-01-22
feat: add `getOrCreateUnique` method to CoMap, CoList, and CoFeed

gdorsi:
Adds `getOrCreateUnique` method to CoMap, CoList, and CoFeed 

This new method provides a "get or create only" semantic - it returns an existing value as-is, and only uses the provided value when creating a new CoValue. Unlike `upsertUnique`, it does NOT update existing values with the provided value.

It also enforces a `first-comes-wins` type of resolution on init payloads and propagates uniqueness in inline values.

Example usage:
```typescript
const billingStatus = await BillingStatus.getOrCreateUnique({
  value:  { status: { type: "pending" }, balance: { value: INITIAL_BALANCE_VALUE } },
  unique: `billing-${user.$jazz.id}`,
  resolve: { status: true, balance: true }
  owner: billingGroup,
});
```

Now, even if one offline user calls the above line offline we can guarantee that it won't override any valid updates.

Changes:
- Added `getOrCreateUnique` method to CoMap, CoList, and CoFeed that provides "get or create only" semantics - returns existing values as-is and only uses the provided value when creating new CoValues
- Implemented first-comes-wins conflict resolution for unique CoValues, ensuring deterministic behavior when concurrent creates occur
- Propagated uniqueness constraints to child CoValues for consistent handling across nested structures
- Deprecated  `upsertUnique` methods in favor of the new `getOrCreateUnique`
- Simplified the multi-cursors example to use the new API pattern

Updated [the docs](https://jazz-homepage-git-feat-getorcreateunique-garden-co.vercel.app/docs/react/core-concepts/covalues/comaps#uniqueness) to only mention this new API, but probably needs a full rework.

WIth these new features it is now possible to handle safely offline migrations with multiple nested values, with the only constraint that the childs must share the same group.

---

> [!NOTE]
> **High Risk**
> Touches core CRDT transaction validation/processing (`coValueCore`) and changes how invalidated init transactions affect content computation and rebuilds, which can impact convergence and data visibility across nodes. Also alters unique-load/create behavior around permissions and nested uniqueness propagation, so regressions could surface in offline or concurrent-creation flows.
> 
> **Overview**
> Introduces `getOrCreateUnique` on `CoMap`, `CoList`, and `CoFeed` to support **create-if-missing without mutating existing** unique values, and wires it through schema wrappers so both class and `co.*` schema APIs can use it.
> 
> Implements **first-writer-wins** handling for competing “init” transactions (via `meta.fww`) in `cojson` core transaction parsing, including content rebuild scheduling when a late-arriving winner invalidates previously-processed transactions; list processing is updated to *exclude* entries from invalid init transactions while still processing subsequent valid updates.
> 
> Adds derived uniqueness propagation for inline/nested CoValues (and warnings when ownership prevents inheritance), tightens unique create/update behavior around write permissions (returning `UNAUTHORIZED` when a caller can’t create), and deprecates `upsertUnique`/`loadUnique` in docs in favor of `getOrCreateUnique`. The `multi-cursors` example is refactored to use the new unique feed pattern (daily rotating feed keyed by origin+date) and Suspense-based account loading, and extensive new tests cover concurrency/offline/permission scenarios.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3683146570c087fc2af244ba43947ae1a37f6b9a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


examples/multi-cursors/src/utils/loadCursorContainer.ts
```diff
@@ -51,28 +51,16 @@ export async function loadCursorContainer(
...
   const cursorUID = { feed: cursorFeedID, origin: location.origin };
 
-  const cursorContainer = await CursorContainer.loadUnique(
-    cursorUID,
-    group?.$jazz.id,
-  );
+  const cursorContainer = await CursorContainer.getOrCreateUnique({
+    value: {
+      cursorFeed: [],
+    },
```
gdorsi: Going to fix this, as it is an example on how not to use unique.

On conflict, a new cursorFeed is created.

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
```diff
@@ -208,6 +208,54 @@ export class CoMapSchema<
...
+   * Get an existing unique CoMap or create a new one if it doesn't exist.
+   *
+   * Unlike `upsertUnique`, this method does NOT update existing values with the provided value.
+   * The provided value is only used when creating a new CoMap.
+   *
+   * @example
+   * ```ts
+   * const settings = await UserSettings.getOrCreateUnique({
+   *   value: { theme: "dark", language: "en" },
+   *   unique: ["user-settings", me.id],
```
nrainhart: This pattern is not the most straightforward for the example, why not just go with:
```suggestion
   *   unique: "user-settings",
```
?

.specs/unique-conflict-strategy/getOrCreateUnique.md
```diff
@@ -0,0 +1,296 @@
...
+    │   Alice's init wins (earlier)     │
+    │   theme: "dark"                   │
+    └───────────────────────────────────┘
+```
+
+### First-Init-Wins Strategy
+
+The system uses a **first-init-wins** strategy for creation conflicts:
+
+1. **Init transactions** are marked with `meta: { init: true }`
```
nrainhart: I really like this idea of `init` transactions!

.specs/unique-conflict-strategy/getOrCreateUnique.md
```diff
@@ -0,0 +1,296 @@
...
+}
+```
+
+### What Gets Resolved
+
+| Scenario | Resolution |
+|----------|------------|
+| Different `madeAt` timestamps | Earlier timestamp wins |
+| Same `madeAt`, different sessions | Lexicographically smaller sessionID wins |
+| Same session, same timestamp | Earlier txIndex wins |
```
nrainhart, line 62: Would it make sense to allow a single init transaction per session, and require it to be the first transaction?

.specs/unique-conflict-strategy/getOrCreateUnique.md
```diff
@@ -0,0 +1,296 @@
...
+// Bob updates avatar
+bobProfile.$jazz.set("avatar", "avatar.png");
+
+// After sync, both see ALL updates merged:
+// { name: "Shared Profile", bio: "Hello world", avatar: "avatar.png" }
+```
+
+### Nested CoValues (References)
+
+Nested CoValues with `sameAsContainer` permission inherit derived uniqueness:
```
nrainhart, line 138: We really need to make this the default setting. Otherwise it's going to be super-hard for folks to discover (and not trivial for us to explain to them)

.specs/unique-conflict-strategy/getOrCreateUnique.md
```diff
@@ -0,0 +1,296 @@
...
+  value: [Task.create({ title: "Task B", done: false }, { owner: group })],
+  unique: "tasks-123",
+  owner: group,
+});
+
+// After sync, list only contains "Task A" (from Alice's winning init)
+
+// But subsequent additions from both are preserved:
+aliceList.push(Task.create({ title: "Task C", done: false }, { owner: group }));
+bobList.push(Task.create({ title: "Task D", done: false }, { owner: group }));
```
nrainhart, line 249: You didn't mention how you'd handle bob updates that reference init transactions (since bob's init is marked as invalid). I think this is important to nail right, because if a user creates and edits their unique colist while offline for a long period of time, all edits will transitively depend on the init transaction, right?

nrainhart (commented):
Did a pass over the spec. It looks great! I loved the spec format btw, super clear and concise

.specs/unique-conflict-strategy/getOrCreateUnique.md
gdorsi: Yes, good point

.specs/unique-conflict-strategy/getOrCreateUnique.md
gdorsi: Wanted to try the implementation first before coming up with a proposal: https://github.com/garden-co/jazz/pull/3396/changes/e2b8a1ee7643a6d51de1158fa18d795b90f2262d

.specs/unique-conflict-strategy/getOrCreateUnique.md
gdorsi: Agree

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoListSchema.ts
```diff
@@ -180,7 +180,7 @@ export class CoListSchema<
     return this.coValueClass;
   }
 
-  /** @deprecated Use `CoList.upsertUnique` and `CoList.loadUnique` instead. */
+  /** @deprecated Use `getOrCreateUnique` instead. */
```
nrainhart:
```suggestion
  /** @deprecated Use `loadUnique` instead. */
```

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoMapSchema.ts
```diff
@@ -199,7 +199,7 @@ export class CoMapSchema<
     );
   }
 
-  /** @deprecated Use `CoMap.upsertUnique` and `CoMap.loadUnique` instead. */
+  /** @deprecated Use `getOrCreateUnique` instead. */
```
nrainhart:
```suggestion
  /** @deprecated Use `loadUnique` instead. */
```

packages/jazz-tools/src/tools/coValues/interfaces.ts
```diff
@@ -628,7 +653,7 @@ export async function internalLoadUnique<
       resolve: options.resolve,
     });
 
-    if (loaded.$isLoaded) {
+    if (loaded.$isLoaded && loadAs.canWrite(options.owner)) {
```
nrainhart, line 715: Mind adding a test for this?

examples/multi-cursors/src/hooks/useCursorsFeed.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ * Loads or creates a cursor feed for the current day.
+ *
+ * The feed is uniquely identified by the cursor group, origin, and current date.
+ * This ensures each day gets a fresh cursor feed, preventing accumulation of old
+ * cursor data over time.
+ */
+export async function loadCursorContainer(date: string): Promise<CursorFeed> {
```
nrainhart: Nit:
```suggestion
export async function loadCursorFeed(date: string): Promise<CursorFeed> {
```

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -137,13 +137,13 @@ export class VerifiedTransaction {
...
   // The decoded meta information of the transaction
   meta: JsonObject | undefined;
-  isValidated: boolean = false;
   // Whether the transaction is valid, as per membership rules
   isValid: boolean = false;
   // The error message that caused the transaction to be invalid
   validationErrorMessage: string | undefined = undefined;
   // The previous verified transaction for the same session
   previous: VerifiedTransaction | undefined;
+  stage: "to-vaildate" | "validated" | "processed" = "to-vaildate";
```
```suggestion
  stage: "to-validate" | "validated" | "processed" = "to-validate";
```

packages/cojson/src/tests/coValueCore.initTransaction.test.ts
```diff
@@ -0,0 +1,560 @@
...
+    const withInvalid = map.core.getValidSortedTransactions({
+      includeInvalidMetaTransactions: true,
+      ignorePrivateTransactions: false,
+    });
+    expect(withInvalid).toHaveLength(2);
+    expect(withInvalid.filter((tx) => tx.isValid)).toHaveLength(1);
+    expect(withInvalid.filter((tx) => !tx.isValid)).toHaveLength(1);
+  });
+
+  test("getValidTransactions does not return permission-invalid transactions", async () => {
```
nrainhart:
```suggestion
  test("getValidTransactions({includeInvalidMetaTransactions: true}) does not return permission-invalid transactions", async () => {
```

packages/cojson/src/coValueCore/coValueCore.ts
nrainhart: Mind adding a comment with a brief explanation of each stage?

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -1804,6 +1895,28 @@ export class CoValueCore {
...
+      return a.madeAt - b.madeAt;
+    }
+
+    // 2. Same session: compare by txIndex
+    if (a.txID.sessionID === b.txID.sessionID) {
+      return a.txID.txIndex - b.txID.txIndex;
+    }
+
+    // 3. Same timestamp, different sessions: use sessionID for determinism
+    return a.txID.sessionID.localeCompare(b.txID.sessionID);
```
nrainhart: ⚠️ Should we set a locale? It seems like if you don't the browser's default locale is used, which could lead to inconsistent behavior across peers: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#locales

packages/jazz-tools/src/tools/tests/getOrCreateUnique.test.ts
```diff
@@ -0,0 +1,1250 @@
...
+
+    assertLoaded(bobCounter);
+    assertLoaded(aliceCounter);
+
+    expect(bobCounter.$jazz.id).toBe(aliceCounter.$jazz.id);
+    expect(bobCounter.value).toBe(0);
+    expect(aliceCounter.value).toBe(0);
+  });
+
+  test("handles unauthorized access gracefully", async () => {
```
nrainhart, line 177: Can we add a test for the case where one user is unauthorized but also offline? So that from their point of view they still have write access and can create the unique covalue, but after network access is restored, that tx should no longer be valid

packages/jazz-tools/src/tools/coValues/coMap.ts
```diff
@@ -442,25 +458,55 @@ export class CoMap extends CoValueBase implements CoValue {
...
+   * @param options The options for creating or loading the CoMap.
+   * @returns Either an existing CoMap (unchanged), or a new initialised CoMap if none exists.
+   * @category Subscription & Loading
+   */
+  static async getOrCreateUnique<
+    M extends CoMap,
+    const R extends RefsToResolve<M> = true,
+  >(
+    this: CoValueClass<M>,
+    options: {
```
nrainhart, line 489: I'm wondering if we should also let folks configure what should happen on network timeout (create new covalue or throw an error). I can see the current behavior being undesirable for some folks, and it could lead to unexpected situations that are hard to debug

nrainhart (approved):
This is amazing!! Excited to start using `getOrCreateUnique` without all the footguns.

I have some comments, but they're all minor things

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: This slipeed in my local reviews.

Dunno why, but Claude adds this method every time I start working on the init transactions!

My take on this is that we should re-use `compareTransactions` and add the session tie-breaker there.

But I want to do it on a separate PR, and I don't want to use localeCompare because it's not the optimal solution.

Sessions always have the same prefix, and starting the comparison from the initial parts could lead to unfairness (the same account always wins vs others) and it takes more checks if the session is owned by the same account.

We should do char-by-char comparison starting in reverse.

packages/jazz-tools/src/tools/coValues/interfaces.ts
gdorsi: There is already one, "reader permission can load but not create".

packages/jazz-tools/src/tools/tests/getOrCreateUnique.test.ts
gdorsi: Good idea, done

packages/jazz-tools/src/tools/coValues/coMap.ts
gdorsi: This is a good point.

We currently do not distinguish between `unavailable because the server said that they don't have it and timeouts.

It would be good to add the `timeout` loadingState, initially internally and after externally.

packages/jazz-tools/src/tools/coValues/interfaces.ts
nrainhart: That's for creating unique covalues, this one's for update

gdorsi: Checked again, the update path is already covered by the "reader permission can load but not update via upsertUnique" test (line 1283 in getOrCreateUnique.test.ts).

It verifies that a reader calling upsertUnique gets back the original values without the update being applied.

PR #3397, opened 2026-01-23
perf: cache groups readkey access

gdorsi:
# Description

This PR improves performance when accessing read keys in groups, by caching their content validation.

Fixes the performance issue in the following case:
Screenshot 2026-01-23 at 12 36 59

Here the group content was never cached because it was called with `ignorePrivateTransactions` which skips caching.

## Changes

- Removed `ignorePrivateTransactions` when accessing group content inside `CoValueCore.getReadKey()`
- Private transactions are now explicitly marked as invalid when attempted directly on groups
- Private transactions remain valid in accounts (for backward compat with private root)

### DIFF
--- a/packages/cojson/src/permissions.ts
+++ b/packages/cojson/src/permissions.ts
@@ -235,6 +235,7 @@ function determineValidTransactionsForGroup(
   const writeOnlyKeys: Record<RawAccountID | AgentID, KeyID> = {};
   const writeKeys = new Set<string>();
   const memberRoleResolver = new MemberRoleResolver();
+  const isGroup = coValue.isGroup();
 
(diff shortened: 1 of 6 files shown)


packages/cojson/src/permissions.ts
```diff
@@ -247,6 +248,11 @@ function determineValidTransactionsForGroup(
     const tx = transaction.tx;
 
     if (tx.privacy === "private") {
+      if (isGroup) {
```
nrainhart, line 251: I got confused by this function's name and this check. May be worth adding a comment that accounts can have private transactions

nrainhart (approved):

antoniomuso (approved):

PR #3400, opened 2026-01-23
test: improve db cleanup and remove maxWorker limit

gdorsi:


vite.config.ts
```diff
@@ -12,6 +12,5 @@ export default defineConfig({
       "crates/cojson-core-napi",
       "crates/cojson-core-wasm",
     ],
-    maxWorkers: 5,
```
gdorsi, line 15: I did set this because we had some flaky tests that were failing when concurrency was too high.

Now that those are fixed we can remove the limit

nrainhart (approved):

PR #3401, opened 2026-01-23
feat: SessionMap in rust

antoniomuso:
# Description

This PR moves `CoValueHeader` and `SessionMap` data structures to Rust while avoiding performance regression from FFI overhead. The key insight is that we must move **algorithms** to Rust, not just data structures, to avoid per-item FFI calls during iteration.
While this feature currently has a minor impact on performance, we expect to mitigate this as we migrate more components to Rust.
Benchmark:
image

## Tests

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Replaces the public session API across NAPI/WASM/RN with a new Rust-backed `SessionMap`, changing method signatures and return shapes (notably `KnownState` as native objects), which may break consumers or subtly affect sync/signature semantics.
> 
> **Overview**
> Introduces a Rust-backed `SessionMap` API (wrapping `SessionMapImpl`) and **replaces the old `SessionLog` bindings** across NAPI, React Native (uniffi), and WASM. The new surface supports batch `addTransactions`, creating private/trusting transactions, session/transaction queries, streaming known-state tracking, deletion flags, and decrypting transaction payloads/meta.
> 
> Updates generated TypeScript typings/bindings to match the new API, returns `KnownState` as native objects/records instead of JSON strings, and adds extensive JS/WASM/NAPI test coverage for header validation/stable serialization, transaction flows, streaming state, and error cases. Also adds `shortHash` exports for NAPI/RN/WASM and pulls in `indexmap` (with `serde`) to preserve stable ordering in Rust.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9094fb8328542d84a8fe69d007442d86a4e4a293. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


crates/cojson-core-napi/src/lib.rs
```diff
@@ -43,172 +134,278 @@ impl From<CojsonCoreError> for String {
...
-    meta: Option<String>,
-  ) -> napi::Result<()> {
+  pub fn get_transaction(&self, session_id: String, tx_index: u32) -> Option<Transaction> {
     self
       .internal
-      .add_existing_trusting_transaction(changes, made_at as u64, meta)
-      .map_err(|e| napi::Error::new(napi::Status::GenericFailure, e.to_string()))
+      .get_transaction(&session_id, tx_index)
+      .and_then(|tx_json| serde_json::from_str::<RustTransaction>(&tx_json).ok())
+      .map(Transaction::from)
```
gdorsi: Do we need all these two de-serialization and seralization steps?

packages/cojson/src/coValueCore/verifiedState.ts
```diff
@@ -68,42 +69,240 @@ export type TrustingTransaction = {
...
+    signerID: SignerID | undefined,
+    newTransactions: Transaction[],
+    newSignature: Signature,
+  ) {
+    const cached = this.sessionLogCache.get(sessionID);
+    const currentTxCount = this.impl.getTransactionCount(sessionID);
+
+    if (cached) {
+      // Append to existing cache
+      cached.transactions.push(...newTransactions);
```
gdorsi:
```suggestion
for (const tx of  newTransactions) {
   cached.transactions.push(tx);
}
```

JavaScript engines have a maximum argument count limit.

In V8 (Chrome/Node), this is typically around ~65,536 items (sometimes up to ~125,000 depending on stack frame size and available stack space). 
In JavaScriptCore (Safari) it's similar. The exact number isn't guaranteed and varies with stack depth at the time of the call.

Beyond that, you get RangeError: Maximum call stack size exceeded.

packages/cojson/src/coValueCore/verifiedState.ts
```diff
@@ -68,42 +69,240 @@ export type TrustingTransaction = {
...
+        if (sig) {
+          cached.signatureAfter[lastCheckpoint] = sig as Signature;
+        }
+      }
+      this.sessionLogCacheValid.set(sessionID, currentTxCount);
+    } else {
+      // Create new cache entry
+      const sessionLog: SessionLog = {
+        signerID,
+        transactions: [...newTransactions],
```
gdorsi:
```suggestion
        transactions: newTransactions.slice(),
```

Dunno if `.slice()` is faster in every case, but I'm 100% sure that's not slower.

packages/cojson/src/coValueCore/verifiedState.ts
```diff
@@ -68,42 +69,240 @@ export type TrustingTransaction = {
...
+        sessionID,
+      };
+      // Check for in-between signature
+      const lastCheckpoint = this.impl.getLastSignatureCheckpoint(sessionID);
+      if (
+        lastCheckpoint !== undefined &&
+        lastCheckpoint !== null &&
+        lastCheckpoint >= 0
+      ) {
+        const sig = this.impl.getSignatureAfter(sessionID, lastCheckpoint);
```
gdorsi: How about we optimize this to 1 call and make `getLastSignatureCheckpoint` return both the index and the signature?

packages/cojson/src/coValueCore/verifiedState.ts
```diff
@@ -68,42 +69,240 @@ export type TrustingTransaction = {
...
+      lastCheckpoint >= 0
+    ) {
+      // We need to find all checkpoints - iterate from 0 to lastCheckpoint
+      for (let i = 0; i <= lastCheckpoint; i++) {
+        const sig = this.impl.getSignatureAfter(sessionID, i);
+        if (sig) {
+          signatureAfter[i] = sig as Signature;
+        }
+      }
+    }
```
gdorsi: This looks like something that could be abstraced away with a private method.

packages/cojson/src/coValueCore/verifiedState.ts
```diff
@@ -148,33 +396,107 @@ export class VerifiedState {
...
+    for (const sessionID of sessionIds) {
+      yield this.getSessionLog(sessionID);
+    }
+  }
+
+  *sessionKeys(): IterableIterator<SessionID> {
+    const sessionIds = this.impl.getSessionIds() as SessionID[];
+    for (const sessionID of sessionIds) {
+      yield sessionID;
+    }
```
gdorsi: What are these methods used for?

packages/cojson/src/crypto/RNCrypto.ts
```diff
@@ -41,14 +39,44 @@ import {
...
+  PrivateTransaction as RNPrivateTransaction,
+  TrustingTransaction as RNTrustingTransaction,
 } from "cojson-core-rn";
 import { WasmCrypto } from "./WasmCrypto.js";
 
+/**
+ * Convert a Uniffi Transaction enum to the TypeScript Transaction type.
+ * Uniffi enums have a `tag` property and `inner` containing the data.
+ */
+function convertRNTransaction(rnTx: RNTransaction): Transaction {
```
gdorsi: Could you add more context on this transform?

Why do we need it?

packages/codemod-upgrade-0-19/tests/fixtures/test-example.tsx
```diff
@@ -3,11 +3,7 @@
 // Test file demonstrating Jazz hook patterns that will be migrated
```
gdorsi, line 3: This file should not change here

packages/codemod-upgrade-0-19/tests/__snapshots__/codemod.test.ts.snap
```diff
@@ -6,9 +6,7 @@ exports[`runTransform 1`] = `
 // Test file demonstrating Jazz hook patterns that will be migrated
 // This file shows examples of what the codemod will transform
 
-import {
```
gdorsi, line 9: This file should not change here

gdorsi (approved):
LGTM, good job!

packages/cojson/src/crypto/RNCrypto.ts
antoniomuso: I preferred to remove it, we don't have a really significant performance bump.

crates/cojson-core-napi/src/lib.rs
antoniomuso: I am going to remove this.

packages/cojson/src/coValueCore/verifiedState.ts
antoniomuso: This means serializing the result to JSON, with both the result. And parsing it.

packages/cojson/src/coValueCore/verifiedState.ts
antoniomuso: There were not used, I remove them.

crates/cojson-core-napi/src/lib.rs
```diff
@@ -1,11 +1,32 @@
...
+}
+
+impl From<RustKnownState> for KnownState {
+  fn from(ks: RustKnownState) -> Self {
+    KnownState {
+      id: ks.id,
+      header: ks.header,
+      sessions: ks.sessions.into_iter().collect(),
+    }
+  }
```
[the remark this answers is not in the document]
antoniomuso: In production, KnownState doesn't need to be deterministic, so this isn't a problem. (We don't have React Native Unit Testing.)

crates/cojson-core-napi/src/lib.rs
```diff
@@ -43,172 +64,267 @@ impl From<CojsonCoreError> for String {
...
     &self,
+    session_id: String,
     tx_index: u32,
-    encryption_key: String,
+    key_secret: String,
   ) -> napi::Result<Option<String>> {
     self
       .internal
-      .decrypt_next_transaction_meta_json(tx_index, KeySecret(encryption_key))
+      .decrypt_transaction_meta(&session_id, tx_index, &key_secret)
```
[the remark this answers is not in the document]
antoniomuso: Done.

crates/cojson-core-rn/src/generated/cojson_core_rn.ts
```diff
@@ -687,6 +708,71 @@ export function x25519PublicKey(
   );
 }
 
+/**
+ * KnownState as a native Record (no JSON serialization needed)
+ */
+export type KnownState = {
+  id: string;
+  header: boolean;
+  sessions: Map<string, /*u32*/ number>;
```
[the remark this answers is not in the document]
antoniomuso: This is not an issue; we do the conversion inside RNCrypto.

PR #3409, opened 2026-01-26
perf: optimize group iteration for getUncachedReadKey

gdorsi:
Doing this change in order to optimize read key resolution on big groups.

The problem is, iterating all the keys when we need to access a given readKey is too slow for big groups, so I've added an incremental cache built at transaction processing time to skip the iteration and resolve the key directly.


nrainhart (approved):
🚀

PR #3413, opened 2026-01-27
docs(spec): full storage sync

nrainhart:
# Description

- Added spec for full storage sync
- Added internal documentation for Jazz's sync protocol

### DIFF
--- a/.specs/full-storage-reconciliation/design.md
+++ b/.specs/full-storage-reconciliation/design.md
@@ -4,31 +4,46 @@
 
 Full Storage Reconciliation ensures all local CoValue content (both in memory and in storage) is uploaded to the server peers. It's implemented by adding a new `SyncManager.ensureCoValuesSync()` function. This is separate from peer reconciliation and focuses on uploading local changes to servers, without subscribing to peer changes.
 
-The key mechanism is a new **"reconcile"** message. The client sends a batch of `[coValueId, sessionsHash]` pairs; the server checks for each whether it is missing that CoValue or has different `KnownStateSessions`. When it does, it responds with a "known" message (its known state for that CoValue), so that the client can upload content as needed.
+The key mechanism is a new **"reconcile"** message. The client sends a batch of `[coValueId, sessionsHash]` pairs; the server checks for each whether it is missing that CoValue or has different `KnownStateSessions`. When it does, it responds with a "known" message (its known state for that CoValue), so that the client can upload content as needed. The server then sends a **"reconcile-ack"** message with the same batch `id` so the client knows when processing of that batch is complete.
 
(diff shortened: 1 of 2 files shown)


.specs/full-storage-reconciliation/design.md
```diff
@@ -0,0 +1,333 @@
...
+## Design Considerations
+
+1. **No Subscription**: When `skipSubscription` is true, the server should not add the client to its subscription list for that CoValue. This prevents the server from pushing future updates.
+
+2. **Known State Comparison**: The client compares its known state with the server's known state to determine if upload is needed. For in-memory CoValues this comparison happens before sending the load request, for CoValues in storage it happens
+after receiving the known state from the server.
+
+3. **Loading Strategy**: For CoValues not in memory, we use `getKnownStateFromStorage` to avoid full loads. Only load full content when we need to send it to the server.
+
+4. **Performance**: For large databases, consider batching the processing of CoValues to avoid blocking.
```
nrainhart: I think this is actually a must. I'll update the spec, but it shouldn't change the rest of the things mentioned in this document

.specs/full-storage-reconciliation/design.md
```diff
@@ -0,0 +1,333 @@
...
+#### 2.1. Add skipSubscription Flag to LoadMessage
+
+**File:** `packages/cojson/src/sync.ts`
+
+Modify the `LoadMessage` type:
+
+```typescript
+export type LoadMessage = {
+  action: "load";
+  skipSubscription?: boolean;  // If true, server should only return known state, not send new content
```
nrainhart: I'm actually wondering if we should introduce a new sync message. `skipSubscription` totally changes the behavior of how `"load"` works. We're not even expecting CoValue content as a response

.specs/full-storage-reconciliation/design.md
```diff
@@ -0,0 +1,333 @@
...
+
+Add new method to `SyncManager`:
+
+```typescript
+/**
+ * Ensures all CoValues (both in memory and in storage) are synced to server peers.
+ * This checks known states and uploads local content to servers without
+ * subscribing to server changes.
+ */
+ensureCoValuesSync(): void {
```
nrainhart, line 260: There's still an open question about when this sync process kicks in. Since it's quite advanced, I'd say we just offer this function now, and tell folks who really need it to call it on app startup based on some condition

gdorsi: We could schedule the thing every 30 days from the last full storage sync.

nrainhart:
Closing in favor of https://github.com/garden-co/jazz/pull/3422

PR #3417, opened 2026-01-29
chore: add skills for SDD and changeset generation

gdorsi:


antoniomuso (approved):

.cursor/skills/spec/SKILL.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Workflow
+
+Follow these steps in order. **Do not skip steps.** Always ask for explicit approval before moving to the next step.
+
+### Step 1: Create Feature Directory
+
+Create a feature directory under `.specs/{feature_name}` using kebab-case for the name.
+
+### Step 2: Create Design Document
```
gdorsi: Removed the requirements step to jump directly into the design.

It is helpful only in a very limited amount of cases, when the project is complicated.

While adding this comment I've realized that I can add it as optional opt-in step

PR #3422, opened 2026-01-30
feat: full storage reconciliation

nrainhart:
## Description

Full Storage Reconciliation ensures all CoValue content in storage is uploaded to the server peers. It's implemented by adding a new `SyncManager.startStorageReconciliation()` function. This is separate from peer reconciliation and focuses on uploading local changes to servers, without subscribing to peer changes. It concerns only CoValues in storage, as in-memory CoValues are already expected to be synced.

The key mechanism is a new **"reconcile"** message. The client sends a batch of `[coValueId, sessionsHash]` pairs; the server checks for each whether it is missing that CoValue or has different `KnownStateSessions`. When it does, it responds with a "load" message, so that the client can upload content as needed. The server then sends a **"reconcile-ack"** message with the same batch `id` so the client knows when processing of that batch is complete.

The reconciliation process runs every 30 days, and uses a simple locking mechanism to prevent multiple tabs/sessions in the same device from re-triggering the process. If the process is interrupted, we'll re-run it the next day.

### Testing

I ran the full storage reconciliation process in an app with 130k CoValues, and the process completed in ~15s without a noticeable performance degradation (as all CoValues were already synced). If the number of unsynced CoValues is high, this will of course introduce and additional cost due to the client reading the full CoValue content from storage.

I also experimented with increasing the batch size to 1000, but then you start noticing slower reads when using the app.

To validate this process won't cause excessive load when syncing between edge and core servers, we plan to roll this feature out incrementally to sync servers using `jazz-run --full-storage-reconciliation`.

---

> [!NOTE]
> **Medium Risk**
> Touches core sync protocol and storage backends with new message types, batching, and persisted locking/migrations; bugs could cause missed uploads or extra load, though changes are gated behind an enable flag and well-tested.
> 
> **Overview**
> Adds a **full storage reconciliation** flow to periodically ensure all CoValues persisted in local storage are present/up-to-date on persistent server peers. This introduces new sync protocol messages `reconcile`/`reconcile-ack`, a low-priority queueing rule for reconciliation traffic, and client/server handlers that compare hashed known-state sessions and trigger targeted `load`/`content` uploads without creating subscriptions.
> 
> Extends the Storage API and backends (SQLite sync/async and IndexedDB) with batched CoValue enumeration (`getCoValueIDs`, `getCoValueCount`) plus a per-peer reconciliation **lock with resumable progress** (`tryAcquireStorageReconciliationLock`, `renewStorageReconciliationLock`, `releaseStorageReconciliationLock`), including new DB tables/object stores and migrations/version bumps. Reconciliation can be enabled via a `LocalNode` flag (wired through `jazz-tools` contexts and `jazz-run --full-storage-reconciliation`), and is automatically attempted when adding persistent server peers, with extensive new tests and docs/specs covering batching, scheduling, interruption/resume, and ack tracking.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0415a3607a691c536e6b4d9da6c2aca313f80893. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/cojson/src/sync.ts
```diff
@@ -281,15 +315,83 @@ export class SyncManager {
...
+      for (const peer of serverPeers) {
+        this.pendingReconciliationAck.add(`${batchId}#${peer.id}`);
+        this.trySendToPeer(peer, msg);
+      }
+    };
+
+    const processStorageBatch = (offset: number) => {
+      this.local.storage!.getCoValueIDs(batchSize, offset, (batch) => {
+        // Skip in-memory CoValues
+        const pending = batch.filter(({ id }) => !this.local.hasCoValue(id));
```
antoniomuso: How do we manage the "onlyKnownState" CoValues? They don't have active subscriptions.

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -352,6 +352,10 @@ export class CoValueCore {
     return this.hasVerifiedContent();
   }
 
+  isPartiallyLoaded(): boolean {
```
antoniomuso: Since `garbageCollected` is also a "partial" state, the name `isPartiallyLoaded` is definitely misleading if it only checks for "onlyKnownState".
Maybe `isOnlyKnownState` is better? WDYT?

.specs/full-storage-reconciliation/design.md
```diff
@@ -0,0 +1,446 @@
...
+export type ReconcileMessage = {
+  action: "reconcile";
+  id: string;
+  values: [coValue: RawCoID, sessionsHash: string][];
+};
+```
+
+- **id**: Unique batch id for this reconcile message (e.g. UUID or monotonic string). Used to correlate the **reconcile-ack** response.
+- **coValue**: CoValue ID (`RawCoID`).
+- **sessionsHash**: Hash of that CoValue's **KnownStateSessions** (the `sessions` field of `CoValueKnownState`: `{ [sessionID]: number }`). Use `CryptoProvider.shortHash(sessions)` so client and server hashes are comparable.
```
gdorsi, line 23: shortHash uses stableStringify inside which can't be ported safely inside Rust.

Probably not a huge deal here since the input is quite strict, @antoniomuso what do you think?

packages/cojson/src/sync.ts
nrainhart: Good catch! I'll change this to skip only available covalues

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -352,6 +352,10 @@ export class CoValueCore {
...
+    return this.loadingState === "onlyKnownState";
+  }
```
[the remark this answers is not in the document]
nrainhart: This is ok. We don't GC covalues until they're synced

packages/cojson/src/sync.ts
nrainhart: Fixed in [`23f4d65` (this PR)](https://github.com/garden-co/jazz/pull/3422/commits/23f4d656e4538ca445676a8399038eefcd3e4e29)

packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -363,6 +389,20 @@ export class IDBClient implements DBClientInterfaceAsync {
...
+  ): Promise<{ id: RawCoID }[]> {
+    const rows = await queryIndexedDbStore<StoredCoValueRow[]>(
+      this.db,
+      "coValues",
+      (store) =>
+        // Include upper bound but not lower bound (offset starts at 0)
+        store.getAll(IDBKeyRange.bound(offset, offset + limit, true, false)),
+    );
+    return rows.map((row) => ({ id: row.id }));
+  }
```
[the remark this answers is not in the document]
nrainhart: This is fine, entries are never deleted from the `coValues` object store

.specs/full-storage-reconciliation/design.md
```diff
@@ -0,0 +1,446 @@
...
+          }
+          sendReconcileMessageWhenDone();
+        });
+      }
+      if (pending.length === 0 && batch.length >= batchSize) {
+        processStorageBatch(offset + batchSize);
+      }
+    });
+  };
+  processStorageBatch(0);
```
gdorsi, line 309: Why are we not resuming from the last interrupted reconciliation?

.specs/full-storage-reconciliation/design.md
```diff
@@ -0,0 +1,446 @@
...
+      const sendReconcileMessageWhenDone = () => {
+        if (++done === pending.length) {
+          sendReconcileBatch(entries);
+          if (batch.length === batchSize) {
+            processStorageBatch(offset + batch.length);
+          }
+        }
+      };
+      for (const { id } of pending) {
+        const coValue = this.local.getCoValue(id);
```
gdorsi, line 293: This allocates an unknown CoValue, this can be a problem on the sync server where we have millions of coValue.

How about we try to do this operation without allocating it?

.specs/full-storage-reconciliation/design.md
```diff
@@ -0,0 +1,446 @@
...
+ * and if no other process/tab holds the lock for this peer, then acquires it.
+ * @param sessionId - Unique ID for this process/tab (e.g. crypto.randomUUID())
+ * @param peerId - The peer ID to reconcile with.
+ * @param callback - Called with result. If acquired, caller must call releaseStorageReconciliationLock on completion.
+ */
+tryAcquireStorageReconciliationLock(
+  sessionId: string,
+  peerId: string,
+  callback: (result: StorageReconciliationAcquireResult) => void
+): void;
```
gdorsi, line 352: How do you feel about offloading some of the lock complexity using `navigator.locks`?

The fallback for non-secure contexts could be that we don't do the reconciliation.

packages/cojson/src/sync.ts
```diff
@@ -727,11 +884,69 @@ export class SyncManager {
...
+  handleReconcile(msg: ReconcileMessage, peer: PeerState): void {
+    let pending = msg.values.length;
+    const sendAckWhenDone = () => {
+      if (--pending === 0) {
+        this.trySendToPeer(peer, { action: "reconcile-ack", id: msg.id });
+      }
+    };
+
+    for (const [coValueId, clientSessionsHash] of msg.values) {
+      const coValue = this.local.getCoValue(coValueId);
```
gdorsi: Again, it would be great if we could do the operation without loading the coValue in memory.

We could maybe add a method on localNode to "maybe" get the coValue?

.specs/full-storage-reconciliation/design.md
nrainhart: Good point, should be easy to do

.specs/full-storage-reconciliation/design.md
nrainhart: I feel like the result we'll end up being more complex, because we'll still need the persistence anyway to track completed and interrupted runs, and we don't win much (just dropping the lock expiration logic, which I think is _ok_ anyway). Happy to discuss it more if you feel strongly about this

.specs/full-storage-reconciliation/design.md
antoniomuso: I just did this in my latest branch, and it's now in main so shortHash is now in Rust. We still have to use stableStringify, but it is faster than before.

.specs/full-storage-reconciliation/design.md
nrainhart: Attempted to avoid keeping the covalue in memory, but it breaks assumptions in our sync protocol. For example, when handling "known" messages we only send new content back if the covalue is available or if its known state is loaded: https://github.com/garden-co/jazz/blob/feat/full-storage-sync/packages/cojson/src/sync.ts#L885

Don't think it's a good idea to change those assumptions and e.g. always fetch known state from storage when receiving a "known" message

nrainhart:
@gdorsi this is ready for another pass. Latest additions:
- when resuming an interrupted reconciliation process, we now pick it up in the offset it was left off 
- made full-storage reconciliation opt-in. It's enabled by default on clients, and can be enabled in sync servers using the `--full-storage-reconciliation` flag

packages/cojson/src/sync.ts
```diff
@@ -734,11 +910,77 @@ export class SyncManager {
...
+    const nextOffset = this.pendingReconciliationAck.get(key);
+    this.pendingReconciliationAck.delete(key);
+    if (nextOffset !== undefined) {
+      this.local.storage?.renewStorageReconciliationLock(
+        this.local.currentSessionID,
+        peer.id,
+        nextOffset,
+      );
+    }
+  }
```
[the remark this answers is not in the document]
nrainhart: I initially went with an implementation that kept track of batch ordering and never ack-ed a batch before ack-ing the previous ones, but it made the code way more complex and required storing additional metadata which could be expensive for sync servers. I think it's fine to make this process' resumption best-effort for now, and reevaluate if we find issues in the future.

packages/cojson/src/sync.ts
```diff
@@ -281,15 +317,148 @@ export class SyncManager {
...
+        onComplete();
+      }
+    };
+
+    const sendReconcileBatch = (
+      entries: [RawCoID, string][],
+      offset: number,
+    ) => {
+      if (entries.length === 0) return;
+      const batchId = crypto.randomUUID();
```
gdorsi: We should find something different, randomUUID is not avail on RN

packages/cojson/src/sync.ts
```diff
@@ -281,15 +317,148 @@ export class SyncManager {
...
+        offset + batchSize,
+      );
+      this.trySendToPeer(peer, msg);
+    };
+
+    const processStorageBatch = (offset: number) => {
+      this.local.storage!.getCoValueIDs(batchSize, offset, (batch) => {
+        // Skip in-memory CoValues
+        const coValues = batch.map(({ id }) => this.local.getCoValue(id));
+        const pending = coValues.filter((coValue) => !coValue.isAvailable());
```
gdorsi: We should filter only the unknown loadingStates here

packages/cojson/src/sync.ts
```diff
@@ -281,15 +317,148 @@ export class SyncManager {
...
+        let done = 0;
+        const entries: [RawCoID, string][] = [];
```
gdorsi: How about we check here if the peer is already too busy sending content?

We can implement this for the WS peer by checking that the OutoingQueue is bigger than 0

packages/cojson/src/sync.ts
```diff
@@ -281,15 +317,148 @@ export class SyncManager {
...
+        values: entries,
+      };
+      this.pendingReconciliationAck.set(
+        `${batchId}#${peer.id}`,
+        offset + batchSize,
+      );
+      this.trySendToPeer(peer, msg);
+    };
+
+    const processStorageBatch = (offset: number) => {
```
gdorsi: Let's discuss observability with Gio.

packages/cojson/src/sync.ts
```diff
@@ -734,11 +910,77 @@ export class SyncManager {
...
+        coValue.getKnownStateFromStorage(maybeSendKnown);
+      }
+    }
+
+    if (msg.values.length === 0) {
+      this.trySendToPeer(peer, { action: "reconcile-ack", id: msg.id });
+    }
+  }
+
+  handleReconcileAck(msg: ReconcileAckMessage, peer: PeerState): void {
```
gdorsi, line 1011: We should add a comment that we accept drift by design

docs/sync-protocol.md
```diff
@@ -0,0 +1,108 @@
...
+- **Response**:
+  - CONTENT missing from the sender
+  - KNOWN state if the sender is not missing any content
+
+### KNOWN Message
+- **Purpose**: Share known state (acknowledgment or state update)
+- **Contains**: Current known state (header + session transaction counts)
+- **Variants**: 
+  - Normal: `{action: "known", ...}`
+  - CoValue dependency: `{action: "known", asDependencyOf: "co_z...", ...}`
```
gdorsi: This field is not used anymore

packages/cojson/src/sync.ts
```diff
@@ -281,15 +317,148 @@ export class SyncManager {
...
+            if (storageKnownState) {
+              entries.push([
+                coValue.id,
+                this.hashKnownStateSessions(storageKnownState.sessions),
+              ]);
+            }
+            sendReconcileMessageWhenDone();
+          });
+        }
+        if (pending.length === 0 && batch.length >= batchSize) {
```
gdorsi: Let's wait for the ack before processing new ids

packages/cojson/src/sync.ts
nrainhart: This is no longer true now that we wait for the previous reconciliation batch ack before sending a new one

packages/cojson/src/sync.ts
```diff
@@ -281,15 +319,165 @@ export class SyncManager {
...
+    };
+    while (!sendNextBatch()) {
+      await new Promise<void>((resolve) =>
+        setTimeout(
+          resolve,
+          STORAGE_RECONCILIATION_CONFIG.BATCH_ACK_POLLING_INTERVAL_MS,
+        ),
+      );
+    }
+  }
```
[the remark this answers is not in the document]
nrainhart: Fixed in [`53e7a6c` (this PR)](https://github.com/garden-co/jazz/pull/3422/commits/53e7a6cb52391a40cdb9ee283ba17ae80e1736b7)

gdorsi (approved):

PR #3427, opened 2026-02-04
perf: optimize FileStream.asBase64 and RawBinaryCoStream transaction processing

gdorsi:
This PR extracts `RawBinaryCoStream` into a standalone module and implement performance optimizations for `FileStream` operations, particularly transaction processing and `asBase64`.

**Benchmark results with toBase64 native support (5MB file):**
- `asBase64`: 732.39 op/sec vs  49.78 op/sec (**+1371.36% faster**)
- `write`: 12.53 op/sec vs 12.19 op/sec (+2.79%)
- `getChunks`: 695.03 op/sec vs 153.89 op/sec  (**+351.64% faster**)

**Benchmark results (5MB file):**
- `asBase64`: 118.62 op/sec vs 47.16 op/sec (**+151.50% faster**)
- `write`: 12.53 op/sec vs 12.19 op/sec (+2.79%)
- `getChunks`: 93.31 op/sec (unchanged)

### 1. Code Refactoring (`cojson`)

- **Extracted `RawBinaryCoStream`** from `coValues/coStream.ts` into a new standalone file `coValues/binaryCoStream.ts`
- **Optimized transactions processing**: We now focus on the RawBinaryCoStream data instead reusing RawCoStream processing

### 2. Base64 Optimization (`cojson`)

- **Optimized base64 encoding/decoding path** for binary stream operations, thanks to @Xelson suggestion

### 3. Inspector Improvements (`jazz-tools`)

- **Created dedicated `co-binary-stream-view.tsx`** component for visualizing binary streams
- **Refactored `co-stream-view.tsx`** to focus on generic streams, reducing complexity
- **Enhanced `use-resolve-covalue.ts`** to better distinguish binary vs. generic streams
- **Updated viewer page** to include the new binary stream viewer

### 4. Benchmarking

- **Added benchmarks for BinaryCoStream** in `bench/jazz-tools/binaryCoStream.jazz-tools.bench.ts`
  - Benchmarks write operations (100KB, 1MB, 5MB)
  - Benchmarks read operations (`getChunks`, `asBase64`)
  - Compares workspace version against published npm packages
  
- **Created benchmarking skill** (`.cursor/skills/benchmarking/SKILL.md`)
  - Provides guidance for writing performance benchmarks
  - Documents cronometro usage and conventions
  - Explains worker thread architecture and comparison strategies


.cursor/skills/benchmarking/SKILL.md
nrainhart, line 1: Nice! How good was the agent on generating the benchmark vs how much did you have to tune it afterwards?

bench/package.json
```diff
@@ -7,13 +7,16 @@
     "cojson": "workspace:*",
     "cojson-core-napi": "workspace:*",
     "cojson-core-wasm": "workspace:*",
+    "cojson-latest": "npm:cojson@0.20.7",
     "cronometro": "^5.3.0",
     "jazz-tools": "workspace:*",
+    "jazz-tools-latest": "npm:jazz-tools@0.20.7",
```
nrainhart: FYI we previously removed this because referencing a specific version of jazz caused issues with our release process: https://github.com/garden-co/jazz/pull/3255.

@antoniomuso do you have any additional context about this?

packages/jazz-tools/src/tools/coValues/coFeed.ts
```diff
@@ -837,22 +837,41 @@ export class FileStream extends CoValueBase implements CoValue {
...
+    let offset = 0;
+    for (const chunk of data.chunks) {
+      merged.set(chunk, offset);
+      offset += chunk.length;
     }
 
-    const base64 = btoa(output.join(""));
+    const CHUNK_SIZE = 32768;
+    const parts: string[] = [];
+    for (let i = 0; i < totalLen; i += CHUNK_SIZE) {
```
nrainhart: Can you leave a comment in the code explaining how merging the chunks and then re-chunking helps with performance? In particular, I don't get why re-chunking is necessary

nrainhart (commented):
This is amazing! Did a high-level pass over the changes and everything looks good. Will wait for the PR to be ready for review before approving

.cursor/skills/benchmarking/SKILL.md
gdorsi: I'm thinking that I still need to tune it because it's not normal that Writing 100KB takes the same op/sec as writing 5MB

packages/jazz-tools/src/tools/coValues/coFeed.ts
gdorsi: Done https://github.com/garden-co/jazz/pull/3427/changes/be9386cc6fa3ecf88821dc3055aa1778c872ba46

packages/jazz-tools/src/tools/coValues/coFeed.ts
```diff
@@ -950,22 +837,40 @@ export class FileStream extends CoValueBase implements CoValue {
...
+    // than building the string byte-by-byte (e.g., `result += String.fromCharCode(byte)`).
+    // Each string concatenation creates a new string object, making byte-by-byte O(n²).
+    // With apply, we convert many bytes at once, reducing string allocations.
+    //
+    // We limit batch size to 32KB because V8 has ~64k argument limit for function calls.
+    const CHUNK_SIZE = 32768;
+    const parts: string[] = [];
+
+    // Process each chunk directly without merging into a single buffer
+    for (const chunk of data.chunks) {
```
antoniomuso, line 968: What if we use [toBase64()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/toBase64)?

bench/package.json
gdorsi: Reverted.

When running the benchmark skill the Agent will install those, but we need to find a better solution to compare old versions.

packages/jazz-tools/src/tools/coValues/coFeed.ts
gdorsi: I tried that in https://github.com/garden-co/jazz/blob/main/packages/cojson/src/base64url.ts

It doesn't bring any performance benefits, and it's not widely supported.
So I reverted it back to our implementation.

I've also tried to port in Rust, but to make it faster than what we have we probably should pull the data directly from inside of Rust which makes the task too complicated.

gdorsi: Ok, found that the problem is that toBase64 is supported only from Node.js v25

Done the upgrade and the performance are around 10x better on the chunks -> base65 conversion

PR #3428, opened 2026-02-05
Global tx for local and sync

Takeno:
This PR adds _atomic transactions_ for CoValue mutations: multiple mutations can be executed inside a single `withTransaction()` callback and are then persisted together in one IndexedDB transaction and synced together in one network `BatchMessage`. Mutations are applied to memory immediately (optimistic); persistence and sync happen once the callback returns, with a single attempt for each. No automatic retry and no rollback of in-memory state.

This is primarily to protect against the client being closed in the middle of a sync/store operation.

It is based on already existing concepts of `DBTransactionInterfaceAsync`.

## DX
```ts
account.$jazz.unstable_withTransaction(() => {
  const map = TestMap.create({ name: "test", value: 0 });
  map.$jazz.set("name", "updated");
  map.$jazz.set("value", 42);
});
```

## Critical changes

- The [addition](https://github.com/garden-co/jazz/pull/3428/changes#diff-cb2c887ce03124928dac24533bd2da3b6163ea137372165eea913ea2dd0f93d2R153-R157) of `upsertCoValue` based on an existing transaction.
- The addition of a [Batch message](https://github.com/garden-co/jazz/pull/3428/changes#diff-c90e0211d2243ec0555557cabc377441df15b3b8ca4b76caa07882039f29da68R71-R74), which is sent to every [remote peer](https://github.com/garden-co/jazz/pull/3428/changes#diff-c90e0211d2243ec0555557cabc377441df15b3b8ca4b76caa07882039f29da68R1118-R1151). The batch message is handled like a simple list of messages. Its role is to ensure messages arrive together; we can't guarantee they are stored successfully. 

## Limitations
- If coValues inside `withTransaction` interact with coValues mutated outside, the indexeddb tx might fail because `upsertCoValue` looks inside the database and may not find the CoValue due to transaction concurrency.
- The `Batch` message's payload sent to the sync server might be bigger than transport limits for a single message.
- In case of failure on store/sync, the in-memory state is inconsistent. But this is already the current behaviour for every coValue
- Callbacks must be synchronous, to avoid nested transactions or endless promises.
- No ACK from sync server for batch messages: since every message in the batch is processed as usual, the server will follow the common flow after a `NewContent` message is received.

### DIFF
--- a/packages/cojson-storage-indexeddb/src/idbClient.ts
+++ b/packages/cojson-storage-indexeddb/src/idbClient.ts
@@ -52,6 +52,39 @@ export class IDBTransaction implements DBTransactionInterfaceAsync {
     );
   }
 
+  async upsertCoValue(
+    id: RawCoID,
+    header?: CojsonInternalTypes.CoValueHeader,
+  ): Promise<number | undefined> {
+    // If no header is provided, just look up the existing rowID within this transaction.
+    if (!header) {
+      const row: StoredCoValueRow | undefined = await this.run((tx) =>
+        tx.getObjectStore("coValues").index("coValuesById").get(id),
+      );
+
+      return row?.rowID;
+    }
+
+    try {
+      // Attempt to insert or update the coValue in the current transaction.
+      const row = await this.run<number>(
+        (tx) =>
+          tx
+            .getObjectStore("coValues")
+            .put({ id, header }) as IDBRequest<number>,
+      );
+
+      return row;
+    } catch (e) {
+      // On failure (e.g. uniqueness constraint), fall back to reading the existing row.
+      const row: StoredCoValueRow | undefined = await this.run((tx) =>
+        tx.getObjectStore("coValues").index("coValuesById").get(id),
+      );
+
+      return row?.rowID;
+    }
+  }
+
   async markCoValueAsDeleted(id: RawCoID): Promise<void> {
     await this.run((tx) =>
       tx.getObjectStore("deletedCoValues").put({
@@ -221,9 +254,6 @@ export class IDBTransaction implements DBTransactionInterfaceAsync {
 export class IDBClient implements DBClientInterfaceAsync {
   private db;
 
-  activeTransaction: CoJsonIDBTransaction | undefined;
-  autoBatchingTransaction: CoJsonIDBTransaction | undefined;
-
   constructor(db: IDBDatabase) {
     this.db = db;
   }
@@ -274,14 +304,7 @@ export class IDBClient implements DBClientInterfaceAsync {
     id: RawCoID,
     header?: CojsonInternalTypes.CoValueHeader,
   ): Promise<number | undefined> {
-    if (!header) {
-      return this.getCoValueRowID(id);
-    }
-
-    return putIndexedDbStore<CoValueRow, number>(this.db, "coValues", {
-      id,
-      header,
-    }).catch(() => this.getCoValueRowID(id));
+    return this.transaction(async (tx) => tx.upsertCoValue(id, header));
   }
 
   async getAllCoValuesWaitingForDelete(): Promise<RawCoID[]> {
@@ -296,17 +319,19 @@ export class IDBClient implements DBClientInterfaceAsync {
     return entries.map((e) => e.coValueID);
   }
 
-  async transaction(
-    operationsCallback: (tx: IDBTransaction) => Promise<unknown>,
+  async transaction<T>(
+    operationsCallback: (tx: IDBTransaction) => Promise<T>,
     storeNames?: StoreName[],
-  ) {
+  ): Promise<T> {
     const tx = new CoJsonIDBTransaction(this.db, storeNames);
 
     try {
-      await operationsCallback(new IDBTransaction(tx));
+      const result = await operationsCallback(new IDBTransaction(tx));
       tx.commit(); // Tells the browser to not wait for another possible request and commit the transaction immediately
+      return result;
     } catch (error) {
       tx.rollback();
+      throw error;
     }
   }
 
--- a/packages/cojson/src/storage/sqliteAsync/client.ts
+++ b/packages/cojson/src/storage/sqliteAsync/client.ts
@@ -320,13 +320,14 @@ export class SQLiteClientAsync implements DBClientInterfaceAsync {
     return rows.map((r) => r.id);
   }
 
-  async transaction(
-    operationsCallback: (tx: DBTransactionInterfaceAsync) => Promise<unknown>,
-  ): Promise<unknown> {
-    return this.enqueueTx(() =>
-      this.db.transaction((tx) =>
-        operationsCallback(new SQLiteTransactionAsync(tx)),
-      ),
+  async transaction<T>(
+    operationsCallback: (tx: DBTransactionInterfaceAsync) => Promise<T>,
+  ): Promise<T> {
+    return this.enqueueTx(
+      () =>
+        this.db.transaction((tx) =>
+          operationsCallback(new SQLiteTransactionAsync(tx)),
+        ) as Promise<T>,
     );
   }
 
--- a/packages/cojson/src/storage/sqliteAsync/client.ts
+++ b/packages/cojson/src/storage/sqliteAsync/client.ts
@@ -44,6 +44,36 @@ export function getErrorMessage(error: unknown) {
 export class SQLiteTransactionAsync implements DBTransactionInterfaceAsync {
   constructor(private readonly tx: SQLiteDatabaseDriverAsync) {}
 
+  async upsertCoValue(
+    id: RawCoID,
+    header?: CoValueHeader,
+  ): Promise<number | undefined> {
+    if (!header) {
+      const row = await this.tx.get<{ rowID: number }>(
+        "SELECT rowID FROM coValues WHERE id = ?",
+        [id],
+      );
+      return row?.rowID;
+    }
+
+    const result = await this.tx.get<{ rowID: number }>(
+      `INSERT INTO coValues (id, header) VALUES (?, ?)
+       ON CONFLICT(id) DO NOTHING
+       RETURNING rowID`,
+      [id, JSON.stringify(header)],
+    );
+
+    if (!result) {
+      const row = await this.tx.get<{ rowID: number }>(
+        "SELECT rowID FROM coValues WHERE id = ?",
+        [id],
+      );
+      return row?.rowID;
+    }
+
+    return result.rowID;
+  }
+
   async getSingleCoValueSession(
     coValueRowId: number,
     sessionID: SessionID,
@@ -251,8 +281,8 @@ export class SQLiteClientAsync implements DBClientInterfaceAsync {
     }
 
     const result = await this.db.get<{ rowID: number }>(
-      `INSERT INTO coValues (id, header) VALUES (?, ?) 
-       ON CONFLICT(id) DO NOTHING 
+      `INSERT INTO coValues (id, header) VALUES (?, ?)
+       ON CONFLICT(id) DO NOTHING
        RETURNING rowID`,
       [id, JSON.stringify(header)],
     );
--- a/packages/cojson/src/storage/storageAsync.ts
+++ b/packages/cojson/src/storage/storageAsync.ts
@@ -313,6 +313,7 @@ export class StorageApiAsync implements StorageAPI {
   private async handleCorrection(
     knownState: CoValueKnownState,
     correctionCallback: CorrectionCallback,
+    tx: DBTransactionInterfaceAsync | undefined,
   ) {
     const correction = correctionCallback(knownState);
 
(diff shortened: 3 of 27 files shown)


packages/jazz-tools/src/tools/tests/account.test.ts
```diff
@@ -527,3 +527,185 @@ describe("createAs", () => {
...
+
+    // Wait for sync to complete
+    await clientAccount.$jazz.waitForAllCoValuesSync({ timeout: 1000 });
+
+    // Verify both mutations were synced to server
+    const loadedMap1 = await serverNode.load(map1.$jazz.raw.id);
+    const loadedMap2 = await serverNode.load(map2.$jazz.raw.id);
+
+    expect(loadedMap1).not.toBe(CoValueLoadingState.UNAVAILABLE);
+    expect(loadedMap2).not.toBe(CoValueLoadingState.UNAVAILABLE);
```
nrainhart: We could actually check the maps' content is correct

packages/cojson/src/tests/localNode.transactions.test.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    try {
+      await node.unstable_withTransaction(() => {
+        map.set("key1", "value1", "trusting");
+        expect(map.get("key1")).toBe("value1");
+
+        throw new Error("Test error");
+
+        // This line won't be reached
+        // map.set("key2", "value2", "trusting");
```
nrainhart: Should this line be commented out? Or was it causing issues with lint?
```suggestion
        map.set("key2", "value2", "trusting");
```

packages/cojson/src/transactionContext.ts
```diff
@@ -0,0 +1,69 @@
...
+    return this.pendingMessages;
+  }
+
+  /**
+   * Check if this transaction context is active.
+   * A TransactionContext is always considered active when it exists.
+   */
+  isActive(): boolean {
+    return true;
+  }
```
nrainhart: Nit: I'd just remove this method, as checking for the existence of a transaction context is enough to know if we're in a transaction context

packages/cojson/src/sync.ts
```diff
@@ -1067,13 +1083,86 @@ export class SyncManager {
...
+    // Track sync state for all CoValues in the batch
+    for (const coValueId of coValueIds) {
+      this.trackSyncState(coValueId);
+    }
+
+    // Send to all server peers for each CoValue
+    const sentToPeers = new Set<PeerID>();
+    for (const coValueId of coValueIds) {
+      for (const peer of this.getServerPeers(coValueId)) {
+        if (!sentToPeers.has(peer.id) && !peer.closed) {
```
nrainhart: I don't understand why `sentToPeers` is necessary. Do we expect getServerPeers to return duplicate peer ids?

packages/cojson/src/sync.ts
```diff
@@ -1067,13 +1083,86 @@ export class SyncManager {
...
+      ? this.local.storage.storeAtomicBatch(messages)
+      : Promise.resolve(undefined);
+
+    const syncPromise = this.syncBatchToServers(messages, coValueIds);
+
+    // Wait for both to complete
+    await Promise.allSettled([storagePromise, syncPromise]);
+  }
+
+  private async syncBatchToServers(
```
nrainhart: It's very important we keep this method's implementation consistent with `syncContent`. I'd add a comment in both places to remind folks of this

packages/cojson/src/sync.ts
```diff
@@ -1067,13 +1083,86 @@ export class SyncManager {
...
+    }
+
+    const coValueIds = new Set(messages.map((m) => m.id));
+
+    // Start both operations in parallel (single attempt for storage and network)
+    const storagePromise = this.local.storage
+      ? this.local.storage.storeAtomicBatch(messages)
+      : Promise.resolve(undefined);
+
+    const syncPromise = this.syncBatchToServers(messages, coValueIds);
```
nrainhart: ⚠️ Why does this return a promise? We're not waiting for sync to complete, and all the code we're running is sync

packages/cojson-storage-indexeddb/src/idbClient.ts
```diff
@@ -274,14 +304,7 @@ export class IDBClient implements DBClientInterfaceAsync {
...
   ): Promise<number | undefined> {
-    if (!header) {
-      return this.getCoValueRowID(id);
-    }
-
-    return putIndexedDbStore<CoValueRow, number>(this.db, "coValues", {
-      id,
-      header,
-    }).catch(() => this.getCoValueRowID(id));
+    return this.transaction(async (tx) => tx.upsertCoValue(id, header));
```
nrainhart, line 326: This is adding a new transaction on a relatively hot path, could you check if this introduces any noticeable performance degradation when doing lots of writes?

packages/cojson/src/storage/sqliteAsync/client.ts
```diff
@@ -44,6 +44,36 @@ export function getErrorMessage(error: unknown) {
 export class SQLiteTransactionAsync implements DBTransactionInterfaceAsync {
   constructor(private readonly tx: SQLiteDatabaseDriverAsync) {}
 
+  async upsertCoValue(
```
nrainhart, line 50: I don't love that we're duplicating upsertCoValue in the DB client and transaction APIs. I think we need to rethink this DB client/transaction split, because in many cases we want to run queries both inside and outside a transaction, and this leads us to either code duplication or creating unnecessary transactions (see also my previous comment) cc @gdorsi 

We don't need to fix this in this PR, thought. Maybe let's just add a comment to keep both implementations from drifting apart

packages/cojson/src/storage/storageAsync.ts
nrainhart, line 1: ⚠️ We should add tests for `storeAtomicBatch` (also for IndexedDB)

packages/cojson/src/storage/storageAsync.ts
```diff
@@ -370,37 +370,20 @@ export class StorageApiAsync implements StorageAPI {
 
     for (const sessionID of Object.keys(msg.new) as SessionID[]) {
       await this.dbClient.transaction(async (tx) => {
-        const sessionRow = await tx.getSingleCoValueSession(
-          storedCoValueRowID,
+        const sessionHadInvalidAssumptions = await this.applyMessageForSession(
```
nrainhart:
```suggestion
        const sessionHadInvalidAssumptions = await this.applyContentFromSession(
```

packages/cojson/src/storage/storageAsync.ts
```diff
@@ -531,6 +574,71 @@ export class StorageApiAsync implements StorageAPI {
...
+    await this.dbClient.transaction(async (tx) => {
+      for (const msg of messages) {
+        await this.storeMessageInTransaction(tx, msg);
+      }
+    });
+
+    // Update known states after successful commit
+    for (const msg of messages) {
+      this.inMemoryCoValues.add(msg.id);
+    }
```
nrainhart: I'd put this inside `storeMessageInTransaction`

packages/cojson/src/storage/storageAsync.ts
```diff
@@ -416,6 +399,66 @@ export class StorageApiAsync implements StorageAPI {
...
   }
 
+  /**
+   * Apply a single message to storage for a specific session within an existing
+   * transaction.
+   *
+   * @returns true if this session had invalid assumptions (only when
+   *          allowInvalidAssumptions is true), otherwise false.
+   */
+  private async applyMessageForSession(
```
nrainhart: This split is kind of awkward (this function has lots or arguments, and not a clearly defined scope). Do you think we could make `storeSingle` use `storeMessageInTransaction` instead? That'd also avoid some code duplication we still have

nrainhart (commented):
The new API looks good! I feel like the final implementation should provide stronger guarantees, such as:
- roll back in-memory changes if an error is thrown during the callback execution
- store batch transactionally on the server
- (optionally) waiting for confirmation from the server before resolving the promise? 

But we can start trying out this version now.

As a sidenote, I wonder if using `withTransaction` actually introduces any noticeable performance improvements when doing several operations at the same time. Would it make sense to use it in methods like `applyDiff` and `CoList.push`?

packages/cojson/src/sync.ts
Takeno: Without sentToPeers, if a peer appears in both getServerPeers(coValueIdA) and getServerPeers(coValueIdB), it would receive the same batchMessage multiple times in this single call.
I add an additional comment in the code.

packages/cojson/src/sync.ts
Takeno: I refactored syncContent to handle BatchMessage too, so there's a single implementation.

Please check 53c08c7acbfec5e22c2493cac3a34958c22d1492

packages/cojson/src/storage/storageAsync.ts
Takeno: I put it outside to store them only if the full transaction resolves successfully. If I put them inside and a later `storeMessageInTransaction` fails, the state would be inconsistent

packages/cojson/src/storage/storageAsync.ts
Takeno: Refactored both methods to have a single implementation. Check the bf28b34335250e168da9f021b052980e60aa472e for the files, then probably it's better to see the final diff main/this branch [here](https://github.com/garden-co/jazz/pull/3428/changes#diff-23c27edc552030839508cb69e98970fd215113c741649b31f902b70ae2e3d4a6R350-R438).

Currently, we still use a single transaction for each msg/session.

Takeno:
> The new API looks good! I feel like the final implementation should provide stronger guarantees, such as:
> 
>     * roll back in-memory changes if an error is thrown during the callback execution
> 
>     * store batch transactionally on the server
> 
>     * (optionally) waiting for confirmation from the server before resolving the promise?
> 
> 
> But we can start trying out this version now.

* 1 Rollback in-memory changes is something much harder with the current paradigm, since the transactions stored/synced are based on the diffs with in-memory. Currently, normal mutations also have no rollbacks if anything fails.
Should we rename `withTransaction` to something more like `batchUpdates`?

* 2/3 The main purpose is to avoid partial storage for deeply coupled mutations. Once it is successfully stored locally, we can rely on the background syncing in case the sync didn't go through.
Would `Promise.race` be better than `Promise.allSettled` in this case? 


> 
> As a sidenote, I wonder if using `withTransaction` actually introduces any noticeable performance improvements when doing several operations at the same time. Would it make sense to use it in methods like `applyDiff` and `CoList.push`?

I observed some performance improvements in browser-perf-tests, creating 50x50 grids, from ~2300ms to ~1900ms.
I would start releasing this under `unstable_` prefix, gather feedback, and later opt-in in some internals like applyDiff / push.

packages/cojson/src/sync.ts
```diff
@@ -1351,40 +1367,90 @@ export class SyncManager {
...
+    }
 
+    for (const peer of peersToSync) {
       // We assume that the peer already knows anything before this content
       // Any eventual reconciliation will be handled through the known state messages exchange
+
       this.trySendToPeer(peer, content);
-      peer.combineOptimisticWith(coValue.id, contentKnownState);
-      peer.trackToldKnownState(coValue.id);
+      for (const content of contents) {
```
nrainhart, line 1450: It's a bit confusing that the two `content` variables here refer to different things. We could rename the param to `contentMsg`

packages/cojson/src/sync.ts
```diff
@@ -1351,40 +1367,90 @@ export class SyncManager {
...
+    }
 
-    this.trackSyncState(coValue.id);
+    // Start both operations in parallel (single attempt for storage and network)
+    const storagePromise = this.local.storage
+      ? this.local.storage.storeAtomicBatch(messages)
+      : Promise.resolve(undefined);
 
-    const contentKnownState = knownStateFromContent(content);
+    const syncPromise = this.syncContent({
```
nrainhart, line 1394: https://github.com/garden-co/jazz/pull/3428#discussion_r2812362949

packages/cojson/src/tests/StorageApiSync.test.ts
nrainhart, line 1: Is there any reason not to have the same test for StorageApiAsync?

packages/cojson/src/storage/storageAsync.ts
```diff
@@ -567,6 +588,31 @@ export class StorageApiAsync implements StorageAPI {
...
+    await this.dbClient.transaction(async (tx) => {
+      for (const msg of messages) {
+        await this.storeSingle(msg, () => undefined, tx);
+      }
+    });
+
+    // // Update known states after successful commit
+    // for (const msg of messages) {
+    //   this.inMemoryCoValues.add(msg.id);
+    // }
```
nrainhart: Nit: remove
```suggestion
```

packages/cojson/src/storage/storageAsync.ts
```diff
@@ -345,23 +350,39 @@ export class StorageApiAsync implements StorageAPI {
...
+    ): Promise<T> => {
+      if (tx) {
+        return callback(tx);
+      }
+
+      return new Promise((resolve, reject) => {
+        this.dbClient.transaction((tx) => {
+          return callback(tx).then(resolve, reject);
+        });
+      });
```
nrainhart, line 368: ⚠️ This will create several promises and cojson transactions if a tx is not provided, which could impact performance.

Can we always pass a transaction instead? That way for regular stores we use one tx per store operation

nrainhart (commented):
Leaving a few more comments, but only one's important.

> Should we rename withTransaction to something more like batchUpdates

Yeah, this actually may be better to avoid setting incorrect expectations. Wanna discuss this with the rest of the team?

>  I would start releasing this under unstable_ prefix, gather feedback, and later opt-in in some internals like applyDiff / push.

Sounds good!

PR #3429, opened 2026-02-05
fix(cojson): improve transaction comparison for deterministic ordering

gdorsi:
Previously, `compareTransactions` returned `0` when transactions had equal timestamps but different sessionIDs, leading to non-deterministic ordering. This PR adds a tie-breaker that compares sessionIDs from the last character, providing a deterministic ordering

### Changes

- Added reverse character comparison for sessionIDs in `compareTransactions`
- Added benchmark to validate the most performant solution between reverse charCodeAt comparison and localeCompare

### Benchmark Results

Sorting 100 transactions with equal timestamps (worst case scenario where all comparisons go through the sessionID tie-breaker):

| Method | ops/sec | Performance |
|--------|---------|-------------|
| **reverse charCodeAt** | **104,261** | **2.5x faster** |
| localeCompare | 41,268 | baseline |

The reverse comparison is faster because sessionIDs have the structure:
```
co_zUwh2n1dhsCxZzZyfhsmCECSfZM_session_zGeHDNSz64X8
└─────────── 30 char prefix ───────────┘└─ 21 char suffix ─┘
```

By comparing from the end, we immediately hit the random suffix where differences occur, avoiding comparison of the shared prefix.


antoniomuso (approved):
LGTM! Nice Idea!

PR #3431, opened 2026-02-05
fix: prevent issues with concurrent async SQLite transactions

nrainhart:
## Description

While working on https://github.com/garden-co/jazz/pull/3422, I noticed that DB operations inside interleaved async js tasks could end up inside the same transaction when using the async SQLite client. I solved that by modifying `SQLiteClientAsync.transaction` to pass a tx object to its callback instead of performing all the operation on the same DB client, and splitting the Client and Transaction APIs. Under the hood, different clients now handle concurrent txs differently:
- `op-sqlite` serializes all transactions: https://github.com/OP-Engineering/op-sqlite/blob/main/src/functions.ts#L352
- `expo-sqlite` creates a new connection to avoid concurrent js tasks from submitting operations to the same client: https://docs.expo.dev/versions/latest/sdk/sqlite/#withexclusivetransactionasynctask
  - it does not prevent concurrent txs though, so running into SQLITE_BUSY errors is still possible (1)
- `libsql` does not solve this (2). I tried switching to https://github.com/tursodatabase/libsql-client-ts, but that turned out to be a rabbit hole, so I'm keeping this one as-is (considering we use it only for testing)

Due to (1) and (2) still making it possible to run into SQLITE_BUSY errors, I also modified `SQLiteClientAsync.transaction` to serialize transactions.

I also made some improvements to DB cleanup in tests:
- prevented stores from being closed before the node is shut down
- `-wal` and `-shm` SQLite files were not being cleaned up 

---

> [!NOTE]
> **Medium Risk**
> Touches core persistence transaction boundaries and driver adapters; serialization and tx-scoping could change concurrency behavior/performance and may surface driver-specific transaction edge cases.
> 
> **Overview**
> Prevents async JS tasks from accidentally sharing a single SQLite transaction by splitting `SQLiteClientAsync` into a client + `SQLiteTransactionAsync`, and changing `SQLiteDatabaseDriverAsync.transaction` to pass a tx-scoped driver into the callback.
> 
> Adds a per-client transaction queue to **serialize concurrent async transactions** (reducing `SQLITE_BUSY`/interleaving issues), updates Expo/OP-SQLite adapters to provide tx-specific adapters (using Expo’s exclusive transactions and OP-SQLite’s transaction object), and introduces a focused regression test for concurrent `transaction()` calls.
> 
> Improves test DB lifecycle cleanup by deferring storage close/delete until after node shutdown hooks and by also removing `-wal`/`-shm` files; test code is updated accordingly (including `Promise.withResolvers` usage).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b38a52649bfb4c131100b3774aef2fee47974567. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/jazz-tools/src/expo/storage/expo-sqlite-adapter.ts
+++ b/packages/jazz-tools/src/expo/storage/expo-sqlite-adapter.ts
@@ -6,6 +6,12 @@ export class ExpoSQLiteAdapter implements SQLiteDatabaseDriverAsync {
   private db: SQLiteDatabase | null = null;
   private dbName: string;
 
+  static withDB(db: SQLiteDatabase): ExpoSQLiteAdapter {
+    const adapter = new ExpoSQLiteAdapter();
+    adapter.db = db;
+    return adapter;
+  }
+
   public constructor(dbName: string = "jazz-storage") {
     this.dbName = dbName;
   }
@@ -56,13 +62,13 @@ export class ExpoSQLiteAdapter implements SQLiteDatabaseDriverAsync {
     await this.db.runAsync(sql, params?.map((p) => p as SQLiteBindValue) ?? []);
   }
 
-  public async transaction(callback: () => unknown) {
+  public async transaction(callback: (tx: ExpoSQLiteAdapter) => unknown) {
     if (!this.db) {
       throw new Error("Database not initialized");
     }
 
-    await this.db.withTransactionAsync(async () => {
-      await callback();
+    await this.db.withExclusiveTransactionAsync(async (tx) => {
+      await callback(ExpoSQLiteAdapter.withDB(tx));
     });
   }
 
--- a/.changeset/fruity-words-worry.md
+++ b/.changeset/fruity-words-worry.md
@@ -1,0 +1,6 @@
+---
+"jazz-tools": patch
+"cojson": patch
+---
+
+fix: prevent conflicts between concurrent async SQLite transactions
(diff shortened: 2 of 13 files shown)


gdorsi (approved):

packages/jazz-tools/src/expo/storage/expo-sqlite-adapter.ts
```diff
@@ -56,13 +62,13 @@ export class ExpoSQLiteAdapter implements SQLiteDatabaseDriverAsync {
...
-  public async transaction(callback: () => unknown) {
+  public async transaction(callback: (tx: ExpoSQLiteAdapter) => unknown) {
     if (!this.db) {
       throw new Error("Database not initialized");
     }
 
-    await this.db.withTransactionAsync(async () => {
-      await callback();
+    await this.db.withExclusiveTransactionAsync(async (tx) => {
+      await callback(ExpoSQLiteAdapter.withDB(tx));
```
gdorsi, line 71: Is the switch to withExclusiveTransactionAsync correct?

nrainhart: Yes, the combination of `withExclusiveTransactionAsync` & transaction serialization prevents both potential issues:
- interleaved db calls from ending up in the same transaction
- concurrent transactions from causing SQLITE_BUSY errors

PR #3432, opened 2026-02-05
feat: add group-level asymmetric key for key revelation

gdorsi:
This commit introduces a new design for group asymmetric keys, allowing non-members to reveal child group keys to parent groups without the need for per-member writeOnly keys.


aeplay (approved):
draft looks good.

Private key derived from readKey is elegant, but this means we need to update groupSealer when we rotate readKey and potentially deal with multiple sealers over the history of it

We could avoid introducing an anonymous box, since we know/can look up the sender id by looking at who made the revelation, but this is more complex

packages/cojson-storage-do-sqlite/src/tests/dosqlite.test.ts
```diff
@@ -64,19 +64,19 @@ describe("Cloudflare DurableObject SQL storage", () => {
...
-        [
-          "client -> CONTENT Group header: true new: After: 0 New: 3",
-          "client -> CONTENT Map header: true new: After: 0 New: 1",
-        ]
-      `);
+      [
+        "client -> CONTENT Group header: true new: After: 0 New: 4",
+        "client -> CONTENT Map header: true new: After: 0 New: 1",
+      ]
+    `);
```
gdorsi, line 71: With groupSealer we are adding a transaction to the group initialization

crates/cojson-core/src/crypto/seal.rs
```diff
@@ -98,6 +98,89 @@ pub fn unseal(
     decrypt_xsalsa20_poly1305(&shared_secret, &nonce, sealed_message)
 }
 
+/// Seal a message for a group using anonymous box pattern.
+/// The sender generates an ephemeral key pair, so no sender authentication is provided.
+/// The ephemeral public key is prepended to the ciphertext.
```
aeplay, line 103: nice, this is exactly how libsodium boxes operate

packages/cojson/src/coValues/group.ts
```diff
@@ -1143,17 +1273,45 @@ export class RawGroup<
...
+              parentGroupSealer,
+              keyID,
+              secret,
+            );
+          }
+        }
+        return;
+      } else {
+        // LEGACY FALLBACK: Create a writeOnly key in the parent group
+        parent.internalCreateWriteOnlyKeyForMember(
```
aeplay, line 1365: so this means that only groups that have a sealer benefit from this, but the assumption is that they will soon be backfilled? What exactly triggers this?

packages/cojson/src/crypto/crypto.ts
```diff
@@ -205,6 +207,39 @@ export abstract class CryptoProvider<Blake3State = any> {
...
+  // This ensures concurrent migrations by different admins produce the same result
+  groupSealerFromReadKey(readKeySecret: KeySecret): {
+    id: SealerID;
+    secret: SealerSecret;
+  } {
+    const sealerBytes = this.blake3HashOnceWithContext(
+      textEncoder.encode(readKeySecret),
+      { context: textEncoder.encode("groupSealer") },
+    );
+    const secret: SealerSecret = `sealerSecret_z${base58.encode(sealerBytes)}`;
```
aeplay, line 226: some assertion that the blake3 output is exactly as long as a secret key would be nice here

aeplay (approved):

packages/cojson/src/coValues/group.ts
gdorsi: Old groups are migrated on-load if the account has "manager" or "admin" rights in the group: https://github.com/garden-co/jazz/pull/3432/changes#diff-54b8bc4665484baffa43e423c89921f5267435e1a01d12e7f313ec959e3cc362R143

packages/cojson/src/crypto/crypto.ts
gdorsi: Is this what you mean? https://github.com/garden-co/jazz/pull/3432/commits/13fc4c505cc02bda8b986af8ae95b098acfe7dbf

packages/cojson/src/coValues/group.ts
```diff
@@ -124,6 +133,41 @@ function healMissingKeyForEveryone(group: RawGroup) {
...
+  }
+
+  const readKeySecret = group.getReadKey(readKeyId);
+  if (!readKeySecret) {
+    return;
+  }
+
+  const groupSealer =
+    group.core.node.crypto.groupSealerFromReadKey(readKeySecret);
+  group.set("groupSealer", groupSealer.id, "trusting");
```
Takeno: `.id` -> `.publicKey`

packages/cojson/src/coValues/group.ts
```diff
@@ -893,6 +937,75 @@ export class RawGroup<
...
+      const sealerTime = sealerEntry.madeAt;
+      let readKeyID: KeyID | undefined;
+      for (let j = readKeyEntries.length - 1; j >= 0; j--) {
+        const readKeyEntry = readKeyEntries[j]!;
+        if (readKeyEntry.change.op !== "set") continue;
+        if (readKeyEntry.madeAt <= sealerTime) {
+          readKeyID = readKeyEntry.change.value as KeyID;
+          break;
+        }
+      }
```
Takeno: Possible conflict: a new readKey has been generated by rotation in a client, while another client generates the group sealer. Merging transactions will retrieve the last readKey, not the one used for groupSealer.

packages/cojson/src/tests/groupSealer.test.ts
```diff
@@ -0,0 +1,1338 @@
...
+    const group = node1.node.createGroup();
+
+    const groupSealerSecret = group.getGroupSealerSecret();
+    expect(groupSealerSecret).toBeDefined();
+
+    const groupSealerID = group.get("groupSealer");
+    expect(groupSealerID).toBeDefined();
+
+    // Verify the secret corresponds to the public ID
+    if (groupSealerSecret) {
```
Takeno: why this if in tests?

packages/cojson/src/tests/groupSealer.test.ts
```diff
@@ -0,0 +1,1338 @@
...
+    expect(sealerFromNode2).toBeDefined();
+
+    await groupOnNode2.core.waitForSync();
+
+    // Node3 loads - migration should be idempotent (groupSealer already set via sync or re-derived)
+    const groupOnNode3 = await loadCoValueOrFail(node3.node, legacyGroup.id);
+    const sealerFromNode3 = groupOnNode3.get("groupSealer");
+    expect(sealerFromNode3).toBeDefined();
+
+    // Both should produce the same groupSealer (deterministic from readKey)
```
Takeno: Here, groupOnNode3 has already been migrated. Check the number of transactions to verify that a new migration is not applied.

PR #3446, opened 2026-02-09
perf(subscription): delay value content parsing to when it's fully downloaded

gdorsi:
In the dashboard we've found that the main perf issue is caused by the group content invalidation during the streaming.

Screenshot 2026-02-09 at 15 05 44
Screenshot 2026-02-09 at 15 05 56

We don't need to parse the content immediately, and delaying the parsing to when the value is fully completed should fix the biggest issue we have there.


nrainhart (approved):
:shipit:

PR #3447, opened 2026-02-09
fix: avoided useless knownState calls

antoniomuso:
This PR avoids useless knownState calls.

### DIFF
--- a/packages/cojson/src/coValueCore/verifiedState.ts
+++ b/packages/cojson/src/coValueCore/verifiedState.ts
@@ -29,7 +29,6 @@ import {
   CoValueKnownState,
   getKnownStateToSend,
   KnownStateSessions,
-  cloneKnownState,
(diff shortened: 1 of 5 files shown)


packages/cojson/src/coValueCore/verifiedState.ts
```diff
@@ -253,63 +255,6 @@ export class VerifiedState {
     return sessionLog;
   }
 
-  clone(): VerifiedState {
```
gdorsi, line 256: Is this method unused?

gdorsi (approved):
Sweet!

packages/cojson/src/coValueCore/verifiedState.ts
antoniomuso: yeah it is not used.

PR #3455, opened 2026-02-11
feat: remove legacy coField schema

gdorsi:
This PR gets rid of the `_schema` based descriptors, moving the descriptors mapping in the co definers side.


packages/jazz-tools/src/tools/coValues/account.ts
```diff
@@ -498,13 +503,29 @@ class AccountJazzApi<A extends Account> extends CoValueJazzApi<A> {
...
+    const accountSchema = this.coValueSchema;
+
+    if (
+      accountSchema.builtin !== "Account" &&
+      accountSchema.builtin !== "CoMap"
+    ) {
+      throw new Error(
+        `[schema-invariant] ${this.account.constructor.name || "Account"}.resolve expected Account/CoMap schema, got ${accountSchema.builtin}.`,
+      );
+    }
```
Takeno: What's the case to have an `accountSchema` different from `Account`?

packages/jazz-tools/src/tools/coValues/coFeed.ts
```diff
@@ -108,31 +110,9 @@ export { CoFeed as CoStream };
...
-    const cls = class CoFeedOf extends CoFeed<Item> {
-      [coField.items] = item;
-    };
-
-    cls._schema ||= {};
-    cls._schema[ItemsSym] = (item as any)[SchemaInit];
-
-    return cls;
-  }
-
```
Takeno, line 135: Should we mark it as deprecated before removing it? It's not mentioned in the doc, but it is in the upgrade guides for 0.12 and 0.14 for CoLists.

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -81,38 +77,10 @@ export class CoList<out Item = any>
...
-   *
-   * @category Declaration
-   */
-  static Of<Item>(item: Item): typeof CoList<Item> {
-    // TODO: cache superclass for item class
-    return class CoListOf extends CoList<Item> {
-      [coField.items] = item;
-    };
-  }
-
```
Takeno, line 108: Should we mark it as deprecated before removing it? It's not mentioned in the doc, but it is in the upgrade guides for 0.12 and 0.14 for CoLists.

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -586,7 +545,7 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
   constructor(
     private coList: L,
     private getRaw: () => RawCoList,
-    private coListSchema?: CoreCoValueSchema,
+    private coListSchema?: CoreCoListSchema,
```
Takeno:
```suggestion
    private coListSchema: CoreCoListSchema,
```

packages/jazz-tools/src/tools/coValues/coList.ts
```diff
@@ -1001,8 +964,20 @@ export class CoListJazzApi<L extends CoList> extends CoValueJazzApi<L> {
...
+      throw new Error(
+        `[schema-invariant] ${this.coList.constructor.name || "CoList"} is missing coValueSchema.`,
+      );
+    }
+
+    if (this.coListSchema.builtin !== "CoList") {
+      throw new Error(
+        `[schema-invariant] ${this.coList.constructor.name || "CoList"}.resolve expected CoList schema, got ${this.coListSchema.builtin}.`,
+      );
+    }
```
Takeno:
```suggestion

```

Not used anymore, with an always-defined schema

packages/jazz-tools/src/tools/coValues/coMap.ts
```diff
@@ -873,11 +837,29 @@ class CoMapJazzApi<M extends CoMap> extends CoValueJazzApi<M> {
...
+    }
+
+    if (
+      this.coMapSchema.builtin !== "CoMap" &&
+      this.coMapSchema.builtin !== "Account"
+    ) {
+      throw new Error(
+        `[schema-invariant] ${this.coMap.constructor.name || "CoMap"}.resolve expected CoMap/Account schema, got ${this.coMapSchema.builtin}.`,
+      );
+    }
```
Takeno: Can an instance of coMap exist from the Account schema?

packages/jazz-tools/src/tools/coValues/schemaUnion.ts
```diff
@@ -35,99 +35,9 @@ export type SchemaUnionDiscriminator<V extends CoValue> = (discriminable: {
...
-        this: CoValueClass<T> & CoValueFromRaw<T>,
-        raw: T["$jazz"]["raw"],
-      ): T {
-        const ResolvedClass = discriminator(
-          raw as RawCoMap,
-        ) as unknown as CoValueClass<T> & CoValueFromRaw<T>;
-        return ResolvedClass.fromRaw(raw);
-      }
-    } as unknown as SchemaUnionConcreteSubclass<V>;
-  }
```
Takeno, line 129: Should we mark it as deprecated before removing it? It's not mentioned in the doc, but it is in an old changelog (0.8.23)

Takeno (approved):

packages/jazz-tools/src/tools/coValues/account.ts
```diff
@@ -58,6 +60,9 @@ import {
   InstanceOfSchemaCoValuesMaybeLoaded,
   LoadedAndRequired,
 } from "../internal.js";
+import type { CoreAccountSchema } from "../implementation/zodSchema/schemaTypes/AccountSchema.js";
+import type { AccountSchema as HydratedAccountSchema } from "../implementation/zodSchema/schemaTypes/AccountSchema.js";
```
nrainhart: Is this rename necessary?

packages/jazz-tools/src/tools/coValues/account.ts
gdorsi: Removed by applying your suggestion from https://github.com/garden-co/jazz/pull/3454/changes/b08126095ce38dd2a14c6b43cb66a2d26ce9a81b

packages/jazz-tools/src/tools/coValues/coMap.ts
```diff
@@ -873,11 +835,29 @@ class CoMapJazzApi<M extends CoMap> extends CoValueJazzApi<M> {
...
+    const descriptorsSchema = this.coMapSchema.getDescriptorsSchema();
+    const descriptor =
+      descriptorsSchema.shape[key] ?? descriptorsSchema.catchall;
+    if (descriptor) {
+      this.descriptorCache.set(key, descriptor);
+      return descriptor;
+    }
+
+    this.descriptorCache.set(key, undefined);
+    return undefined;
```
nrainhart:
```suggestion
    this.descriptorCache.set(key, descriptor);
    return descriptor;
```

packages/jazz-tools/src/tools/coValues/coFeed.ts
gdorsi: I think it's safe to remove, I can't think of any good reason for people to stick with this API.

packages/jazz-tools/src/tools/coValues/coMap.ts
gdorsi: Removed!

gdorsi:
@Takeno 

On the remove of the Class.Of static methods.

These would be helpful only if people were still using class-based schemas, which have been deprecated since we moved to Zod schemas.

I think it's ok to remove them now.

PR #3458, opened 2026-02-13
Autogenerate BETTER_AUTH_SECRET

joeinnes:
Currently, building the monorepo fails because the `betterauth-svelte` example imports `BETTER_AUTH_SECRET` from `$env/static/private`, which requires a `.env` file to exist. Since `.env` is gitignored, fresh clones and CI always hit this error:

```
Error: Module '"$env/static/private"' has no exported member 'BETTER_AUTH_SECRET'.
```

This PR adds a small Node script (`scripts/ensure-env.js`) that runs before `build` and `dev`. It checks for a `.env` file and, if one doesn't exist, generates it with:

- `BETTER_AUTH_SECRET` — a random 32-byte hex string
- `BETTER_AUTH_URL` — defaulting to `http://localhost:5173`

If a `.env` file already exists, then it will not be overwritten.

### DIFF
--- a/examples/betterauth-svelte/package.json
+++ b/examples/betterauth-svelte/package.json
@@ -5,8 +5,9 @@
   "type": "module",
   "packageManager": "pnpm@10.16.1",
   "scripts": {
-    "dev": "vite dev",
-    "build": "pnpm check && vite build",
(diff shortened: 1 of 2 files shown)


nrainhart (approved):
Thanks for adding this Joe!

PR #3460, opened 2026-02-13
Schema validation with superRefine

Takeno:
While exploring the possibility of using zod transforms during schema validation, I encountered a case not handled by the validation `z.instanceOf(CoMap).or(plainSchema)`.

If the data is a coValue but not a CoMap, it was validated against the plain schema. The result wasn't wrong: the validation threw an error, but it made no sense to validate the `coValue.toJSON`.

This PR replaces the `z.instanceOf(CoMap).or(plainSchema)` pattern with `z.unknown().superRefine` and uses zod metadata to store a reference to the plainSchema. Same output, better readability.

### DIFF
--- a/packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoFeedSchema.ts
+++ b/packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoFeedSchema.ts
@@ -71,11 +71,7 @@ export class CoFeedSchema<
       generateValidationSchemaFromItem(this.element),
     );
 
-    this.#validationSchema = coValueValidationSchema(
-      validationSchema,
-      CoFeed,
-      "CoFeed",
-    );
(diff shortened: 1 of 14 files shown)


gdorsi (approved):

packages/jazz-tools/src/tools/implementation/zodSchema/schemaTypes/CoFeedSchema.ts
```diff
@@ -65,9 +67,15 @@ export class CoFeedSchema<
...
-      .instanceof(CoFeed)
-      .or(z.array(generateValidationSchemaFromItem(this.element)));
+    const validationSchema = z.array(
+      generateValidationSchemaFromItem(this.element),
+    );
+
+    this.#validationSchema = coValueValidationSchema(
+      validationSchema,
+      CoFeed,
+      "CoFeed",
```
nrainhart: Nit: you could get the name directly from the class using `.name`

nrainhart (approved):

PR #3461, opened 2026-02-13
docs: separate deployment config from server-side setup

joeinnes:
Move crypto implementation details (WASM edge runtimes, Node-API) and deployment patterns (single-instance vs multi-region) into a dedicated deployment page. Simplify setup.mdx to focus on worker runtime basics and jazz-rpc.mdx to focus on the RPC API itself.


gdorsi (approved):
Makes sense

PR #3466, opened 2026-02-16
appendOnly coList

Takeno:
This PR introduces an append-only coList based on a new property in the cojson ruleset. Append-only co-lists allow writers to append new items, but only admins and managers can remove items. 

## jazz-tools API
```ts
const List = co.list(z.string())
  .withPermissions({ writer: "appendOnly" });
 ```
 
Only CoListSchema has the new `restrictDeletion` option in the `withPermissions` method, to enable the feature.

## cojson API

A new `restrictDeletion` property has been added to the `ownedByGroup` permission ruleset. It is part of the covalue's header, is signed, and remains immutable over time. 

In the cojson colist, transactions with a `del` operator are [validated](https://github.com/garden-co/jazz/pull/3466/changes#diff-b4653113263df14a0faab5201f5cc192ea3859fc229a3e751c32e9a60e34f60fR247-R259) against the role of the author at that time.

## Missing
- [ ] docs
- [ ] stress tests

### CODE REVIEW

File: .specs/colist-restricted-deletion/design.md [hunk 1]
```diff
@@ -0,0 +1,173 @@
...
+
+  for (const tx of coValue.toValidateTransactions) {
+    // ...existing checks...
+
+    const role = groupAtTime.roleOfInternal(effectiveTransactor);
+    if (!isWriteCapable(role)) markInvalid();
+
+    if (isRestrictedCoList && tx.changes.some((c) => isObject(c) && c.op === "del")) {
+      if (role !== "admin" && role !== "manager") markInvalid();
+    }
```
gdorsi, line 141: Changes aren't parsed at this step.

We do this to avoid invalid decrypted data.

I think this validation step should live in RawCoList directly, where data is already decrypted.

packages/cojson/src/coValues/coList.ts
```diff
@@ -208,6 +208,16 @@ export class RawCoList<
...
   }
 
+  private transactionContainsDeletion(changes: JsonValue[]) {
+    return changes.some(
+      (change) =>
+        typeof change === "object" &&
+        change !== null &&
+        "op" in change &&
+        (change as { op?: string }).op === "del",
+    );
```
gdorsi: No need for all these checks here, we can optimistically expect that this metches `ListOpPayload<Item>`

packages/cojson/src/coValues/coList.ts
```diff
@@ -221,6 +231,10 @@ export class RawCoList<
       return;
     }
 
+    const ruleset = this.core.verified.header.ruleset;
+    const isDeletionRestricted =
+      ruleset.type === "ownedByGroup" && ruleset.restrictDeletion === true;
```
gdorsi: Can we move this parsing in the constructor?

gdorsi (approved):
LGTM, two minor comments.

Could you add the changeset?

PR #3468, opened 2026-02-16
Fix Hermes V1 Proxy invariant enforcement for React Native 0.84+

boorad:
# Fix Hermes V1 Proxy invariant enforcement for React Native 0.84+

## Problem

React Native 0.84 ships with **Hermes V1**, which strictly enforces ES2015 Proxy invariants that earlier versions silently allowed. This causes a fatal `TypeError: HasProperty trap result is not configurable` error when any CoMap is rendered in a React component.

### Reproduction

Any React Native 0.84+ app using Jazz with `__DEV__=true` crashes on first render with:

```
TypeError: HasProperty trap result is not configurable
    at addObjectDiffToProperties
    at logComponentRender
    at commitPassiveMountOnFiber
    ...
```

followed by a cascading `Error: Should not already be working.` from React's scheduler.

### Root Cause

React 19's dev-mode `addObjectDiffToProperties` (used for render performance logging) iterates CoMap properties to diff props between renders:

```js
for (key in prev) key in next || (...)  // line ~10796 in RN bundle
```

This calls the CoMap proxy's `has` trap for every key. Two issues cause invariant violations:

**1. `has` trap returns `false` inconsistently with `ownKeys`/`getOwnPropertyDescriptor`**

The `has` trap returns `false` for schema keys whose raw value is `undefined`:

```ts
// Before (coMap.ts proxy handler)
has(target, key) {
  const descriptor = target.$jazz?.getDescriptor(key);
  if (target.$jazz?.raw && typeof key === "string" && descriptor) {
    return target.$jazz.raw.get(key) !== undefined;  // false for undefined values!
  } else {
    return Reflect.has(target, key);
  }
}
```

However, `ownKeys` can return those same keys (via `raw.keys()`), and `getOwnPropertyDescriptor` returns a valid descriptor for them. Per the ES spec (and enforced by Hermes V1), if `ownKeys` reports a key and `getOwnPropertyDescriptor` reports it as existing, the `has` trap must not return `false` for it.

**2. `Object.defineProperties` calls omit `configurable: true`**

Properties like `$jazz`, `$isLoaded`, `[TypeSym]`, and `_instanceID` are defined on CoValue instances via `Object.defineProperties` without `configurable: true`. Since `configurable` defaults to `false`, these become non-configurable own properties of the proxy target.

Per the ES spec §10.5.7: *"If the target object has a non-configurable own property with the same name, the proxy's `has` trap cannot return `false` for that property."* While the `has` trap delegates to `Reflect.has` for these keys (returning `true`), the non-configurable state creates problems when other trap operations interact with them.

### 1. Make `has` trap consistent with `ownKeys`

Change the `has` trap to return `true` for all keys that have a schema descriptor, regardless of whether the raw value is currently `undefined`:

```ts
has(target, key) {
  const descriptor = target.$jazz?.getDescriptor(key);
  if (target.$jazz?.raw && typeof key === "string" && descriptor) {
    return true;
  } else {
    return Reflect.has(target, key);
  }
}
```

**Why this is safe:**

- `$jazz.has(key)` (the public API for checking if a key is set) is a regular method on `CoMapJazzApi`, NOT the proxy trap. It is unaffected by this change and continues to return `false` for unset/deleted keys.
- No Jazz internal code uses `"key" in coMap` (the `in` operator on a CoMap proxy) to check if a key has been set — they all use `$jazz.has(key)`.
- The `in` operator on a CoMap was already inconsistent: `Object.keys(coMap)` could include a key that `"key" in coMap` reported as absent (e.g., when set to `undefined`). This fix makes `in` consistent with `Object.keys` and `getOwnPropertyDescriptor`.

**Test adjustments needed:** A few tests assert `"age" in john` is `false` for unset/deleted optional properties. These tests should be updated to reflect the new behavior where `in` returns `true` for any schema-defined key (and use `$jazz.has()` to test set/unset semantics, which is the correct API for that purpose).

### 2. Add `configurable: true` to all `Object.defineProperties` calls

All `Object.defineProperties`/`Object.defineProperty` calls that define properties on CoValue instances (proxy targets) must include `configurable: true`. This prevents Hermes from throwing when proxy traps interact with these properties.

**Files changed:**

| File | Properties |
|------|-----------|
| `CoValueBase.ts` | `$isLoaded`, `_instanceID` |
| `coMap.ts` | `$jazz` (constructor + `_createCoMap`), `has` trap |
| `coList.ts` | `$jazz`, `$isLoaded` (constructor + `create`) |
| `coFeed.ts` | `$jazz` |
| `account.ts` | `[TypeSym]`, `$jazz` |
| `group.ts` | `$jazz` |
| `coPlainText.ts` | `[TypeSym]`, `$jazz`, `$isLoaded` (2 constructor paths) |
| `coVector.ts` | `[TypeSym]`, `$jazz`, `$isLoaded`, `_isVectorLoaded`, `_requiredDimensionsCount` |
| `FileStream` (in coFeed.ts) | `[TypeSym]`, `$jazz` |

## Impact

- **React Native 0.84+**: Fixes fatal crash on any component render involving CoMaps
- **React Native <0.84**: No change in behavior (older Hermes didn't enforce these invariants)
- **Web/Node**: No change in behavior (V8/SpiderMonkey don't enforce these invariants as strictly)
- **`$jazz.has()` API**: Completely unaffected — continues to work as documented
- **`"key" in coMap`**: Now returns `true` for all schema-defined keys. This is more consistent with `Object.keys()` behavior and aligns the proxy traps per the ES spec.

## References

- [ES2015 Proxy Invariants §10.5.7](https://tc39.es/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-hasproperty-p) — `[[HasProperty]]` invariant
- [Hermes V1 JSProxy.cpp](https://github.com/nicolo-ribaudo/tc39-proposal-hermes-static-hermes/blob/main/lib/VM/JSProxy.cpp#L787) — strict invariant enforcement
- React 19 `addObjectDiffToProperties` in `logComponentRender` — dev-mode prop diffing that triggers the `has` trap

---

> [!NOTE]
> **Medium Risk**
> Changes fundamental `Proxy`/property-existence semantics for `CoMap` (and adds configurability to many internal properties), which can affect downstream code that relied on `"key" in coMap` for unset/deleted optional fields.
> 
> **Overview**
> Fixes React Native 0.84+ (Hermes V1) crashes by making the `CoMap` proxy `has` trap always return `true` for *schema-defined* keys, even when unset/`undefined`/deleted, so it stays consistent with `ownKeys`/`getOwnPropertyDescriptor`.
> 
> Adds `configurable: true` to internal properties (`$jazz`, `$isLoaded`, `[TypeSym]`, `_instanceID`, and a few CoVector internals) across multiple CoValue types to satisfy ES2015 proxy invariants, and updates/extends tests to assert the new `in` semantics and to use `$jazz.has()` for “value is set” checks (including new proxy-invariant consistency tests).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 796c65ba36ce0a2d3ea36d8e9dfae18c4786e32e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/fix-hermes-proxy-invariants.md
+++ b/.changeset/fix-hermes-proxy-invariants.md
@@ -1,0 +1,11 @@
+---
+"jazz-tools": patch
+---
+
+**BREAKING:** The `in` operator on CoMap instances now returns `true` for all schema-defined keys, even if the value is `undefined` or has been deleted. This fixes a fatal `TypeError` on React Native 0.84+ (Hermes V1) caused by proxy invariant violations.
+
+Previously, `"key" in coMap` returned `false` for unset/deleted optional properties. Now it returns `true` for any key with a schema descriptor, consistent with `Object.keys()` and `Object.getOwnPropertyDescriptor()`.
+
+To check whether a key has an actual value set, use `coMap.$jazz.has("key")` instead of the `in` operator.
+
+Also adds `configurable: true` to all internal property definitions (`$jazz`, `$isLoaded`, `[TypeSym]`, `_instanceID`) across all CoValue types to satisfy ES2015 proxy invariants enforced by Hermes V1.
--- a/packages/jazz-tools/src/tools/coValues/CoValueBase.ts
+++ b/packages/jazz-tools/src/tools/coValues/CoValueBase.ts
@@ -25,7 +25,7 @@ export abstract class CoValueBase implements CoValue {
 
   constructor() {
     Object.defineProperties(this, {
-      $isLoaded: { value: true, enumerable: false },
+      $isLoaded: { value: true, enumerable: false, configurable: true },
     });
   }
 
@@ -56,6 +56,7 @@ export abstract class CoValueJazzApi<V extends CoValue> {
   constructor(private coValue: V) {
     Object.defineProperty(this, "_instanceID", {
       value: `instance-${Math.random().toString(36).slice(2)}`,
+      configurable: true,
       enumerable: false,
     });
   }
--- a/packages/jazz-tools/src/tools/coValues/account.ts
+++ b/packages/jazz-tools/src/tools/coValues/account.ts
@@ -103,10 +103,11 @@ export class Account extends CoValueBase implements CoValue {
     );
 
     Object.defineProperties(this, {
-      [TypeSym]: { value: "Account", enumerable: false },
+      [TypeSym]: { value: "Account", enumerable: false, configurable: true },
       $jazz: {
         value: new AccountJazzApi(proxy, options.fromRaw),
         enumerable: false,
+        configurable: true,
       },
     });
 
--- a/packages/jazz-tools/src/tools/coValues/coFeed.ts
+++ b/packages/jazz-tools/src/tools/coValues/coFeed.ts
@@ -208,6 +208,7 @@ export class CoFeed<out Item = any> extends CoValueBase implements CoValue {
       $jazz: {
         value: new CoFeedJazzApi(this, options.fromRaw),
         enumerable: false,
+        configurable: true,
       },
     });
 
@@ -823,10 +824,15 @@ export class FileStream extends CoValueBase implements CoValue {
     }
 
     Object.defineProperties(this, {
-      [TypeSym]: { value: "BinaryCoStream", enumerable: false },
+      [TypeSym]: {
+        value: "BinaryCoStream",
+        enumerable: false,
+        configurable: true,
+      },
       $jazz: {
         value: new FileStreamJazzApi(this, raw),
         enumerable: false,
+        configurable: true,
       },
     });
   }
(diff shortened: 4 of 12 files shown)


antoniomuso (approved):
LGTM, should we add a changeset with a breaking change? @gdorsi

boorad:
> LGTM, should we add a changeset with a breaking change? @gdorsi

I added a changeset that is `patch` - whether you want to make it `minor` or not is up to you guys...  lmk

gdorsi:
No, patch looks good to me.

The breaking change on keys can actually be considered a bugfix, because the prev logic didn't make sense IMHO.

gdorsi:
Thanks @boorad !

PR #3472, opened 2026-02-17
fix: hanging in-flight load on gc values

gdorsi:
Track a load message as complete when the server replies with a known state message and the value has been garbage collected in the meantime

### DIFF
--- a/packages/cojson/src/coValueCore/coValueCore.ts
+++ b/packages/cojson/src/coValueCore/coValueCore.ts
@@ -345,6 +345,14 @@ export class CoValueCore {
     return this.hasVerifiedContent();
   }
 
+  isKnownStateAvailable(): boolean {
+    return (
+      this.loadingState === "available" ||
+      this.loadingState === "onlyKnownState" ||
+      this.loadingState === "garbageCollected"
+    );
+  }
+
   isCompletelyDownloaded(): boolean {
     if (!this.hasVerifiedContent()) {
       return false;
--- a/packages/cojson/src/coValueCore/coValueCore.ts
+++ b/packages/cojson/src/coValueCore/coValueCore.ts
@@ -312,16 +312,16 @@ export class CoValueCore {
       return "available";
     }
 
+    // Check for lastKnownStateSource (garbageCollected or onlyKnownState)
+    if (this.#lastKnownStateSource) {
+      return this.#lastKnownStateSource;
+    }
+
     // Check for pending peers FIRST - loading takes priority over other states
     for (const peer of this.loadingStatuses.values()) {
       if (peer.type === "pending") {
         return "loading";
       }
-    }
-
-    // Check for lastKnownStateSource (garbageCollected or onlyKnownState)
-    if (this.#lastKnownStateSource) {
-      return this.#lastKnownStateSource;
     }
 
     if (this.loadingStatuses.size === 0) {
--- a/packages/cojson/src/coValueCore/coValueCore.ts
+++ b/packages/cojson/src/coValueCore/coValueCore.ts
@@ -36,12 +36,7 @@ import {
   getDependenciesFromGroupRawTransactions,
   getDependenciesFromHeader,
 } from "./utils.js";
-import {
-  CoValueHeader,
-  Transaction,
-  Uniqueness,
-  VerifiedState,
-} from "./verifiedState.js";
+import { CoValueHeader, Transaction, VerifiedState } from "./verifiedState.js";
 import {
   MergeCommit,
   BranchPointerCommit,
--- a/packages/cojson/src/queue/OutgoingLoadQueue.ts
+++ b/packages/cojson/src/queue/OutgoingLoadQueue.ts
@@ -11,6 +11,11 @@ interface PendingLoad {
   sendCallback: () => void;
 }
 
+interface InFlightLoad {
+  value: CoValueCore;
+  sentAt: number;
+}
+
 /**
  * Mode for enqueuing load requests:
  * - "high-priority" (default): high priority, processed in order
@@ -29,7 +34,7 @@ export type LoadMode = "low-priority" | "immediate" | "high-priority";
  * - Manages timeouts for in-flight loads with a single timer
  */
 export class OutgoingLoadQueue {
-  private inFlightLoads: Map<CoValueCore, number> = new Map();
+  private inFlightLoads: Map<RawCoID, InFlightLoad> = new Map();
   private inFlightCounter: UpDownCounter;
   private highPriorityPending: LinkedList<PendingLoad> = meteredList(
     "load-requests-queue",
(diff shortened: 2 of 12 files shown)


packages/cojson/src/sync.ts
gdorsi: @nrainhart should we do the same for garbage collected values?

nrainhart: It's not necessary for storage reconciliation, because if the value was garbage collected, it means it's already in-sync with the server

nrainhart (approved):

antoniomuso (approved):

packages/cojson/src/coValueCore/coValueCore.ts
```diff
@@ -312,18 +312,18 @@ export class CoValueCore {
       return "available";
     }
 
+    // Check for lastKnownStateSource (garbageCollected or onlyKnownState)
+    if (this.#lastKnownStateSource) {
+      return this.#lastKnownStateSource;
+    }
```
gdorsi, line 313: Before "loading" state had the precedence over garbageCollected or onlyKnownState

This was making many values stuck in the loading state, because we resolve it either when content arrives or when the value is not found.

packages/cojson/src/queue/OutgoingLoadQueue.ts
```diff
@@ -79,13 +84,13 @@ export class OutgoingLoadQueue {
    */
   private trackSent(coValue: CoValueCore): void {
     const now = performance.now();
-    this.inFlightLoads.set(coValue, now);
+    this.inFlightLoads.set(coValue.id, { value: coValue, sentAt: now });
```
gdorsi, line 88: Switching to use the id as key to handle the cases where the coValue instance changes

packages/cojson/src/sync.ts
gdorsi: Updated the logic because we should assume that between the reconciliation request/response anything can happen.

The value might be loaded+collected.

Added also a test to cover the case

packages/cojson/src/sync.ts
```diff
@@ -960,12 +960,7 @@ export class SyncManager {
       }
     }
 
-    // KNOWN with header=true does not guarantee that content arrived.
-    // Keep the request in-flight until content (or a terminal state) lands,
-    // so queue timeout logic can mark this peer as unavailable if needed.
-    if (!availableOnPeer || coValue.isAvailable()) {
-      peer.trackLoadRequestComplete(coValue);
-    }
+    peer.trackLoadRequestComplete(coValue, "known");
```
gdorsi, line 963: Reverted this change because I found the source of the issue I was trying to fix with this change.

I've introduced these checks because many values were stuck in `lodingState: loading` but I realized in this PR that the reason was this one https://github.com/garden-co/jazz/pull/3472/changes#r2821277092

packages/cojson/src/coValueCore/coValueCore.ts
nrainhart: But if the order of events is:
- known state was loaded (or covalue was garbage collected)
- covalue is loaded

Isn't it correct for the covalue to be in "loading" state?

packages/cojson/src/queue/OutgoingLoadQueue.ts
```diff
@@ -124,7 +130,7 @@ export class OutgoingLoadQueue {
             id: coValue.id,
             peerId: this.peerId,
           });
-          coValue.markNotFoundInPeer(this.peerId);
+          coValue.node.getCoValue(coValue.id).markNotFoundInPeer(this.peerId);
```
nrainhart, line 134: Isn't `coValue.node.getCoValue(coValue.id)` the same as just `coValue`? Or are we concerned about stale object instances? if so, we should leave a comment here

gdorsi: Yes, that's to avoid sending the update on a stale object instance

gdorsi: Adding a comment

packages/cojson/src/tests/sync.garbageCollection.test.ts
```diff
@@ -205,6 +205,8 @@ describe("sync after the garbage collector has run", () => {
         "server -> storage | CONTENT Group header: true new: After: 0 New: 4",
         "server -> client | KNOWN Map sessions: header/1",
         "server -> storage | CONTENT Map header: true new: After: 0 New: 1",
+        "client -> storage | LOAD Map sessions: empty",
+        "storage -> client | CONTENT Map header: true new: After: 0 New: 1",
```
nrainhart, line 209: ⚠️ Is this additional load from storage ok?

packages/cojson/src/tests/sync.concurrentLoad.test.ts
```diff
@@ -729,4 +734,399 @@ describe("concurrent load", () => {
...
+
+    client.connectToSyncServer();
+
+    const groupOnServer = jazzCloud.node.createGroup();
+    const groupOnClient = await loadCoValueOrFail(
+      client.node,
+      groupOnServer.id,
+    );
+
+    expect(groupOnClient.core.isAvailable()).toBe(true);
```
nrainhart: We're loading a new group? I thought we'd load the same group we just gc-ed

packages/cojson/src/tests/sync.concurrentLoad.test.ts
```diff
@@ -729,4 +734,399 @@ describe("concurrent load", () => {
...
+
+    expect(onlyKnownGroup.loadingState).toBe("onlyKnownState");
+    expect(onlyKnownMap1.loadingState).toBe("onlyKnownState");
+    expect(onlyKnownMap2.loadingState).toBe("onlyKnownState");
+
+    SyncMessagesLog.clear();
+
+    client.connectToSyncServer();
+
+    await waitFor(() => {
```
nrainhart, line 889: Can't we just wait for the 3 covalues to be available and then assert the whole message log (in the right order) is ok?

packages/cojson/src/tests/sync.garbageCollection.test.ts
gdorsi: Good spot!

My initial answer would have been "yes" but after reading this more carefully I've noticed that [this](https://github.com/garden-co/jazz/blob/fix/in-flight-gc-values/packages/cojson/src/sync.ts#L956) always return false here, because we return false [here](https://github.com/garden-co/jazz/blob/main/packages/cojson/src/SyncStateManager.ts#L146) when there is no verified content

Fixing

packages/cojson/src/coValueCore/coValueCore.ts
gdorsi: No, because we have the content locally, it's just not loaded in memory.

The loading happens in background but it is just for reconciliation, shouldn't surface in the coValue state.

packages/cojson/src/tests/sync.concurrentLoad.test.ts
gdorsi: This is to test that the load queue is free now.

Adding a comment

packages/cojson/src/tests/sync.concurrentLoad.test.ts
gdorsi: No, because their state in both nodes shouldn't change.

Cleaned up the waitFor to make it more readable: https://github.com/garden-co/jazz/pull/3472/changes/2f7155eeadf47249910d3441fffb6ee8d58fb07b

packages/cojson/src/tests/sync.garbageCollection.test.ts
gdorsi: Fix here: https://github.com/garden-co/jazz/pull/3472/changes/2f7155eeadf47249910d3441fffb6ee8d58fb07b

packages/cojson/src/tests/sync.concurrentLoad.test.ts
gdorsi: Done!

nrainhart (approved):
Thanks for addressing all my comments!

PR #3477, opened 2026-02-19
feat(create-jazz-app): seed AGENTS.md and .skills in scaffolded projects

joeinnes:
## Summary

- Projects scaffolded by `create-jazz-app` now receive an `AGENTS.md` (fetched from `jazz.tools/AGENTS.md`) with a URL-based docs index and skills listing, plus a `.skills/` directory containing the five `jazz-*` skill files
- Extracts shared docs-index utilities from `add-docs-links-to-agents.mjs` into `docs-index-utils.mjs` so both the monorepo AGENTS.md and the user-facing AGENTS.md can be generated from the same code
- Adds the `jazz-*` skills to `packages/cursor-docs/.skills/` so they're cloned alongside `.cursor/` during scaffolding
- Adds `AGENTS_MD_URL` constant with unit tests in `docsFetch.test.ts`, plus live URL integration tests gated behind `TEST_LIVE_URLS` env var

### DIFF
--- a/.changeset/create-jazz-app-agents-md.md
+++ b/.changeset/create-jazz-app-agents-md.md
@@ -1,0 +1,5 @@
+---
+"create-jazz-app": patch
+---
+
+Scaffold AGENTS.md and .skills/ directory in new projects, providing an AI agent-ready docs index and Jazz skill files out of the box.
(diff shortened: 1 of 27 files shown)


gdorsi (approved):
LGTM!

Could you add a changeset?

PR #3482, opened 2026-02-23
feat(storage-reconciliation): wait for coValues' sync before ack-ing batch

nrainhart:
### Description

After releasing https://github.com/garden-co/jazz/pull/3422 to clients, we started noticing spikes in edge servers' outgoing queues due to excessive concurrent loads.

To avoid this, we decided to implement stronger backpressure by not acknowledging a reconciliation batch until all CoValues in that batch have been synced. This means storage reconciliation will be slower, but less intensive on servers.

### DIFF
--- a/packages/cojson/src/OngoingStorageReconciliationTracker.ts
+++ b/packages/cojson/src/OngoingStorageReconciliationTracker.ts
@@ -1,19 +1,12 @@
 import { RawCoID } from "./ids.js";
 import { PeerID, ReconcileBatchID } from "./sync.js";
 
-type ReconcileBatchTracker = {
-  coValues: Set<RawCoID>;
-  pending: Set<RawCoID>;
-};
-
 /**
  * Tracks ongoing storage reconciliation batches in a server.
  */
 export class OngoingStorageReconciliationTracker {
-  private reconcileBatches: Map<
-    PeerID,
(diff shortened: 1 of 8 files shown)


packages/cojson/src/OngoingStorageReconciliationTracker.ts
```diff
@@ -0,0 +1,138 @@
...
+        coValuesToBatches.set(coValueId, batchIds);
+      }
+      batchIds.add(batchId);
+    }
+  }
+
+  /**
+   * Marks a coValue as reconciled and returns the batch IDs that completed syncing.
+   */
+  markItemComplete(peerId: PeerID, coValueId: RawCoID): ReconcileBatchID[] {
```
gdorsi, line 51: I struggle a bit to follow the cleanup logic.

The risk here is to end up with some memory leaks.

How about we merge this and clearBatch?

gdorsi (approved):
LGTM

packages/cojson/src/OngoingStorageReconciliationTracker.ts
nrainhart: Done! I also added tests for OngoingStorageReconciliationTracker to ensure we're doing the cleanup properly

PR #3494, opened 2026-03-12
fix(cojson): avoid wasm streaming state aliasing

gdorsi:
## Summary
- clone WASM known-state values before serializing them back to JS
- cache the streaming known-state result in `VerifiedState` so `isStreaming()` and `newContentSince()` reuse the same FFI result
- add regression coverage for both direct `isStreaming()` calls and the `newContentSince()` path, then regenerate the checked-in WASM artifacts

## Testing
- `pnpm -C crates/cojson-core-wasm build:wasm`
- `pnpm -C crates/cojson-core-wasm test -- --runInBand`
- `pnpm exec biome check packages/cojson/src/coValueCore/verifiedState.ts packages/cojson/src/tests/coValueCore.isStreaming.test.ts`
- `pnpm exec vitest --run --root . --project cojson packages/cojson/src/tests/coValueCore.isStreaming.test.ts -t "cached streaming known state"`
- `pnpm exec vitest --run --root . --project cojson packages/cojson/src/tests/sync.multipleServers.test.ts -t "uploading a large update between two clients"`


antoniomuso (approved):
LGTM!

PR #3503, opened 2026-03-17
fix(cojson-core-wasm): surface wasm panics as JS errors

gdorsi:
I'm adding a WASM panic hook that rethrows panics as JavaScript errors with the panic text, to make this kind of errors debuggable.

We've got one user finding a "unreachable" error in their error tracker, and we are adding this change to debug the problem.

Testing: we manually verified the flow by triggering an explicit panic from the WASM module and confirming the panic surfaced through the browser error path.


nrainhart (approved):
I'd normally wait for @antoniomuso's review, but approving since he's out today and the changes look reasonable.

Did you confirm Rust panics are properly surfaced on the JS side?

gdorsi:
> I'd normally wait for @antoniomuso's review, but approving since he's out today and the changes looks reasonable.
> 
> Did you confirm Rust panics are properly surfaced on the JS side?

Yes, tested with a manual trigger in an example app.

Issue #321, opened 2024-08-17
Indexing and search


aeplay:

CanRau:
Hey how you doing?

what's the current situation regarding search? just filtering the complete array for example?

aeplay:
> Hey how you doing?
> 
> what's the current situation regarding search? just filtering the complete array for example?

Hey Can, good to hear from you!

yes! But we're starting to have adopters with "lots of data" so we'll probably prioritise indexing soon

CanRau:
Sounds great, I have a (mobile) app in mind that would need good full-text and potentially vector search, when and if I get the time for it 🤓

Gabrola:
@gdorsi Hiya! Is this on your radar soon by any chance?

gdorsi:
> @gdorsi Hiya! Is this on your radar soon by any chance?

Hey, what's your use case?

Gabrola:
@gdorsi all of them 😆

On a serious note though, we want to be able to search, filter and aggregate by anything, just shy of full-text searching. We want to be able to do this easily and repeatably across different schemas and access patterns. So its either having indexing built into Jazz, or syncing data with an external DB.

The thing is that there is no solution that abstracts syncing data from Jazz either. So we'd have to do lots of work just to be able to orchestrate subscribing to changes on the relevant covalues and managing sync state across the different schemas. Either that or we build a proper sync solution.

From my perspective, I feel like if I take a stab at this, I might be able to build a simple indexing and searching solution that could cover the majority of use cases, but in less time and effort than doing any of these "workarounds". I didn't want to put too much thought or work into this, without consulting with you first.

gdorsi:
I think we need a good subscription layer to make incremental computation possible.

I took a stab at implementing a CDC system on jazz-tools (spec [here](https://github.com/garden-co/jazz/pull/3504/)) but I think we need to include some rewrite of SubscriptionScope to make that possible/maintainable.

I'm going to tweak the design a little bit more, to see if I can reduce the scope while keeping it flexible and useful.

We also have some expermients around indexes (draft [here](https://github.com/garden-co/jazz/pull/2946)) but it's nowhere near to what you are asking for.

So, my take is that we should try to first make it possible to subscribe to changes, and then try to build local indexes.

[...]

PR #3504, opened 2026-03-17
docs: design a possible implementation for CDC

gdorsi:


Gabrola:
@gdorsi this looks really good. I think this would work really well for local indexing on a client level. On a bigger scale, which is most likely on servers, I think the subscribe-to-everything problem I mentioned here https://github.com/garden-co/jazz/issues/321#issuecomment-4101386561 would still be an issue. The entirety of the data will need to be loaded (and actually fit) into memory, and the subscribe call itself would take a really long time to finish loading and executing.

I was thinking... since on this scale, we'll probably need it on servers, we can actually do this functionality in the sync servers themselves, and since sync servers receive all changes. We can do the CDC in `SyncManager.handleNewContent` on the sync server without having to "subscribe" to everything. Obviously, the sync server needs to be able to decrypt the data, but that'll be the case anyway for a server worker wanting to do the same. If we want a clear separation of concerns, we can instead setup a "CDC server worker" and configure that as a persistent server peer for the sync server to make sure all data reaches the worker.

gdorsi:
@Gabrola 

We can make it so that the Changes data capture doesn't actually load values that haven't changed.

At sync level we already have all the primitives we need (the onlyKnownState path) that makes us fetch only the cursor data.

This makes the implementation a little more complicated, but it's still possible and should be quite scalable.

Gabrola:
@gdorsi Yeah, I was thinking along something quite similar. I was thinking on the sync level, if we want to "subscribe" to any CoValue, we can simply store the non-opaque cursor in sqlite so its persisted across reboots, and then compare any incoming transactions if they match a "tracked" cursor, and emit a `Change` event, which we could easily build just from the cursor data + transaction without even having to load the `knownState` for the entirety of the data in the cursor up front, and then on Changes emit we can update the cursor incrementally. So overall the CDC is handled a bit more end-to-end by Jazz itself rather than user code. I think it's actually quite similar to idea in the `jazz-webhook` package, but we track a "cursor" instead of a single covalue.

gdorsi:
@Gabrola Update the design to aim for building the feature inside cojson for 1-level deep collections.

We can use this as building block for a more complex feature based on resolve inside jazz-tools.

One-shotted a POC and complexity-wise is ok.

The only tradeoff for not changing the sync protocol is that syncing content requires a correction if the client doesn't have a storage.

The POC has probably some bugs (e.g. it doesn't take into account local changes/storage state) but I feel that this can be a good path to follow.

PR #3512, opened 2026-03-26
feat: add ability to create cursors and load by cursor

Gabrola:
A couple of divergences from from spec in #3504 

- Children error loading states are stored in the cursor to ensure we can replay the same exact state at the time of cursor creation.
- The recommended method to create a cursor for a CoValue is to load it with `{ cursor: { useCurrentCursor: true } }`. This will load the current state of the value, create a cursor, and then return the same value loaded (and as a result frozen) with that same cursor. This ensures we do not encounter any race conditions if a CoValue is changed after the cursor is created.

From the spec:
```ts
const tasks = await TaskList.load(listId, { resolve: { $each: true } });
for (const task of tasks ?? []) {
  const id = await insert(db, {
    title: task.title ?? "",
    body: task.body ?? "",
  });
  oramaIds.set(task.id, id);
}

// if anything changes between the above and below lines, these changes would not be captured by the index

// Snapshot the current frontier so incremental indexing
// only replays changes made after this point.
const cursor = await currentCursor(TaskList, listId, {
  resolve: { $each: true },
});
```

A more serious case:
```ts
const product = await Product.load(productId);
const price = product.price * (1 + product.taxes);

const cursor = await currentCursor(Product, productId, {
  resolve: true,
});

// at a later time
const product = await Product.load(productId, {
  cursor,
});

// not guaranteed to match the above price
const price = product.price * (1 + product.taxes);
```

With the change:
```ts
const tasks = await TaskList.load(listId, {
  resolve: { $each: true },
  cursor: { useCurrentCursor: true },
});

for (const task of tasks ?? []) {
  const id = await insert(db, {
    title: task.title ?? "",
    body: task.body ?? "",
  });
  oramaIds.set(task.id, id);
}

// this cursor is guaranteed to be consistent with the state of `tasks`
const cursor = tasks.$jazz.cursor;
```

A cursor can still be created using `.$jazz.createCursor()` in case this consistency is not needed

### DIFF
--- a/packages/cojson/src/coValues/coList.ts
+++ b/packages/cojson/src/coValues/coList.ts
@@ -1,11 +1,15 @@
 import { CoID, RawCoValue } from "../coValue.js";
-import { AvailableCoValueCore } from "../coValueCore/coValueCore.js";
+import {
+  AvailableCoValueCore,
+  DecryptedTransaction,
+} from "../coValueCore/coValueCore.js";
 import { AgentID, SessionID, TransactionID, RawCoID } from "../ids.js";
 import { JsonObject, JsonValue } from "../jsonValue.js";
 import { accountOrAgentIDfromSessionID } from "../typeUtils/accountOrAgentIDfromSessionID.js";
 import { isCoValue } from "../typeUtils/isCoValue.js";
 import { RawAccountID } from "./account.js";
 import { RawGroup } from "./group.js";
+import { CoValueFrontier } from "../knownState.js";
 
 export type OpID = TransactionID & { changeIdx: number };
 
@@ -78,6 +82,9 @@ export class RawCoList<
   /** @internal */
   atTimeFilter?: number = undefined;
 
+  /** @internal */
+  atFrontierFilter?: CoValueFrontier = undefined;
+
   /** @category 6. Meta */
   readonly _item!: Item;
 
@@ -106,7 +113,13 @@ export class RawCoList<
   }
 
   /** @internal */
-  constructor(core: AvailableCoValueCore, atTimeFilter?: number) {
+  constructor(
+    core: AvailableCoValueCore,
+    options?: {
+      atTimeFilter?: number;
+      atFrontierFilter?: CoValueFrontier;
+    },
+  ) {
     this.id = core.id as CoID<this>;
     this.core = core;
 
@@ -115,13 +128,14 @@ export class RawCoList<
     this.afterStart = [];
     this.beforeEnd = [];
     this.knownTransactions = { [core.id]: 0 };
-    this.atTimeFilter = atTimeFilter;
+    this.atTimeFilter = options?.atTimeFilter;
+    this.atFrontierFilter = options?.atFrontierFilter;
 
     const ruleset = this.core.verified.header.ruleset;
     this.#isDeletionRestricted =
       ruleset.type === "ownedByGroup" && ruleset.restrictDeletion === true;
 
-    this._processNewTransactions();
+    this.processNewTransactions();
   }
 
   private getInsertionsEntry(opID: OpID) {
@@ -208,17 +222,6 @@ export class RawCoList<
   }
 
   processNewTransactions() {
-    if (this.isTimeTravelEntity()) {
-      throw new Error("Cannot process transactions on a time travel entity");
-    }
-    this._processNewTransactions();
-  }
-
-  private transactionContainsDeletion(changes: ListOpPayload<Item>[]) {
-    return changes.some((change) => change.op === "del");
-  }
-
-  private _processNewTransactions() {
     // Get all transactions including invalid ones, so that items referencing
     // entries from invalid init transactions can still find their references.
     const transactions = this.core.getValidSortedTransactions({
@@ -235,10 +238,12 @@ export class RawCoList<
     let oldestValidTransaction: number | undefined = undefined;
     this._cachedEntries = undefined;
 
-    for (const { txID, changes, madeAt, isValid } of transactions) {
-      if (this.isFilteredOut(madeAt)) {
+    for (const tx of transactions) {
+      if (this.isFilteredOut(tx)) {
         continue;
       }
+
+      const { txID, changes, madeAt, isValid } = tx;
 
       if (
         isValid &&
@@ -338,6 +343,10 @@ export class RawCoList<
     this.totalValidTransactions += transactions.length;
   }
 
+  private transactionContainsDeletion(changes: ListOpPayload<Item>[]) {
+    return changes.some((change) => change.op === "del");
+  }
+
   rebuildFromCore() {
     this.version += 1;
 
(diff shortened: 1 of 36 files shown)


packages/cojson/src/coValues/coList.ts
```diff
@@ -208,17 +222,6 @@ export class RawCoList<
   }
 
   processNewTransactions() {
-    if (this.isTimeTravelEntity()) {
-      throw new Error("Cannot process transactions on a time travel entity");
-    }
```
Gabrola, line 213: Should be fine since the transactions are going to be filtered by `isFilteredOut` anyway. Had to make this change since we are now time traveling inside a `SubscriptionScope` which calls `.processNewTransactions()` and was throwing

packages/jazz-tools/src/tools/subscribe/cursor.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export const encodeCursor = (
+  decodedCursor: Omit<DecodedCoValueCursor, "resolveFingerprint"> & {
+    resolveFingerprint: DecodedCoValueCursor["resolveFingerprint"] | boolean;
+  },
+): CoValueCursor => {
+  const textEncoder = new TextEncoder();
+  return `cursor_z${base58.encode(
+    textEncoder.encode(
+      cojsonInternals.stableStringify({
```
gdorsi, line 51: Why stableStringify?

packages/jazz-tools/src/tools/subscribe/cursor.ts
```diff
@@ -0,0 +1,110 @@
...
+  CoValueLoadingState,
+  DecodedCoValueCursor,
+} from "./types.js";
+import { z } from "zod/v4";
+import type { RefsToResolve } from "../coValues/deepLoading.js";
+
+const cursorSchema = z.object({
+  version: z.literal(1),
+  rootId: z.string<RawCoID>(),
+  resolveFingerprint: z.record(z.string(), z.any()),
```
gdorsi, line 11: This is the resolve query snapshot right?

packages/jazz-tools/src/tools/subscribe/cursor.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  const normalizedResolveFingerprint = cojsonInternals.stableStringify(
+    decodedCursor.resolveFingerprint,
+  );
+  const normalizedResolve = cojsonInternals.stableStringify(
+    resolve === true ? {} : resolve,
+  );
+
+  if (normalizedResolveFingerprint !== normalizedResolve) {
```
gdorsi: What's the reason for this check?

packages/jazz-tools/src/tools/subscribe/cursor.ts
```diff
@@ -0,0 +1,110 @@
...
+} from "./types.js";
+import { z } from "zod/v4";
+import type { RefsToResolve } from "../coValues/deepLoading.js";
+
+const cursorSchema = z.object({
+  version: z.literal(1),
+  rootId: z.string<RawCoID>(),
+  resolveFingerprint: z.record(z.string(), z.any()),
+  frontiers: z.record(z.string<RawCoID>(), z.record(z.string(), z.number())),
+  valueErrors: z
```
gdorsi: Why valueErrors are part of the schema?

gdorsi (commented):
Thanks for the contribution!

The implementation looks good to me, just left some questions around encoding/decoding.

packages/jazz-tools/src/tools/subscribe/cursor.ts
Gabrola: So it normalizes the object to ensure that it produces a stable cursor string. Not sure if it's preferred to not use specifically stableStringify outside of the crypto, but I thought I'd use it instead of implementing the same thing again.

packages/jazz-tools/src/tools/subscribe/cursor.ts
Gabrola: Yes

packages/jazz-tools/src/tools/subscribe/cursor.ts
Gabrola: To ensure when loading with a cursor that the subscriber does not resolve more data than what is referenced by the cursor. Basically this part of the spec https://github.com/garden-co/jazz/pull/3504/changes/BASE..dc873c46e754e5c83ad8efb6e8913dcf1232cac6#diff-6ae3710f89317c968fb3baa5205e6a8f2ab6270842f61187ba55302d3fc473beR414-R418

I think the true functional requirement would be to actually ensure that the `resolve` is a strict subset of `resolveFingerprint` rather than check for equality.

packages/jazz-tools/src/tools/subscribe/cursor.ts
Gabrola: This wouldn't be needed if we were only doing CDC, but if we want to replay the actual cursor, I THINK we need to replay the errors too in case they are intermittent. I'm thinking of something like this test for example.

```ts
const CartItem = co.map({
  price: z.number(),
  quantity: z.number(),
});

const Cart = co.list(CartItem);

const originalCart = Cart.create([
  { price: 10, quantity: 1 },
  { price: 20, quantity: 2 },
]);

const cartSnapshot = await originalCart.$jazz.ensureLoaded({
  resolve: { $each: { $onError: "catch" } },
  cursor: { useCurrentCursor: true },
});

const total = cartSnapshot.reduce(
  (acc, item) =>
    item.$isLoaded ? acc + item.price * item.quantity : acc,
  0,
);

// at a later time
const loadedSnapshot = await Cart.load(originalCart.$jazz.id, {
  resolve: { $each: { $onError: "catch" } },
  cursor: cartSnapshot.$jazz.cursor,
});

assertLoaded(loadedSnapshot);

const loadedTotal = loadedSnapshot.reduce(
  (acc, item) =>
    item.$isLoaded ? acc + item.price * item.quantity : acc,
  0,
);

expect(loadedTotal).toBe(total);
```

If for some reason, such as a syncing issue, one of the cart items failed to load when `cartSnapshot` was created, but then it becomes available when `loadedSnapshot` is loaded. If we didn't replay the error we'd get a different result and this test would fail.

I think we can get away with transmitting `UNAVAILABLE` instead of replaying the actual error for children with a missing frontier, and that would allow us to remove `valueErrors`

Gabrola: For example, this test (`handles nested values when access is granted after cursor creation`) fails if we don't replay errors https://github.com/garden-co/jazz/pull/3512/changes#diff-7a6be4f53f512c70e71d541efd6527679a97a6f02f9ec3907336bf3484581a00R711

Gabrola: I removed it in this commit https://github.com/garden-co/jazz/pull/3512/commits/55230a667b06146d80732e783dfe37879d8246e5

I think its better to shrink what we keep in the cursor than needing to replay the same exact error. We still get the safety of not loading any extra and unexpected data, so I think that's what matters the most.

packages/jazz-tools/src/tools/subscribe/cursor.ts
Gabrola: @gdorsi I changed it to do a subset check instead of equality

packages/jazz-tools/src/tools/subscribe/cursor.ts
gdorsi: Yours was actually a good point.

The same cursor, loaded by different users might give a different result.

What happens when the cursor is created by a user with restricted access and then loaded by someone with more access rights?

Are the "new values" discovered by the admin loaded without a cursor?

gdorsi: Also, does cursors cover groups?

What happens if permissions change?

gdorsi: > What happens when the cursor is created by a user with restricted access and then loaded by someone with more access rights?

Checked myself, the answer is in `  this.frontier = decodedCursor.frontiers[id] ?? CoValueLoadingState.UNAVAILABLE;`

So anything that was unavailable at snapshot time remains unavailable.

gdorsi (approved):
Stellar job!

PR #3515, opened 2026-03-31
chore: remove axios by installing wasm-pack via cargo

joeinnes:
## Summary

- Remove the npm `wasm-pack` package from `cojson-core-wasm` devDependencies, eliminating `axios` (and `binary-install`) from the dependency tree
- Install `wasm-pack` via `cargo install` in `scripts/setup-cojson-core.sh` instead, matching the existing `cargo-ndk` pattern
- Remove `wasm-pack` from `onlyBuiltDependencies` in root `package.json`

`wasm-pack` is a build-time dependency — anyone building the WASM crate already needs a full Rust toolchain, so installing it via cargo is the natural fit. This avoids exposing us to additional transitive dependency risks through npm (the npm package pulled in `binary-install` → `axios`).

No CI changes needed — CI already installs `wasm-pack` via `taiki-e/install-action`.

## Test plan

- [ ] `pnpm build` succeeds (requires `wasm-pack` on PATH via cargo)
- [ ] `pnpm why axios` returns nothing
- [ ] Lockfile no longer contains `axios`, `binary-install`, or `wasm-pack`

### DIFF
--- a/.changeset/remove-axios.md
+++ b/.changeset/remove-axios.md
@@ -1,5 +1,0 @@
----
-"cojson-core-wasm": patch
----
-
-Remove npm wasm-pack dependency in favour of cargo-installed wasm-pack, eliminating axios from the dependency tree
--- a/.changeset/remove-axios.md
+++ b/.changeset/remove-axios.md
@@ -1,0 +1,5 @@
+---
+"cojson-core-wasm": patch
+---
+
+Remove npm wasm-pack dependency in favour of cargo-installed wasm-pack, eliminating axios from the dependency tree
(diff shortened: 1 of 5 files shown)


.changeset/remove-axios.md
```diff
@@ -0,0 +1,5 @@
+---
+"cojson-core-wasm": patch
+---
+
+Remove npm wasm-pack dependency in favour of cargo-installed wasm-pack, eliminating axios from the dependency tree
```
gdorsi: This change is not user-facing, so we should not need a changeset here.

gdorsi (approved):

PR #3517, opened 2026-04-01
feat: add new co.snapshotRef covalue

Gabrola:
## Summary
Adds `co.snapshotRef()` — a new CoValue primitive that captures an immutable, point-in-time reference to another CoValue. This allows developers more easily leverage Jazz's history functionality and makes it trivial to build use cases like message replies, quoting/citing content, pinned versions of documents where you need to reference a specific state of collaborative data without it changing under you.

### How it works
A `SnapshotRef` stores a `targetId` and a `cursor. When loaded through the subscription system, the SnapshotRef is **transparently resolved** — consumers receive the inner CoValue directly with a `$snapshotRef` handle attached, rather than the wrapper. This passthrough design means existing resolve queries and deep loading work seamlessly on snapshot-referenced data.
Deduplication is built in: two snapshots of the same CoValue at the same cursor state and assigned the same owner produce the same underlying CoMap (via blake3-hashed uniqueness), so repeated snapshots for unchanged data are free. However, this needs an owner to be passed explicitly.

### Key changes
- **`co.snapshotRef(schema, { cursorResolve })`** — new schema definer that wraps any CoValue schema. The `cursorResolve` option controls how deep the cursor captures (e.g. `{ pet: true }` to include nested refs in the snapshot boundary).
- **`SnapshotRef` class + `SnapshotRefSchema`** — full CoValue implementation with `create()`, `load()`, `loadRef()`, `loadReferencedValue()`, plus standard schema methods (`.resolved()`, `.withPermissions()`, `.optional()`).
- **`SubscriptionScope` passthrough** — when the subscription target is a SnapshotRef, the scope loads the ref map first, then transparently redirects to the inner CoValue with the captured cursor applied. Child ID mapping is handled so parent scopes find their children correctly.
- **Type system integration** — `CoFieldInit`, `InstanceOfSchema`, `InstanceOfSchemaCoValuesMaybeLoaded`, and `CoFieldSchemaInit` all updated to support SnapshotRef passthrough types.
- **Chat example updated** — demonstrates the feature with message replying (snapshot the replied-to message), message editing (`applyDiff`), reply preview bubbles with scroll-to-original, and hover action buttons.
- **Test suite** covering creation, deduplication, loading, passthrough, deep nesting, cross-node sync, and error cases

### DIFF
--- a/.changeset/stale-cobras-tap.md
+++ b/.changeset/stale-cobras-tap.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Add new co.snapshotRef() CoValue that abstracts away cursor-based snapshotting and allows snapshots of values to be seamlessly referenced
--- a/examples/chat/src/chatScreen.tsx
+++ b/examples/chat/src/chatScreen.tsx
@@ -1,9 +1,10 @@
 import { Account } from "jazz-tools";
 import { createImage } from "jazz-tools/media";
 import { useSuspenseAccount, useSuspenseCoState } from "jazz-tools/react";
-import { useState } from "react";
-import { Chat, Message } from "./schema.ts";
+import { useRef, useState } from "react";
+import { Chat, Message, MessageSnapshot } from "./schema.ts";
 import {
+  BubbleActions,
   BubbleBody,
   BubbleContainer,
   BubbleImage,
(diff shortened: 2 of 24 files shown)


gdorsi (approved):
Awesome, thanks!

PR #3519, opened 2026-04-10
Fix Better Auth race condition by tracking auth generation

joeinnes:
## What

This PR adds an authGeneration counter to the Better Auth client plugin that increments on each authenticateOnJazz call. The generation is attached to every outgoing request via an x-jazz-auth-generation header and checked when a /get-session response returns null — a logout is only triggered if the generation on the request matches the current generation, preventing stale in-flight session responses from logging out a user who has since re-authenticated.

## Why

A race condition existed where a slow /get-session request could return null after the user had already signed in via a different flow. Without generation tracking, the plugin would interpret the stale null as a genuine session loss and call jazzContext.logOut(), undoing the authentication that had just succeeded. This is especially likely during OAuth flows where multiple session checks may fire concurrently.

### DIFF
--- a/packages/jazz-tools/src/better-auth/auth/client.ts
+++ b/packages/jazz-tools/src/better-auth/auth/client.ts
@@ -91,10 +91,12 @@ export const jazzPluginClient = () => {
         name: "jazz-plugin",
         hooks: {
           async onRequest(context) {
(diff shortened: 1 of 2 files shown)


Takeno (approved):
LGTM

PR #3520, opened 2026-04-10
fix: createContext hangs when migration loads a CoValue with a schema mismatch

joeinnes:
## Problem

When a custom `AccountSchema` migration calls `load()` on a CoValue that fails `fromRaw` instantiation (e.g. a `co.discriminatedUnion()` encountering a stored value with no matching variant), `createJazzContext` hangs forever rather than resolving or rejecting.

The root cause is in `CoValueCoreSubscription.subscribe()`. When a CoValue arrives asynchronously from a peer, its subscription callback fires inside cojson's internal update loop. If `this.emit()` throws — which it does when `instantiateRefEncodedFromRaw` throws on a schema mismatch — the error is silently swallowed by cojson. `handleUpdate` is never called, the `load()` promise never settles, and the migration awaiting it hangs indefinitely.

## Fix

Add a `syncCallback` flag in `CoValueCoreSubscription.subscribe()` to distinguish between synchronous and asynchronous callback invocations:

- **Sync** (value already in local cache): errors propagate naturally through the call stack, preserving existing behaviour — `load()` rejects, `subscribe()` throws. The CoVector schema-mismatch tests rely on this.
- **Async** (value arriving from a peer): errors are caught and re-emitted as `UNAVAILABLE`, so the `load()` promise settles and context creation can complete.

## Tests

- Adds a regression test for the hang: creates a `TypeA` CoValue, then opens a new context whose migration tries to load it as a `discriminatedUnion` that only knows `TypeB`. Previously timed out; now resolves.
- Adds a test confirming synchronous migration errors still reject `createContext` immediately.

### DIFF
--- a/packages/jazz-tools/src/tools/subscribe/CoValueCoreSubscription.ts
+++ b/packages/jazz-tools/src/tools/subscribe/CoValueCoreSubscription.ts
@@ -178,12 +178,29 @@ export class CoValueCoreSubscription {
   private subscribe(value: CoValueCore): void {
     if (this.unsubscribed) return;
 
-    // Subscribe to the value and store the unsubscribe function
+    // Track whether we're still in the synchronous call to value.subscribe().
+    // Callbacks fired synchronously (value already in cache) should let errors
+    // propagate naturally so callers like load() and subscribe() can surface them.
+    // Callbacks fired asynchronously (value arriving from peers) run inside
+    // cojson's internal update loop, which swallows thrown errors — so we must
+    // catch them ourselves and emit UNAVAILABLE instead, or the promise hangs.
+    let syncCallback = true;
+
     this._unsubscribe = value.subscribe((value) => {
       if (value.isAvailable()) {
-        this.emit(value);
+        if (syncCallback) {
+          this.emit(value);
+        } else {
+          try {
+            this.emit(value);
+          } catch {
+            this.emit(CoValueLoadingState.UNAVAILABLE);
+          }
+        }
       }
     });
+
+    syncCallback = false;
 
     if (!value.isAvailable()) {
       this.load(value);
--- a/packages/jazz-tools/src/tools/subscribe/CoValueCoreSubscription.ts
+++ b/packages/jazz-tools/src/tools/subscribe/CoValueCoreSubscription.ts
@@ -178,29 +178,12 @@ export class CoValueCoreSubscription {
   private subscribe(value: CoValueCore): void {
     if (this.unsubscribed) return;
 
-    // Track whether we're still in the synchronous call to value.subscribe().
-    // Callbacks fired synchronously (value already in cache) should let errors
-    // propagate naturally so callers like load() and subscribe() can surface them.
-    // Callbacks fired asynchronously (value arriving from peers) run inside
-    // cojson's internal update loop, which swallows thrown errors — so we must
-    // catch them ourselves and emit UNAVAILABLE instead, or the promise hangs.
-    let syncCallback = true;
-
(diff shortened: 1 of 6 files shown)


packages/jazz-tools/src/tools/tests/createContext.test.ts
```diff
@@ -144,6 +145,69 @@ describe("createContext methods", () => {
...
+          peers: [getPeerConnectedToTestSyncServer()],
+          crypto: Crypto,
+          AccountSchema: CustomAccount,
+          sessionProvider: randomSessionProvider,
+          asActiveAccount: true,
+        }),
+      ).rejects.toThrow("migration failed");
+    }, 2000);
+
+    test("does not hang when a CoValue loaded during migration fails fromRaw due to a schema mismatch", async () => {
```
nrainhart: This test feels overly specific. Can't we simplify this to just any async migration error?

packages/jazz-tools/src/tools/subscribe/CoValueCoreSubscription.ts
```diff
@@ -178,13 +178,30 @@ export class CoValueCoreSubscription {
...
       if (value.isAvailable()) {
-        this.emit(value);
+        if (syncCallback) {
+          this.emit(value);
+        } else {
+          try {
+            this.emit(value);
+          } catch {
+            this.emit(CoValueLoadingState.UNAVAILABLE);
+          }
```
nrainhart: ⚠️ It feels weird to always swallow async errors and silently treat the covalue as unavailable. It could be confusing for folks, as we're not surfacing the cause of why that covalue is unavailable.

Maybe this is the best we can do right now, as we don't currently expose the reason why a covalue is unavailable in the response object. But should we at least log the error?

nrainhart (commented):
It feels like the fix is more general than the comment and tests indicate, I'd tweak them a bit to make this clearer. I also left a comment about how we should provide context to users about async errors

packages/jazz-tools/src/tools/tests/createContext.test.ts
```diff
@@ -144,6 +146,47 @@ describe("createContext methods", () => {
...
+          peers: [getPeerConnectedToTestSyncServer()],
+          crypto: Crypto,
+          AccountSchema: CustomAccount,
+          sessionProvider: randomSessionProvider,
+          asActiveAccount: true,
+        }),
+      ).rejects.toThrow("migration failed");
+    }, 2000);
+
+    test("load() settles as unavailable when a discriminated union has no matching variant", async () => {
```
nrainhart: I'd move this into the discriminated union tests, and maybe keep another test here for an account migration that loads an invalid discriminated union

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
```diff
@@ -173,7 +174,28 @@ export class SubscriptionScope<D extends CoValue> {
           }
 
           this.migrating = true;
-          const instance = instantiateRefEncodedFromRaw(this.schema, value);
+          let instance: CoValue;
+          try {
```
nrainhart, line 178: Nit: we could just merge both try-catch blocks

nrainhart (approved):
The new solutions looks better! Thanks for iterating on it ✨

packages/jazz-tools/src/tools/subscribe/SubscriptionScope.ts
joeinnes: Tried this but ended up deciding to keep the two blocks. The rethrow and the different error messages I think justify keeping them separate

PR #3524, opened 2026-04-14
Experimental clock-sync for clients with drifting system clocks

joeinnes:
## Summary
- Adds an optional `experimental_clockSyncFromServerPings` flag on `createJazzContext` / `startWorker` / `JazzReactProvider` / React Native / Svelte / browser guest contexts. When enabled, a client's locally-authored `madeAt` is corrected toward server time using the sync server's existing WebSocket ping traffic as a one-way offset signal — well-behaving clients with a drifting system clock pull their own stamps back to truth instead of stamping transactions 20 s in the future.
- New `ClockOffset` estimator (median over a sliding window, outlier-gated, magnitude-capped) consumes samples from a new `onPingReceived` callback on `cojson-transport-ws`. Samples are always collected; only the flag decides whether `LocalNode.stampNow()` applies the correction.
- Default is off. When off, transaction stamping is unchanged.

## Notes
- `LocalNode` exposes `getClockOffsetDiagnostics()` returning `{ currentOffset, sampleCount }` for debug harnesses. Not re-exported as a public API.
- Includes an end-to-end regression that reproduces a reported `OneToOneChat` flow (client 20 s ahead, RPC to a worker that publishes and removes itself) and confirms it works under the flag with a seeded offset pulling the client back to real time.
- Known follow-up: `ClockOffset`'s first-accepted-sample path only passes the absolute magnitude gate, not the outlier gate. A transient delay on the very first ping can seed a poor baseline. Worth revisiting once real-world sample distributions are observable — not a blocker for an experimental flag.

## Test plan
- [ ] `pnpm -F cojson test` — 1368 / 13 skipped / 0 failed
- [ ] `pnpm -F cojson build` — clean
- [ ] `pnpm -F cojson-transport-ws test` — 48 / 0 failed
- [ ] `pnpm -F jazz-tools exec vitest run createContext ContextManager worker` — 56 / 0 failed, no type errors
- [ ] Smoke-test a staging worker with `experimental_clockSyncFromServerPings: true` and verify `getClockOffsetDiagnostics()` surfaces a sensible offset against real traffic

### DIFF
--- a/packages/cojson/src/ClockOffset.ts
+++ b/packages/cojson/src/ClockOffset.ts
@@ -41,6 +41,11 @@ export class ClockOffset {
       return;
     }
 
+    // Outlier gate: reject samples that are too far from the current estimate.
+    // This prevents a single bad ping from poisoning the median.
+    // Note: on the first sample this gate is skipped — only the magnitude cap applies.
+    // A transient delay on that first ping can seed a poor baseline, causing subsequent
+    // legitimate samples to be rejected. Worth revisiting once we have real-world data.
     if (
       this.samples.length > 0 &&
       Math.abs(impliedOffset - this.cachedOffset) > this.outlierThresholdMs
--- a/packages/cojson/src/ClockOffset.ts
+++ b/packages/cojson/src/ClockOffset.ts
@@ -1,0 +1,78 @@
+export type ClockOffsetOptions = {
+  windowSize?: number;
+  outlierThresholdMs?: number;
+  maxAbsOffsetMs?: number;
+};
+
+export type ClockOffsetSample = {
+  serverTime: number;
+  localReceiveTime: number;
+};
+
+const DEFAULT_WINDOW_SIZE = 32;
+const DEFAULT_OUTLIER_THRESHOLD_MS = 2000;
+const DEFAULT_MAX_ABS_OFFSET_MS = 24 * 60 * 60 * 1000;
+
+export class ClockOffset {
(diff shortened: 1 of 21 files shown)


gdorsi (approved):

packages/cojson/src/ClockOffset.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    if (Math.abs(impliedOffset) > this.maxAbsOffsetMs) {
+      return;
+    }
+
+    if (
+      this.samples.length > 0 &&
+      Math.abs(impliedOffset - this.cachedOffset) > this.outlierThresholdMs
+    ) {
+      return;
```
gdorsi, line 53: Suggestion: it would help to add a brief comment on the outlier gate explaining its purpose and the first-sample bypass. Something like:

```ts
// Outlier gate: reject samples that are too far from the current estimate.
// This prevents a single bad ping from poisoning the median.
// Note: on the first sample this gate is skipped — only the magnitude cap applies.
// A transient delay on that first ping can seed a poor baseline, causing subsequent
// legitimate samples to be rejected. Worth revisiting once we have real-world data.
if (
  this.samples.length > 0 &&
  Math.abs(impliedOffset - this.cachedOffset) > this.outlierThresholdMs
) {
  return;
}
```

This would make the intent self-documenting since the code will outlive the PR description.

joeinnes: Done

PR #3527, opened 2026-07-02
fix: SQLite transaction errors during storage reconciliation on expo-sqlite

### LINKED ISSUE #1046: SQLite transaction errors during storage reconciliation on React Native (expo-sqlite)
We're seeing multiple SQLite transaction errors during Jazz's storage reconciliation on React Native (Expo) with physical iOS devices. These errors appear in the `StoreQueue` and `expo-sqlite` adapter during initial sync and can cause the app to hang on the loading state for 2+ minutes.

## Environment
- jazz-tools: 0.20.17
- expo-sqlite: ~16.0.10
- cojson (bundled with jazz-tools 0.20.17)
- Platform: iOS 26.5 (physical iPhone 17,3)
- Expo SDK: 56

## Errors observed

Three distinct SQLite errors appear during storage reconciliation of ~193 CoValues:

### 1. Cannot rollback — no transaction is active
```
Error: Calling the 'execAsync' function has failed
→ Caused by: cannot rollback - no transaction is active
```

### 2. UNIQUE constraint failed
```
Error processing message in store queue
Error: Calling the 'finalizeAsync' function has failed
→ Caused by: Error code 19: UNIQUE constraint failed: transactions.ses, transactions.idx
```
Call stack points to `cojson/dist/queue/StoreQueue.js:51`. The `INSERT INTO transactions (ses, idx, tx)` in `sqliteAsync/client.js:36` has no `ON CONFLICT` clause, so duplicate writes throw.

### 3. Cannot start a transaction within a transaction
```
Error: Calling the 'execAsync' function has failed
→ Caused by: cannot start a transaction within a transaction
```

## Reproduction

Inconsistent — we triggered the 2+ minute hang once on a cold start (fresh app install, empty SQLite database). After force-closing and reopening, the app loaded normally. Subsequent reinstalls did not reproduce the hang, though errors #2 and #3 continued to appear on every launch (without blocking the app).

Console output during the hang:
```
INFO  Starting storage reconciliation {"peerId": "wss://cloud.jazz.tools/...", "startOffset": 0, "total": 193}
LOG   migrating account
ERROR [cannot rollback - no transaction is active]
```

The account migration (`ConsembleAccount.withMigration`) runs concurrently with reconciliation, which may contribute to the transaction conflicts.

## Analysis

Looking at the code in `cojson/dist/storage/sqliteAsync/client.js`:

- `SQLiteClientAsync` serializes write transactions via `txQueue` (line 84), but read operations (`getCoValue`, `getCoValueSessions`, `getCoValueIDs`, `getCoValueKnownState`) bypass the queue and hit SQLite directly
- During reconciliation, `buildStorageReconciliationEntries` fires concurrent `loadKnownState` reads while the `StoreQueue` processes incoming sync messages with write transactions
- The `expo-sqlite` `withTransactionAsync` may not handle the case where SQLite auto-rolls-back a transaction due to a conflict — it then tries to explicitly ROLLBACK, causing error #1

## Impact

When the errors are non-blocking (errors #2 and #3), the app functions normally. When error #1 occurs during a cold start reconciliation, it can cause the app to hang indefinitely on the loading state, requiring a force-close.

## Versions checked

Verified that jazz-tools 0.20.18 has an identical `expo-sqlite-adapter.ts` — the fix is not in the latest release.

gdorsi:
Fixes https://github.com/garden-co/jazz/issues/1046

## Problem

expo-sqlite's `withTransactionAsync` is not exclusive, and Jazz was serializing transactions per storage client rather than per connection. Since the shared `ExpoSQLiteAdapter` (via `getInstance`) is reused across multiple contexts on the same device — auth transitions, anonymous account migration, multiple providers — several of these contexts can end up issuing transactions on the same underlying connection at once.

When that happens, a second `BEGIN` fails with "cannot start a transaction within a transaction", and expo-sqlite's own catch handler then issues a `ROLLBACK` — which aborts the *other* caller's active transaction ("cannot rollback - no transaction is active"). That mid-flight rollback breaks the atomicity of `storeSingle`, leaving orphan rows with `idx >= ` the session's `lastIdx`. From that point on, every subsequent store for that session fails permanently with `UNIQUE constraint failed: transactions.ses, transactions.idx`.

## Fix

- **Connection-level transaction queue in `ExpoSQLiteAdapter`**: transactions issued against the shared adapter are now serialized at the connection level instead of relying on each storage client to serialize its own calls, so concurrent callers can no longer interleave `BEGIN`/`ROLLBACK` on the same connection.
- **`ON CONFLICT(ses, idx) DO UPDATE` upserts** in `addTransaction` and `addSignatureAfter` of the async SQLite client, so databases that were already corrupted by this bug self-heal instead of failing forever.

### Why the upserts cannot break signature consistency

Session logs are signed, so overwriting rows would be dangerous if it could touch content a stored signature covers. It can't, structurally:

1. **A conflict can only happen outside the signed range.** `putNewTxs` computes its write range inside the same DB transaction from the committed session row: it slices off everything below `lastIdx` and starts inserting exactly at `idx = lastIdx`. Rows at `idx < lastIdx` — the ones covered by `sessions.lastSignature` and `signatureAfter` — are never written at all (this also means a divergent fork of a session is sliced away, as before). A conflict at `idx >= lastIdx` means the row exists but the committed session state says it shouldn't — by definition an orphan no stored signature accounts for.
2. **The signature state commits atomically with the overwritten rows.** `storeSingle` writes the session row (`lastIdx` + `lastSignature`), `signatureAfter`, and the transaction rows in one driver transaction, so there is no window where an overwritten row coexists with a signature computed over the old orphan content.
3. **The replacement content is already signature-verified.** Storage is only fed from the sync manager after per-session verification, so what overwrites the orphan is the verified chain, not raw peer input.

Overwriting orphan `signatureAfter` rows is additionally load-bearing: `loadCoValue` reads *all* signature rows for a session unbounded by `lastIdx`, so an orphan signature row is actually served to readers on load until it is repaired.

The one case the self-heal deliberately does not touch: a hypothetical mismatched row *below* `lastIdx`. That still surfaces as a verification failure at load and gets corrected from peers, exactly as before.

## Testing

- New adapter tests reproduce expo-sqlite's exact `withTransactionAsync` semantics and fail without the connection-level serialization fix.
- New `SQLiteClientAsync` tests cover orphan-row recovery for both `addTransaction` and `addSignatureAfter`.
- Full async storage/reconciliation test suites pass (106 tests).

### DIFF
--- a/.changeset/heavy-moons-repair.md
+++ b/.changeset/heavy-moons-repair.md
@@ -1,0 +1,5 @@
+---
+"cojson": patch
+---
+
+Recover from partially-written storage rows in async SQLite storage. Interrupted write transactions could leave orphan `transactions`/`signatureAfter` rows that made every subsequent store of the same session fail with "UNIQUE constraint failed: transactions.ses, transactions.idx"; those rows are now overwritten instead.
--- a/.changeset/spicy-pandas-refuse.md
+++ b/.changeset/spicy-pandas-refuse.md
@@ -1,0 +1,5 @@
+---
+"jazz-tools": patch
+---
+
+Serialize Expo SQLite transactions at the connection level. The adapter is shared across providers/contexts, so concurrent storage clients could nest `BEGIN` statements on the same connection, causing "cannot start a transaction within a transaction" and "cannot rollback - no transaction is active" errors during storage reconciliation.
--- a/packages/cojson/src/storage/sqliteAsync/client.ts
+++ b/packages/cojson/src/storage/sqliteAsync/client.ts
@@ -96,10 +96,12 @@ export class SQLiteTransactionAsync implements DBTransactionInterfaceAsync {
     newTransaction: Transaction,
   ): Promise<void> {
     // An interrupted write transaction (e.g. a concurrent ROLLBACK on a shared
-    // connection) can leave orphan rows with idx >= the session's lastIdx.
-    // A session's transaction at a given idx is canonical, so overwriting is
-    // safe and lets the session recover instead of failing on the primary key
-    // forever.
+    // connection) can leave orphan rows at idx >= the session's lastIdx, which
+    // would otherwise block the session on this primary key forever.
+    // Overwriting is safe: writes always start at the committed lastIdx (see
+    // putNewTxs), so a conflicting row is never covered by any stored
+    // signature, and the verified replacement commits atomically with the
+    // session row and signature that cover it.
     await this.tx.run(
       `INSERT INTO transactions (ses, idx, tx) VALUES (?, ?, ?)
        ON CONFLICT(ses, idx) DO UPDATE SET tx=excluded.tx`,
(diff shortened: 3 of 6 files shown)


nrainhart (approved):
Thanks for the fixes!!

PR #3529, opened 2026-07-03
Prevent session forking when a crash races local persistence

gdorsi:
## Summary

A client that crashes after a transaction reached the sync server but before the local (IndexedDB / async SQLite) write completed would, on restart, reuse the same session from a stale position and produce a conflicting transaction at an already-taken index — the server rejects the signature and the session can never sync again.

This PR closes that window without coupling remote send latency to local persistence:

- **cojson**: `StorageAPI.store` gains an optional `done` callback that fires only after a durable store (including correction round-trips). `SyncManager` counts sent-but-unpersisted local content and notifies `onLocalStoreDurabilityChange(hasPending, sessionID)` synchronously *before* the first peer send; the listener is supplied via `LocalNode` creation options so it is live before account migration can sync. Sync storage completes inline, so the whole feature is a no-op there.
- **jazz-tools**: a `SessionDurabilityMarker` (per-session crash-surviving dirty flag) is set from that listener (immediate set, debounced clear). `BrowserSessionProvider` (localStorage marker) and `ReactNativeSessionProvider` (KvStore marker) skip dirty sessions at acquire time and mint a fresh session instead, reclaiming the abandoned session's storage slot. Abandoning a session loses no data: the server re-sends the crashed transactions on the next connect.

Design doc: `docs/superpowers/specs/2026-07-02-session-durability-marker-design.md`

## Notes / accepted trade-offs

- Marker durability is best-effort (localStorage is not fsync-guaranteed on hard OS crashes; RN KV writes are async-initiated before the send). The vulnerable window shrinks from "any crash during any write" to "power loss within milliseconds of an edit".
- A tab starting while another live tab is inside its ~200ms write window can reclaim that tab's slot — worst case one extra session, no divergence possible.
- Closing a tab within the 200ms clear debounce burns one session on next launch.

## Test plan

- [x] New cojson suite `sync.localStoreDurability.test.ts` (13 tests): store `done` semantics (async, sync, correction path), window open-before-send ordering, sync-storage/no-peer/remote-content negative cases, close-then-reopen, crash-window stays-open, and an end-to-end crash-recovery regression (fresh session over same storage recovers the server-held transaction and syncs cleanly).
- [x] jazz-tools: listener adapter unit tests (debounce, reopen-cancels-clear, error swallowing), context wiring tests, browser provider skip/reclaim + concurrent-tab test, RN provider dirty-session tests.
- [x] Full suites green: cojson (1382), jazz-tools (2118 incl. typecheck), cojson-storage-sqlite, cojson-storage-do-sqlite, cojson-storage-indexeddb.


packages/jazz-tools/src/react-native-core/ReactNativeSessionDurabilityMarker.ts
```diff
@@ -0,0 +1,40 @@
...
+// provider that reads this marker uses the "jazz-tools" one
+import {
+  KvStoreContext,
+  sessionDurabilityMarkerKey as markerKey,
+} from "jazz-tools";
+
+/**
+ * KvStore-backed durability marker. KvStore writes are async, so `set` can
+ * only be *initiated* before the network send — a best-effort guarantee
+ * (accepted limitation, see the design spec's residual-risk section).
```
nrainhart:
```suggestion
 * (accepted limitation).
```

nrainhart (approved):
Nice fix!!

PR #3534, opened 2026-07-04
feat(cojson): native NodeCore for the wasm binding

gdorsi:
Stacked on #3533 (unified validateTransactions) → #3532 → #3531. Part A of the cleanup phase.

## Summary

- Ports the `NodeCore` registry + unified validation surface to `cojson-core-wasm` (wasm-bindgen mirror of the napi binding, 27-member parity) and gives `WasmCrypto` a native adapter — browsers now run the same Rust permission engine as Node, replacing the TypeScript shim. The wire types deliberately match `NodeCorePendingTx`/`NodeCoreGroupVerdict` verbatim (serde renames, including the nested `sourceTxId.sessionID` casing) so the adapter is pure pass-through, unlike napi's bridged casing.
- Rebuilds and commits the checked-in wasm artifacts, which had been stale since before the migration began — note: wasm has **no CI recompile gate**, so the artifacts in this PR are the shipped binary (follow-up CI work suggested).
- **Fixes a latent correctness bug that also affected the live napi path**: native verdict application re-marked all transactions each pass, clobbering first-writer-wins invalidations on owned covalues; the naive delta-only fix would in turn have starved group verdict flips (a late-arriving revocation leaving a revoked admin's grants valid — caught in review with a concrete scenario). Landed fix scopes application per ruleset (groups → full history; owned/unsafeAllowAll → delta) with a red/green regression test reproducing the out-of-order revocation flip, which no prior test could catch.

## Manual browser verification (mandatory gate for this PR)

Performed against a local `jazz-run sync` server with the freshly built wasm binary, two isolated browser sessions per app:

- **Chat**: room creation, bidirectional live sync between two identities, reload persistence (session-reuse path confirmed in console), zero page errors, zero wasm-panic/`Unknown CoValue`/`CoValue not loaded` signatures.
- **Organization**: org + project creation, invite-link capture, writer-invite acceptance by a second identity, role-gated UI (member lacks admin controls; Members list renders native-resolved roles), writer's permitted write syncing to admin, then **live revocation**: the removed member's UI flipped to "You don't have access to this organization" while their pre-revocation write correctly remained valid — and post-revocation writes are impossible. Zero page errors across all sessions.

## Test Plan

- [x] cojson suite: 1431 passed / 13 skipped (incl. the new out-of-order revocation regression test, verified red on the unfixed code and green kill-switched)
- [x] Fixture exporter's 25 scenarios now execute through wasm-native validation (the exporter instantiates WasmCrypto) — all green
- [x] jazz-tools: 2120 passed / 8 skipped on the new binary
- [x] cargo 148/148; tsc clean; clippy no new warnings
- [x] Manual browser verification as above


gdorsi:
CI triage: the Lint failure was a `cargo fmt --check` debt dating back to the stage-1 napi commits — now fixed with a mechanical `chore: rustfmt crates workspace` commit pushed to every branch in the stack (#3531–#3535), so cascade merges stay clean. The test (1/2) failure is the known-flaky `SessionLock.cross-browser` Playwright strict-mode race (passes on #3533, which contains this PR's entire diff); the push retriggers it.

PR #3536, opened 2026-07-04
feat(cojson): migrate cojson core and comap to rust

gdorsi:
Stacked on #3535 (RN NodeCore) → #3534 → #3533 → #3532 → #3531. Originally Part C of the permissions cleanup (TS engine deletion); this PR has since grown to also carry the opening stages of the follow-on migration — moving CoValueCore's content materialization into Rust (spec: `docs/superpowers/specs/2026-07-04-covaluecore-in-rust.md`).

## Part 1 — TS permission engine deletion (original scope, unchanged)

- Deletes the TypeScript permission-validation engine (`determineValidTransactionsForGroup`, `MemberRoleResolver`, `isHigherRole`, the ownedByGroup/unsafeAllowAll fallback branches — ~656 lines from permissions.ts), the `COJSON_DISABLE_NATIVE_VALIDATION` kill switch, the `ShimNodeCore` capability shim, the differential harness, and four dead classifier exports. `validateTransactions`/`roleOf`/`resetValidation` become required on `NodeCoreImpl`; `createNodeCore()` becomes abstract. Every provider (napi, wasm, RN) now validates natively with no fallback.
- Oracle inversion handled explicitly: the 25 committed group-engine fixtures are frozen golden files from the deleted TS engine; two independent regenerate-and-diff runs confirmed the native engine reproduces them byte-for-byte.
- Manually verified in the browser (wasm, no fallback): chat sync, org invite/writer-write/live revocation — zero error signatures.

## Part 2 — CoValueCore-in-Rust: R0-R3.5 (new)

Goal: move CRDT content materialization (starting with coMap) into `cojson-core`, with a hard performance bar (beat the TS baseline, measured against a pinned end-to-end yardstick vs `origin/main`) and a hard backward-compat bar (public API/wire/storage unchanged).

- **R0** — prototyped three read boundaries for a Rust-resident coMap view; benchmarked a Rust-view-plus-delta-cache design as the clear winner on both napi and wasm (1.16-2.2x on cached reads), per-key/snapshot boundaries disqualified. This pattern is now the template for every CRDT.
- **R1** — per-node key store in `NodeCore`; native decryption of private transactions feeds the same views (secrets never cross back out); read wins hold on private data (up to ~4x).
- **R2** — safe incremental permission validation (suffix-only ingests extend cached verdicts instead of a full O(n²) recompute per ingest, with a conservative any-doubt-recompute fallback) plus native meta/merge parsing. An adversarial review caught a real gap (a key arriving mid-ingest could leave stale verdicts cached); fixed and closed with a red→green regression plus a widened 200-seed property test. Write-path control case: **17-23x faster**.
- **R2.5** — closed the remaining O(n²) ingest costs (whole-session re-parsing, full verdict re-marshaling); local-edit write path went from 0.04x to **0.95x** of the TS baseline (its ceiling — signing cost shared identically with `main`).
- **R3** — rich per-op coMap materialization (frontier reads, content fixtures frozen from the current TS implementation before any deletion), plus a single-call native ingest kernel. Stage "2b" (always routing `RawCoMap` through native materialization) was fully implemented and functionally correct, but its own yardstick run showed it **regressed** ingest performance (~0.4x vs `main`, worse than the pre-2b baseline) — the per-op JSON transfer for a cold/full materialization costs more than the JS loop it replaces. **Disabled by default** behind `nativeCoMapMaterializationEnabled` (off), preserving the capability for future work without shipping a regression.
- **R3.5** — attempted a cheaper two-tier transfer (eager cheap latest-values + lazy per-key op history). Isolated benchmarking showed a real 3-17x improvement over the rich transfer and a win on op-dense maps, but confirmed a genuine floor: plain trusting-content bulk ingest has nothing for Rust to amortize (no decryption, no validation), so any transfer-boundary redesign loses to TS's in-process walk on that specific shape. Left unwired, flag stays off — this is a physics-bound limit on this workload, not a defect.

**Net effect on the pinned yardstick** (`bench/cojson/yardstick-compare.sh`, current branch vs `origin/main`): permission-validation-heavy and private-data paths substantially faster (see above); plain local edits at parity; plain-content bulk ingest still ~0.5-0.6x pending a CRDT where Rust has real work to amortize (R4 target).

## Test Plan

- [x] cargo (cojson-core): 212 passed
- [x] cojson: 1427 passed / 13 skipped
- [x] jazz-tools: 2120 passed / 8 skipped (hard backward-compat bar)
- [x] tsc clean
- [x] 25 permission-engine golden fixtures + new coMap content fixtures, byte-for-byte
- [x] Manual browser re-verification (wasm, permission-engine deletion)
- [x] Yardstick vs `origin/main` run at each phase (see spec for full tables)


gdorsi:
Final phase review verdict: READY, no blocking issues. One provenance note: the committed wasm artifacts (built at 80198a2f5 in #3534) predate the workspace-wide rustfmt commit, which reformatted cojson-core sources that compile into the binary. Whitespace-only — the binary is semantically identical — so no rebuild was done here; the next functional wasm change should rebuild as usual.

PR #3541, opened 2026-07-22
fix(jazz-tools): support Better Auth 1.6 client types

joeinnes:
## Summary

- prevent `jazzPluginClient` declarations from leaking Better Auth's version-specific Better Fetch callback types
- preserve standard Better Auth client inference while retaining Jazz account and action inference
- add a pinned Better Auth 1.6.23 declaration consumer check and deterministic browser provider test setup

## Testing

- `pnpm --filter jazz-tools test:better-auth-compat`
- `pnpm --filter jazz-tools exec vitest run src/better-auth/auth/tests --project jazz-tools`
- `pnpm --filter jazz-tools run check`


gdorsi (approved):
LGTM, thanks!

